PR #445, opened 2016-12-05
Allows couchjs to run under snap confinement

mhall119:
Add snapd browser-support interface to allow couchjs to control process priority


rnewson:
+1

PR #446, opened 2016-12-05
Provides access to check disk usage by calling the df command

mhall119:
The AppArmor rules provided by the mount-observe interface have been expanded to cover /bin/df


rnewson:
+1

PR #448, opened 2016-12-26
Support pass command line arguments to couchdb when it is called by bash script

iuridiniz:
couchdb accepts command line arguments like -pidfile and -couch_ini, but they are ignored by bin/couchdb, this patch fixes it.


janl:
nice one!

+1

PR #456, opened 2017-03-17
Count JS test pass/fails; return proper exit code

wohali:
Currently `test/javascript/run` always returns an exit code of `0`,
regardless of whether any tests fail or not. This makes any sort of
automated testing that expects to inspect the results of the run
invalid.

`make javascript` should now properly fail if any of the JS tests fail.
It also prints a numerical summary of passed/failed tests, similar to
the eunit results.

Fixes COUCHDB-3328

### DIFF
--- a/test/javascript/run
+++ b/test/javascript/run
@@ -92,6 +92,7 @@ def run_couchjs(test, fmt):
             sys.stderr.write(line)
     p.wait()
     fmt(p.returncode == 0)
+    return p.returncode
 
 
 def options():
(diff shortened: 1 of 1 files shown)


wohali:
Ironically, the failed Travis CI run shows exactly what I mean. I don't know if the failing JS test (`test/javascript/tests/replication.js`) fails only on R16B03 because we have a bug there, or because the test case is flaky. This demands further investigation - _after_ we merge this change. :)

michellephung:
+1 \o/ passing tests

PR #460, opened 2017-03-20
Change snap packaging to build from source

mhall119:
## Overview

Rather than creating a snap package from pre-build binaries, this allows building the snap package directly from source. Makes it possible to use external build automation such as build.snapcraft.io 

## Testing recommendations

On any Ubuntu 16.04 or newer system, apt install "snapcraft" and then run "snapcraft snap" or "make snap". The resulting .snap file can be installed with "snap install --dangerous *.snap"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ config.erl
 install.mk
 rel/*.config
 rel/dev*
-rel/snap*
 rel/tmpdata
 src/
 /log
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,81 @@
+name: couchdb
+version: 2.0
+summary: Document based database
+description: CouchDB is a database that completely embraces the web. Store your data with JSON documents. Access your documents and query your indexes with your web browser, via HTTP. Index, combine, and transform your documents with JavaScript. 
+confinement: strict
+grade: stable
+
+apps:
+    server:
+        daemon: simple
+        command: rel/couchdb/bin/snap_run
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -4,7 +4,6 @@ config.erl
 install.mk
 rel/*.config
 rel/dev*
-rel/snap*
```
wohali: You probably need to add *.snap now, though.

rnewson: ooh I was too hasty. @mhall119 can file another PR if that's the case.

snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+          - libnspr4-dev
+          - libicu-dev
+          - icu-devtools
+          # For erlang/rebar processing
+          - erlang-nox
+          - erlang-os-mon
+          - erlang-syntax-tools
+          # For fauxton
+          - nodejs-dev
+          - nodejs-legacy
```
kxepal: Wonder why two nodejs here.

.gitignore
wohali: You also need to add `rel/snapcraft.yaml`.

mhall119: That hasn't changed from before, but it's easy enough to add if that's desirable

snapcraft.yaml
mhall119: nodejs-legacy is needed for the 'node' executable, which IIRC is being called from somewhere in the build process

kxepal: Thanks!

PR #462, opened 2017-03-21
Tell Git to ignore snap files so they don't clutter things

mhall119:
Per Wohali's request on https://github.com/apache/couchdb/pull/460


wohali:
+1, thanks

PR #469, opened 2017-04-04
Choose index based on fields match

garrensmith:
## Overview

If two indexes can be used, then instead of choosing the index based on
alphabet order of index names, rather choose based on a score. This
score is calculated by determining which index has the least number of
fields

## Testing recommendations

There is a test that proves the issue and passes based on this fix.

## JIRA issue number

https://issues.apache.org/jira/browse/COUCHDB-3357

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -34,7 +34,7 @@
 create(Db, Indexes, Selector, Opts) ->
     FieldRanges = mango_idx_view:field_ranges(Selector),
     Composited = composite_indexes(Indexes, FieldRanges),
-    {Index, IndexRanges} = choose_best_index(Db, Composited),
+    {Index, IndexRanges, _ColPrefixLength} = choose_best_index(Db, Composited),
 
     Limit = couch_util:get_value(limit, Opts, mango_opts:default_limit()),
     Skip = couch_util:get_value(skip, Opts, 0),
@@ -107,10 +107,14 @@ execute(#cursor{db = Db, index = Idx} = Cursor0, UserFun, UserAcc) ->
 % check FieldRanges for a, b, c, and d and return
 % the longest prefix of columns found.
 composite_indexes(Indexes, FieldRanges) ->
-    lists:foldl(fun(Idx, Acc) ->
+    FieldKeys = [Key || {Key, _} <- FieldRanges],
+    SortedIndexes = lists:foldl(fun(Idx, Acc) ->
         Cols = mango_idx:columns(Idx),
         Prefix = composite_prefix(Cols, FieldRanges),
-        [{Idx, Prefix} | Acc]
+        % create a ColPrefixLength based on how close the number of fields
+        % the index has to the number of Prefixes the index can use
+        ColPrefixLength = abs(length(Cols) - length(FieldKeys)),
+        [{Idx, Prefix, ColPrefixLength} | Acc]
     end, [], Indexes).
 
 
@@ -135,14 +139,13 @@ composite_prefix([Col | Rest], Ranges) ->
 % reduce view read on each index with the ranges to find
 % the one that has the fewest number of rows or something.
 choose_best_index(_DbName, IndexRanges) ->
-    Cmp = fun({A1, A2}, {B1, B2}) ->
-        case length(A2) - length(B2) of
+    Cmp = fun({IdxA, PrefixA, ColPrefixLengthA}, {IdxB, PrefixB, ColPrefixLengthB}) ->
+        case ColPrefixLengthA - ColPrefixLengthB of
             N when N < 0 -> true;
             N when N == 0 ->
-                % This is a really bad sort and will end
-                % up preferring indices based on the
-                % (dbname, ddocid, view_name) triple
-                A1 =< B1;
+                % We have no other way to choose, so at this point 
+                % select index based on (dbname, ddocid, view_name) triple
+                IdxA =< IdxB;
             _ ->
                 false
         end
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -34,7 +34,7 @@
 create(Db, Indexes, Selector, Opts) ->
     FieldRanges = mango_idx_view:field_ranges(Selector),
     Composited = composite_indexes(Indexes, FieldRanges),
-    {Index, IndexRanges, _KeysPrefixLength} = choose_best_index(Db, Composited),
+    {Index, IndexRanges} = choose_best_index(Db, Composited),
 
     Limit = couch_util:get_value(limit, Opts, mango_opts:default_limit()),
     Skip = couch_util:get_value(skip, Opts, 0),
@@ -107,14 +107,14 @@ execute(#cursor{db = Db, index = Idx} = Cursor0, UserFun, UserAcc) ->
 % check FieldRanges for a, b, c, and d and return
 % the longest prefix of columns found.
 composite_indexes(Indexes, FieldRanges) ->
-    FieldKeys = [Key || {Key, _} <- FieldRanges],
     lists:foldl(fun(Idx, Acc) ->
         Cols = mango_idx:columns(Idx),
         Prefix = composite_prefix(Cols, FieldRanges),
-        % create a KeysPrefixLength based on how close the prefix of
-        % the index is to the number of keys in the selector
-        KeysPrefixLength = length(FieldKeys) - length(Prefix),
-        [{Idx, Prefix, KeysPrefixLength} | Acc]
+        % Calcuate the difference between the FieldRanges/Selector
+        % and the Prefix. We want to select the index with a prefix
+        % that is as close to the FieldRanges as possible
+        PrefixDifference = length(FieldRanges) - length(Prefix),
+        [{Idx, Prefix, PrefixDifference} | Acc]
     end, [], Indexes).
 
 
@@ -128,36 +128,41 @@ composite_prefix([Col | Rest], Ranges) ->
             []
     end.
 
-% Low and behold our query planner
-% First pass: sort the IndexRanges by KeysPrefixLength,
-% if that is equal, choose the index with the least number of 
+
+% The query planner
+% First choose the index with the lowest difference between its 
+% Prefix and the FieldRanges. If that is equal, then
+% choose the index with the least number of 
 % fields in the index. If we still cannot break the tie, 
 % then choose alphabetically based on ddocId.
-% Return the first element.
+% Return the first element's Index and IndexRanges.
 %
 % In the future we can look into doing a cached parallel
 % reduce view read on each index with the ranges to find
 % the one that has the fewest number of rows or something.
 choose_best_index(_DbName, IndexRanges) ->
-    Cmp = fun({IdxA, PrefixA, KeysPrefixLengthA}, {IdxB, PrefixB, KeysPrefixLengthB}) ->
-        case KeysPrefixLengthA - KeysPrefixLengthB of
+    Cmp = fun({IdxA, _PrefixA, PrefixDifferenceA}, {IdxB, _PrefixB, PrefixDifferenceB}) ->
+        case PrefixDifferenceA - PrefixDifferenceB of
             N when N < 0 -> true;
             N when N == 0 ->
                 ColsLenA = length(mango_idx:columns(IdxA)),
                 ColsLenB = length(mango_idx:columns(IdxB)),
                 case ColsLenA - ColsLenB of
-                    M when M < 0 -> true;
+                    M when M < 0 -> 
+                        true;
                     M when M == 0 ->
                         % We have no other way to choose, so at this point 
-                        % select index based on (dbname, ddocid, view_name) triple
+                        % select the index based on (dbname, ddocid, view_name) triple
                         IdxA =< IdxB;
-                    _ -> false
+                    _ -> 
+                        false
                 end;
             _ ->
                 false
         end
     end,
-    hd(lists:sort(Cmp, IndexRanges)).
+    {SelectedIndex, SelectedIndexRanges, _} = hd(lists:sort(Cmp, IndexRanges)),
+    {SelectedIndex, SelectedIndexRanges}.
 
 
 handle_message({meta, _}, Cursor) ->
@@ -281,5 +286,3 @@ is_design_doc(RowProps) ->
         <<"_design/", _/binary>> -> true;
         _ -> false
     end.
-
- 
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -34,7 +34,7 @@
 create(Db, Indexes, Selector, Opts) ->
     FieldRanges = mango_idx_view:field_ranges(Selector),
     Composited = composite_indexes(Indexes, FieldRanges),
-    {Index, IndexRanges, _ColPrefixLength} = choose_best_index(Db, Composited),
+    {Index, IndexRanges, _KeysPrefixLength} = choose_best_index(Db, Composited),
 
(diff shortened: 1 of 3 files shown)


garrensmith:
I ran the pouchdb-find tests locally against this fix and all tests passed.

src/mango/rebar.config.script
```diff
@@ -12,16 +12,16 @@
 
 DreyfusAppFile  = filename:join(filename:dirname(SCRIPT),
     "../dreyfus/src/dreyfus.app.src"),
-RenameFile = filename:join(filename:dirname(SCRIPT),
-    "src/mango_cursor_text.erl"),
```
davisp: And there's a weird breakage down below with a file rename that looks not right to me.

src/mango/src/mango_cursor_view.erl
```diff
@@ -135,14 +139,16 @@ composite_prefix([Col | Rest], Ranges) ->
...
 choose_best_index(_DbName, IndexRanges) ->
-    Cmp = fun({A1, A2}, {B1, B2}) ->
-        case length(A2) - length(B2) of
+    Cmp = fun({IdxA, PrefixA, ScoreA}, {IdxB, PrefixB, ScoreB}) ->
+        case length(PrefixA) - length(PrefixB) of
             N when N < 0 -> true;
             N when N == 0 ->
                 % This is a really bad sort and will end
                 % up preferring indices based on the
                 % (dbname, ddocid, view_name) triple
```
davisp: You'll want to remove this comment since its no longer relevant.

src/mango/src/mango_cursor_view.erl
```diff
@@ -135,14 +139,16 @@ composite_prefix([Col | Rest], Ranges) ->
...
-                A1 =< B1;
+                %IdxA =< IdxB;
+                %prefer using the index with the lower score
+                ScoreA =< ScoreB;
```
davisp: Score is a bit opaque. I'd just call it length and say that we prefer using the index with the least number of fields. Also I wouldn't bother calculating it in composite indices just to use it here and then have to ignore the return value. You've got all the info here so I'd just calclulate it on the fly.

src/mango/src/mango_cursor_view.erl
```diff
@@ -270,4 +276,4 @@ is_design_doc(RowProps) ->
     case couch_util:get_value(id, RowProps) of
         <<"_design/", _/binary>> -> true;
         _ -> false
-    end.
```
davisp: Another unrelated change we should try to avoid.

src/mango/src/mango_cursor_view.erl
```diff
@@ -107,10 +107,14 @@ execute(#cursor{db = Db, index = Idx} = Cursor0, UserFun, UserAcc) ->
...
 composite_indexes(Indexes, FieldRanges) ->
-    lists:foldl(fun(Idx, Acc) ->
+    FieldKeys = [Key || {Key, _} <- FieldRanges],
+    SortedIndexes = lists:foldl(fun(Idx, Acc) ->
         Cols = mango_idx:columns(Idx),
         Prefix = composite_prefix(Cols, FieldRanges),
-        [{Idx, Prefix} | Acc]
+        % create a score based on how close the number of fields
+        % the index has to the number of fields in the selector
+        Score = length(Cols) - length(FieldKeys),
```
davisp: Seems like we should be looking at the length of the prefix, no?

davisp: To expand on this, there are three things we're contemplating here:

1. Number of {Field, Range} pairs in the selector.
2. Number of columns in the index
3. The shared prefix between those two.

The important case to remember here is that the selector may have more fields than are being satisfied by the index. Ie, a selector may only use the first one or two out of a three column index.

Which means that if you have a selector that has more fields than the index has columns it ends up having a negative score which seems.... odd. However, what we're really caring about here is that we're being the least specific once we have some part of an index satisfied. Its bit odd given that we also want the longest prefix. So the sort in words is basically something like:

I want to use the index with the most matching fields to my selector, but with the least extra fields not used by my selector (ie, the smallest length(Cols) - length(Prefix)).

Though that leaves us with the same current connundrum when we have two indexes with the same number of columns but sharing a common prefix with the selector. (the [a, b, c], [a, d, e] example I mentioned).

For that case I'd still change it away from the docid sort since that's rather ambiguous and generally not supplied by the user and as such is essentially random (ie, its a hash which is hard to predict, its not uuid random though (if memory serves)).

Anyway, so I'd use the length preference you've got and then break ties with the field name past the end of the sort and then I guess fall back to docid if we have two indices that have the same set of columns cause it shouldn't matter?

tonysun83:
Paul hit the nail for the gist of this, so the only thing I'm going to add is that for the tests, we should probably use the _explain endpoint and make sure we're choosing the right. So

1) Create a bunch of indexes with various columns lengths. 

2) Create selectors that use various combinations of fields

3) Execute those queries but with _explain, and ensure that the correct index with the right prefix was used by checking the index name.

src/mango/src/mango_cursor_view.erl
garrensmith: I've made an attempt at fixing this. I'm not 100% I fully understood this. In the situation of [a, b, c] and [a, d, e] I've defaulted to the indices names. The only problem with that is there is a possibility of a doc being in one index but not in another and this could lead to unexpected results for a user.

garrensmith:
This got lost:
I've made an attempt at fixing this. I'm not 100% I fully understood this. In the situation of [a, b, c] and [a, d, e] I've defaulted to the indices names. The only problem with that is there is a possibility of a doc being in one index but not in another and this could lead to unexpected results for a user.

tonysun83:
@garrensmith : @davisp was just saying that if you have the following indexes:

Idx1 - [a, b, c, d]
Idx2 - [a, b, c, e]
Idx3 - [a, b, c]

and you issued  a query with selectors using a,b,c, then we would choose Idx3. 

If Idx3 didn't exist, then you need some sort of tiebreaker. The old way used a combination of
[dbname, ddocid, viewname]. He was just saying you could still use this crude method as a tiebreaker, and use the ddocid (not all three elements).

You are right that this won't solve the issue when Idx1 is always chosen, and then a new Idx4 [a, b, c, f] which is alphabetically ahead of Idx1is added, then the user could possibly get different results. 

Paul did mention:

"In the future we could get fancy and cache selectors and a reduce query with the prefix range and prefer the view that has more documents."

so that we could possibly choose the index with more documents and be a little more consistent. However, it's still probably some indexes will contain different docs.  I'll think about it some more, but o you think that's more of index design on their part?

garrensmith:
@tonysun83 great. That makes sense. I think I have it to the point now where it will cover the above situations. Then in terms of choosing indexes with different docs, I think for now the best we can do is add documentation to explain that part.

src/mango/src/mango_cursor_view.erl
```diff
@@ -107,10 +107,14 @@ execute(#cursor{db = Db, index = Idx} = Cursor0, UserFun, UserAcc) ->
...
 composite_indexes(Indexes, FieldRanges) ->
-    lists:foldl(fun(Idx, Acc) ->
+    FieldKeys = [Key || {Key, _} <- FieldRanges],
+    SortedIndexes = lists:foldl(fun(Idx, Acc) ->
         Cols = mango_idx:columns(Idx),
         Prefix = composite_prefix(Cols, FieldRanges),
-        [{Idx, Prefix} | Acc]
+        % create a ColPrefixLength based on how close the number of fields
+        % the index has to the number of Prefixes the index can use
+        ColPrefixLength = abs(length(Cols) - length(FieldKeys)),
```
tonysun83: 1) This should be ``` length(Prefix)``` instead of ```length(FieldKeys)```?

2) Not sure abs would work here because of the following scenario:
IdxA has columns [a,b,c]
IdxB has columns [a,b,c,d]
Now are selector queries on a,b,c,f,g
ColPrefixLength for IdxA becomes 2, ColPrefLength for IdxB becomes 1, then that would lead to IdxB being preferred, whereas IdxA should be used.

src/mango/src/mango_cursor_view.erl
```diff
@@ -135,14 +139,13 @@ composite_prefix([Col | Rest], Ranges) ->
...
+        case ColPrefixLengthA - ColPrefixLengthB of
             N when N < 0 -> true;
             N when N == 0 ->
-                % This is a really bad sort and will end
-                % up preferring indices based on the
-                % (dbname, ddocid, view_name) triple
-                A1 =< B1;
+                % We have no other way to choose, so at this point 
+                % select index based on (dbname, ddocid, view_name) triple
+                IdxA =< IdxB;
```
tonysun83: So with the old code (which is also what you're leaving with here), I ran a simple test, and here's what happens:

IdxA: [a,b,c]
IdxB: [a,d,e]

I insert two docs, {"a":1, "b":1, "c":1} and {"a":1000, "d" : 1000, "e": 1000}

then I run the query with _explain:

```
{
  "selector": {
    "a": {
      "$gt": 0
    },
    "b": {
      "$gt": 0
    },
    "c": {
      "$gt": 0
    }
  }
}
```
and I get back IdxB
```
"index": {
        "ddoc": "_design/ccf468609d8ca7ef1744797c6b685d82cb552335",
        "name": "IdxB",
        "type": "json",
        "def": {
            "fields": [
                {
                    "a": "asc"
                },
                {
                    "d": "asc"
                },
                {
                    "e": "asc"
                }
            ]
        }
    }
```
I think we've been wrong all this time, and it should be IdxA right?
Can you try out this scenario in the tests? We might have to change this comparison.

tonysun83:
Good worjk @garrensmith, +1 for me.  @davisp  way want to double check

src/mango/src/mango_cursor_view.erl
```diff
@@ -124,25 +128,31 @@ composite_prefix([Col | Rest], Ranges) ->
...
             N when N < 0 -> true;
             N when N == 0 ->
-                % This is a really bad sort and will end
-                % up preferring indices based on the
-                % (dbname, ddocid, view_name) triple
-                A1 =< B1;
+                ColsLenA = length(mango_idx:columns(IdxA)),
+                ColsLenB = length(mango_idx:columns(IdxB)),
+                case ColsLenA - ColsLenB of
+                    M when M < 0 -> true;
```
davisp: true needs to be on a new line

src/mango/src/mango_cursor_view.erl
```diff
@@ -124,25 +128,31 @@ composite_prefix([Col | Rest], Ranges) ->
...
+                    M when M == 0 ->
+                        % We have no other way to choose, so at this point 
+                        % select index based on (dbname, ddocid, view_name) triple
+                        IdxA =< IdxB;
+                    _ -> false
```
davisp: false needs to be on a new line

src/mango/src/mango_cursor_view.erl
```diff
@@ -270,4 +280,6 @@ is_design_doc(RowProps) ->
     case couch_util:get_value(id, RowProps) of
         <<"_design/", _/binary>> -> true;
         _ -> false
-    end.
\ No newline at end of file
+    end.
+
+
```
davisp: Delete these last two empty lines

src/mango/src/mango_cursor_view.erl
```diff
@@ -124,25 +128,31 @@ composite_prefix([Col | Rest], Ranges) ->
             []
     end.
 
-
```
davisp: Add this empty line back

src/mango/src/mango_cursor_view.erl
```diff
@@ -107,10 +107,14 @@ execute(#cursor{db = Db, index = Idx} = Cursor0, UserFun, UserAcc) ->
 % check FieldRanges for a, b, c, and d and return
 % the longest prefix of columns found.
 composite_indexes(Indexes, FieldRanges) ->
+    FieldKeys = [Key || {Key, _} <- FieldRanges],
     lists:foldl(fun(Idx, Acc) ->
         Cols = mango_idx:columns(Idx),
         Prefix = composite_prefix(Cols, FieldRanges),
-        [{Idx, Prefix} | Acc]
+        % create a KeysPrefixLength based on how close the prefix of
```
davisp: Capitalize "Create" and remove "a".

src/mango/src/mango_cursor_view.erl
```diff
@@ -107,10 +107,14 @@ execute(#cursor{db = Db, index = Idx} = Cursor0, UserFun, UserAcc) ->
...
+        % the index is to the number of keys in the selector
+        KeysPrefixLength = length(FieldKeys) - length(Prefix),
```
davisp: Creating FieldKeys is unecessary. Just use length(FieldRanges) here.

davisp: I'm not following the logic on subtracting the prefix from the total length. At least, the name seems to be misleading. Shouldn't it be something like "ColumnsAfterPrefix"?

src/mango/src/mango_cursor_view.erl
```diff
@@ -124,25 +128,31 @@ composite_prefix([Col | Rest], Ranges) ->
...
+% Return the first element.
 %
 % In the future we can look into doing a cached parallel
 % reduce view read on each index with the ranges to find
 % the one that has the fewest number of rows or something.
 choose_best_index(_DbName, IndexRanges) ->
-    Cmp = fun({A1, A2}, {B1, B2}) ->
-        case length(A2) - length(B2) of
+    Cmp = fun({IdxA, PrefixA, KeysPrefixLengthA}, {IdxB, PrefixB, KeysPrefixLengthB}) ->
+        case KeysPrefixLengthA - KeysPrefixLengthB of
```
davisp: I have to stare at the above code but I'm pretty sure this logic boils down to "prefer the index with the fewest extra columns" which while legit is hard to follow based on the variable names.

src/mango/src/mango_cursor_view.erl
davisp: "NumExtraColumns" seems like it'd be better. Also the comment should be updated to specify that we're looking to use the index with the fewest extra columns. And we should add a comment below in choose_best_index.

src/mango/src/mango_cursor_view.erl
```diff
@@ -124,25 +128,31 @@ composite_prefix([Col | Rest], Ranges) ->
             []
     end.
 
-
-% Low and behold our query planner. Or something.
-% So stupid, but we can fix this up later. First
-% pass: Sort the IndexRanges by (num_columns, idx_name)
-% and return the first element. Yes. Its going to
-% be that dumb for now.
+% Low and behold our query planner
```
davisp: Why'd you remove ". Or something." Seems an odd change to make. Removing the snark is fine but I'd just remove the whole line. Also I think it should actually be "Lo and behold" but my Google is too far away to check for certain.

src/mango/src/mango_cursor_view.erl
```diff
@@ -124,25 +128,31 @@ composite_prefix([Col | Rest], Ranges) ->
...
+% First pass: sort the IndexRanges by KeysPrefixLength,
```
davisp: "First pass:" is kind of odd since we're not labeling anything else. I would rephrase this sentence to avoid the colon there.

davisp: Or alternatively, create a numbered list of the steps of the algorithm.

src/mango/src/mango_cursor_view.erl
garrensmith: What is the general formatting rule, when should it be on a new line?
Does the true on line 144 also need to be on a new line?

davisp: Yes, missed that one. The general rule is to not mix inline and multiline. If they any of the clauses can't be inlined then make all clauses start on a new line.

PR #473, opened 2017-04-06
Reorganize snap files into ./snap/ directory

mhall119:
This cleans up the snap configurations, putting them all into just one ./snap/ directory


wohali:
@mhall119 Based on discussion with @janl we'd like to move this snap/ directory to the new couchdb-pkg repository along with all other packaging work (which includes my ongoing .deb, .rpm, BSD and Windows work).

I know this will make things slightly more difficult for you and possibly Launchpad but I'm hoping we can make a recipe work. Can we discuss?

PR #475, opened 2017-04-08
COUCHDB-3367: Require server admin user for db/_compact and db/_view_cleanup endpoints

fkaempfer:
## Overview

This PR adds a one-line admin security check to the db/_compact and db/_view_cleanup endpoints, which are currently missing. This implements only server level admin checks according to this blog: http://blog.mattwoodward.com/2012/03/definitive-guide-to-couchdb.html

## Testing recommendations
Currently all members of a database can trigger db compaction by sending a POST request to /db/_compact. This should be reserved to admins. Test by sending a POST request as a db member before and after applying this patch.

## JIRA issue number

COUCHDB-3367

### DIFF
--- a/src/chttpd/src/chttpd_auth_request.erl
+++ b/src/chttpd/src/chttpd_auth_request.erl
@@ -61,7 +61,7 @@ authorize_request_int(#httpd{path_parts=[_DbName], method='PUT'}=Req) ->
     require_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName], method='DELETE'}=Req) ->
     require_admin(Req);
-authorize_request_int(#httpd{path_parts=[_DbName, <<"_compact">>]}=Req) ->
+authorize_request_int(#httpd{path_parts=[_DbName, <<"_compact">>|_]}=Req) ->
     require_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName, <<"_view_cleanup">>]}=Req) ->
     require_admin(Req);
--- a/src/chttpd/src/chttpd_auth_request.erl
+++ b/src/chttpd/src/chttpd_auth_request.erl
@@ -61,6 +61,10 @@ authorize_request_int(#httpd{path_parts=[_DbName], method='PUT'}=Req) ->
     require_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName], method='DELETE'}=Req) ->
     require_admin(Req);
+authorize_request_int(#httpd{path_parts=[_DbName, <<"_compact">>]}=Req) ->
+    require_admin(Req);
+authorize_request_int(#httpd{path_parts=[_DbName, <<"_view_cleanup">>]}=Req) ->
+    require_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName|_]}=Req) ->
     db_authorization_check(Req).
 
--- a/src/chttpd/src/chttpd_auth_request.erl
+++ b/src/chttpd/src/chttpd_auth_request.erl
@@ -62,9 +62,9 @@ authorize_request_int(#httpd{path_parts=[_DbName], method='PUT'}=Req) ->
 authorize_request_int(#httpd{path_parts=[_DbName], method='DELETE'}=Req) ->
     require_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName, <<"_compact">>|_]}=Req) ->
-    require_admin(Req);
+    require_db_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName, <<"_view_cleanup">>]}=Req) ->
-    require_admin(Req);
+    require_db_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName|_]}=Req) ->
     db_authorization_check(Req).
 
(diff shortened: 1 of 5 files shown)


wohali:
Hi @fkaempfer thank you for your contribution to CouchDB! Do you think you could add a test to confirm this? Either EUnit or JavaScript would be fine.

fkaempfer:
OK, this is my first erlang code ever written, so perhaps it could still be improved ;).

The test shows (fails on master) that it is possible for unauthorized users to trigger compaction when no security is in place. 

When security is in place for a db, members can also trigger compaction, but should not be able to (only admins should). I did not set up a test for this special case, because it is fairly complicated to create a user in the _users db and set up _security in the test code. This is also fixed by the PR.

I only wonder if it should be possible for db admins to trigger _compaction/_view_cleanup or just for server admins. This PR allows it only for server admins (and not DB admins), which would be much more complex to implement.

Thanks for reviewing this PR.

wohali:
@fkaempfer Do you have a CouchDB 1.6 lying around? Can you test what it does? Your PR should, for now, match 1.6's behaviour when it comes to db admins vs. server admins. If you don't have time I can take a look for you.

fkaempfer:
I verified that the behaviour of 1.6.1 is indeed that db admins are also allowed to perform compaction:
- anonymous: disallowed
- member: disallowed
- db admin: allowed
- server admin: allowed

I guess the same is true for view compaction, which I have not tested.

This would mean the pull request needs to be changed to get the actual permission response from the corresponding shard via RPC as far as I understand. This is a bit more tricky to implement and requires changing a couple of method signatures.

If you think it's worth persuing I can try doing it. Or perhaps you prefer that somebody from the CouchDB team does it.

wohali:
Thanks for doing the legwork! Indeed I do think it's worth pursuing, and it'd be great if you could give it a go. We'll be sure to get someone to review your work.

fkaempfer:
I have updated the PR. 

For the three cases db compaction, view compaction and view cleanup, the calls in fabric now receive a response from the nodes and return an error in case of failure, instead of just sending the request and returning OK. 

I have also added the necessary security checks in couch_db.erl. These security checks cause tests to fail (for example couchdb_view_tests:compact_db), because they do not supply an admin context.

Let me know what you think of the fixes. I'm new to erlang and couchdb source, so it's not unlikely there are some things I have missed.

wohali:
Hi @fkaempfer.  A couple of things:

1. You need to fix the EUnit test cases you just broke: https://travis-ci.org/apache/couchdb/builds/229280412 You'll also want to check the JavaScript test cases, you can run these with `make javascript`.
2. We probably want a double-check from someone like @nickva to be sure we're not breaking things downstream from us (like Cloudant's alternative authentication layer.)

fkaempfer:
Hi @wohali and thanks again.
1. I have fixed the 3 failing calls to compact by opening the db as admin.  The Javascript test cases also pass fine.
2. Yes absolutely. In any case it's good to have a third set of eyes on the code. Let me know if you want me to change something.

wohali:
Ping @nickva @eiri can one of you +1 this PR before we merge it? Thanks!

rnewson:
This isn't quite the right approach but I really appreciate the effort and the tests. the chttpd_auth_request.erl module controls this stuff. It's my fault, I think, that I added clustered _compact without restricting it to admins. Here's how I think it should be done;

```
diff --git a/src/chttpd/src/chttpd_auth_request.erl b/src/chttpd/src/chttpd_auth_request.erl
index 90176c824..ab160ee79 100644
--- a/src/chttpd/src/chttpd_auth_request.erl
+++ b/src/chttpd/src/chttpd_auth_request.erl
@@ -61,6 +61,10 @@ authorize_request_int(#httpd{path_parts=[_DbName], method='PUT'}=Req) ->
     require_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName], method='DELETE'}=Req) ->
     require_admin(Req);
+authorize_request_int(#httpd{path_parts=[_DbName, <<"_compact">>]}=Req) ->
+    require_admin(Req);
+authorize_request_int(#httpd{path_parts=[_DbName, <<"_view_cleanup">>]}=Req) ->
+    require_admin(Req);
 authorize_request_int(#httpd{path_parts=[_DbName|_]}=Req) ->
     db_authorization_check(Req).
```

wohali:
@fkaempfer we'd really like to fix this for 2.1, which is imminent. Can you respin your changes to match what @rnewson shows above? Keep the tests as they are, they are valuable.

fkaempfer:
Thanks for the comments! I have updated the PR according to @rnewson's request.

If I understand correctly, this change now limits  _compact to server admins, and disallows it for db admins (contrary to 1.6.1, see the lengthy discussion above). To do this right, something like db_authorization_check_admin could be implemented in chttpd_auth_request.erl. But I don't know if this feature would even be useful - I certainly don't need it and I'm happy with this smaller change going into 2.1.

rnewson:
Ah true enough. My change might not be sufficient but it does show you the right path/module.

wohali:
@fkaempfer If you're willing to go the extra mile I'd be happy to see the full fix land. We have a couple of weeks before 2.1 freezes anyway, so you have time.

fkaempfer:
@wohali okay I'll give it a shot in the next days.

fkaempfer:
The latest commit adds a require_db_admin method to chttpd_auth_request alongside a few helper methods.

The two commits before add tests to check if the db compaction is allowed for db admins in addition to which was already tested (only server admin/anonymous).

I'm in a hurry, so let's hope the last test checks out, I'll have an eye on it again tomorrow.

wohali:
@rnewson everything is passing, changes were made per your request, can you approve this PR? Thanks!

fkaempfer:
@rnewson @wohali all seems good from my side. Let me know if you still want me to make changes.

wohali:
Robert is busy so I'm going to +1 this on his behalf. Thank you so much @fkaempfer !

wohali:
I should have squashed that, my mistake.

fkaempfer:
Thanks!

PR #477, opened 2017-04-14
Add LRU for view groups

sagelywizard:
## Overview

This PR changes how CouchDB manages view indexes. Specifically, it 1) adds an LRU cache to couch_index_server for view groups 2) causes idle view groups to close if more than a configurable soft maximum number of indexes are open and 3) adds additional bookkeeping for couch_index_server to track if a view index is eligible for closing. The PR also has a few other changes in CouchDB for bits of code which assume view groups are never closed.

## Testing recommendations

This is primarily a performance improvement and shouldn't affect the API. This PR includes eunit tests, which can be run with `make check`.

## JIRA issue number

COUCHDB-3377

## Related Pull Requests

None

## Checklist

### DIFF
--- a/src/couch/src/couch_lru.erl
+++ b/src/couch/src/couch_lru.erl
@@ -11,34 +11,36 @@
 % the License.
 
 -module(couch_lru).
--export([new/0, insert/2, update/2, close/1]).
+-export([new/1, insert/2, update/2, close/1]).
 
 -include_lib("couch/include/couch_db.hrl").
 
-new() ->
+new(CloseFun) ->
     Updates = ets:new(couch_lru_updates, [ordered_set]),
-    Dbs = ets:new(couch_lru_dbs, [set]),
-    {0, Updates, Dbs}.
+    Counts = ets:new(couch_lru_counts, [set]),
+    #couch_lru{updates=Updates, counts=Counts, close_fun=CloseFun}.
 
-insert(DbName, {Count, Updates, Dbs}) ->
-    update(DbName, {Count, Updates, Dbs}).
+insert(Name, Lru) ->
+    update(Name, Lru).
 
-update(DbName, {Count, Updates, Dbs}) ->
-    case ets:lookup(Dbs, DbName) of
+update(Name, Lru) ->
+    #couch_lru{counts=Counts, updates=Updates, count=Count} = Lru,
+    case ets:lookup(Counts, Name) of
         [] ->
-            true = ets:insert(Dbs, {DbName, Count});
-        [{DbName, OldCount}] ->
-            true = ets:update_element(Dbs, DbName, {2, Count}),
-            true = ets:delete(Updates, {OldCount, DbName})
+            true = ets:insert(Counts, {Name, Count});
+        [{Name, OldCount}] ->
+            true = ets:update_element(Counts, Name, {2, Count}),
+            true = ets:delete(Updates, {OldCount, Name})
     end,
-    true = ets:insert(Updates, {{Count, DbName}}),
-    {Count + 1, Updates, Dbs}.
+    true = ets:insert(Updates, {{Count, Name}}),
+    Lru#couch_lru{count=Count+1}.
 
 
-close({Count, Updates, Dbs}) ->
-    case close_int(ets:next(Updates, {-1, <<>>}), Updates, Dbs) of
+close(Lru) ->
+    #couch_lru{updates=Updates} = Lru,
+    case close_int(ets:next(Updates, {-1, <<>>}), Lru) of
         true ->
-            {true, {Count, Updates, Dbs}};
+            {true, Lru};
         false ->
             false
     end.
@@ -46,26 +48,19 @@ close({Count, Updates, Dbs}) ->
 
 %% internals
 
-close_int('$end_of_table', _Updates, _Dbs) ->
+close_int('$end_of_table', _Lru) ->
     false;
-close_int({_Count, DbName} = Key, Updates, Dbs) ->
-    case ets:update_element(couch_dbs, DbName, {#db.fd_monitor, locked}) of
-    true ->
-        [#db{main_pid = Pid} = Db] = ets:lookup(couch_dbs, DbName),
-        case couch_db:is_idle(Db) of true ->
-            true = ets:delete(couch_dbs, DbName),
-            true = ets:delete(couch_dbs_pid_to_name, Pid),
-            exit(Pid, kill),
+close_int({_Count, Name} = Key, Lru) ->
+    #couch_lru{updates=Updates, counts=Counts, close_fun=CloseFun} = Lru,
+    {Stop, Remove} = CloseFun(Name),
+    case Remove of
+        true ->
             true = ets:delete(Updates, Key),
-            true = ets:delete(Dbs, DbName),
-            true;
+            true = ets:delete(Counts, Name);
         false ->
-            true = ets:update_element(couch_dbs, DbName, {#db.fd_monitor, nil}),
-            couch_stats:increment_counter([couchdb, couch_server, lru_skip]),
-            close_int(ets:next(Updates, Key), Updates, Dbs)
-        end;
-    false ->
-        true = ets:delete(Updates, Key),
-        true = ets:delete(Dbs, DbName),
-        close_int(ets:next(Updates, Key), Updates, Dbs)
+            ok
+    end,
+    case Stop of
+        false -> close_int(ets:next(Updates, Key), Lru);
+        true -> true
     end.
--- a/src/couch_index/src/couch_index_compactor.erl
+++ b/src/couch_index/src/couch_index_compactor.erl
@@ -95,6 +95,7 @@ handle_info({'EXIT', Pid, Reason}, #st{pid = Pid} = State) ->
     IdxName = Mod:get(idx_name, IdxState),
     Args = [DbName, IdxName, Reason],
     couch_log:error("Compaction failed for db: ~s idx: ~s reason: ~p", Args),
+    ok = couch_index_server:set_compacting(State#st.idx, false),
     {noreply, State#st{pid = undefined}};
 handle_info({'EXIT', _Pid, normal}, State) ->
     {noreply, State};
--- a/src/couch_index/src/couch_index_compactor.erl
+++ b/src/couch_index/src/couch_index_compactor.erl
@@ -98,8 +98,6 @@ handle_info({'EXIT', Pid, Reason}, #st{pid = Pid} = State) ->
     {noreply, State#st{pid = undefined}};
 handle_info({'EXIT', _Pid, normal}, State) ->
     {noreply, State};
-handle_info({'EXIT', Pid, _Reason}, #st{idx=Pid}=State) ->
-    {stop, normal, State};
 handle_info(_Mesg, State) ->
     {stop, unknown_info, State}.
 
--- a/src/couch_index/src/couch_index_compactor.erl
+++ b/src/couch_index/src/couch_index_compactor.erl
@@ -64,12 +64,14 @@ handle_call({compact, _}, _From, #st{pid=Pid}=State) when is_pid(Pid) ->
     {reply, {ok, Pid}, State};
 handle_call({compact, IdxState}, _From, #st{idx=Idx}=State) ->
     Pid = spawn_link(fun() -> compact(Idx, State#st.mod, IdxState) end),
+    ok = couch_index_server:set_compacting(Idx, true),
     {reply, {ok, Pid}, State#st{pid=Pid}};
 handle_call(cancel, _From, #st{pid=undefined}=State) ->
     {reply, ok, State};
 handle_call(cancel, _From, #st{pid=Pid}=State) ->
     unlink(Pid),
     exit(Pid, kill),
+    ok = couch_index_server:set_compacting(State#st.idx, false),
     {reply, ok, State#st{pid=undefined}};
 handle_call(get_compacting_pid, _From, #st{pid=Pid}=State) ->
     {reply, {ok, Pid}, State};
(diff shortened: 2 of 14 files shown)


src/couch/src/couch_lru.erl
```diff
@@ -11,61 +11,56 @@
...
-            close_int(ets:next(Updates, Key), Updates, Dbs)
-        end;
-    false ->
-        true = ets:delete(Updates, Key),
-        true = ets:delete(Dbs, DbName),
-        close_int(ets:next(Updates, Key), Updates, Dbs)
+            ok
+    end,
+    case Stop of
+        false -> close_int(ets:next(Updates, Key), Lru);
```
tonysun83: Do we need to do a
```
 ets:delete(Updates, Key),
ets:delete(Dbs, DbName),
```
here?

sagelywizard: That case is handled above on line 58.

sagelywizard: The return value of `CloseFun` indicates if it should close the DB (or view) and/or stop iterating over the LRU. This change just explicitly pulled those two considerations apart.

tonysun83: wups, I missed it on 58, yup i see it now

src/couch_index/src/couch_index_compactor.erl
```diff
@@ -80,6 +82,7 @@ handle_cast(_Mesg, State) ->
 
 
 handle_info({'EXIT', Pid, normal}, #st{pid=Pid}=State) ->
+    ok = couch_index_server:set_compacting(State#st.idx, false),
```
tonysun83: do we also need to add this to the 
```
handle_info({'EXIT', Pid, _Reason}, #st{idx=Pid}=State) ->
    {stop, normal, State};
```
clause?

src/couch_index/src/couch_index_server.erl
```diff
@@ -34,8 +35,76 @@
...
+    gen_server:call(?MODULE, {compacting, Idx, IsCompacting}, infinity).
+
+set_committing(Pid, IsCommitting) ->
+    gen_server:call(?MODULE, {committing, Pid, IsCommitting}, infinity).
+
+maybe_close_index({DbName, DDocId, Sig}) ->
+    case ets:update_element(?BY_SIG, {DbName, Sig}, {#entry.locked, true}) of
+        true ->
+            case ets:lookup(?BY_SIG, {DbName, Sig}) of
+                [#entry{pid=Pid, committing=false, compacting=false}] ->
```
tonysun83: For my own understanding, I notice that we're checking comitting or compacting for views whereas for dbs we're only checking whether the file descriptor is locked. Is it correct to assume that a locked fd for dbs encompasses commits and compaction?

sagelywizard: We do the same thing for DBs, it's just not explicit and the flags are set in the `#db{}` record. https://github.com/apache/couchdb/blob/master/src/couch/src/couch_db.erl#L131

src/couch_index/test/couch_index_lru_tests.erl
```diff
@@ -0,0 +1,226 @@
...
+        ok = meck:expect(couch_index_server, set_compacting, fun(Idx, IsCompacting) ->
+            meck:passthrough([Idx, IsCompacting]),
+            Self ! {compact, IsCompacting, self()},
+            receive finish ->
+                ok
+            end,
+            ok
+        end),
+
+        ok = meck:expect(couch_index_server, set_committing, fun(Idx, IsCommitting) ->
```
tonysun83: seems like only a naming thing, but doesn't set_committing take in a PID as an argument?

src/couch_index/src/couch_index_compactor.erl
sagelywizard: Yep, good catch!

src/couch_index/test/couch_index_lru_tests.erl
sagelywizard: Nope, it's a boolean flag.

src/couch_index/test/couch_index_lru_tests.erl
```diff
@@ -0,0 +1,226 @@
...
+                receive close ->
+                    ok
+                end,
+                couch_index_server:close(Mon)
+            end)
+        end, DDocIds),
+
+        % check that all indexes are open
+        ToClose = wait_all_compacting(true, [], ?MAX_INDICES_OPEN+10),
+        ?assertEqual(?MAX_INDICES_OPEN+10, gen_server:call(couch_index_server, get_open_count)),
```
tonysun83: a little confused here, since ?MAX_INDICES_OPEN is 10, and we create 20 indexes, shouldn't get_open_count return 10?

tonysun83:
hmmm got  failure for the test soft

```
  Test the view index LRU
    couch_index_lru_tests:61: test_close_while_compacting...[0.265 s] ok
    couch_index_lru_tests:130: test_soft_max...*failed*
in function gen_server:call/3 (gen_server.erl, line 190)
in call from couch_index_server:get_index/2 (src/couch_index_server.erl, line 182)
in call from lists:map/2 (lists.erl, line 1237)
in call from couch_index_lru_tests:'-test_soft_max/1-fun-11-'/1 (test/couch_index_lru_tests.erl, line 156)
**exit:{badarg,
    {gen_server,call,
        [couch_index_server,
         {get_index,
             {test_index,
                 {test_idx,<<"eunit-test-db-1492651861"...>>,
                     {<<"_design/i1">>},
                     <<"CzF??"...>>},
                 <<"eunit-test-db-14926518618861"...>>,
                 <<67,122,70,243,220,127,...>>}},
         infinity]}}
```
Will take a look a little later tonight

src/couch_index/test/couch_index_lru_tests.erl
sagelywizard: It's a soft maximum, so since all 20 are being held open concurrently, all 20 should be open.

sagelywizard:
Yeah, I saw that. I think it's a problem with meck. Let me know if you figure it out.

src/couch_index/test/couch_index_lru_tests.erl
```diff
@@ -0,0 +1,226 @@
...
+                Seq
+        end),
+
+        ok = meck:reset(test_index),
+
+        IdxOpens = lists:map(fun(I) ->
+            BI = integer_to_binary(I),
+            % hack: use tuple as index name so couch_index_server won't try to open
+            % it as a design document.
+            IndexName = {<<"_design/i", BI/binary>>},
```
tonysun83: seems like the tuple here is causing badargs down the line. Wonder if ddoc_to_mrst is where it crashes, but need to debug more in detail

davisp:
Mulling this over it looks more or less like the obvious approach to adding the same LRU that couch_server has. For the most part it reads fine. Though I'm mulling over the new monitor that's returned that has to be closed. I keep thinking there must be something a bit cleaner here with exposing a new API function to ask the module if its index is idle.

sagelywizard:
@davisp What do you think about doing something similar to the `#db{}` record for views (and holding the ref in its state, like `#db{}`? It'd involve changing a lot of LOC, but it seems cleaner on the surface and would be consistent.

iilyak:
I noticed that we often do something like the following:
```
  {ok, Pid, Mon, Args} = get_index(...),
  <something>
  couch_index_server:close(Mon)
```
Did you consider implementing following helper functions
```
with_index_process(Db, DDoc, ViewName, InArgs, Fun) ->
    case get_view_index_pid(Db, DDoc, ViewName, Args0) of
        {ok, Pid, Mon, Args} ->
            try
                Fun(Pid, Args)
            after
                ok = couch_index_server:close(Mon)
            end;
        {error, _Reason} = Error ->
            Error
    end.

with_index(Module, DbName, DDocId, Fun) ->
    case get_index(Module, DbName, DDocId) of
        {ok, Pid, Mon} ->
            try
                Fun(Pid)
            after
                ok = couch_index_server:close(Mon)
            end;
        {error, _Reason} = Error ->
            Error
    end.
```

sagelywizard:
Could add that. Not necessary though. We'd still need to do the manual `get_index` and `close` in some places.

src/couch_index/src/couch_index_server.erl
```diff
@@ -187,18 +285,20 @@ code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
 
-handle_config_change("couchdb", "index_dir", RootDir, _, RootDir) ->
-    {ok, RootDir};
-handle_config_change("couchdb", "view_index_dir", RootDir, _, RootDir) ->
-    {ok, RootDir};
+handle_config_change("couchdb", "index_dir", RootDir, _, #st{root_dir=RootDir}=State) ->
```
iilyak: It is not quite right. The state here is the value we pass to [`config:listen_for_changes/2`](https://github.com/apache/couchdb/pull/477/files#diff-3f723e40663b52d70b47cbd302216517R193)

src/couch_index/src/couch_index_server.erl
```diff
@@ -187,18 +285,20 @@ code_change(_OldVsn, State, _Extra) ->
...
 handle_config_change("couchdb", "index_dir", _, _, _) ->
     exit(whereis(couch_index_server), config_change),
     remove_handler;
 handle_config_change("couchdb", "view_index_dir", _, _, _) ->
     exit(whereis(couch_index_server), config_change),
     remove_handler;
-handle_config_change(_, _, _, _, RootDir) ->
-    {ok, RootDir}.
+handle_config_change("couchdb", "max_indices_open", Max, _, State) ->
+    {ok, State#st{max_open=list_to_integer(Max)}};
```
iilyak: Two questions here:
  1. Do we want to support a possibility to have unlimited number of opened indexes?
  2. How would we want it to fail if max_indices_open doesn't parse to integer (typo has been made).

iilyak:
> @davisp
 I keep thinking there must be something a bit cleaner here with exposing a new API function to ask the module if its index is idle.

Ideally we would want to minimize the interaction with couch_index_server since it is a bottlneck. 
If we can ensure that it is impossible to have at any given time multiple index processes for the same index we could move locking out of couch_index_server. In such case the maybe_close_index function would be simpler (maybe).

I also experimented with some other possibilities a little bit to make it more generic.  The goal was to make couch_server and couch_index_server similar. Most likely it is overkill. I tend to over engineer at times.
  
couch_lock.erl
---------------
```
new(Table, Key, Pos) ->
    {Table, Key, Pos}.
with_lock({Table, Key, Pos}, Default, Fun) ->
    case ets:update_element(Table, Key, {Pos, true}) of
         true ->
              try 
                   [Entry] = ets:lookup(Table, Key),
                  Fun(Entry)
              after
                  ets:update_element(Table, Key, Pos, false)
              end;
         false ->
             Default
   end.
```

couch_index.erl
----------------
```
txn({DbName, DDocId, Sig}, Default, Fun) ->
    Lock = couch_lock:lock(?BY_SIG, {DbName, Sig}, #entry.locked),
    couch_lock:with_lock(Lock, Default, Fun). %% we don't have it yet

is_idle(Pid) ->
         gen_server:call(Pid, is_idle).

set_committing(Pid, Value) ->
         gen_server:call(Pid, {set_committing, Value}).

set_compacting(Pid, Value) ->
         gen_server:call(Pid, {set_compacting, Value}).
```

couch_index_server.erl
-----------------------
```
maybe_close_index({DbName, DDocId, Sig} = Idx) ->
   couch_index:txn(Idx, {false, true}, fun(#entry{pid =Pid}) ->
        case couch_index:is_idle(Pid) of
            true -> 
                 rem_from_ets(DbName, Sig, DDocId, Pid),
                 couch_index:stop(Pid),
                 {true, true};
            false ->
                couch_stats:increment_counter([couchdb, couch_index_server, lru_skip]),
                {false, false}
        end
   end).
```

couch_server.erl
----------------
```
txn(DbName, Default, Fun) ->
    Lock = couch_lock:lock(couch_dbs, DbName, #db.fd_monitor),
    couch_lock:with_lock(Lock, Default, Fun). %% we don't have it yet

maybe_close_db(DbName) ->
   couch_server:txn(Idx, {false, true}, fun(#db{main_pid =Pid} = Db) ->
        case couch_db:is_idle(Db) of
            true -> 
                true = ets:delete(couch_dbs, DbName),
                true = ets:delete(couch_dbs_pid_to_name, Pid),
                exit(Pid, kill),
                {true, true};
            false ->
                couch_stats:increment_counter([couchdb, couch_server, lru_skip]),
                {false, false}
        end
   end).
```

davisp:
Re-reading this I guess I must have written down my concerns somewhere else and never got them on to this PR. My biggest concern with this change is a lack of profiling of what sort of bottleneck the new LRU might be. Given that its modeled off couch_server, and on a number of deployments we have more views than databases it seems like this is a prime candidate for another run away queue process.

If we do want to pursue this I'd really want to see some serious benchmarking that shows its not just creating a second couch_server issue for us.

Also, the original impetuous for this was to work on closing out views since its possible that large production environments can have some high memory usage requirements for all views to be open at once. An alternative approach that doesn't use an LRU would be to close idle views like we've added for databases and os processes. The recent PR for couch_db_updaters is here:

https://github.com/apache/couchdb/commit/ba0c10bcf49be66b042d8f0deaf14bcfda1a49bf

wohali:
bumpety bump bump...is this useful still or should we close it out? /cc @sagelywizard

davisp:
I don't want to merge this till we get an idea what the performance of the LRU is given the couch_server similarity which already bites us quite hard.

sagelywizard:
What sort of performance stats do you think would show that these changes aren't problematic from a performance perspective, Paul? Comparing open/close throughput before/after? Comparing the index open latency on new/old CouchDB when more than the max number of indexes are open? It's worth noting that this change works completely differently than the existing couch_server.

davisp:
@sagelywizard I'd like to at least have some tests that show how many open indexes we can have before this becomes a bottleneck. And then looking at the rates of open/closing we can sustain when we're close to that threshold.

I'm confused when you say its completely different than couch_server. Its going so far as to reuse couch_lru and has similar idle/not-idle states that it uses to decide when to close views along with the public ets table cache pattern. Am I missing something else?

sagelywizard:
Oh, yeah, you're right. I wrote a few different versions, a couple of which were completely different, but this one is pretty similar.

janl:
Closing until this moves forward.

PR #483, opened 2017-04-19
New couchup 1.x -> 2.x database migration tool

wohali:
This commit adds a new Python-based database migration tool, `couchup`.
It is intended to be used at the command-line on the server being
upgraded, before bringing the node (or cluster) into service.

`couchup` provides 4 sub-commands to assist in the migration process:

* `list` - lists all CouchDB 1.x databases
* `replicate` - replicates one or more 1.x databases to CouchDB 2.x
* `rebuild` - rebuilds one or more CouchDB 2.x views
* `delete` - deletes one or more CouchDB 1.x databases

A typical workflow for a single-node upgrade process would look like:

```sh
$ couchup list
$ couchup replicate -a
$ couchup rebuild -a
$ couchup delete -a
```

A clustered upgrade process would be the same, but must be preceded by
setting up all the nodes in the cluster first.

Various optional arguments provide for admin login/password, overriding
ports, quiet mode and so on.

Of special note is that `couchup rebuild` supports an optional flag,
`-f`, to filter deleted documents during the replication process.

I struggled some with the naming convention. For those in the know, a
'1.x database' is a node-local database appearing only on port 5986, and
a '2.x database' is a clustered database appearing on port 5984, and in
raw, sharded form on port 5986.

## Testing recommendations

1. Copy a bunch of CouchDB 1.x `.couch` files into the CouchDB 2.x `data/` directory.
2. Start CouchDB 2.x.
3. Use the 4 `couchup` commands to list, replicate, rebuild and delete the databases as desired.
4. Test your application against the migrated data.
5. Report any bugs on this PR please 👍

### DIFF
--- a/rel/overlay/bin/couchup
+++ b/rel/overlay/bin/couchup
@@ -1,0 +1,480 @@
+#!/usr/bin/env python
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+import argparse
+import base64
+import json
+import textwrap
+import threading
+import time
+import sys
+try:
+    from urllib import quote
+except ImportError:
+    from urllib.parse import quote
+import requests
+try:
+    import progressbar
+    HAVE_BAR = True
+except ImportError:
+    HAVE_BAR = False
+
+def _tojson(req):
+    """Support requests v0.x as well as 1.x+"""
+    if requests.__version__[0] == '0':
+        return json.loads(req.content)
+    return req.json()
+
+def _args(args):
+    args = vars(args)
+    if args['password']:
+        args['creds'] = (args['login'], args['password'])
+    else:
+        args['creds'] = None
+    return args
+
+def _do_list(args):
+    port = str(args['local_port'])
+    req = requests.get('http://127.0.0.1:' + port + '/_all_dbs',
+        auth=args['creds'])
+    req.raise_for_status()
+    dbs = _tojson(req)
+    local_dbs = [x for x in dbs if "shards" not in x
+        and x not in ['_dbs', '_nodes']]
+    clustered_dbs = list(set(
+        [x.split('/')[2].split('.')[0] for x in dbs if "shards" in x]
+    ))
+    if not args['include_system_dbs']:
+        # list comprehension to eliminate dbs starting with underscore
+        local_dbs = [x for x in local_dbs if x[0] != '_']
+        clustered_dbs = [x for x in clustered_dbs if x[0] != '_']
+    local_dbs.sort()
+    clustered_dbs.sort()
+    if args.get('clustered'):
+        return clustered_dbs
+    return local_dbs
+
+def _list(args):
+    args = _args(args)
+    ret = _do_list(args)
+    print(", ".join(ret))
+
+def _watch_replication(db,
+        local_port=5986,
+        clustered_port=5984,
+        creds=None,
+        hide_progress_bar=False,
+        quiet=False,
+        timeout=30):
+    """Watches replication, optionally with a progressbar."""
+    time.sleep(1)
+    if not quiet:
+        print("Replication started.")
+    url = "http://127.0.0.1:{}/{}".format(local_port, db)
+    try:
+        req = requests.get(url, auth=creds)
+        req.raise_for_status()
+        req = _tojson(req)
+        # here, local means node-local, i.e. source (1.x) database
+        local_docs = req['doc_count']
+        local_size = req['data_size']
+    except requests.exceptions.HTTPError:
+        raise Exception('Cannot retrieve {} doc_count!'.format(db))
+    if local_size == 0:
+        return
+    if HAVE_BAR and not hide_progress_bar and not quiet:
+        widgets = [
+            db,
+            ' ', progressbar.Percentage(),
+            ' ', progressbar.Bar(marker=progressbar.RotatingMarker()),
+            ' ', progressbar.ETA(),
+            ' ', progressbar.FileTransferSpeed(),
+        ]
+        progbar = progressbar.ProgressBar(widgets=widgets,
+                maxval=local_size).start()
+    count = 0
+    stall_count = 0
+    url = "http://127.0.0.1:{}/{}".format(clustered_port, db)
+    while count < local_docs:
+        try:
+            req = requests.get(url, auth=creds)
+            req.raise_for_status()
+            req = _tojson(req)
+            # here, cluster means clustered port, i.e. port 5984
+            clus_count = req['doc_count']
+            clus_size = req['data_size']
+        except requests.exceptions.HTTPError as exc:
+            if exc.response.status_code == 404:
+                clus_count = 0
+                clus_size = 0
+            else:
+                raise Exception('Cannot retrieve {} doc_count!'.format(db))
+        if count == clus_count:
+            stall_count += 1
+        else:
+            stall_count = 0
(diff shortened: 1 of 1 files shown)


rel/overlay/bin/couchup
```diff
@@ -0,0 +1,474 @@
...
+            'http://127.0.0.1:{}/{}/_design/repl_filters'.format(
+            args['local_port'], db),
+            auth=args['creds'])
+        req.raise_for_status()
+        doc = _tojson(req)
+        del doc['_rev']
+        if doc != ddoc:
+            if not args['quiet']:
+                print('Source replication filter does not match! Aborting.')
+            exit(1)
```
janl: 🎨 I’m impartial to returning different exit codes depending on the condition, and then use named constants to refer to the different exit codes. E.g. here it could be `exit(E_DDOC_MISMATCH)` that is set to `1` further up and then line [167](https://github.com/apache/couchdb/pull/483/files#diff-5ef2b13ddb2104387d351635732e3b1aR167) gets anther, e.g. `E_DDOC_PUT_ERROR` set to `2). And similar for all other lines with `exit(1)` that I’m not referencing here individually.

Absolutely not a blocker.

rel/overlay/bin/couchup
```diff
@@ -0,0 +1,474 @@
...
+        description=textwrap.dedent('''\
+            Examples:
+              couchup list
+              couchup list -c -i -p mysecretpassword
+         '''))
+    parser_list.add_argument('-c', '--clustered', action='store_true',
+        help='show clustered (2.x) databases instead')
+    parser_list.add_argument('-i', '--include-system-dbs',
+        action='store_true',
+        help='include system databases (_users, _replicator, etc.)')
```
janl: 🤔 Might be worth considering doing separate options for `_users` and `_replicator` (the only system dbs in 1.x), I might wanna do `_users` but not `_replicator`.

Not a blocker, but would raise as open issue after merge. Would make a good first time contributor issue.

rel/overlay/bin/couchup
```diff
@@ -0,0 +1,474 @@
...
+except ImportError:
+    from urllib.parse import quote
+import requests
+try:
+    import progressbar
+    HAVE_BAR = True
+except ImportError:
+    HAVE_BAR = False
+
+def _tojson(req):
```
wohali: Yup, this is how I support python-requests 0.x. Will do.

rel/overlay/bin/couchup
```diff
@@ -0,0 +1,474 @@
...
+    args = vars(args)
+    if args['password']:
+        args['creds'] = (args['login'], args['password'])
+    else:
+        args['creds'] = None
+    return args
+
+def _do_list(args):
+    port = str(args['local_port'])
+    req = requests.get('http://127.0.0.1:' + port + '/_all_dbs',
```
wohali: I presumed this might be one of the first requests, and agree that it'd be a good thing for a first timer to tackle, though it's a trivial refactor.

rel/overlay/bin/couchup
```diff
@@ -0,0 +1,474 @@
...
+def _list(args):
+    args = _args(args)
+    ret = _do_list(args)
+    print(", ".join(ret))
+
+def _watch_replication(db,
+        local_port=5986,
+        clustered_port=5984,
+        creds=None,
+        no_progress_bar=False,
```
wohali: This is because of how argparse works; it's annoying to get it to handle "--flag={true,false}" but very easy to get it to handle "--flag" (and store either True or False depending) I want the default for the progress bar to be on, so the CLI flag needs to turn it off.

rel/overlay/bin/couchup
```diff
@@ -0,0 +1,474 @@
...
+    """Watches replication, optionally with a progressbar."""
+    time.sleep(1)
+    if not quiet:
+        print("Replication started.")
+    url = "http://127.0.0.1:{}/{}".format(local_port, db)
+    try:
+        req = requests.get(url, auth=creds)
+        req.raise_for_status()
+        req = _tojson(req)
+        local_docs = req['doc_count']
```
wohali: Good point, will change. I meant "node-local" docs. Could just make this more generic with source/target rather than local_/clus_.

rel/overlay/bin/couchup
```diff
@@ -0,0 +1,474 @@
...
+            stall_count = 0
+        if stall_count == timeout:
+            print("Replication is stalled. Increase timeout or reduce load.")
+            exit(1)
+        if HAVE_BAR and not no_progress_bar and not quiet:
+            if clus_size > local_size:
+                clus_size = local_size
+            progbar.update(clus_size)
+        count = clus_count
+        time.sleep(1)
```
wohali: Disagree...this is only used for the --timeout=XX option, and needs to stay at 1s granularity otherwise the timeout needs to be forced to be a multiple of the per-loop value.

rel/overlay/bin/couchup
wohali: I did consider this, yeah. Initially I was raising unique Exceptions but switched over to this to make it easier for shell scripting. Going on memory I think the only function that has multiple exit()s is this one, so this might be the only place that needs more exit codes.

rel/overlay/bin/couchup
wohali: 🤔 Not sure I agree, but maybe. Line 352 is just for `couchup list` which probably doesn't need explicit flags for those 2 DBs.

I suspect you mean line 387 or line 427 where action can be taken on them instead. Right now you'd do that with `couchup {replicate|rebuild} _users`, the only time `-i` makes sense is in `couchup -a -i`.  Is it worth adding `--users` and `--replicator` flags when `couchup replicate -a && couchup replicate _users` would suffice? I'm unconvinced, but not super opinionated.

wohali:
Thanks @janl. I'll correct a few of the more straightforward tweaks today.

I'd like to get feedback from at least a couple of people trying the script (other than me!) before merging, but if no one responds on user@ or dev@ by middle of next week, I'll merge anyway.

rel/overlay/bin/couchup
janl: gotcha

rel/overlay/bin/couchup
janl: aye.

rel/overlay/bin/couchup
janl: ah sorry, yes, ignore me for now :)

rel/overlay/bin/couchup
gesellix: What about `hide_progress_bar` instead?

wohali: Sure, ok.

jvabob:
Trying to use the replicate function to migrate my 686GB collection of databases...

First 3-4 databases (about 5GB) work great... but now I get something like this when I try it
first got it using the -a feature and now am trying individual db to see if I can find a pattern

/usr/local/couchdb/bin/couchup replicate -l USER -p PASSWORD -t 3600 -f eridu
Starting replication for eridu...
{"error":"not_found","reason":"Could not open source database `http://127.0.0.1:5986/eridu/`: {db_not_found,<<\"http://127.0.0.1:5986/eridu/\">>}"}

if I curl http://127.0.0.1:5986/eridu/ ( and add the user and pass ) I get

{"db_name":"eridu","doc_count":7438871,"doc_del_count":511625,"update_seq":8168757,"purge_seq":0,"compact_running":false,"disk_size":81804304630,"other":{"data_size":109},"data_size":12273332816,"sizes":{"file":81804304630,"active":12273332816,"external":109},"instance_start_time":"1502734780758722","disk_format_version":6,"committed_update_seq":8168757,"compacted_seq":0,"uuid":"4a84be53b812cac9b71915eab09f3be9"}

So the db sure seems to be there... I can go to 5986/_utils and see the db... can view data etc

Any pointers?

wohali:
Please file this as a [new issue](https://github.com/apache/couchdb/issues/new) so we can properly follow this request, not as a comment on the pull request.

Please be sure to paste the exact results of the output. If possible please wrap your paste in triple-backticks, like this:

```
    ```
    /usr/local/couchdb/bin/couchup replicate...
    ```
```

jvabob:
Sorry for the confusion on my part…

I have submitted the new issue.

Thanks!

Bob Jones
850.251.7485

PR #487, opened 2017-04-24
Document that Fauxton requires npm

ianpartridge:
Fixes this build failure on Ubuntu 16.04:

```
$ make release
[snip...]
==> couchdb (compile)
Building Fauxton
/bin/sh: npm: command not found
Makefile:366: recipe for target 'share/www' failed
make: *** [share/www] Error 127
```

### DIFF
--- a/INSTALL.Unix.md
+++ b/INSTALL.Unix.md
@@ -64,7 +64,7 @@ You can install the dependencies by running:
 
     sudo apt-get --no-install-recommends -y install \
         build-essential pkg-config erlang erlang-reltool \
-        libicu-dev libmozjs185-dev libcurl4-openssl-dev
+        libicu-dev libmozjs185-dev libcurl4-openssl-dev npm
 
 Be sure to update the version numbers to match your system's available
 packages.


ianpartridge:
Passing `--disable-fauxton` to `configure` also works around this problem.

wohali:
Hi @ianpartridge,

Unfortunately, even on 16.04, `apt-get install npm` isn't new enough to bring in our recommended version of NPM, which is 3.x (comes with Node 6.x). See https://github.com/apache/couchdb-fauxton/blob/master/readme.md for this specification.

Our CI workflow explicitly calls out this Node 6.x requirement: https://github.com/apache/couchdb-ci/blob/master/ansible/debian-8-base.yml#L22

If you want to update your PR (and squash, please) so that it points readers at Fauxton's `readme.md` for specific requirements, and/or the use of `--disable-fauxton`, I'd be happy to merge your change.

Thanks for your contribution to Apache CouchDB!

ianpartridge:
Thanks @wohali! Thinking about this a bit more, shouldn't `configure` check for Fauxton pre-requisites and fail if they aren't present?

wohali:
We don't presently check for all of our dependencies in the configure script.

For instance, we'll error out if you don't have Erlang installed, or if you are missing Sphinx, you'll get a similar error down in the documentation build.

So, for now, no.

ianpartridge:
OK thanks I didn't realise that the `configure` script isn't designed to check for prerequisites (like they do in Autotools based projects).

wohali:
We actually moved away from GNU Autotools intentionally for many reasons, not the least of which was that it made the Windows port very difficult. ;)

PR #494, opened 2017-04-27
couchdb2.0 is not sending charset=utf-8 in response headers of json

FgfdCBVCbnsebtgf:
HTTP/1.1 200 OK
Server: CouchDB/1.6.0 (Erlang OTP/19)
ETag: "GG61GZEDIDA2MOVW8O6204F7"
Date: Wed, 26 Apr 2017 07:53:28 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 0

HTTP/1.1 200 OK
Content-Length: 0
Date: Wed, 26 Apr 2017 07:53:34 GMT
Server: CouchDB/2.0.0 (Erlang OTP/19)
X-Couch-Request-ID: undefined
X-CouchDB-Body-Time: 0


kxepal:
Thank you for your contribution and interest to this problem!

Unfortuanelly, I have two things to notice:
1. You need to create a new feature branch for your PR from the branch to which you want it get merged (it's `2.0.x` now), but...
2. `application/json` is always UTF-8 by default and his MIME type doesn't accepts any parameters per [RFC-7159](https://tools.ietf.org/html/rfc7159). See sections [8.1](https://tools.ietf.org/html/rfc7159#section-8.1) and [11](https://tools.ietf.org/html/rfc7159#section-11) for more information. 

I think there is nothing need to do and everything should works fine. Do you have any issues caused by missed charset on MIME type? And why?

FgfdCBVCbnsebtgf:
Hello, the Chromium version 57.0.2987.98 Built on Ubuntu , running on Ubuntu 17.04 (64-bit) does not show correctly the accent characters.

rnewson:
utf-8 is the defined charset when interpreting application/json, this is unnecessary.

https://www.ietf.org/rfc/rfc4627.txt

"3.  Encoding

   JSON text SHALL be encoded in Unicode.  The default encoding is
   UTF-8."

PR #496, opened 2017-04-27
Couchdb 3287 pluggable storage engines

davisp:
## Overview

Pluggable Storage Engines (Oh my!)

I've finally covered enough bases to start asking for reviews on the pluggable storage engine work. I know that this is a fairly large change so I don't expect this to actually be merged for a number of weeks. So everyone that wants to review any part of this feel free to take your time and be thorough.

I highly suggest reviewing this work a commit at a time as there are a few fairly large commits. A few signposts for the intrepid reader:

1. Add couch_db_engine module

This is a single file addition that just introduces the new pluggable storage engine API. Its fairly thoroughly documented and hopefully makes sense as to the level that this API is geared at. When I initially started this work I spent a lot of time trying to figure out what level of the API would be the best place to separate the storage engine from database logic. Go too high and every engine is reimplementing core bits of logic. Go to low and there's not enough room for interesting changes to the actual storage engine algorithms. I believe this is a happy middle ground that gives storage engines the room to play and invent alternative implementations while also not requiring an extremely large amount of reimplementation for various bits of behavior that are required.

2. Add legacy storage engine implementation

This commit is quite big but its important to note that all its doing is copying the existing parts of couch_db.erl and couch_db_updater.erl and moving them to a new set of modules named couch_bt_engine_*.erl. Nothing uses this code yet. I have this in its own commit since its fairly large. However, if you mostly just want to check it you should be able to see that the implementation is just pulled from existing functions. The behavior of this engine is identical to the "pre-PSE" engine because that's what it is. Its just been reformatted a bit and had its name changed.

Also, I've kept this and couch_db_engine.erl in the main couch application since we'll always want to have at least one storage engine. Though this was pre-monorepo when adding another git repo to our builds seemed awfully heavy. Now that we're monorepo we're just creating a folder which is easy enough. For any reviewer I'd like to have them keep this in mind as one thing I expect to discuss is whether we should split this out into its own application.

3. Implement pluggable storage engines

This is the doozy. There are two main bits of work going on in this commit. First, the removal of all the code that was in the previous commit and second, the addition of all the code to have couch_db and couch_db_updater start using the couch_db_engine APIs. Its long and big but really if you take your time there's nothing magical going on here. For the core bits to study I'd recommend spending some extra time on couch_server to see how engines are configured and chosen at runtime.

4. Add storage engine test suite

This is a fairly complete test suite for the entire storage engine API. I don't remember what coverage of couch_bt_engine was but I know its fairly high. The useful bit about this is when devs are creating new storage engines they can just pull this in with a single eunit case to test their implementations. I'll show a few examples on this below.

5. Ensure deterministic revisions for attachments

I may end up squashing this into the implementation commit. This was a fix while I was developing PSE that I ended up refixing after the fact slightly differently. The original goal was to make sure everything compiled and all tests suites ran for each commit. I think I've just convinced myself to squash this. So if its missing its because I did that and then forgot to edit this description to remove this paragraph...

## Testing recommendations

$ make check

## JIRA issue number

https://issues.apache.org/jira/browse/COUCHDB-3287

## Related Pull Requests

This PR is based on the previous mixed cluster upgrade PR. However when merge time comes around we'll merge #495 first and then this will be merged into master so that users can deploy the first changes before these changes. (Uptime is fun time).

https://github.com/apache/couchdb/pull/495

## Checklist

(We should probably remove the todo item for updating rebar.config since we're doing almost all work on the monorepo now)


davisp:
Also for documentation I'm not sure what we want to do there. Should I open a PR for the docs now or should we wait and let this soak on master and just let the more adventurous folk take cracks at running and writing new storage engines to make sure there's nothing major that needs reworked?

davisp:
Also, I forgot to link out to the first complete alternative storage engines.

https://github.com/cloudant/couch_ngen

The couch_ngen engine (its a super terrible name I know) was one that I wrote fairly quickly just to try and prove that the API wasn't artificially restrictive. The biggest thing here is that I'm using multiple files for a single database to try and make things like compaction faster. This is the same general shape as the existing legacy engine with a few tweaks that we've considered over the years but never actually tried to implement due to upgrade concerns. This also uses a NIF for file I/O with dirty schedulers (thus your Erlang VM needs to have dirty schedulers enabled to run this engine). Initial benchmarks show this engine to be significantly faster than the legacy engine at the local node interface. Unfortunately those gains don't show up nearly as well at the clustered interface (so we've got more work to do there :)

https://github.com/cloudant-labs/lethe/tree/initial-branch

Lethe is the first engine written by !me. This work is by @chewbranca trying to investigate the storage engine API by actually using it to write a storage engine. Lethe is an ephemeral/in-memory storage engine. Think ramdisk database thing. Its not been benchmarked to the best of my knowledge but last I heard it was passing the storage engine test suite and could run most of the existing test suite (it doesn't support attachments which are optional in the API).

There are some other obvious candidates for storage engines as well that I've not tried (on purpose to let others play with something new and interesting). LevelDB/RocksDB, LMDB, PostgreSQL, SQLite, Bitcask, etc all come to mind as obvious things to play with. Also something I've been contemplating is writing a custom storage engine in C that is behaviorally similar to the existing code but hopefully significantly faster (though remember if speed is your goal we've got more work to do in the clustering layer :)

wohali:
@davisp IMO write the docs and have that PR land at the same time as this one, or they might get left behind.

src/couch/test/couch_stream_tests.erl
```diff
@@ -14,10 +14,11 @@
 
 -include_lib("couch/include/couch_eunit.hrl").
 
+-define(ENGINE, {couch_bt_engine_stream, {Fd, []}}).
```
iilyak: The variable name is hard-coded in a define. Could we do something about it? For example:

`-define(ENGINE(Var), {couch_bt_engine_stream, {Var, []}}).`

src/couch/src/couch_att.erl
```diff
@@ -233,6 +234,14 @@ transform(Field, Fun, Att) ->
     store(Field, NewValue, Att).
 
 
+copy(Att, DstStream) ->
```
iilyak: I couldn't find any calls to this function.

src/couch/src/couch_util.erl
```diff
@@ -62,6 +63,37 @@ normparts(["." | RestParts], Acc) ->
 normparts([Part | RestParts], Acc) ->
     normparts(RestParts, [Part | Acc]).
 
+
```
iilyak: This looks very similar to https://github.com/erlang/otp/blob/master/lib/stdlib/src/filelib.erl#L103. Is there a reason to write own implementation? filelib's implementation considered slow (based on info from erlang mailing list) due to use of regexp. However this implementation uses it as well.

I'll read through the implementation some more tomorrow (maybe there is a reason).

garrensmith:
Here are the steps to use this branch with a [couch_ngen](https://github.com/cloudant/couch_ngen)
Most of this is obvious for someone experienced with Erlang development. But for someone like me who is not that familiar with erlang, I needed to complete all these steps.

# Erlang dirty scheduler
`couch_ngen` requires erlang with a dirty scheduler. So to install that on a mac:
```
brew install kerl
KERL_CONFIGURE_OPTIONS=--enable-dirty-schedulers kerl build 19.3 19.3-dirty
kerl install 19.3-dirty /Users/username/kerl/19.3-dirty
. /Users/username/kerl/19.3-dirty/activate
```
# Setup dependancies 
In the CouchDB repo add these two lines to the `rebar.config.script` in the `DepDescs` section:

```
%% PSE
{couch_ngen,  {url, "https://github.com/cloudant/couch_ngen"}, "45c73d3d822e7cf42fbe60cd3e316f4ff12c2399"},
{nifile,  {url, "https://github.com/cloudant/nifile"}, "d5f3b909846137f616cd5e754b810246802394e1"},
```

Then run `rebar get-deps`. 

# Edit local.ini

Edit `rel/overlay/etc/local.ini and add these settings in:
```
[couchdb]
default_engine = couch_bt_engine

[couchdb_engines]
couch = couch_bt_engine
ngen = couch_ngen
```
This sets up the list of supported engines and which is the default engine. If you change `default_engine=ngen` then that will be the default storage engine used.

*Important to note*, don't add these settings to the bottom of the file if you are using `dev/run` it will break `dev/run`. Rather add it somewhere in the middle.

# Build CouchDB

`make clean && make` to build a new version

# Run CouchDB

`dev/run --with-haproxy -a tester:testerpass`

# Create database with couch_ngen
If `couch_ngen` isn't the default storage engine. To create a database using that engine:

```
curl -X PUT http://testuser:testpass@127.0.0.1:5984/testdbname?engine=ngen
```

Boom! You now have multiple storage engines running.

davisp:
Awesome, thanks for the notes @garrensmith!

src/couch/src/couch_util.erl
iilyak: Answering my own question:

This implementation doesn't do useless call to `file:read_file_info(File)`.

src/couch/test/couchdb_views_tests.erl
```diff
@@ -543,23 +543,27 @@ has_doc(DocId1, Rows) ->
     lists:any(fun({R}) -> lists:member({<<"id">>, DocId}, R) end, Rows).
 
 backup_db_file(DbName) ->
-    DbDir = config:get("couchdb", "database_dir"),
-    DbFile = filename:join([DbDir, ?b2l(DbName) ++ ".couch"]),
-    {ok, _} = file:copy(DbFile, DbFile ++ ".backup"),
-    ok.
+    {ok, Db} = couch_db:open_int(DbName, []),
```
iilyak: Should we abstract it away into `get_filepath(DbName)`?
In this case it could be used from both `backup_db_file` and `restore_backup_db_files`.

src/couch/src/couch_db.erl
```diff
@@ -414,37 +390,34 @@ get_before_doc_update_fun(#db{before_doc_update = Fun}) ->
 get_committed_update_seq(#db{committed_update_seq=Seq}) ->
     Seq.
 
-get_update_seq(#db{update_seq=Seq})->
-    Seq.
+get_update_seq(#db{} = Db)->
```
iilyak: Minor nitpick. Some of the `get_` functions return `{ok, Result} and some just `Result`.

src/couch_mrview/src/couch_mrview.erl
```diff
@@ -531,17 +541,25 @@ map_fold({{Key, Id}, Val}, _Offset, Acc) ->
         user_acc=UAcc1,
         last_go=Go
     }};
-map_fold({<<"_local/",_/binary>> = DocId, {Rev0, Body}}, _Offset, #mracc{} = Acc) ->
+map_fold(#doc{id = <<"_local/", _/binary>>} = Doc, _Offset, #mracc{} = Acc) ->
```
iilyak: I had to stop here for quite some time. Leaving comment for future reviewers. This change would work because we open local btree differently. We pass split and join callbacks [to couch_btree:open for local_tree]( https://github.com/apache/couchdb/blob/a41c3b12896e027b1b154139f2c164551b4c5b2b/src/couch/src/couch_bt_engine.erl#L695:L696) now.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -236,7 +237,9 @@ init_state(Db, Fd, State, Header) ->
         view_states=ViewStates
     } = Header,
 
-    IdBtOpts = [{compression, couch_db:compression(Db)}],
+    IdBtOpts = [
+        {compression, couch_compress:get_compression_method()}
```
iilyak: This would work since old implementation calls to `couch_compress:get_compression_methods()` from `couch_db_updater:init_db`. However this would mean that we wouldn't be able to support different compression methods for different databases. Also it wouldn't be possible to delegate the decision on a compression algorithm to the storage engine (if we are not dispatching through `couch_engine`).

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -840,6 +843,8 @@ maybe_convert_reductions({KVs0, UserReductions}) ->
 
 maybe_convert_kv({<<"_local/", _/binary>> = DocId, _}) ->
```
iilyak: Shouldn't we convert this clause to `maybe_convert_kv(#doc{id = <<"_local/", _/binary>>} = Doc) ->`?

src/couch/src/couch_server.erl
```diff
@@ -114,6 +117,35 @@ create(DbName, Options0) ->
 delete(DbName, Options) ->
     gen_server:call(couch_server, {delete, DbName, Options}, infinity).
 
+
+exists(DbName) ->
+    RootDir = config:get("couchdb", "database_dir", "."),
+    Engines = get_configured_engines(),
+    Possible = lists:foldl(fun({Extension, Engine}, Acc) ->
```
iilyak: We have exactly same lines in `get_engine(Server, DbName)` function. We can create a common function such as:
- `select_candidate_engines(DbName, RootDir, Engines)`
- `select_possible_engines(DbName, RootDir, Engines)`

src/couch/src/couch_att.erl
davisp: Strange. Looks like I wrote it and then never used it. Will remove it.

src/couch/test/couch_stream_tests.erl
davisp: Yeap, that's better.

src/couch/src/couch_util.erl
davisp: Hah. So, it looks like I'm adding a comment to explain this cause it took me at least 10m of staring at this and the original to remember why I did this. While it does avoid the extra read_file_info, that's ancillary to the main change which is to accept both directories and files when matching the regular expression. This way a pluggable storage engine can be a directory or single file (or group of files not utilizing a directory structure for some odd reason).

src/couch/test/couchdb_views_tests.erl
davisp: I'm not see what you're suggesting to abstract away for this.

src/couch/src/couch_db.erl
davisp: This is a nit for me too but its historical and if I were to change it here the diff would spider out to all places that call these functions so I left them all as they were.

src/couch_mrview/src/couch_mrview_util.erl
davisp: I'm a bit confused by this comment but I think you have it right. To make sure we agree this was the intent:

The pluggable storage engine API avoids dictating a compression algorithm to use so that each storage engine can choose its own approach to compression. For instance a LevelDB engine could just rely on LevelDB applying snappy under the covers.

However, since we don't have a notion of "database compression" we now have to change the scheme for view compression. Given that the existing approach is to just set it to the result of couch_compress:get_compression_method() (via what is returned when a database is opened, and more to the point *not* when it was created) this is effectively not a behavior change.

src/couch_mrview/src/couch_mrview_util.erl
davisp: Its an upgrade clause for the time being. The #doc version will already be caught on line 846.

src/couch/src/couch_server.erl
davisp: Done.

nickva:
Let's rebase it on master when you have a chance.

davisp:
Rebased. Forgot to push earlier. Just have that one function I'm not sure about where it went. I'm guessing the couch_server revert but will have to look harder on Monday. This branch is rebased on master but only loosely. I haven't done a full make check yet so there might be a lingering issue or two.

nickva:
I ran couchdyno's replication tests against PSE and they all passed:

```
/test.sh -x
==================================================================================== test session starts =====================================================================================
platform darwin -- Python 2.7.13, pytest-3.0.3, py-1.4.33, pluggy-0.4.0 -- /Users/nvatama/src/couchdyno/venv/bin/python
cachedir: .cache
rootdir: /Users/nvatama/src/couchdyno, inifile: pytest.ini
collected 47 items

tests/test_rep_00_basics.py::test_basic_continuous PASSED
tests/test_rep_00_basics.py::test_basic_10_continuous PASSED
tests/test_rep_00_basics.py::test_basic_10_continuous_1000_docs PASSED
tests/test_rep_00_basics.py::test_basic_normal PASSED
tests/test_rep_00_basics.py::test_basic_10_normal PASSED
tests/test_rep_00_basics.py::test_basic_attachment PASSED
tests/test_rep_00_basics.py::test_basic_js_filter PASSED
tests/test_rep_01_patterns.py::test_n_to_n_pattern[False-False] PASSED
tests/test_rep_01_patterns.py::test_n_to_n_pattern[False-True] PASSED
tests/test_rep_01_patterns.py::test_n_to_n_pattern[True-False] PASSED
tests/test_rep_01_patterns.py::test_n_to_n_pattern[True-True] PASSED
tests/test_rep_01_patterns.py::test_1_to_n_pattern PASSED
tests/test_rep_01_patterns.py::test_n_to_1_pattern PASSED
tests/test_rep_01_patterns.py::test_n_chain_pattern PASSED
tests/test_rep_01_patterns.py::test_all_pattern_continuous PASSED
tests/test_rep_02_filter.py::test_n_to_n_view_filter[False-None] PASSED
tests/test_rep_02_filter.py::test_n_to_n_view_filter[False-1] PASSED
tests/test_rep_02_filter.py::test_n_to_n_view_filter[True-None] PASSED
tests/test_rep_02_filter.py::test_n_to_n_view_filter[True-1] PASSED
tests/test_rep_02_filter.py::test_n_to_n_js_filter[False-None] PASSED
tests/test_rep_02_filter.py::test_n_to_n_js_filter[False-1] PASSED
tests/test_rep_02_filter.py::test_n_to_n_js_filter[True-None] PASSED
tests/test_rep_02_filter.py::test_n_to_n_js_filter[True-1] PASSED
tests/test_rep_02_filter.py::test_n_to_n_doc_ids_filter[False-None] PASSED
tests/test_rep_02_filter.py::test_n_to_n_doc_ids_filter[False-1] PASSED
tests/test_rep_02_filter.py::test_n_to_n_doc_ids_filter[True-None] PASSED
tests/test_rep_02_filter.py::test_n_to_n_doc_ids_filter[True-1] PASSED
tests/test_rep_02_filter.py::test_n_to_n_mango_filter[False-None] PASSED
tests/test_rep_02_filter.py::test_n_to_n_mango_filter[False-1] PASSED
tests/test_rep_02_filter.py::test_n_to_n_mango_filter[True-None] PASSED
tests/test_rep_02_filter.py::test_n_to_n_mango_filter[True-1] PASSED
tests/test_rep_03_attachments.py::test_attachments[False-64-1] PASSED
tests/test_rep_03_attachments.py::test_attachments[False-64-100] PASSED
tests/test_rep_03_attachments.py::test_attachments[False-100000-1] PASSED
tests/test_rep_03_attachments.py::test_attachments[False-100000-100] PASSED
tests/test_rep_03_attachments.py::test_attachments[False-attachments4-1] PASSED
tests/test_rep_03_attachments.py::test_attachments[False-attachments5-100] PASSED
tests/test_rep_03_attachments.py::test_attachments[True-64-1] PASSED
tests/test_rep_03_attachments.py::test_attachments[True-64-100] PASSED
tests/test_rep_03_attachments.py::test_attachments[True-100000-1] PASSED
tests/test_rep_03_attachments.py::test_attachments[True-100000-100] PASSED
tests/test_rep_03_attachments.py::test_attachments[True-attachments10-1] PASSED
tests/test_rep_03_attachments.py::test_attachments[True-attachments11-100] PASSED
tests/test_rep_04_cluster.py::test_migration_on_node_failure SKIPPED
tests/test_rep_05_migration.py::test_migration_error PASSED
tests/test_rep_05_migration.py::test_migration_triggered PASSED
tests/test_rep_05_migration.py::test_migration_downgrade_failed SKIPPED

======================================================================================= 45 passed, 2 skipped in 13390.16 seconds ========================================================================================
```

tonysun83:
@davisp: I followed @garrensmith 's  setup and created a db using the ngen engine. It's probably documented somewhere, but how do I verify that the db created is using another engine or are we hiding that abstraction?

garrensmith:
@tonysun83 each storage engine should have a different file extension. So if you do `http://localhost:15984/testdb?engine=ngen` and then look at the files you will see a few db files with a `.ngen` extension. 

From what I understand of this code https://github.com/apache/couchdb/blob/6375cd41afed07818ff77713fcab7f7e72e5040c/src/couch/src/couch_server.erl#L629-L642
If you do `?engine=blahgen` and that engine is not supported it will fallback to the configured default engine.

garrensmith:
With the `/` endpoint we have a `features` field that will show if the CouchDB instance supports the new replication api. I think it might be worthwhile adding another field that shows which storage engines are supported.

davisp:
@tonysun83 The shard's db info blob should include the engine, however this wasn't included in the clustered db info because I had originally had plans on being able to upgrade the storage engine for a database. However that's not been written yet but if it were then a clustered db could have multiple engines involved for different shards and we don't currently surface shard specific information.

@garrensmith That's correct. It currently just defaults to the configured default engine. And if nothing is configured the default is the legacy couch engine.

Good call on the features list. That was added after the majority of the PSE dev but that'd be a good thing to support and as I recall it was fairly simple. I'll take a look at adding that today.

tonysun83:
@davisp: given the negative scenario above with ```http://localhost:15984/testdbblah?engine=blahgen```, a 200 was returned saying the db was created successfully, but it seems it didn't default to use the couch_ngen engine. I get a 500 saying internal_server_error : No DB shards could be opened. I checked and there were no .couch files for that particular db. Upon retrying to create the database, I get a 412 saying the db is already created.

tonysun83:
@davisp @nickva : I'm not sure if there has been work to rebase the latest scheduler replicator with PSE.  but I'm seeing some function clauses in my weird scenario. 
1) Create a _replicator database using ngen engine (it worked, but not sure we support this?)
2) Create a normal db and add some documents.
3) Run a replication with create_target:true that creates a new ngen db
```
{
  "_id": "ngen_test",
  "source": "http://adm:pass@localhost:15984/mydb",
  "target": "http://adm:pass@localhost:15984/repngen?engine=ngen",
  "create_target": true
}
```
So after I did this, the odd behavior was that my original mydb had 7 documents, but the repngen keeps increasing in number of docs. Currently, it's at 20. and the original mydb has not changed at all. Must be something to do with the update_seq. Log is provided below. 

https://gist.github.com/tonysun83/b600e855ea0d3ae0e6ef92e388cf05c0

davisp:
@tonysun83 Hrm, I'll add a test for that case and try and track down what's causing the 500's with the bad engine. For that and for the replicator function clause its probably a bad rebase or refactor somewhere.

For the replication you're trying I'd expect that to not work I think. Adding URL strings to replication URLs seems like it'd not work so hot. Can you check what documents have been added? My guess is that the URL string is screwing things up and the replicator ends up posting random data that gets a UUID since the target database would interpret everything after the ? as a query parameter. Once its created any post would probably be interpreted as a request to create a new document. And that's not PSE specific. A similar test on not-PSE would be to try and specify a Q or N parameter in the target URL. I'd expect similar results there.

tonysun83:
@davisp:  yup, it was the URL posting random data and not engine related.  I tried with a n parameter and it produced the same errors.

src/couch/src/couch_server.erl
```diff
@@ -559,6 +591,110 @@ db_closed(Server, Options) ->
...
+    end, [], Engines).
+
+
+get_default_engine(Server, DbName) ->
+    #server{
+        root_dir = RootDir,
+        engines = Engines
+    } = Server,
+    Default = {couch_bt_engine, make_filepath(RootDir, DbName, "couch")},
+    case config:get("couchdb", "default_engine") of
```
tonysun83: hmmm, something might be weird with my config, but I set the local.ini to use ```couch_ngen``` as the default_engine, but creating a db still defaulted to a couch_bt_engine db. I checked my env:
```
(node1@127.0.0.1)1> rpc:multicall(config, get, ["couchdb", "default_engine"]).
{["couch_ngen","couch_ngen","couch_ngen"],[]}
```
and it seems like it's using couch_ngen.

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+    Engine:delete(RootDir, DbPath, DelOpts).
+
+
+delete_compaction_files(Engine, RootDir, DbPath, DelOpts)
+        when is_list(DelOpts) ->
+    Engine:delete_compaction_files(RootDir, DbPath, DelOpts).
+
+
+init(Engine, DbPath, Options) ->
+    case Engine:init(DbPath, Options) of
```
tonysun83: hmmm, when I set the default engine to be ```couch_ngen```, it complains about a security object and Engine becomes undefined:

```
hmmm, looks like also I can't create a couch_ngen db when the default is couch_ngen. It's complaining about a security object that's undefined

```
[error] 2017-05-19T09:29:20.961350Z node2@127.0.0.1 <0.8696.0> -------- CRASH REPORT Process  (<0.8696.0>) with 1 neighbors crashed with reason: call to undefined function undefined:init("/Users/tonysun/github/apache/couchdb/dev/lib/node2/data/shards/e0000000-ffffffff/testdb2.149518...", [{default_security_object,[]},create,{user_ctx,{user_ctx,null,[<<"_admin">>],undefined}},{engine,...}]) at couch_db_engine:init/3(line:637) <= couch_db_updater:init/1(line:30) <= proc_lib:init_p_do_apply/3(line:247); initial_call: {couch_db_updater,init,['Argument__1']}, ancestors: [<0.8695.0>], messages: [], links: [<0.8695.0>], dictionary: [{io_priority,{db_update,<<"shards/e0000000-ffffffff/testdb2.1495186160">>}}], trap_exit: false, status: running, heap_size: 987, stack_size: 27, reductions: 181
```

File: src/couch/src/couch_server.erl (hunk 10 above)
Comment (davisp) [inline src/couch/src/couch_server.erl]: Whats the contents of couch_engines?

File: src/couch/src/couch_db_engine.erl (hunk 11 above)
Comment (davisp) [inline src/couch/src/couch_db_engine.erl]: I'm not seeing anything about a security object. But that looks like a bug thats not handled properly wigh a bad config. Somehow its getting an undefined atom back from couch_server:get_engine

File: src/couch/src/couch_server.erl (hunk 10 above)
Comment (eiri) [inline src/couch/src/couch_server.erl]: @davisp Tony right about "default_engine" been ignored, but the problem not here, but in [chttpd_db.erl](https://github.com/apache/couchdb/blob/7162a24ba6b6341a9c015a5fc6b6f4d4306ba432/src/chttpd/src/chttpd_db.erl#L287l) - if query "engine" is missing it defaults to "couch", so "default_engine" is essentially always ignored.

**Comment by davisp** [comment]
Aha! Good catch.

File: src/couch/src/couch_server.erl (hunk 10 above)
Comment (eiri) [inline src/couch/src/couch_server.erl]: @tonysun83 So I've made the same mistake you did and it took me some time to figure it out. An engine name in "default_engine" is not a name of a module, but one of "couchdb_engines" section key aka db file's extension. In your case you are setting default engine to "couch_ngen", but it should be "ngen".

To be fair @garrensmith did the same mistake in his otherwise very helping setup instruction and set default engine to "couch_bt_engine", which worked only because it _is_ default in case of misconfiguration.

File: src/couch/src/couch_server.erl [hunk 12]
```diff
@@ -559,6 +591,110 @@ db_closed(Server, Options) ->
...
+
+get_engine(Server, DbName, Options) ->
+    #server{
+        root_dir = RootDir,
+        engines = Engines
+    } = Server,
+    case couch_util:get_value(engine, Options) of
+        Ext when is_binary(Ext) ->
+            ExtStr = binary_to_list(Ext),
+            case couch_util:get_value(ExtStr, Engines) of
```
eiri: Here `couch_util:get_value/2` returns `undefined` which _is_ an atom, so we are ending up with engine set as `{undefined, "/path/dbname.cowabunga"}` (or whatever came in `Options`), responding "ok" to http api and crashing db updater a moment later.

src/couch/src/couch_server.erl
garrensmith: Thanks @eiri good spot. I've updated my setup instructions.

tonysun83: @eiri @garrensmith : even when the default_engine is set to ngen, I'm still getting db's created as .couch file extension. It could be that I messed my evnironment up too much by testing various configurations.

tonysun83: For reference, once I got past the default_engine issue, I decided to set `default_engine` to `bt_engine` on one node, and the rest as `ngen`. This configuration also did not allow the db to be created successfully.

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+% database is guaranteed to only happen in a single process).
+% However, multiple process may be trying to open different
+% databases concurrently so if a database requires a shared
+% resource that will require concurrency control at the storage
+% engine layer.
+%
+% The returned DbHandle should be a term that can be freely
+% copied between processes and accessed concurrently. However
+% its guaranteed that the handle will only ever be mutated
+% in a single threaded context (ie, within the couch_db_updater
```
nickva: Just to clarify, the handle itself is not mutated, just the database. Handle is just an opaque context and once returned from `init` it is not mutated outside the engine module implementation?

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+    {stop, Reason::any(), NewDbHandle::db_handle()}.
+
+
+% These functions are called by any process opening or closing
+% a database. As such they need to be able to handle being
+% called concurrently. For example, the legacy engine uses these
+% to add monitors to the main engine process.
+-callback incref(DbHandle::db_handle()) -> {ok, NewDbHandle::db_handle()}.
+-callback decref(DbHandle::db_handle()) -> ok.
+-callback monitored_by(DbHande::db_handle()) -> [pid()].
```
nickva: `monitored_by` feels a bit awkward in relation to incref and decref.  It exposes the implementation that ref counting is done using monitors and pids. Any chance to turn this into `refcount() -> integer()`. Not a big deal though. If it requires a significant rewrite keep it as is.

Also it might be worth explaining how `couch_stats_process_tracker` monitoring is involved here. The pid list we get is expected to have that one extra monitor. Should we hide that from the user of the engine and always subtract process tracker before returning here. As with previous point, if it involved a significant change, let's skip it.

src/couch/src/couch_bt_engine.erl
```diff
@@ -0,0 +1,956 @@
...
+        active = S1#size_info.active + S2#size_info.active,
+        external = S1#size_info.external + S2#size_info.external
+    };
+reduce_sizes(S1, S2) ->
+    US1 = couch_db_updater:upgrade_sizes(S1),
+    US2 = couch_db_updater:upgrade_sizes(S2),
+    reduce_sizes(US1, US2).
+
+
+active_size(#st{} = St, Size) when is_integer(Size) ->
```
eiri: We don't actually need this clause because the only place `active_size/2` called is on [line 243](https://github.com/apache/couchdb/blob/7162a24ba6b6341a9c015a5fc6b6f4d4306ba432/src/couch/src/couch_bt_engine.erl#L243) and we are converting size to #size_info right before of that.

src/couch/src/couch_bt_engine.erl
```diff
@@ -0,0 +1,956 @@
...
+    lists:map(fun
+        ({ok, FDI}) -> FDI;
+        (not_found) -> not_found
+    end, Results).
+
+
+open_local_docs(#st{} = St, DocIds) ->
+    Results = couch_btree:lookup(St#st.local_tree, DocIds),
+    lists:map(fun
+        ({ok, #doc{} = Doc}) -> Doc;
```
eiri: We are guarding on return to be `{ok, #doc}` here, but passing `FDI` as it is right above in `open_docs/2`. Is there something subtle about such distinction? I thought `open_docs` _has_ to return `{ok, #full_doc_info}`, so should we guard there too?

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+% persistent UUID for identification in case the shard should
+% ever need to be moved between nodes in a cluster.
+-callback get_uuid(DbHandle::db_handle()) -> UUID::binary().
+
+
+% These functions are only called by couch_db_updater and
+% as such are guaranteed to be single threaded calls. The
+% database should simply store these values somewhere so
+% they can be returned by the corresponding get_* calls.
+
```
nickva: Remove space between comment and function (for consistency with the rest of the module).

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+    UserAcc::any(),
+    changes_fold_options()) ->
+        {ok, LastUserAcc::any()}.
+
+
+% This function may be called by many processes concurrently.
+%
+% This function is called to count the number of documents changed
+% since they given UpdateSeq (ie, not including the possible change
+% at exactly UpdateSeq). It is currently only used internally to
```
nickva: `they given` -> `the given`

src/couch/src/couch_db_engine.erl
davisp: Well it also refers to any calls that would require mutating the handle like writing docs or setting the revs limit. And yes, nothing outside the engine will mutate the handle as well.

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+    UserFold::changes_fold_fun(),
+    UserAcc::any(),
+    changes_fold_options()) ->
+        {ok, LastUserAcc::any()}.
+
+
+% This function may be called by many processes concurrently.
+%
+% This function is called to count the number of documents changed
+% since they given UpdateSeq (ie, not including the possible change
```
nickva: the given

src/couch/src/couch_db_engine.erl
davisp: Yeah, This is one of the weird bits in how we track idleness and so on. Your refcount idea seems valid. That might be a good way to go but I'll have to go back and see what all uses monitored_by. If memory serves I think its mostly just test code that looks at the contents of the monitors.

src/couch/src/couch_bt_engine.erl
davisp: That's possible. Pretty sure I just pulled this from the old implementation without changing it.

src/couch/src/couch_bt_engine.erl
davisp: Its a local tree which returns doc records. The id and seq trees return full doc infos.

eiri: @davisp yes, I understand that, what I meant: we are guarding on `Doc` to be `#doc` record here, but we are not doing the same above and just passing `FDI` as it is. It is not a big of a deal, just seems inconsistent and implies that there are more into it.

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+% is called from inside the couch_server process which
+% means that the storage engine does not have to guarantee
+% its own consistency checks when executing in this
+% context. Although since this is executed in the context
+% of couch_server it should return relatively quickly.
+-callback delete(
+            RootDir::filepath(),
+            DbPath::filepath(),
+            DelOpts::delete_options()) ->
+        ok | {error, Reason::atom()} | any().
```
nickva: When will it be `any()`. Dialyzer might say something here about "any() subsumes `ok` and error term since it's more general.

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+% and as such has the same properties for init/2. It's guaranteed
+% to be consistent for a given database but may be called by many
+% databases concurrently.
+-callback terminate(Reason::any(), DbHandle::db_handle()) -> Ignored::any().
+
+
+% This is called in the context of couch_db_updater:handle_call/3
+% for any message that is unknown. It can be used to handle messages
+% from asynchronous processes like the engine's compactor if it has one.
+-callback handle_call(Msg::any(), DbHandle::db_handle()) ->
```
nickva: I wonder if it is better to use a different name/arity than gen_server's callback. In case someone for whatever reason would want to implement a gen_server in the same module. `handle_unknown_call/2` or `updater_handle_call/2` ?

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+        {reply, Resp::any(), NewDbHandle::db_handle()} |
+        {stop, Reason::any(), Resp::any(), NewDbHandle::db_handle()}.
+
+
+% This is called in the context of couch_db_updater:handle_info/2
+% and has the same properties as handle_call/3.
+-callback handle_info(Msg::any(), DbHandle::db_handle()) ->
```
nickva: Wonder if it is better to use a different name/arity here in case someone, for whatever reason wanted to implement a gen_server in the engine implementation module. `handle_call/2` is fine here since it's arity is 2 not 3.

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+    {noreply, NewDbHandle::db_handle(), Timeout::timeout()} |
+    {stop, Reason::any(), NewDbHandle::db_handle()}.
+
+
+% These functions are called by any process opening or closing
+% a database. As such they need to be able to handle being
+% called concurrently. For example, the legacy engine uses these
+% to add monitors to the main engine process.
+-callback incref(DbHandle::db_handle()) -> {ok, NewDbHandle::db_handle()}.
+-callback decref(DbHandle::db_handle()) -> ok.
```
nickva: `decref` should decref also return an {ok, `NewDbHandle} just to be consistent?

 Could imagine for example some implementation where decref produces an updated handle just like incref.

src/couch/src/couch_db_updater.erl
```diff
@@ -14,74 +14,42 @@
...
+            false ->
+                couch_stats_process_tracker:track([couchdb, open_databases]);
+            true ->
+                ok
+        end,
+        % Don't load validation funs here because the fabric query is
+        % liable to race conditions. Instead see
+        % couch_db:validate_doc_update, which loads them lazily.
+        NewDb = Db#db{main_pid = self()},
+        proc_lib:init_ack({ok, NewDb}),
```
nickva: Just curious, why use init_ack pattern +  enter_loop and instead regular gen_server init pattern?

src/couch/src/couch_bt_engine.erl
davisp: Ahh! Gotchya. I probably added #doc when i changed the split/join functions on the local tree to assert that I'd done that correctly. Will remove it.

src/couch/src/couch_db_engine.erl
davisp: I was probably being defensive when I first made up the type specs. I'll remove any() from the list.

src/couch/src/couch_db_engine.erl
davisp: I'm kind of meh on this. If someone wants to use a gen_server for their engine they could always just put it into a different module. The functions in the main engine module are all intended to be called directly from couch_db_updater in lots of places and adding a gen_server message pass overhead seems at first glance to be something to discourage.

Granted its an API so making as few assumptions is probably for the better. Perhaps we rename these to something like handle_db_updater_call and handle_db_updater_info?

src/couch/src/couch_db_engine.erl
davisp: I originally had something along those lines but something like 99% of the places are calling decref via couch_db:close so it doesn't really make sense to return anything because anything that was returned would be thrown away. I pushed that into this function so that engine authors would have to make sure their decref didn't require cleanup down the line.

src/couch/src/couch_db_updater.erl
davisp: This is the proper pattern to use when a gen_server may not successfully complete its init/1 function. You can see on line 46 below here where we return errors instead of starting the gen_server.

The anti-pattern is what couch_file does when it takes in a parent pid, returns ignore, and then sends an OOB message back to the parent with the error message.

src/couch/src/couch_db_engine.erl
nickva: It's up to you. It wouldn't be about message passing overhead, as much as about someone deciding to make one large module which includes engine callbacks and a gen_server, ... which is probably a silly idea and they shouldn't do anyway.

src/couch/src/couch_db_updater.erl
nickva: Ah makes sense. That `ignore` thing was hackish

src/couch/src/couch_db.erl
```diff
@@ -397,10 +376,7 @@ get_full_doc_info(Db, Id) ->
     Result.
 
 get_full_doc_infos(Db, Ids) ->
```
eiri: Let's remove or update [this comment]
(https://github.com/apache/couchdb/blob/7162a24ba6b6341a9c015a5fc6b6f4d4306ba432/src/couch/src/couch_db.erl#L373), 5 lines above since now we return either `#full_doc_info` or `not_found`

src/couch/src/couch_db.erl
```diff
@@ -1568,9 +1456,8 @@ open_doc_revs_int(Db, IdRevs, Options) ->
         IdRevs, LookupResults).
 
 open_doc_int(Db, <<?LOCAL_DOC_PREFIX, _/binary>> = Id, Options) ->
-    case couch_btree:lookup(Db#db.local_tree, [Id]) of
-    [{ok, {_, {Rev, BodyData}}}] ->
-        Doc = #doc{id=Id, revs={0, [?l2b(integer_to_list(Rev))]}, body=BodyData},
+    case couch_db_engine:open_local_docs(Db, [Id]) of
+    [#doc{} = Doc] ->
```
eiri: Just a suggestion: if `[not_found]` moved on a first match here, then there will be no need to guard on `Doc` to be `#doc` record.

src/couch/src/couch_db.erl
```diff
@@ -1589,7 +1476,7 @@ open_doc_int(Db, #full_doc_info{id=Id,rev_tree=RevTree}=FullDocInfo, Options) ->
         {ok, Doc#doc{meta=doc_meta_info(DocInfo, RevTree, Options)}}, Options);
 open_doc_int(Db, Id, Options) ->
     case get_full_doc_info(Db, Id) of
-    {ok, FullDocInfo} ->
+    #full_doc_info{} = FullDocInfo ->
```
eiri: Same as above: if `not_found` will be a first match, we can trust `get_full_doc_info/2` to do the right thing and drop the guard, seems less defensive. Though I don't feel too strong about it, either way is fine with me.

src/couch/src/couch_db_engine.erl
```diff
@@ -0,0 +1,877 @@
...
+    ].
+
+-type purge_info() :: [{docid(), revs()}].
+-type epochs() :: [{Node::atom(), UpdateSeq::non_neg_integer()}].
+-type size_info() :: [{Name::atom(), Size::non_neg_integer()}].
+
+-type write_stream_options() :: [
+        {buffer_size, Size::pos_integer()} |
+        {encoding, atom()} |
+        {compression_level, pos_integer()}
```
eiri: This should be `non_neg_integer()`, as in specs `pos_integer()` defined as `1..` and we use 0 to disable compression.

src/couch/src/couch_db.erl
```diff
@@ -1138,8 +1068,8 @@ update_docs(Db, Docs0, Options, interactive_edit) ->
         Options2 = if AllOrNothing -> [merge_conflicts];
                 true -> [] end ++ Options,
         DocBuckets3 = [[
-                doc_flush_atts(set_new_att_revpos(
-                        check_dup_atts(Doc)), Db#db.fd)
+                doc_flush_atts(Db, set_new_att_revpos(
```
eiri: There are an old bit in [`set_new_att_revpos/1`](https://github.com/apache/couchdb/blob/7162a24ba6b6341a9c015a5fc6b6f4d4306ba432/src/couch/src/couch_db.erl#L1210) function, when it expects `couch_att:fetch(data, Att)` to return `{Fd, Sp}` instead of new response `{stream, Engine}`. Vars there set with `_` to be ignored, so it works, but it'd be nice to update for sake of the future readers.

src/couch/src/couch_db.erl
davisp: I assume you're thinking in terms of performance? AFAIK matching records is fairly well optimized so I'm not too worried there. Also we're basically doing the same operation as matching {ok, any()} so its not any slower than the old way.

I also don't think of this as defensive so much as assertive. Defensive to me would be if we added an Error clause that was then thrown. Here was just say "these two things may come back and then vomit if we see anything else". And using the record match tells readers what we're getting back there.

src/couch/src/couch_db_updater.erl
```diff
@@ -963,481 +677,37 @@ commit_data(Db, true) ->
...
-    EnumBySeqFun =
-    fun(DocInfo, _Offset,
-            {AccNewDb, AccUncopied, AccUncopiedSize, AccCopiedSize}) ->
-
-        Seq = case DocInfo of
-            #full_doc_info{} -> DocInfo#full_doc_info.update_seq;
-            #doc_info{} -> DocInfo#doc_info.high_seq
-        end,
+finish_engine_swap(_OldDb, _NewEngine, _CompactFilePath) ->
+    erlang:error(explode).
```
eiri: I believe the traditional exception would be `kaboom` 😄 

Seriously though, maybe we go for something less creative, but more descriptive, like `not_implemented`?

src/couch/src/couch_db_engine.erl
davisp: +1 will update

src/couch/src/couch_db.erl
davisp: Will fix

src/couch/src/couch_db_updater.erl
davisp: Actually I'm going to just remove the reference to finish_engine_swap as it was only half of an idea to support. I've since booted that to a future v2 storage engine API to support. However if we get around to supporting database renames first I'm not even sure it'll be a particularly necessary feature.

src/mem3/src/mem3_util.erl
```diff
@@ -153,6 +153,10 @@ build_ordered_shards(DbName, DocProps) ->
 build_shards_by_node(DbName, DocProps) ->
     {ByNode} = couch_util:get_value(<<"by_node">>, DocProps, {[]}),
     Suffix = couch_util:get_value(<<"shard_suffix">>, DocProps, ""),
+    EngineOpt = case couch_util:get_value(<<"engine">>, DocProps) of
+        Engine when is_binary(Engine) -> [{engine, Engine}];
+        _ -> []
+    end,
```
nickva: Minor nit. Notice that bit repeates below. Is it worth making a `engine_from_doc_props` function for it?

davisp: Done.

davisp:
@tonysun83 @eiri I've fixed up the input validation for invalid engines and added a log statement for when the default engine isn't set to an extension. Wanna take the new bits for a spin?

nickva:
If anyone is curious here is the code coverage analysis for unit tests:

To get this had to upgrade to latest meck repo. Otherwise meck and cover don't play well with each other.

Also this includes tests themselves which makes number look a lot better than they are. But the good new is`couch_bt*`,  and `couch_db_engine` are covered pretty well

```
Cover analysis: asf-pse/src/couch/.eunit/index.html

Code Coverage:
chttpd_endpoints_tests          :  98%
couch                           :   0%
couch_app                       :  75%
couch_att                       :  76%
couch_auth_cache                :  73%
couch_base32                    :  96%
couch_bt_engine                 :  89%
couch_bt_engine_compactor       :  88%
couch_bt_engine_header          :  82%
couch_bt_engine_stream          :  88%
couch_btree                     :  88%
couch_changes                   :  65%
couch_compaction_daemon         :  19%
couch_compress                  :  84%
couch_crypto                    :  41%
couch_db                        :  66%
couch_db_doc_tests              : 100%
couch_db_engine                 :  76%
couch_db_epi                    : 100%
couch_db_header                 :  82%
couch_db_mpr_tests              : 100%
couch_db_plugin                 :  91%
couch_db_updater                :  66%
couch_debug                     :   0%
couch_doc                       :  81%
couch_doc_json_tests            :  95%
couch_drv                       :  37%
couch_ejson_compare             :  18%
couch_emsort                    :  89%
couch_etag_tests                : 100%
couch_event_sup                 :   0%
couch_external_manager          :  17%
couch_external_server           :   0%
couch_file                      :  84%
couch_hotp                      :  85%
couch_httpd                     :  65%
couch_httpd_auth                :  40%
couch_httpd_db                  :  36%
couch_httpd_external            :   0%
couch_httpd_handlers            : 100%
couch_httpd_misc_handlers       :  10%
couch_httpd_multipart           :  73%
couch_httpd_oauth               :  41%
couch_httpd_proxy               :  75%
couch_httpd_rewrite             :  52%
couch_httpd_vhost               :  80%
couch_key_tree                  :  88%
couch_lru                       : 100%
couch_multidb_changes           :  93%
couch_native_process            :   0%
couch_os_daemons                :  70%
couch_os_process                :  52%
couch_passwords                 :  78%
couch_primary_sup               : 100%
couch_proc_manager              :  72%
couch_query_servers             :  50%
couch_secondary_sup             : 100%
couch_server                    :  73%
couch_stream                    :  90%
couch_sup                       :  63%
couch_task_status               :  90%
couch_totp                      : 100%
couch_users_db                  :   4%
couch_util                      :  68%
couch_uuids                     :  79%
couch_work_queue                :  77%
couchdb_attachments_tests       :  99%
couchdb_auth_tests              :  96%
couchdb_compaction_daemon_tests :  95%
couchdb_cors_tests              : 100%
couchdb_csp_tests               : 100%
couchdb_design_doc_tests        : 100%
couchdb_file_compression_tests  :  99%
couchdb_http_proxy_tests        :  90%
couchdb_location_header_tests   : 100%
couchdb_mrview_cors_tests       : 100%
couchdb_mrview_tests            :  98%
couchdb_os_daemons_tests        :  98%
couchdb_os_proc_pool            :  95%
couchdb_update_conflicts_tests  :  97%
couchdb_vhosts_tests            :  90%
couchdb_views_tests             :  94%
global_changes_tests            :  98%
json_stream_parse_tests         : 100%
test_engine_attachments         :  93%
test_engine_compaction          :  96%
test_engine_fold_changes        :  93%
test_engine_fold_docs           : 100%
test_engine_get_set_props       : 100%
test_engine_open_close_delete   : 100%
test_engine_purge_docs          : 100%
test_engine_read_write_docs     : 100%
test_engine_ref_counting        :  89%
test_engine_util                :  91%
test_request                    :  67%
test_util                       :  71%
test_web                        :  63%

Total                           : 73%
==> rel (eunit)
==> asf-pse (eunit)
```

nickva:
PR to fix local docs compatibility with master:

https://github.com/apache/couchdb/pull/558

Issue was found by @eiri

src/couch/src/couch_bt_engine.erl
```diff
@@ -0,0 +1,954 @@
...
+    #doc{
+        id = Id,
+        revs = {0, [Rev]},
+        body = BodyData
+    };
+
+local_tree_join(Id, {Rev, BodyData}) when is_integer(Rev) ->
+    #doc{
+        id = Id,
+        revs = {0, [list_to_binary(integer_to_list(Rev))]},
```
nickva: Minor nit: can use `integer_to_binary/1` here

src/couch/src/couch_server.erl
```diff
@@ -559,6 +596,115 @@ db_closed(Server, Options) ->
...
+
+
+make_filepath(RootDir, DbName, Extension) when is_binary(RootDir) ->
+    make_filepath(binary_to_list(RootDir), DbName, Extension);
+make_filepath(RootDir, DbName, Extension) when is_binary(DbName) ->
+    make_filepath(RootDir, binary_to_list(DbName), Extension);
+make_filepath(RootDir, DbName, Extension) when is_binary(Extension) ->
+    make_filepath(RootDir, DbName, binary_to_list(Extension));
+make_filepath(RootDir, DbName, Extension) ->
+    filename:join([RootDir, "./" ++ DbName ++ "." ++ Extension]).
```
eiri: I think this should be just `filename:join([RootDir, DbName ++ "." ++ Extension])`.

If the current directory bit here just for a case when `RootDir` is an empty string, to make it into a relative path, then it's not going to work, because `filename:join/1` will overwrite that:
```erlang
1> filename:join(["", "./" ++ "koi" ++ "." ++ "couch"]).
"/koi.couch"
```

nickva:
Tested mixed cluster upgrade. Started a cluster from `COUCHDB-3288-mixed-cluster-upgrade` with dev/run. Killed node3 and then started it with PSE code.

Noticed that internal replication pushing documents with attachments from node1 to node3 was failing with a function_clause in `couch_att:flush_data`.

```
rpc couch_db:update_docs/4 function_clause [{couch_att,flush_data, [{db,1,<<"shards/...
```

The reason is that flush_data was called with `Db, {Fd, StreamPointer}, Att` arguments instead of the new format of `Db, {stream, StreamEngine}, Att`

Probably need to add a clause to upgrade old stream to new couch_bt_engine's stream format and vice versa.

nickva:
+1

Very nice work. This is a significant step forward. It opens a lot of possibilities for CouchDB.

Things I tested:
 
 * Read the source. Had some API comments. Those were addressed
 * Ran extended replication tests which test a good amount of APIs: crud operations, change feeds, bulk docs, attachments, filters
  * Tested crud operations by hand. Building & querying views.
  * Tested mixed cluster scenarios. Started a dev cluster in with the `COUCHDB-3288-mixed-cluster-upgrade branch` then stopped one node and started with the PSE branch. Then watched logs. Performed more tests. Noticed a few issues there and submitted PRs

Didn't test:
 * Switching engines. Using a non-default engine. Saw other reviewers focused on that so skipped that.

Great job. Looking forward to seeing this merged.

src/couch/src/couch_bt_engine.erl
```diff
@@ -0,0 +1,953 @@
...
+        local_tree = LocalTree,
+        compression = Compression
+    },
+
+    % If this is a new database we've just created a
+    % new UUID and default security object which need
+    % to be written to disk.
+    case Header /= Header0 of
+        true ->
+            {ok, NewSt} = commit_data(St),
```
mayya-sharipova: As Paul D. suggested in slack to really commit data here, we may need `{ok, NewSt} = commit_data(St#st{needs_commit = true})` this statement instead

src/couch/src/test_engine_fold_changes.erl
```diff
@@ -0,0 +1,204 @@
...
+    {ok, NewSt} = test_engine_util:apply_actions(Engine, St, Actions),
+    NewAcc = DocSeqAcc ++ [{DocId, Seq}],
+    Expected = Rest ++ NewAcc,
+    {ok, RevOrder} = Engine:fold_changes(NewSt, 0, fun fold_fun/2, [], []),
+    ?assertEqual(Expected, lists:reverse(RevOrder)),
+    do_mutation_ordering(Engine, NewSt, Seq + 1, Rest, NewAcc).
+
+
+shuffle(List) ->
+    random:seed(os:timestamp()),
```
garrensmith: It seems that the `random` module is deprecated. Could we use the rand module instead?

davisp:
Looks like not because it doesn't appear to exist in 17.x

garrensmith:
+1 this is really nice. To test the api I built a [RocksDB adapter](https://github.com/garrensmith/couch_rocks/tree/initial-work). The api is really nice to use, there are two slightly rough edges - include_reductions on `fold_docs` and `monitored_by` functions. Paul already new those two sections are a little rough and said fixing those now will increase the scope of this PR even further. 

I tested CouchDB using `ngen`, `rocks` and `couch_bt_engine`. It all worked smoothly.

davisp:
For more color since I was talking to @garrensmith off list about his work on couch_rocks, those two places are a bit rough because I wanted to match the existing API/behavior exactly.

The `include_reductions/fold_docs` issue is for _all_docs when we list a total_rows and offset value. Specifically, the offset is difficult to calculate for anything that's not tree shaped. For engines that aren't tree shaped I recommend just faking a reduce value there and ignoring that the offset would be wrong (since we already recommend against using it for paging and other such things anyway). This edge case would be fixable if we ever drop the offset calculation.

The monitoring functions are also a bit funky. I exposed them in the API in case a storage engine wanted to be fancier there. But that's all how it interacts with couch_server so that we can make the guarantees about single writers per shard and so on. Its possible that in the future we remove those if we spend much time on reworking couch_server to change its semantics. For now its easy enough to implement, its just confusing *why* we implement it the way we do unless you're familiar with the whole couch_server notion of idle and how that's implemented with couch_file pids vs couch_db_updater pids.

src/couch/src/couch_db_updater.erl
```diff
@@ -345,15 +271,14 @@ handle_info({'EXIT', _Pid, normal}, Db) ->
...
-    {stop, normal, Db#db{fd=undefined, fd_monitor=closed}};
-handle_info(timeout, #db{fd=Fd, name=DbName} = Db) ->
+handle_info(timeout, #db{name=DbName} = Db) ->
     IdleLimitMSec = update_idle_limit_from_config(),
     case couch_db:is_idle(Db) of
         true ->
-            MSecSinceLastRead = couch_file:msec_since_last_read(Fd),
-            case MSecSinceLastRead > IdleLimitMSec of
+            LastActivity = couch_db_engine:last_activity(Db),
+            DtMSec = time:now_diff(os:timestamp(), LastActivity) div 1000,
```
jiangphcn: Should we need to use `timer:now_diff/2` instead of `time:now_diff/2`?

src/couch/src/couch_bt_engine.erl
```diff
@@ -0,0 +1,959 @@
...
+    case erlang:process_info(St#st.fd, monitored_by) of
+        {monitored_by, Pids} ->
+            Pids;
+        _ ->
+            []
+    end.
+
+
+last_activity(#st{fd = Fd}) ->
+    couch_file:get_last_read(Fd).
```
jiangphcn: There is no `couch_file:get_last_read/1`. Should we need to change to `couch_file:last_read/1`?

davisp:
Has anyone got any reason for me to not merge this in 72h or whenever? I've rebased it against master and its passing CI checks. As far as I remember everyone was happy with the current state for the V1 pluggable storage engines.

jiangphcn:
+1 from my side. All tests passed locally.

davisp:
Last rebase. Will merge once CI comes back green.

PR #497, opened 2017-04-27
notify couch_index_processes on all shards when ddoc updated

mayya-sharipova:
Before when a design doc is updated/deleted, only one couch_index process
was notified - the one which shard contained a design doc.
couch_index processes from other shards still continued to exist,
and indexing activities for these processes were still be going on.

The patch notifies couch_index_processes on all shards

## Testing recommendations
`make check apps=couch_index tests=ddoc_update_test_`

## JIRA issue number
[COUCHDB-3400](https://issues.apache.org/jira/browse/COUCHDB-3400)

## Checklist

### DIFF
--- a/src/couch_index/src/couch_index_server.erl
+++ b/src/couch_index/src/couch_index_server.erl
@@ -251,6 +251,22 @@ handle_db_event(DbName, created, St) ->
 handle_db_event(DbName, deleted, St) ->
     gen_server:cast(?MODULE, {reset_indexes, DbName}),
     {ok, St};
+handle_db_event(<<"shards/", _/binary>> = DbName, {ddoc_updated,
+        DDocId}, St) ->
+    DDocResult = couch_util:with_db(DbName, fun(Db) ->
+        couch_db:open_doc(Db, DDocId, [ejson_body, ?ADMIN_CTX])
+    end),
+    DbShards = [mem3:name(Sh) || Sh <- mem3:local_shards(mem3:dbname(DbName))],
+    lists:foreach(fun(DbShard) ->
+        lists:foreach(fun({_DbShard, {_DDocId, Sig}}) ->
+            case ets:lookup(?BY_SIG, {DbShard, Sig}) of
+                [{_, IndexPid}] -> (catch
+                    gen_server:cast(IndexPid, {ddoc_updated, DDocResult}));
+                [] -> []
+            end
+        end, ets:match_object(?BY_DB, {DbShard, {DDocId, '$1'}}))
+    end, DbShards),
+    {ok, St};
 handle_db_event(DbName, {ddoc_updated, DDocId}, St) ->
     lists:foreach(fun({_DbName, {_DDocId, Sig}}) ->
         case ets:lookup(?BY_SIG, {DbName, Sig}) of
--- a/src/couch_index/test/couch_index_ddoc_updated_tests.erl
+++ b/src/couch_index/test/couch_index_ddoc_updated_tests.erl
@@ -1,0 +1,144 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(couch_index_ddoc_updated_tests).
+
+-include_lib("couch/include/couch_eunit.hrl").
+-include_lib("couch/include/couch_db.hrl").
+
+
+start() ->
+    fake_index(),
+    Ctx = test_util:start_couch([mem3, fabric]),
+    DbName = ?tempdb(),
+    ok = fabric:create_db(DbName, [?ADMIN_CTX]),
+    {Ctx, DbName}.
+
+
+stop({Ctx, DbName}) ->
+    (catch meck:unload(test_index)),
+    ok = fabric:delete_db(DbName, [?ADMIN_CTX]),
+    DbDir = config:get("couchdb", "database_dir", "."),
+    WaitFun = fun() ->
+        filelib:fold_files(DbDir, <<".*", DbName/binary, "\.[0-9]+.*">>,
+            true, fun(_F, _A) -> wait end, ok)
+    end,
+    ok = test_util:wait(WaitFun),
+    test_util:stop_couch(Ctx),
+    ok.
+
+
+ddoc_update_test_() ->
+    {
+        "Check ddoc update actions",
+        {
+            setup,
+            fun start/0, fun stop/1,
+            fun check_all_indexers_exit_on_ddoc_change/1
+        }
+    }.
+
+
+check_all_indexers_exit_on_ddoc_change({_Ctx, DbName}) ->
+    ?_test(begin
+        [DbShard1 | RestDbShards] = lists:map(fun(Sh) ->
+           {ok, ShardDb} = couch_db:open(mem3:name(Sh), []),
+            ShardDb
+        end, mem3:local_shards(mem3:dbname(DbName))),
+
+        % create a DDoc on Db1
+        DDocID = <<"idx_name">>,
+        DDocJson = couch_doc:from_json_obj({[
+           {<<"_id">>, DDocID},
+           {<<"value">>, 1}
+        ]}),
+        {ok, _Rev} = couch_db:update_doc(DbShard1, DDocJson, []),
+        {ok, DbShard} = couch_db:reopen(DbShard1),
+        {ok, DDoc} = couch_db:open_doc(
+            DbShard, DDocID, [ejson_body, ?ADMIN_CTX]),
+        DbShards = [DbShard | RestDbShards],
+        N = length(DbShards),
+
+        % run couch_index process for each shard database
+        ok = meck:reset(test_index),
+        lists:foreach(fun(ShardDb) ->
+            couch_index_server:get_index(test_index, ShardDb, DDoc)
+        end, DbShards),
+
+        IndexesBefore = get_indexes_by_ddoc(DDocID, N),
+        ?assertEqual(N, length(IndexesBefore)),
+
+        AliveBefore = lists:filter(fun erlang:is_process_alive/1, IndexesBefore),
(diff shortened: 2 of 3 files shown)


src/couch_index/src/couch_index_server.erl
```diff
@@ -252,14 +252,20 @@ handle_db_event(DbName, deleted, St) ->
...
-                ok
-        end
-    end, ets:match_object(?BY_DB, {DbName, {DDocId, '$1'}})),
+    DDocResult = couch_util:with_db(DbName, fun(Db) ->
+        couch_db:open_doc(Db, DDocId, [ejson_body, ?ADMIN_CTX])
+    end),
+    DbNames = [mem3:name(Sh) || Sh <- mem3:local_shards(mem3:dbname(DbName))],
+    lists:foreach(
+        fun(DBName) ->
+        lists:foreach(fun({_DBName, {_DDocId, Sig}}) ->
```
iilyak: I think the indentation is wrong here (it should not be lined up with `fun(DbName) ->` IMO).

src/couch_index/src/couch_index_server.erl
```diff
@@ -252,14 +252,20 @@ handle_db_event(DbName, deleted, St) ->
...
-            [{_, IndexPid}] ->
-                (catch gen_server:cast(IndexPid, ddoc_updated));
-            [] ->
-                ok
-        end
-    end, ets:match_object(?BY_DB, {DbName, {DDocId, '$1'}})),
+    DDocResult = couch_util:with_db(DbName, fun(Db) ->
+        couch_db:open_doc(Db, DDocId, [ejson_body, ?ADMIN_CTX])
+    end),
+    DbNames = [mem3:name(Sh) || Sh <- mem3:local_shards(mem3:dbname(DbName))],
```
iilyak: DbNames name is not very clear. It could be named:
- Shards
- ShardNames
- DbShards

iilyak: Just few examples ^^^. You might find a better name.

iilyak:
Please remove `BugzID: 85773.` and provide JIRA reference instead.

src/couch_index/test/couch_index_ddoc_updated_tests.erl
```diff
@@ -0,0 +1,134 @@
...
+
+-include_lib("couch/include/couch_eunit.hrl").
+-include_lib("couch/include/couch_db.hrl").
+
+setup() ->
+    DbName = ?tempdb(),
+    fake_mem3(DbName),
+    fake_index(),
+
+    DbNames = [mem3:name(Sh) || Sh <- mem3:local_shards(mem3:dbname(DbName))],
```
iilyak: This is unusual way of doing things. You could use following if you use `test_util:start_couch([mem3, fabric])` in `ddoc_update_test_()`:
```
ok = fabric:create_db(DbName, [?ADMIN_CTX]),
```
However I see a value in doing it in your way. Since you don't have to start mem3 and fabric and tests are more isolated. 
I think your way is better.

src/couch_index/test/couch_index_ddoc_updated_tests.erl
```diff
@@ -0,0 +1,134 @@
...
+    {
+        "Check ddoc update actions",
+        {
+            setup,
+            fun() -> test_util:start_couch([]) end, fun test_util:stop_couch/1,
+            {
+                foreach,
+                fun setup/0, fun teardown/1,
+                [
+                    fun ddoc_update/1
```
iilyak: The name of the test doesn't represent what it is testing. It could be 
- `notify_indexers_on__ddoc_change`
- `check_all_exit_on_ddoc_change`
or something

src/couch_index/test/couch_index_ddoc_updated_tests.erl
```diff
@@ -0,0 +1,134 @@
...
+            {<<"_id">>, DDocID},
+            {<<"value">>, 2},
+            {<<"_rev">>, couch_doc:rev_to_str(DDoc#doc.revs)}
+        ]}),
+        {ok, _} = couch_db:update_doc(Db, DDocJson2, []),
+
+        % assert that all index processes exited after ddoc updated
+        couch_index_server:handle_db_event(
+            Db#db.name, {ddoc_updated, DDocID}, {st, ""}),
+        timer:sleep(1000),
```
iilyak: Is there a way to avoid using timer:sleep?

mayya-sharipova: @iilyak thanks for your comments. I will address them all in the next commit. The only thing I am not sure is how to avoid timer:sleep? If I check `couchdb_indexes_by_db` ETS immediately  after `ddoc_updated` notification, the ETS will still contain indexes. We need to allow some time pass for `couch_index` process to exit,  and `couch_index_server` process to clean its ETS tables.

iilyak: Here is what works locally:
```
--- a/src/couch_index/test/couch_index_ddoc_updated_tests.erl
+++ b/src/couch_index/test/couch_index_ddoc_updated_tests.erl
@@ -72,13 +72,17 @@ ddoc_update({[Db | _] = Dbs, DDoc}) ->
         DDocID = DDoc#doc.id,

         % run couch_index process for each Db
+        ok = meck:reset(test_index),
         lists:foreach(fun(DB) ->
             couch_index_server:get_index(test_index, DB, DDoc)
         end, Dbs),
-        IndexesBefore = ets:match_object(
-            couchdb_indexes_by_db, {'$1', {DDocID, '$2'}}),
+        IndexesBefore = get_indexes_by_ddoc(DDocID),
+
         ?assertEqual(N, length(IndexesBefore)),

+        AliveBefore = lists:filter(fun erlang:is_process_alive/1, IndexesBefore),
+        ?assertEqual(N, length(AliveBefore)),
+
         % update ddoc
         DDocJson2 = couch_doc:from_json_obj({[
             {<<"_id">>, DDocID},
@@ -88,15 +92,30 @@ ddoc_update({[Db | _] = Dbs, DDoc}) ->
         {ok, _} = couch_db:update_doc(Db, DDocJson2, []),

         % assert that all index processes exited after ddoc updated
+        ok = meck:reset(test_index),
         couch_index_server:handle_db_event(
             Db#db.name, {ddoc_updated, DDocID}, {st, ""}),
-        timer:sleep(1000),
-        IndexesAfter = ets:match_object(
-            couchdb_indexes_by_db, {'$1', {DDocID, '$2'}}),
+        ok = meck:wait(N, test_index, init, ['_', '_'], 5000),
+        IndexesAfter = get_indexes_by_ddoc(DDocID),
         ?assertEqual(0, length(IndexesAfter)),
+
+        %% make sure previously running indexes are gone
+        AliveAfter = lists:filter(fun erlang:is_process_alive/1, IndexesBefore),
+        ?assertEqual(0, length(AliveAfter)),
+
         ok
     end).

+get_indexes_by_ddoc(DDocID) ->
+    Indexes = ets:match_object(
+        couchdb_indexes_by_db, {'$1', {DDocID, '$2'}}),
+    [get_index(DbName, Sig) || {DbName, {_DDocID, Sig}} <- Indexes].
+
+get_index(DbName, Sig) ->
+    case ets:lookup(couchdb_indexes_by_sig, {DbName, Sig}) of
+        [{_, Pid}] -> Pid;
+        _ -> false
+    end.
 ```
Although I would move get_indexes_by_ddoc and get_index into couch_index_server module.

File: src/couch_index/src/couch_index_server.erl (hunk 1 above)
Comment (davisp) [inline src/couch_index/src/couch_index_server.erl]: The `fun(DbName) ->` should be moved up a line with `lists:foreach(` and the rest should be at the same indentation level as it is now.

File: src/couch_index/test/couch_index_ddoc_updated_tests.erl (hunk 5 above)
Comment (davisp) [inline src/couch_index/test/couch_index_ddoc_updated_tests.erl]: +1 for avoiding the sleep call. There's been a lot of work lately trying to track down broken tests in CI on ASF Jenkins and Travis-CI because those build hosts are so slow that our timing stuff goes out the window.

File: src/couch_index/test/couch_index_ddoc_updated_tests.erl [hunk 6]
```diff
@@ -0,0 +1,134 @@
...
+        ?assertEqual(0, length(IndexesAfter)),
+        ok
+    end).
+
+
+fake_mem3(DbName0) ->
+    ok = meck:expect(mem3, dbname, fun(_DbName) -> DbName0 end),
+    ok = meck:expect(mem3, name, fun(DbName) -> DbName end),
+    ok = meck:expect(mem3, local_shards, fun(DbName) -> [
+        <<DbName/binary, <<"/00000000-1fffffff">>/binary>>,
```
davisp: I'm a bit leery of mocking these functions to return different values than what would happen in real life. Perhaps adding a make_shard(Name) to return actual #shard records would be sufficient to remove the extra mocks?

davisp:
+1 to the logic. Just a few minor style issues to clean up and then this should be good to go.

src/couch_index/test/couch_index_ddoc_updated_tests.erl
```diff
@@ -0,0 +1,135 @@
...
+            setup,
+            fun start/0, fun stop/1,
+            fun check_all_indexers_exit_on_ddoc_change/1
+        }
+    }.
+
+
+check_all_indexers_exit_on_ddoc_change({_Ctx, DbName}) ->
+    ?_test(begin
+        fake_index(),
```
iilyak: This should be moved to start function for consistency. Since you do `meck:unload` in a `stop/1`.

src/couch_index/test/couch_index_ddoc_updated_tests.erl
```diff
@@ -0,0 +1,135 @@
...
+        [Db1 | RestDbs] = lists:map(fun(Sh) ->
```
iilyak: These are not `Dbs`. They are `DbOpenShards` or something. Feel free to ignore since it is minor.

src/couch_index/test/couch_index_ddoc_updated_tests.erl
```diff
@@ -0,0 +1,135 @@
...
+            Seq
+    end).
+
+
+get_indexes_by_ddoc(DDocID) ->
+    Indexes = ets:match_object(
+        couchdb_indexes_by_db, {'$1', {DDocID, '$2'}}),
+    lists:foldl(fun({DbName, {_DDocID, Sig}}, Acc) ->
+        case ets:lookup(couchdb_indexes_by_sig, {DbName, Sig}) of
+            [{_, Pid}] -> [Pid| Acc];
```
iilyak: Strange space. You either do `Pid|Acc` or `Pid | Acc`.

PR #501, opened 2017-04-30
Fix error on race condition in mem3 startup

wohali:
During mem3 startup, 2 paths attempt to call `couch_server:create/2` on
`_dbs`:

```
gen_server:init_it/6
  -> mem3_shards:init/1
    -> mem3_shards:get_update_seq/0
      -> couch_server:create/2
```

and

```
mem3_sync:initial_sync/1
  -> mem3_shards:fold/2
    -> couch_server:create/2
```

Normally, the first path completes before the second. If the second path
finishes first, the first path fails because it does not expect a
`file_exists` response.

This patch simply retries mem3_util:ensure_exists/1 once if it gets back
a `file_exists` response. Any failures past this point are not handled.

Fixes COUCHDB-3402.

/cc @davisp @nickva

### DIFF
--- a/src/mem3/src/mem3_util.erl
+++ b/src/mem3/src/mem3_util.erl
@@ -214,11 +214,12 @@ shard_info(DbName) ->
 ensure_exists(DbName) when is_list(DbName) ->
     ensure_exists(list_to_binary(DbName));
 ensure_exists(DbName) ->
-    case couch_db:open(DbName, [nologifmissing, sys_db | [?ADMIN_CTX]]) of
+    Options = [nologifmissing, sys_db, {create_if_missing, true}, ?ADMIN_CTX],
+    case couch_db:open(DbName, Options) of
     {ok, Db} ->
         {ok, Db};
-    _ ->
-        couch_server:create(DbName, [?ADMIN_CTX])
+    file_exists ->
+        couch_db:open(DbName, [sys_db, ?ADMIN_CTX])
     end.
 
 


davisp:
I think the bigger issue here is that `mem3_util:ensure_exists/1` is racey in that if multiple concurrent clients attempt to ensure a missing database exists simultaneously only one of them will succeed. A better approach would be to allow for an expected file_exists return and for the client that receives it to try re-opening a second time. Something along these lines:

```diff
diff --git a/src/mem3/src/mem3_util.erl b/src/mem3/src/mem3_util.erl
index 4e1b5fd7d..deb2c6640 100644
--- a/src/mem3/src/mem3_util.erl
+++ b/src/mem3/src/mem3_util.erl
@@ -224,11 +224,21 @@ shard_info(DbName) ->
 ensure_exists(DbName) when is_list(DbName) ->
     ensure_exists(list_to_binary(DbName));
 ensure_exists(DbName) ->
-    case couch_db:open(DbName, [nologifmissing, sys_db | [?ADMIN_CTX]]) of
-    {ok, Db} ->
-        {ok, Db};
-    _ ->
-        couch_server:create(DbName, [?ADMIN_CTX])
+    Options = [nologifmissing, sys_db, ?ADMIN_CTX],
+    case couch_db:open(DbName, Options) of
+        {ok, Db} ->
+            {ok, Db};
+        _ ->
+            case couch_server:create(DbName, Options) of
+                {ok, Db} ->
+                    {ok, Db};
+                file_exists ->
+                    % Try opening again in case there was a race to
+                    % create the database.
+                    couch_db:open(DbName, Options);
+                Error ->
+                    Error
+            end
     end.
```

wohali:
@davisp I agree, I was worried about making that change but I like the parallelism between your approach and what is done in e.g. https://github.com/apache/couchdb/blob/350a67b3bc2372b4a88649768805deb896c86451/src/couch/src/couch_server.erl#L83-L88 .

Question: Why call couch_db:open/2 instead of couch_server:open/2 ?

davisp:
That'd be more concise yeah.

iilyak:
> Question: Why not call couch_db:open/2 with Options including create_if_missing once, and if that fails with file_exists, just call it again without that option?

Thank you @wohali your suggestion is great. Following works locally:
```
diff --git a/src/mem3/src/mem3_util.erl b/src/mem3/src/mem3_util.erl
index 2cd444d..71ef5b6 100644
--- a/src/mem3/src/mem3_util.erl
+++ b/src/mem3/src/mem3_util.erl
@@ -214,11 +214,12 @@ shard_info(DbName) ->
 ensure_exists(DbName) when is_list(DbName) ->
     ensure_exists(list_to_binary(DbName));
 ensure_exists(DbName) ->
-    case couch_db:open(DbName, [nologifmissing, sys_db | [?ADMIN_CTX]]) of
+    Options = [nologifmissing, sys_db, {create_if_missing, true}, ?ADMIN_CTX],
+    case couch_db:open(DbName, Options) of
     {ok, Db} ->
         {ok, Db};
-    _ ->
-        couch_server:create(DbName, [?ADMIN_CTX])
+    file_exists ->
+        couch_db:open(DbName, [sys_db, ?ADMIN_CTX])
     end.
```

wohali:
Yup @iilyak I got busy with something else yesterday but will redo my PR in the next hour and merge it along the lines of what you show assuming tests pass.

PR #523, opened 2017-05-14
add debugging for failing test COUCHDB-3343

janl:
## Overview

Added debug out to failing test.

## JIRA issue number

https://issues.apache.org/jira/browse/COUCHDB-3343

## Checklist

### DIFF
--- a/test/javascript/couch_test_runner.js
+++ b/test/javascript/couch_test_runner.js
@@ -321,6 +321,11 @@ function TEquals(expected, actual, testName) {
     "', got '" + repr(actual) + "'", testName);
 }
 
+function TNotEquals(expected, actual, testName) {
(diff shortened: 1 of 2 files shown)


test/javascript/couch_test_runner.js
```diff
@@ -335,6 +340,10 @@ function equals(a,b) {
   }
 }
 
+function notEquals(a,b) {
+  if (a != b) return true;
```
janl: that is obviously faulty, will fix soon

wohali: ...is it?

janl: It turned out to be okay because in the equal case it returned `undefined`, which is falsy in JS, but I added a clearer return false now.

PR #533, opened 2017-05-19
Whait db close a little bit before failing test

iilyak:
## Overview

Databases are closed asynchronously. However it was not accounted for in
the test case. This commit waits database to close before it makes a
decision to abort the test. It uses default timeout of 5 sec.
I.e. if test case detects that database is still opened after 5 sec
it considers it as failure.

## Testing recommendations

make eunit apps=couch_index tests=compaction_test
## Checklist

### DIFF
--- a/src/couch_index/test/couch_index_compaction_tests.erl
+++ b/src/couch_index/test/couch_index_compaction_tests.erl
@@ -15,6 +15,8 @@
 -include_lib("couch/include/couch_eunit.hrl").
 -include_lib("couch/include/couch_db.hrl").
 
+-define(WAIT_TIMEOUT, 1000).
+
 setup() ->
     DbName = ?tempdb(),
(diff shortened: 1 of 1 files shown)


eiri:
Looks good to me, didn't know we have `test_util:wait/1,2,3` function. Except maybe pass timeout explicitly? " 'wait' for how long?" was my first question when I've seen the change.

iilyak:
@eiri: What timeout value would you recommend? I am considering 500.

eiri:
@iilyak hard to guess, to be honest, 500 or 1000 both looks good to me. test_util's `wait` default is 5000, but it's more fitting for the tests around os_process related modules, it's common default there.

iilyak:
@wohali: Sorry for the delay.

wohali:
@iilyak Hey I'm just glad people are starting to help ;)

PR #538, opened 2017-05-23
Catch Database Do Not Exist Error

tonysun83:
If dbs do not exist, we catch the error for mem3_sync_security
so that it can continue for databases that do exist.

## Testing recommendations

    Eunit Tests added.

## GitHub issue number

    
    COUCHDB-3423
## Checklist

### DIFF
--- a/src/mem3/src/mem3_sync_security.erl
+++ b/src/mem3/src/mem3_sync_security.erl
@@ -44,10 +44,20 @@ maybe_sync_int(#shard{name=Name}=Src, Dst) ->
 
 go() ->
     {ok, Dbs} = fabric:all_dbs(),
-    lists:foreach(fun handle_db/1, Dbs).
+    lists:foreach(fun handle_existing_db/1, Dbs).
 
 go(DbName) when is_binary(DbName) ->
-    handle_db(DbName).
+    handle_existing_db(DbName).
+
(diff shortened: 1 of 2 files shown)


src/mem3/src/mem3_sync_security.erl
```diff
@@ -44,10 +44,20 @@ maybe_sync_int(#shard{name=Name}=Src, Dst) ->
...
 go(DbName) when is_binary(DbName) ->
-    handle_db(DbName).
+    handle_existing_db(DbName).
+
+handle_existing_db(DbName) ->
+    try handle_db(DbName) of
+        _ -> ok
+    catch
+        error:database_does_not_exist->
+            couch_log:error("Db was deleted while getting security"
```
iilyak: Does it have to be an error? Would the following work?:
`couch_log:info("Skipping deleted database '~p' while syncing security object", [DbName])`

tonysun83: @iilyak : I think that depends on whether or not info level would be helpful for the user or if it will just add to logging clutter.

iilyak: @tonysun83: Agree. Let's keep it error. However please add DbName to the message.

PR #556, opened 2017-05-31
Improve JS test harness restartServer() support fn

### LINKED ISSUE #553: js restartServer() sometimes fails
## Current & Expected Behaviour
Failure is on master.

The JS test `test/javascript/tests/delayed_commits.js` has a bunch of calls to `restartServer()` in it. So do some other JS tests. Sometimes, the restart fails:

```

FAIL restart
```

The uploaded couch.log file after restart looks normal - no problems in startup that I can see (_id jenkins-couchdb-3-2017-05-30T21:27:03.085057). Perhaps it just took too long to restart.

## Possible Solution
Increase timeout?

wohali:
restartServer() is still erroring out sometimes in Travis/Jenkins. This
PR both bumps the timeout to 15s as well as changes the detection
mechanism for restart to look for the uptime in _system to reset to a
low number.

This PR also removes the eclipsed redundant restartServer() definition
in couch_test_runner.js.

Closes #553

## Checklist

- [ ] Documentation reflects the changes; (does not apply)

### DIFF
--- a/test/javascript/couch_test_runner.js
+++ b/test/javascript/couch_test_runner.js
@@ -436,37 +436,6 @@ function waitForSuccess(fun, tag) {
   }
 }
 
-function getCurrentToken() {
-  var xhr = CouchDB.request("GET", "/_restart/token");
-  return JSON.parse(xhr.responseText).token;
-};
-
-
-function restartServer() {
-  var token = getCurrentToken();
-  var token_changed = false;
-  var tDelta = 5000;
-  var t0 = new Date();
-  var t1;
-
-  CouchDB.request("POST", "/_restart");
-
-  do {
-    try {
-      if(token != getCurrentToken()) {
-        token_changed = true;
-      }
-    } catch (e) {
-      // Ignore errors while the server restarts
-    }
-    t1 = new Date();
-  } while(((t1 - t0) <= tDelta) && !token_changed);
-
-  if(!token_changed) {
-    throw("Server restart failed");
-  }
-}
-
 // legacy functions for CouchDB < 1.2.0
 // we keep them to make sure we keep BC
 CouchDB.user_prefix = "org.couchdb.user:";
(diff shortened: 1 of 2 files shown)


janl:
I might be missing something, but I can’t see any call to `/_restart` any more, or any other way of actually issuing a restart command.

On top of that, I think 2.0 dropped the `_restart` API. I can’t find it in `chttpd` and I believe we’ve disabled all tests that rely on it.

For tests we want (re-)enable, we need to port them to not rely on restart any more.

For the stats tests, the call to `/_restart` just not working, could be the cause for the off-by-one/a-few errors we are seeing.

janl:
I stand corrected:

```
[22:20:29]  <+Wohali>	so restart isn't done that way anymore
[22:21:33]  <+Wohali>	this took me a bit to figure out
[22:21:35]  <+Wohali>	https://github.com/apache/couchdb/blob/master/dev/run#L531-L540
[22:21:49]  <+Wohali>	so the javascript test suite is invoked as the run_command argument to dev/run
[22:22:07]  <+Wohali>	anything the test suite prints out to stdout is parsed by L538
[22:23:02]  <+Wohali>	so that's how this works: https://github.com/apache/couchdb/pull/556/files#diff-10d21021a1d1424fbe10267262a4d4a7L75
[22:23:19]  <+Wohali>	JS's restartServer() prints restart, which is eval()'d by dev/run and restarts the nodes
[22:23:48]  <+Wohali>	that's why I completely removed the old /_restart endpoint call - not only was it eclipsed by the version in test_setup.js, it simply doesn't work with 2.0
```

wohali:
So restart isn't done that way anymore. This took me a bit to figure out...

https://github.com/apache/couchdb/blob/master/dev/run#L531-L540

So the javascript test suite is invoked as the `run_command` argument to `dev/run`. Anything the test suite prints out to stdout is parsed by L538. 

So that's how this works:
https://github.com/apache/couchdb/pull/556/files#diff-10d21021a1d1424fbe10267262a4d4a7L75

JS's `restartServer()` prints `restart`, which is `eval()`'d by `dev/run` and restarts the nodes

That's why I completely removed the old `/_restart` endpoint call - not only was it eclipsed by the version in `test_setup.js`, it simply doesn't work with 2.0!

PR #560, opened 2017-06-01
Fix broken eunit test in changes_since_test_ test suite

jiangphcn:
## Overview

Prior to the change in this PR, most test cases in two test suites `changes_since_test_` and `changes_index_test_` failed to execute. They were disabled. Considering different scope and historical reason, this PR is aimed to address failure from `changes_since_test_` test suites. The test suite `changes_index_test_` will be another story.

The change not only includes test code, but also includes minor change of couch_mrview_util.erl to satisfy required pattern matches.

## Testing recommendations
```
module 'couch_mrview_changes_since_tests'
  changes_since tests
    couch_mrview_changes_since_tests:109: test_basic...ok
    couch_mrview_changes_since_tests:132: test_basic_since...ok
    couch_mrview_changes_since_tests:145: test_basic_count...ok
    couch_mrview_changes_since_tests:154: test_basic_count_since...ok
    couch_mrview_changes_since_tests:164: test_compact...ok
    couch_mrview_changes_since_tests:164: test_compact...ok
    [done in 0.666 s]
  changes_since_range tests
    couch_mrview_changes_since_tests:120: test_range...ok
    couch_mrview_changes_since_tests:141: test_range_since...ok
    [done in 0.236 s]
  changes_since_range_count tests
    couch_mrview_changes_since_tests:150: test_range_count...ok
    couch_mrview_changes_since_tests:159: test_range_count_since...ok
    couch_mrview_changes_since_tests:196: test_remove_key...ok
    couch_mrview_changes_since_tests:197: test_remove_key...ok
    couch_mrview_changes_since_tests:198: test_remove_key...ok
    couch_mrview_changes_since_tests:199: test_remove_key...ok
    [done in 0.423 s]
  [done in 3.173 s]
  All 14 tests passed.
==> rel (eunit)
```
Furthermore, the whole test suites from `couch_mrview` can execute successfully.
```
make check apps=couch_mrview
...
    couch_mrview_all_docs_tests:126: should_query_with_include_docs...ok
    couch_mrview_all_docs_tests:133: should_query_empty_views...ok
    [done in 0.537 s]
  [done in 0.635 s]
  All 129 tests passed.
==> rel (eunit)
```

## JIRA issue number
COUCHDB-3360 

## Related Pull Requests
N/A

## Checklist

- [] Documentation reflects the changes;
- [X] I will not forget to update [rebar.config.script](https://github.com/apache/couchdb/blob/master/rebar.config.script)

### DIFF
--- a/src/couch_mrview/src/couch_mrview_test_util.erl
+++ b/src/couch_mrview/src/couch_mrview_test_util.erl
@@ -49,26 +49,27 @@ make_docs(local, Count) ->
 make_docs(_, Count) ->
     [doc(I) || I <- lists:seq(1, Count)].
 
-ddoc(changes) ->
+ddoc({changes, Opts}) ->
+    ViewOpts = case Opts of
+        seq_indexed ->
+            [{<<"seq_indexed">>, true}];
+        keyseq_indexed ->
+            [{<<"keyseq_indexed">>, true}];
+        seq_indexed_keyseq_indexed ->
+            [
+                {<<"seq_indexed">>, true},
+                {<<"keyseq_indexed">>, true}
+            ]
+    end,
     couch_doc:from_json_obj({[
         {<<"_id">>, <<"_design/bar">>},
-        {<<"options">>, {[
-            {<<"seq_indexed">>, true}
-        ]}},
+        {<<"options">>, {ViewOpts}},
         {<<"views">>, {[
             {<<"baz">>, {[
-                {<<"map">>, <<"function(doc) {emit(doc.val, doc.val);}">>}
-            ]}},
-            {<<"bing">>, {[
-                {<<"map">>, <<"function(doc) {}">>}
-            ]}},
-            {<<"zing">>, {[
-                {<<"map">>, <<
-                    "function(doc) {\n"
-                    "  if(doc.foo !== undefined)\n"
-                    "    emit(doc.foo, 0);\n"
-                    "}"
-                >>}
+                {
+                    <<"map">>,
+                    <<"function(doc) {emit(doc.val.toString(), doc.val);}">>
+                }
             ]}}
         ]}}
     ]});
--- a/src/couch_mrview/src/couch_mrview_util.erl
+++ b/src/couch_mrview/src/couch_mrview_util.erl
@@ -343,7 +343,7 @@ get_view_changes_count(View) ->
             {ok, 0};
         {#btree{}, nil} ->
             couch_btree:fold_reduce(SBtree, CountFun, 0, []);
-        {nil, #btree{}} ->
+        {_, #btree{}} ->
             couch_btree:fold_reduce(KSBtree, CountFun, 0, [])
     end,
     case {SBtree, KSBtree} of
@@ -813,9 +813,9 @@ changes_expand_dups([{{[Key, Seq], DocId}, {dups, Vals}} | Rest], Acc) ->
 changes_expand_dups([{{Seq, Key}, {DocId, {dups, Vals}}} | Rest], Acc) ->
     Expanded = [{{Seq, Key, DocId}, Val} || Val <- Vals],
     changes_expand_dups(Rest, Expanded ++ Acc);
-changes_expand_dups([{{[Key, Seq], DocId}, Val} | Rest], Acc) ->
+changes_expand_dups([{{[Key, Seq], DocId}, {Val, _}} | Rest], Acc) ->
     changes_expand_dups(Rest, [{{Seq, Key, DocId}, Val} | Acc]);
-changes_expand_dups([{{Seq, Key}, {DocId, Val}} | Rest], Acc) ->
+changes_expand_dups([{{Seq, Key}, {DocId, Val, _}} | Rest], Acc) ->
     changes_expand_dups(Rest, [{{Seq, Key, DocId}, Val} | Acc]).
 
 maybe_load_doc(_Db, _DI, #mrargs{include_docs=false}) ->
--- a/src/couch_mrview/test/couch_mrview_changes_since_tests.erl
+++ b/src/couch_mrview/test/couch_mrview_changes_since_tests.erl
@@ -17,19 +17,12 @@
 
 -define(TIMEOUT, 1000).
 
-
-
-setup() ->
-    {ok, Db} = couch_mrview_test_util:init_db(?tempdb(), changes),
-    Db.
-
 teardown(Db) ->
     couch_db:close(Db),
     couch_server:delete(Db#db.name, [?ADMIN_CTX]),
     ok.
 
-
-changes_since_test() ->
+changes_since_basic_test_() ->
     {
         "changes_since tests",
         {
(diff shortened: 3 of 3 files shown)


src/couch_mrview/src/couch_mrview_test_util.erl
```diff
@@ -53,7 +53,8 @@ ddoc(changes) ->
     couch_doc:from_json_obj({[
         {<<"_id">>, <<"_design/bar">>},
         {<<"options">>, {[
-            {<<"seq_indexed">>, true}
+            {<<"seq_indexed">>, true},
+            {<<"keyseq_indexed">>, true}
```
eiri: It is valid to have only seq btree enabled on a view, i.e. all tests should pass even with key-seq btree not built. If anything we should expand the tests to assert both variants.

Can you explain, please, why do you think changing a scope of `couch_mrview_changes_since_tests.erl` to test `key-seq` instead of `seq` is the right fix?

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -344,6 +344,9 @@ get_view_changes_count(View) ->
         {#btree{}, nil} ->
             couch_btree:fold_reduce(SBtree, CountFun, 0, []);
         {nil, #btree{}} ->
+            couch_btree:fold_reduce(KSBtree, CountFun, 0, []);
+        {#btree{}, #btree{}} ->
+            couch_btree:fold_reduce(SBtree, CountFun, 0, []),
```
jiangphcn: Good suggestion, and already applied.

src/couch_mrview/src/couch_mrview_test_util.erl
jiangphcn: If query only contains "since", it should be enough to enable "seq_indexed". However, in some of test cases in this test suite, the query contains both "start_key" and "end_key".  This needs to enable "keyseq_indexed". Thus, the test can check `key_byseq_btree` instead of `seq_btree` to get expected results. I sorted out the test cases and only enable these two options when necessary.

src/couch_mrview/src/couch_mrview.erl
```diff
@@ -708,5 +708,5 @@ make_view_changes_opts(StartSeq, Options, Args) ->
         true ->
             couch_mrview_util:changes_key_opts(StartSeq, Args);
         false ->
-            [[{start_key, {StartSeq+1, <<>>}}] ++ Options]
+            [[{start_key, {StartSeq, <<>>}}] ++ Options]
```
jiangphcn: Eric, thanks for your comments. 

This is based on current understanding of query with only since.
```
test_basic_since(_, Db) ->
    Result = run_query(Db, 5, []),
    Expect = {ok, [
                {{6, 4, <<"4">>}, 4},
                {{7, 5, <<"5">>}, 5},
                {{8, 6, <<"6">>}, 6},
                {{9, 7, <<"7">>}, 7},
                {{10, 8, <<"8">>}, 8},
                {{11, 9, <<"9">>}, 9}
    ]},
    ?_assertEqual(Result, Expect).
```
If since is specified with 5, the expected returned document sequence should start from 6 instead of 7. Thus, there is change from [[{start_key, {StartSeq+1, <<>>}}] ++ Options] to [[{start_key, {StartSeq, <<>>}}] ++ Options]

Today, I test using one local cluster, and can find some test result related to `startkey` and `endkey`.
```
jiangphs-MacBook-Pro:geotest jiangph$ curl -u foo:bar -X GET 'http://localhost:15984/sdb1/_design/bar/_view/baz?startkey=3&endkey=4'|jq .
100   101    0   101    0     0   8824      0 --:--:-- --:--:-- --:--:--  9181
{
  "total_rows": 6,
  "offset": 2,
  "rows": [
    {
      "id": "3",
      "key": 3,
      "value": 3
    },
    {
      "id": "4",
      "key": 4,
      "value": 4
    }
  ]
}
```

However, for `since`, I still work from eunit test case.

jiangphcn:
Hi @eiri Thanks so much for reviewing codes and providing suggestion. I almost addressed your comments. However, regarding whether to set StartSeq with or without plus 1 for query only with `since`, we still need some discussion/test.

src/couch_mrview/src/couch_mrview_test_util.erl
```diff
@@ -72,6 +72,53 @@ ddoc(changes) ->
             ]}}
         ]}}
     ]});
+ddoc({changes, keyseq_indexed}) ->
```
eiri: This is rather repetitive, to be honest, when we only want to change options. How changing original `doc(changes)` to this:
```
ddoc({changes, Opts}) ->
    ViewOpts = case Opts of
        seq_indexed ->
            [{<<"seq_indexed">>, true}];
        keyseq_indexed ->
            [{<<"keyseq_indexed">>, true}];
        seq_indexed_keyseq_indexed ->
            [
                {<<"seq_indexed">>, true},
                {<<"keyseq_indexed">>, true}
            ]
    end,
    couch_doc:from_json_obj({[
        {<<"_id">>, <<"_design/bar">>},
        {<<"options">>, {ViewOpts}},
        {<<"views">>, {[
            {<<"baz">>, {[
                {<<"map">>, <<"function(doc) {emit(doc.val, doc.val);}">>}
            ]}}
        ]}}
    ]});
```

Same result, way less typing and it's a bit clearer on what's going on.
Also we don't need `bing` and `zing` since we don't use them in the tests.

src/couch_mrview/src/couch_mrview.erl
```diff
@@ -267,7 +267,7 @@ view_changes_since(View, StartSeq, Fun, Opts0, Acc) ->
     Wrapper = fun(KV, _, Acc1) ->
         Fun(KV, Acc1)
     end,
-    Opts = [{start_key, {StartSeq + 1, <<>>}}] ++ Opts0,
+    Opts = [{start_key, {StartSeq, <<>>}}] ++ Opts0,
```
eiri: Revert this change. See the explanation below.

src/couch_mrview/test/couch_mrview_changes_since_tests.erl
```diff
@@ -92,37 +129,37 @@ test_basic_since(Db) ->
...
 
-test_range_count_since(Db) ->
+test_range_count_since(_, Db) ->
     Result = run_count_query(Db, 5, [{start_key, 3}, {end_key, 5}]),
     ?_assertEqual(Result, 2).
 
-test_compact(Db) ->
+test_compact(_, Db) ->
     Result = couch_mrview:compact(Db, <<"_design/bar">>),
     ?_assertEqual(Result, ok),
```
eiri: There are for warnings for all those tests that looks like "Warning: a term is constructed, but never used". I'll explain what's wrong taking this line as an example.

In erlang unit tests all the macros starting with underscore not executed immediately, but instead return function wrapping the assertion. Function `test_compact` here is what called generator and it supposed to return either function to execute as a test or list of such function. We are returning test function generated with last line `?_assertEqual(Count, 10)`, so it got tested.

But call `?_assertEqual(Result, ok)` is not actually testing anything, it returns function that we are throwing away and this is what the warning about - we are constructing a function, but never use it. To fix it we need to return list of tests, like this: `[?_assertEqual(Result, ok), ?_assertEqual(Count, 10)]`

There are four such places in this test module, please fix all of them.

jiangphcn:
Thanks so much @eiri for your review and suggestions. I have addressed all of your current comments. You also mentioned that we want to have another PR to introduce "customer less". So don't change in this PR. 

I run test suites below and they can pass. Finally, thanks again! 

make eunit apps=couch_mrview suites=couch_mrview_changes_since_tests
```
module 'couch_mrview_changes_since_tests'
  changes_since tests
    couch_mrview_changes_since_tests:109: test_basic...ok
    couch_mrview_changes_since_tests:132: test_basic_since...ok
    couch_mrview_changes_since_tests:145: test_basic_count...ok
    couch_mrview_changes_since_tests:154: test_basic_count_since...ok
    couch_mrview_changes_since_tests:164: test_compact...ok
    couch_mrview_changes_since_tests:164: test_compact...ok
    [done in 0.666 s]
  changes_since_range tests
    couch_mrview_changes_since_tests:120: test_range...ok
    couch_mrview_changes_since_tests:141: test_range_since...ok
    [done in 0.236 s]
  changes_since_range_count tests
    couch_mrview_changes_since_tests:150: test_range_count...ok
    couch_mrview_changes_since_tests:159: test_range_count_since...ok
    couch_mrview_changes_since_tests:196: test_remove_key...ok
    couch_mrview_changes_since_tests:197: test_remove_key...ok
    couch_mrview_changes_since_tests:198: test_remove_key...ok
    couch_mrview_changes_since_tests:199: test_remove_key...ok
    [done in 0.423 s]
  [done in 3.173 s]
  All 14 tests passed.
==> rel (eunit)
```
make eunit apps=couch_mrview
```
    [done in 0.589 s]
  [done in 0.734 s]
  All 129 tests passed.
==> rel (eunit)
==> couchdb (eunit)
```

src/couch_mrview/src/couch_mrview.erl
jiangphcn: Yes, this makes sense. This is really subtle bug. Need to take care of different situations. We can use a different PR to address this.

jiangphcn:
@eiri thanks again for review and approval. Because I didn't have write access to  `https://github.com/apache/couchdb`, so currently I am working on `https://github.com/cloudant/couchdb`. I rebased from `https://github.com/cloudant/couchdb` and  squash commits. Right now, I can see `All Checks have passed`. Would you please help check and merge to  `https://github.com/apache/couchdb` if applicable? Thanks.

Issue #548, opened 2017-05-30
[Jenkins] 500 error on getting view output


wohali:
## Expected Behavior
In `reduce_builtin.js` we call a view that has a built in reduce. It should return fine, and usually does.

## Current Behavior
Once we saw a failure where a 500 comes back instead. The `couch.log` shows a gen_server exit: https://paste.apache.org/QfOz

## Possible Solution
```irc
 <+davisp> Its doing "exit(ExitReasonOfProcess) where ExitReasonOfProcess was {normal, stuff...}
 <+davisp> Looks like something isn't properly monitoring the index and it closes as a race condition
 <+davisp> I *think* this line should go after line 49 instead:
           https://github.com/apache/couchdb/blob/master/src/couch_mrview/src/couch_mrview_util.erl#L54
```

## Your Environment
[Jenkins automated run](https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/jenkins-pipeline/5/pipeline/44), `jenkins-pipeline` build 5, CentOS 7, Erlang 18.3. Logfiles uploaded to central logger as `jenkins-couchdb-5-2017-05-29T01:55:35.826219`.

wohali:
Seen again in the same automated run, Ubuntu 16.04, Erlang 18.3. Log excerpt here: https://paste.apache.org/8cF9

janl:
Let’s see if https://github.com/apache/couchdb/pull/562 fixes this.

@wohali IIRC this has to be merged to master before we see it in Jenkins, right?

janl:
Hah, ok, looking closer now, this can’t work.

@davisp line 54 relies on `State` which is only bound in line 50.

I’m not too deep in the code, is what you meant adding the `erlang:monitor` on the `Pid` that’s returned by `get_index_pid()`?

davisp:
Oh derp. Yeah, thats not right. This is a race condition when deleting the database out from under the view.

davisp:
Random idea. I wonder if curl is using keep-alive on the connection and we're getting a mesage back from an old rpc worker which looks like an error and closes the connection.

davisp:
Maybe try adding connection: close hereish?

https://github.com/apache/couchdb/blob/master/test/javascript/couch.js#L438

janl:
testing: https://github.com/apache/couchdb/pull/562

janl:
@davisp running this locally in a while loop `make javascript suites=builtin_reduce.js` still gives me occasional errors like:

```
[...]

PR #562, opened 2017-06-02
monitor index before getting its state #548

janl:
## Overview

Implemented suggestion in #548 

## Checklist


janl:
First iteration didn’t work, now switching to monitoring the doc index `Pid`, instead of the `State`’s `fd`, as that is only available after getting the state.

janl:
Second iteration didn’t fail, but [@davisp said](https://github.com/apache/couchdb/issues/548#issuecomment-305759652) isn’t helping.

Trying latest suggestion now.

janl:
@eiri is working a proper patch for this, closing.

Issue #552, opened 2017-05-30
[Jenkins] view monitor failure in couchdb_views_tests/couchdb_1283


wohali:
## Expected & Current Behaviour
Normally, `couchdb_views_tests`'s `couchdb_1283` test passes. Today, it failed in Jenkins, master branch, Ubuntu 14.04, default Erlang, logs uploaded as _id `jenkins-couchdb-3-2017-05-30T21:10:32.202064`.

```
  View group shutdown
    couchdb_views_tests:315: couchdb_1283...*failed*
in function gen_server:call/2 (gen_server.erl, line 180)
in call from couchdb_views_tests:'-couchdb_1283/0-fun-22-'/0 (test/couchdb_views_tests.erl, line 368)
**exit:{normal,{gen_server,call,[<0.17099.1>,compact]}}
```

What I see in the couch.log is pid `0.17099.1` closing immediately after creation:

```
[info] 2017-05-30T21:10:27.825327Z nonode@nohost <0.17031.1> -------- Apache CouchDB has started on http://127.0.0.1:39367/
[notice] 2017-05-30T21:10:27.825734Z nonode@nohost <0.16878.1> -------- config: [couchdb] max_dbs_open set to 3 for reason nil
[info] 2017-05-30T21:10:27.825761Z nonode@nohost <0.7.0> -------- Application couch started on node nonode@nohost
[notice] 2017-05-30T21:10:27.826162Z nonode@nohost <0.16878.1> -------- config: [couchdb] delayed_commits set to false for reason nil
[info] 2017-05-30T21:10:27.931491Z nonode@nohost <0.17099.1> -------- Opening index for db: eunit-test-db-1496178627826424 idx: _design/foo sig: "0963a19eb3ef007218f1e11f0aefa2d9"
[info] 2017-05-30T21:10:27.934223Z nonode@nohost <0.17102.1> -------- Starting index update for db: eunit-test-db-1496178627826424 idx: _design/foo
[info] 2017-05-30T21:10:27.987657Z nonode@nohost <0.17102.1> -------- Index update finished for db: eunit-test-db-1496178627826424 idx: _design/foo
[notice] 2017-05-30T21:10:27.987993Z nonode@nohost <0.17050.1> -------- 127.0.0.1 - - GET /eunit-test-db-1496178627826424/_design/foo/_view/foo 200
[...]

PR #568, opened 2017-06-03
feat(test): wait for db and index shutdown

janl:
Test branch/PR for https://github.com/apache/couchdb/issues/552#issuecomment-305909995


janl:
@wohali would be great to give this a spin on Jenkins, but we can equally merge to master and see what happens :)

wohali:
I'd rather not mess with the Jenkins config to build more than just the release branches. Best to just run this specific test in a loop for a few hundred times and see if the error goes away.

janl:
I’ve done a bit, but will set up the loop again and go for dinner, will report back.

janl:
no fail since, +1ing myself :)

PR #569, opened 2017-06-03
chore: remove bogus hint in `make javascript`

janl:


wohali:
+1. If we start getting emfile errors then we'll need to revisit - but I've never seen emfile errors when running `make javascript`.

janl:
Yeah, that was from before we deleted the databases we created in each test.

PR #575, opened 2017-06-05
Test if deleted docs are included/excluded when folding

garrensmith:
Adds tests to fold_changes and fold_docs to check if deleted are
included or excluded.


davisp:
+1, Thanks for the tests.

PR #577, opened 2017-06-05
Don't wrap real error in `badmatch` exception

eiri:
## Overview

A change in apache/couchdb#576 made any unexpected return from functions called in `get_view_index_state/4` to be presented as `badmatch` and hide the real error.

This unwraps the `badmatch` and re-throw it as an original response.

## Testing recommendations

A test suite `make javascript suites=view_errorsjs`should pass.

## Related Pull Requests

Bug introduced in apache/couchdb#576

## Checklist


wohali:
You fixed it! +1

PR #590, opened 2017-06-08
Wait for listener's exit during restart test

### LINKED ISSUE #581: eunit couch_log_config_test failed with get_listener() found
## Current & Expected Behaviour
Sometimes, the `couch_log_config_listener_test` test `couch_log_config_test`_ fails. It should always pass

Makefile output:
```
module 'couch_log_config_listener_test'
  couch_log_config_listener_test: couch_log_config_test_...*failed*
in function couch_log_config_listener_test:'-check_restart_listener/0-fun-2-'/1 (test/couch_log_config_listener_test.erl, line 38)
in call from couch_log_config_listener_test:check_restart_listener/0 (test/couch_log_config_listener_test.erl, line 38)
**error:{assertEqual_failed,[{module,couch_log_config_listener_test},
                     {line,38},
                     {expression,"get_handler ( )"},
                     {expected,not_found},
                     {value,{config_listener,{couch_log_sup,<0.3334.0>}}}]}

  couch_log_config_listener_test: couch_log_config_test_...[1.002 s] ok
  [done in 1.008 s]
```

couch.log from the uploader is **EMPTY**.

## Possible Solution
Looks like this test is expecting `gen_event:delete_handler` to occur immediately. Should there be a `timer:sleep()` call prior to checking `get_handler()`?

## Context
This used to be known as https://issues.apache.org/jira/browse/COUCHDB-3341 but is now on GH Issues.

eiri:
## Overview

To test a config listener's restart we are deleting event handler in config_event and then immediately checking event manager.

This creates race when we can be a slightly early and catch old handler yet to be removed or slightly late and get new handler that already been installed.

This patch trying to address this by doing the check immediately after old handler exits and then incrementally waiting for new handler to be installed.

## Testing recommendations

Run for an isolated test:

```bash
make eunit apps=couch_log suites=couch_log_config_listener_test tests=couch_log_config_test_
```

## GitHub issue number

Fixes #581 

## Checklist

### DIFF
--- a/src/couch_log/test/couch_log_config_listener_test.erl
+++ b/src/couch_log/test/couch_log_config_listener_test.erl
@@ -16,6 +16,7 @@
 -include_lib("couch_log/include/couch_log.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
+-define(TIMEOUT, 1000).
 
 couch_log_config_test_() ->
     {setup,
@@ -34,13 +35,27 @@ check_restart_listener() ->
 
     Handler1 = get_handler(),
     ?assertNotEqual(not_found, Handler1),
+    Ref = erlang:monitor(process, Listener1),
     ok = gen_event:delete_handler(config_event, get_handler(), testing),
-    ?assertEqual(not_found, get_handler()),
 
-    timer:sleep(100),
-    ?assertNot(is_process_alive(Listener1)),
+    receive
+        {'DOWN', Ref, process, _, _} ->
+            ?assertNot(is_process_alive(Listener1))
+        after ?TIMEOUT ->
+            erlang:error({timeout, config_listener_mon_death})
+    end,
 
-    ?assert(is_process_alive(get_listener())),
+    NewHandler = test_util:wait(fun() ->
+        case get_handler() of
+            not_found -> wait;
+            Reply -> Reply
+        end
+    end, ?TIMEOUT, 20),
+    ?assertEqual(Handler1, NewHandler),
+
+    Listener2 = get_listener(),
+    ?assert(is_process_alive(Listener2)),
+    ?assertNotEqual(Listener1, Listener2),
     ok.
 
 check_ignore_non_log() ->


wohali:
Ironically your test for this build failed on the couch_log_config_listener_test:

https://travis-ci.org/apache/couchdb/jobs/240997782

eiri:
Ironically indeed :/

Since it's guaranteed that the old handler was gone, it sent `'DOWN'` after all, that means this is a new registered handler already spawned by sup. I'll have to bash my head now to figure out what can be done about this.

eiri:
So I've been pondering on this for a while and I don't see a good solution on how we can fix this race without altering the test.

The issue here is that when we are trying to confirm that handler got removed from `config_event` manager we are aiming in a pretty narrow gap between `config_listener`'s terminate and `couch_log_sup` spawning and registering new `config_listener_mon`.

We can ensure that we are not checking handlers before old listener dies, but we can't delay supervisor without mocking `couch_log_sup` to set delay in its `handle_config_terminate`.
And mocking the very module we are testing doesn't seems right to me.

So I removed assertion on that short moment when handler is not in `config_event` and used it as a flag to wait on new listener to start. Then I've added assertion on that it is indeed a new listener and it is registered as the same handler id, which should be enough to confirm that listener's restart is working as expected.

@davisp I believe you are the original author of this test. Do you think this change is acceptable and doesn't compromise integrity of the test?

davisp:
+1

I didn't even remember writing this test so I went back to look at history. The reason this got all mixed up was the big config API changes that happened around the time I wrote this test. Originally it was its own gen_server which would wait 5s between reloads of the config listener but then we changed the config listener to be in the supervision tree which as you point out makes it quite the race. Looks like we hit that and then just changed the timeout without actually rethinking the test.

Your changes to wait for the handler to exit, wait for a new one wiht the same id and a different pid I agree perform the same logical assertion that the restart is working.

Good work on this!

PR #605, opened 2017-06-21
Add Prefer: return=minimal support

garrensmith:
## Overview
Adding this header allows the client to request that only Non-Essential
or All headers are removed from the response. This will decrease the
size of the response.

The idea is to remove all headers that a non-browser client would not need. 
I've had to leave `Server` header in otherwise Mochiweb adds its very lame/snarky header in its place.

There are two options `Non-Essential` which removes everything except caching headers. And then `All` which removes everything except Content-length, Content-Type, Server, Date.

## Testing recommendations

curl -v --header "X-Couch-Exclude-Headers: Non-Essential" 'http://dev:5984/_all_dbs

## GitHub issue number
This PR will be the number

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -59,7 +59,7 @@ socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
 require_valid_user = false
 ; List of headers that will be kept when the header Prefer: return=minimal is included in a request.
 ; If Server header is left out, Mochiweb will add its own one in.
-prefer_minimal = Cache-Control, Content-Length, Content-Type, ETag, Server, Vary
+prefer_minimal = Cache-Control, Content-Length, Content-Range, Content-Type, ETag, Server, Transfer-Encoding, Vary
 
 [database_compaction]
 ; larger buffer sizes can originate smaller files
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -194,10 +194,8 @@ credentials = false
 ; List of accepted methods
 ; methods =
 
-[exclude_headers]
-; List of headers that will be kept when the header option X-Couch-Exclude-Headers is included in a request
-; 'X-Couch-Exclude-Headers' : "All"
-all = Content-Length, Content-Type, Server, Transfer-Encoding
+[prefer_header]
+; List of headers that will be kept when the header Prefer: return=minimal is included in a request
 minimal = Cache-Control, Content-Length, Content-Type, ETag, Server, Vary
 
 [x_frame_options]
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -194,6 +194,12 @@ credentials = false
 ; List of accepted methods
 ; methods =
 
+[exclude_headers]
+; List of headers that will be kept when the header option X-Couch-Exclude-Headers is included in a request
+; 'X-Couch-Exclude-Headers' : "All"
+all = Server
+minimal = Cache-Control, Content-Length, Content-Type, ETag, Server, Vary
+
 [x_frame_options]
 ; Settings same-origin will return X-Frame-Options: SAMEORIGIN.
 ; If same origin is set, it will ignore the hosts setting
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -59,7 +59,7 @@ socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
 require_valid_user = false
 ; List of headers that will be kept when the header Prefer: return=minimal is included in a request.
 ; If Server header is left out, Mochiweb will add its own one in.
-prefer_minimal = Cache-Control, Content-Length, Content-Range, Content-Type, ETag, Server, Transfer-Encoding, Vary
+; prefer_minimal = Cache-Control, Content-Length, Content-Range, Content-Type, ETag, Server, Transfer-Encoding, Vary
 
 [database_compaction]
 ; larger buffer sizes can originate smaller files
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -57,6 +57,9 @@ backlog = 512
 docroot = {{fauxton_root}}
 socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
 require_valid_user = false
+; List of headers that will be kept when the header Prefer: return=minimal is included in a request.
+; If Server header is left out, Mochiweb will add its own one in.
+prefer_minimal = Cache-Control, Content-Length, Content-Type, ETag, Server, Vary
 
 [database_compaction]
 ; larger buffer sizes can originate smaller files
@@ -193,10 +196,6 @@ credentials = false
 ; headers =
 ; List of accepted methods
 ; methods =
-
-[prefer_header]
-; List of headers that will be kept when the header Prefer: return=minimal is included in a request
-minimal = Cache-Control, Content-Length, Content-Type, ETag, Server, Vary
 
 [x_frame_options]
 ; Settings same-origin will return X-Frame-Options: SAMEORIGIN.
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -59,7 +59,7 @@ socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
 require_valid_user = false
 ; List of headers that will be kept when the header Prefer: return=minimal is included in a request.
 ; If Server header is left out, Mochiweb will add its own one in.
-; prefer_minimal = Cache-Control, Content-Length, Content-Range, Content-Type, ETag, Server, Transfer-Encoding, Vary
+prefer_minimal = Cache-Control, Content-Length, Content-Range, Content-Type, ETag, Server, Transfer-Encoding, Vary
 
 [database_compaction]
 ; larger buffer sizes can originate smaller files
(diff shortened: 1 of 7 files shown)


src/chttpd/src/chttpd_exclude_headers.erl
```diff
@@ -0,0 +1,45 @@
...
+
+-module(chttpd_exclude_headers).
+-include_lib("couch/include/couch_db.hrl").
+
+-export([maybe_exclude_headers/2]).
+
+
+
+% List of essential headers for an http request. If we leave out server then Mochiweb adds in its default one
+-define(ESSENTIAL_HEADERS, ["Cache-Control", "ETag", "Content-Type", "Content-Length", "Vary", "Server"]).
```
rnewson: list these in alphabetical order.

rnewson: and probably best to lay them out one line per header, so we can modify it over time.

rnewson: perhaps it should even come from config?

rnewson:
forgot to add, the structure of the change looks good to me. My comments are mainly about configurability. I like the use of a custom header to modify the response, rather than a query parameter. The recommendation these days is _not_ to add `X-` but there's so much precedence in CouchDB that it might be weird not to always have it. I do think the request header, and its permitted values, are a bit wordy at the same time as being a bit unclear. Perhaps if the filtering were controlled by config this problem goes away.

garrensmith:
Thanks @rnewson I've changed it to support using a config setting.

rel/overlay/etc/default.ini
```diff
@@ -187,6 +187,12 @@ credentials = false
 ; List of accepted methods
 ; methods =
 
+[exclude_headers]
+; List of headers that will be kept when the header option X-Couch-Exclude-Headers is included in a request
+; 'X-Couch-Exclude-Headers' : "All"
+all = Server
+minimal = Cache-Control, Content-Length, Content-Type, ETag, Server, Vary
```
rnewson: will need Transfer-Encoding and perhaps more

garrensmith: Good point. I've added Content-Length, Content-Type and Transfer-Encoding to the all list

rnewson:
Take a look at RFC 7240 (https://tools.ietf.org/html/rfc7240).

`Prefer: return=minimal`

src/chttpd/src/chttpd_prefer_header.erl
```diff
@@ -0,0 +1,49 @@
...
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(chttpd_prefer_header).
+
+
+
```
davisp: Two blank lines between sections and functions please.

src/chttpd/src/chttpd_prefer_header.erl
```diff
@@ -0,0 +1,49 @@
...
+-export([maybe_return_minimal/2]).
```
davisp: Exports should go one per line so diff's in the future are nicer.

```erlang
-export([
    maybe_return_minimal/2
]).
```

src/chttpd/test/chttpd_prefer_header_test.erl
```diff
@@ -0,0 +1,96 @@
...
+        {"Content-Length", "100"},
+        {"ETag","\"12343\""},
+        {"Vary", "Accept-Encoding"},
+        {"Server","CouchDB/2.1.0-f1a1d7f1c (Erlang OTP/19)"}
+    ].
+
+
+
+default_no_exclude_header_test() ->
+    Headers = chttpd_prefer_header:maybe_return_minimal(mock_request([]), default_headers()),
```
davisp: Line length too long.

src/chttpd/test/chttpd_prefer_header_test.erl
```diff
@@ -0,0 +1,96 @@
...
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(chttpd_prefer_header_test).
+-include_lib("couch/include/couch_db.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+
+
```
davisp: Only two blank lines between sections and functions please.

rel/overlay/etc/default.ini
```diff
@@ -188,6 +188,10 @@ credentials = false
 ; List of accepted methods
 ; methods =
 
+[prefer_header]
```
davisp: Bike shed: Is this the best place for this, or would it be better as a key in the chttpd section. Something like "prefer_minimal = List-Of-Headers"

You might also add a comment that a user doesn't specify Server here that Mochiweb will add its own regardless.

src/chttpd/src/chttpd_prefer_header.erl
```diff
@@ -0,0 +1,49 @@
...
+            filter_headers(Headers, get_header_list());
+        _ -> 
+            Headers
+    end.
+
+
+
+filter_headers(Headers, IncludeList) ->
+    lists:filter(fun({HeaderName, _}) -> 
+        lists:member(HeaderName, IncludeList)
```
davisp: Seems like you probably want to look into mochiweb's header handling to make this comparison case insensitive.

davisp: Because users may or may not give headers in the same case as we have them specified internally.

davisp: Also that'd make for a good unit test as well.

rel/overlay/etc/default.ini
```diff
@@ -57,6 +57,9 @@ backlog = 512
 docroot = {{fauxton_root}}
 socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
 require_valid_user = false
+; List of headers that will be kept when the header Prefer: return=minimal is included in a request.
+; If Server header is left out, Mochiweb will add its own one in.
+prefer_minimal = Cache-Control, Content-Length, Content-Type, ETag, Server, Vary
```
rnewson: we'll need `Transfer-Encoding` otherwise views, changes, _all_dbs will be broken.
we'll need `Set-Cookie` so we can login.
what about CORS related headers?
we'll need `Content-Range` for range reqs on attachments.
we'll need `WWW-Authenticate` for users that enable the standard login prompting setting.

src/chttpd/src/chttpd_prefer_header.erl
```diff
@@ -0,0 +1,54 @@
...
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(chttpd_prefer_header).
+
+
+-export([
+    maybe_return_minimal/2
+    ]).
```
rnewson: this line should be at column 0.

rel/overlay/etc/default.ini
davisp: Agree on Transfer-Encoding and Content-Range. For cooking and authentication headers I think not since this is aimed at non-browser clients and those are both for browsers (and or, non browsers would just have to be smart enough to not request it when they need to get a cookie.

garrensmith: I don't think CORS is required since this is for non-browser clients. I agree with Paul around needing the auth headers, don't use this header when you need a cookie.

garrensmith:
Thanks @rnewson and @davisp for the reviews on this. I've updated the `default.ini` and added in the two missing headers. I've also fixed the spacing issue.

rel/overlay/etc/default.ini
```diff
@@ -57,6 +57,9 @@ backlog = 512
 docroot = {{fauxton_root}}
 socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
 require_valid_user = false
+; List of headers that will be kept when the header Prefer: return=minimal is included in a request.
+; If Server header is left out, Mochiweb will add its own one in.
+prefer_minimal = Cache-Control, Content-Length, Content-Range, Content-Type, ETag, Server, Transfer-Encoding, Vary
```
rnewson: should be commented out by default.

src/chttpd/src/chttpd_prefer_header.erl
```diff
@@ -0,0 +1,54 @@
...
+
+
+filter_headers(Headers, IncludeList) ->
+    lists:filter(fun({HeaderName, _}) -> 
+        lists:member(HeaderName, IncludeList)
+    end, Headers).
+
+
+get_header_list() ->
+    SectionStr = config:get("chttpd", "prefer_minimal", []),
```
rnewson: the default needs to be ""?

garrensmith: I've changed it to `""`. It seems` re:split("", "\\s*,\\s*", [trim, {return, list}]).` and ` re:split([], "\\s*,\\s*", [trim, {return, list}]).` return the same result of `[[]]`

rel/overlay/etc/default.ini
```diff
@@ -57,6 +57,9 @@ backlog = 512
 docroot = {{fauxton_root}}
 socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
 require_valid_user = false
+; List of headers that will be kept when the header Prefer: return=minimal is included in a request.
+; If Server header is left out, Mochiweb will add its own one in.
+; prefer_minimal = Cache-Control, Content-Length, Content-Range, Content-Type, ETag, Server, Transfer-Encoding, Vary
```
rnewson: I'm an idiot. I forgot that this is opt-in (you have to pass 'Prefer: return=minimal'). So please uncomment this again. :)

rnewson:
+1 to merge but please squash down from the 7 commits as appropriate.

garrensmith:
Docs https://github.com/apache/couchdb-documentation/pull/157

PR #606, opened 2017-06-21
Use Ejson Body Instead of Compressed Body for External size

tonysun83:
## Overview

In two places where we calculate the ExternalSize of the document body,
we use the Summary which is a compressed version of the doc body. We
change this to use the actual ejson body. In copy_docs we don't have
access to the #doc record so we can't access the meta where we store
the ejson body. Unfortunately, this means we have to decompress the
document body after reading it from disk.

## Testing recommendations

Performance tests were conducted to compare pre-fix External Size values and post-fix External Size Values. Here are the results:
```
Version        Compression    DB Dize      DB External Size

Pre-Fix         snappy            70 MB        46249702       
Pre-Fix         snappy            700 MB      462495533       
Pre-Fix         snappy            2.1 GB        1387486406      
                                        
Post-Fix       snappy            70 MB        83950133    
Post-Fix       snappy            700 MB      839500133        
Post-Fix       snappy            2.1 GB        2518500133         
                                        
Pre-Fix         deflate_6        70 MB        46249702         
Pre-Fix         deflate_6        700 MB      413395945      
Pre-Fix         deflate_6        2.1 GB       994598087         
                                        
Post-Fix       deflate_6        70 MB        83950133   
Post-Fix       deflate_6        700 MB      430495496  
Post-Fix       deflate_6        2.1 GB       1291486342
```
## GitHub issue number
PR is number.

## Checklist

### DIFF
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -1112,8 +1112,9 @@ prepare_doc_summaries(Db, BucketList) ->
             end,
             SummaryChunk = couch_db_updater:make_doc_summary(Db, {Body, DiskAtts}),
             Meta = Doc#doc.meta,
+            ExternalSize = ?term_size(Body),
             Doc#doc{body = {summary, SummaryChunk, SizeInfo, AttsFd},
-                meta = [{ejson_body, Body} | Meta]}
+                meta = [{ejson_size, ExternalSize} | Meta]}
         end,
         Bucket) || Bucket <- BucketList].
 
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -1112,9 +1112,10 @@ prepare_doc_summaries(Db, BucketList) ->
             end,
             SummaryChunk = couch_db_updater:make_doc_summary(Db, {Body, DiskAtts}),
             Meta = Doc#doc.meta,
-            ExternalSize = ?term_size(Body),
-            Doc#doc{body = {summary, SummaryChunk, SizeInfo, AttsFd},
-                meta = [{ejson_size, ExternalSize} | Meta]}
+            Doc#doc{
+                body = {summary, SummaryChunk, SizeInfo, AttsFd},
+                meta = [{ejson_size, ?term_size(Body)} | Meta]
+            }
         end,
         Bucket) || Bucket <- BucketList].
 
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -1111,7 +1111,9 @@ prepare_doc_summaries(Db, BucketList) ->
                 nil
             end,
             SummaryChunk = couch_db_updater:make_doc_summary(Db, {Body, DiskAtts}),
-            Doc#doc{body = {summary, SummaryChunk, SizeInfo, AttsFd}}
+            Meta = Doc#doc.meta,
+            Doc#doc{body = {summary, SummaryChunk, SizeInfo, AttsFd},
+                meta = [{ejson_body, Body} | Meta]}
         end,
         Bucket) || Bucket <- BucketList].
 
--- a/src/couch/src/couch_db_updater.erl
+++ b/src/couch/src/couch_db_updater.erl
@@ -1076,7 +1076,6 @@ check_md5(_, _) -> throw(md5_mismatch).
 
 copy_docs(Db, #db{fd = DestFd} = NewDb, MixedInfos, Retry) ->
     DocInfoIds = [Id || #doc_info{id=Id} <- MixedInfos],
-    Compress = NewDb#db.compression,
     LookupResults = couch_btree:lookup(Db#db.id_tree, DocInfoIds),
     % COUCHDB-968, make sure we prune duplicates during compaction
     NewInfos0 = lists:usort(fun(#full_doc_info{id=A}, #full_doc_info{id=B}) ->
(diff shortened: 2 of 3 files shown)


davisp:
Not sure what your test was there, but that's a pretty big red flag that the external sizes don't match between different compression algorithms as it should not be affected by that. You also probably want to do a pre/post compaction measurement as well to make sure compaction isn't affecting it either.

src/couch/src/couch_db.erl
```diff
@@ -1111,7 +1111,9 @@ prepare_doc_summaries(Db, BucketList) ->
                 nil
             end,
             SummaryChunk = couch_db_updater:make_doc_summary(Db, {Body, DiskAtts}),
-            Doc#doc{body = {summary, SummaryChunk, SizeInfo, AttsFd}}
+            Meta = Doc#doc.meta,
+            Doc#doc{body = {summary, SummaryChunk, SizeInfo, AttsFd},
+                meta = [{ejson_body, Body} | Meta]}
```
davisp: You should just store the size of the body here rather than pass the whole body. By doing that you're effectively nullifying the summary optimization of only passing binaries in and out of couch_db_updater.

src/couch/src/couch_db_updater.erl
```diff
@@ -1086,8 +1088,15 @@ copy_docs(Db, #db{fd = DestFd} = NewDb, MixedInfos, Retry) ->
         {NewRevTree, FinalAcc} = couch_key_tree:mapfold(fun
             (_Rev, #leaf{ptr=Sp}=Leaf, leaf, SizesAcc) ->
                 {Body, AttInfos} = copy_doc_attachments(Db, Sp, DestFd),
+                IsComp = couch_compress:is_compressed(Body, Compress),
+                EJsonBody = case IsComp of
```
davisp: This check isn't quite right as we might have changed the compression and we're looking at something that's not been compressed with the same method. Just use is_binary and then pass it to couch_compress:decompress if so.

davisp: Ah, though I see that's based off the make_doc_summary code. The difference here is that we don't care what its compressed as, where as make_doc_summary is converting a possibly compressed document to a different compression scheme.

tonysun83:
@davisp: I made the changes you suggested. Also, I think the performance benchmarks were using port 5984 instead of 5986, so it was influenced by the clustered nature of things. Let me know if this looks good, and I'll create a temporary build to check the performance numbers before merging.

src/couch/src/couch_db.erl
```diff
@@ -1111,7 +1111,10 @@ prepare_doc_summaries(Db, BucketList) ->
                 nil
             end,
             SummaryChunk = couch_db_updater:make_doc_summary(Db, {Body, DiskAtts}),
-            Doc#doc{body = {summary, SummaryChunk, SizeInfo, AttsFd}}
+            Meta = Doc#doc.meta,
+            ExternalSize = ?term_size(Body),
+            Doc#doc{body = {summary, SummaryChunk, SizeInfo, AttsFd},
```
davisp: Move attributes to their own lines when you wrap records. Something like:

```erlang
Doc#doc{
    body = {summary, ...},
    meta = [{ejson_size, ?term_size(Body)} | Meta]
}
```

src/couch/src/couch_db_updater.erl
```diff
@@ -1076,6 +1076,7 @@ check_md5(_, _) -> throw(md5_mismatch).
 
 copy_docs(Db, #db{fd = DestFd} = NewDb, MixedInfos, Retry) ->
     DocInfoIds = [Id || #doc_info{id=Id} <- MixedInfos],
+    Compress = NewDb#db.compression,
```
davisp: Not seeing where this is used.

src/couch/src/couch_db_updater.erl
```diff
@@ -1467,6 +1474,16 @@ make_doc_summary(#db{compression = Comp}, {Body0, Atts0}) ->
     SummaryBin = ?term_to_bin({Body, Atts}),
     couch_file:assemble_file_chunk(SummaryBin, couch_crypto:hash(md5, SummaryBin)).
 
+
+get_meta_body_size(Meta, Summary) ->
+    case lists:keyfind(ejson_size, 1, Meta) of
+        {ejson_size, ExternalSize} ->
+            ExternalSize;
+        false ->
+            couch_compress:decompress(Summary)
```
davisp: This is a bug right? You should be doing `?term_size(couch_compress:decompress(Summary))`.

src/couch/src/couch_db_updater.erl
```diff
@@ -1086,8 +1087,14 @@ copy_docs(Db, #db{fd = DestFd} = NewDb, MixedInfos, Retry) ->
         {NewRevTree, FinalAcc} = couch_key_tree:mapfold(fun
             (_Rev, #leaf{ptr=Sp}=Leaf, leaf, SizesAcc) ->
                 {Body, AttInfos} = copy_doc_attachments(Db, Sp, DestFd),
+                EJsonBody = case is_binary(Body) of
```
davisp: I see why we need to do this for existing shards that may have an incorrect value however I'm not a huge fan of this being a permanent change and not just around for an upgrade. Would be nice if we could figure out how to special case this so it only runs the decompression when we need to upgrade.

src/couch/src/couch_db_updater.erl
tonysun83: I call ExternalSize = ?term_size(EJsonBody), further down

davisp: Then your ejson_size you calculated is going to reduce all doc bodies to 4 bytes or whatever since you're returning an integer?

davisp: Also looking at the code I don't see where you call it.

tonysun83: wups, I was looking at the other place

PR #608, opened 2017-06-21
Calculate ExternalSize Correctly for Views

tonysun83:
## Overview

Previously, we were calculating the ExternalSize for views by summing
up all the nodes in the btree. Furthermore, this was the compressed
size. Now we modify the reduce function to return an ExternalSize for
uncompressed values in the KVList.

## Testing recommendations
Performance tests were run to compare External Size Info pre-fix and post-fix.

```
Version        Compression    DB Dize     View External Size

Pre-Fix         snappy            70 MB        26123231     
Pre-Fix         snappy            700 MB      493542711    
Pre-Fix         snappy            2.1 GB       795168531   
                                        
Post-Fix       snappy            70 MB        81775342
Post-Fix       snappy            700 MB      818752126        
Post-Fix       snappy            2.1 GB       2457701744       
                                        
Pre-Fix         deflate_6        70 MB        26109645         
Pre-Fix         deflate_6        700 MB      249677553     
Pre-Fix         deflate_6        2.1 GB       797076702        
                                        
Post-Fix       deflate_6        70 MB        81775244
Post-Fix       deflate_6        700 MB      818752412
Post-Fix       deflate_6        2.1 GB       2478167958
```
## GitHub issue number
PR is number

## Checklist

### DIFF
--- a/src/couch_mrview/src/couch_mrview_util.erl
+++ b/src/couch_mrview/src/couch_mrview_util.erl
@@ -338,7 +338,10 @@ temp_view_to_ddoc({Props}) ->
 
 
 get_row_count(#mrview{btree=Bt}) ->
-    {ok, {Count, _Reds}} = couch_btree:full_reduce(Bt),
+    Count = case couch_btree:full_reduce(Bt) of
+        {ok, {Count0, _Reds, _}} -> Count0;
+        {ok, {Count0, _Reds}} -> Count0
+    end,
     {ok, Count}.
 
 
@@ -786,27 +789,33 @@ changes_ekey_opts(_StartSeq, #mrargs{end_key=EKey,
     end.
 
 
+reduced_external_size(Tree) ->
+    case couch_btree:full_reduce(Tree) of
+        {ok, {_, _, Size}} -> Size;
+        % return 0 for versions of the reduce function without Size
+        {ok, {_, _}} -> 0
+    end.
 
 
 calculate_external_size(Views) ->
     SumFun = fun(#mrview{btree=Bt, seq_btree=SBt, key_byseq_btree=KSBt}, Acc) ->
-        Size0 = sum_btree_sizes(Acc, couch_btree:size(Bt)),
+        Size0 = sum_btree_sizes(Acc, reduced_external_size(Bt)),
         Size1 = case SBt of
             nil -> Size0;
-            _ -> sum_btree_sizes(Size0, couch_btree:size(SBt))
+            _ -> sum_btree_sizes(Size0, reduced_external_size(SBt))
         end,
         case KSBt of
             nil -> Size1;
-            _ -> sum_btree_sizes(Size1, couch_btree:size(KSBt))
+            _ -> sum_btree_sizes(Size1, reduced_external_size(KSBt))
         end
     end,
     {ok, lists:foldl(SumFun, 0, Views)}.
 
 
 sum_btree_sizes(nil, _) ->
-    null;
+    0;
 sum_btree_sizes(_, nil) ->
-    null;
+    0;
 sum_btree_sizes(Size1, Size2) ->
     Size1 + Size2.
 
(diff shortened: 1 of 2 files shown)


src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -609,7 +612,6 @@ make_header(State) ->
         log_btree=LogBtree,
         views=Views
     } = State,
-
```
davisp: Remove this WS only hunk

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -786,26 +788,32 @@ changes_ekey_opts(_StartSeq, #mrargs{end_key=EKey,
...
             nil -> Size1;
-            _ -> sum_btree_sizes(Size1, couch_btree:size(KSBt))
+            _ -> sum_btree_sizes(Size1, reduced_external_size(KSBt))
         end
     end,
     {ok, lists:foldl(SumFun, 0, Views)}.
 
 
-sum_btree_sizes(nil, _) ->
+sum_btree_sizes(null, _) ->
```
davisp: This nil/null change strikes me as possibly problematic for the non KV trees.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -1130,3 +1148,9 @@ get_view_queries({Props}) ->
         _ ->
             throw({bad_request, "`queries` member must be a array."})
     end.
+
+
+kv_external_size(KVList, Reduction) ->
+    lists:foldl(fun([[Key, _], Value], Acc) ->
+        size(term_to_binary(Key)) + size(term_to_binary(Value)) + Acc
```
davisp: Use ?term_size/1 instead of size(term_to_binary(Term)) for these.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -786,26 +788,32 @@ changes_ekey_opts(_StartSeq, #mrargs{end_key=EKey,
     end.
 
 
+reduced_external_size(Tree) ->
+    case couch_btree:full_reduce(Tree) of
+        {ok, {_, _, Size}} -> Size;
+        % return null for versions of the reduce function without Size
+        {ok, {_, _}} -> null
```
davisp: You're crossing streams a bit here. This returns null when there's no external size. But below your reduce function returns 0 when there's no external size. This means that with no write, we'll get null, but the first change will give us a number. We should pick one or the other.

src/couch_mrview/src/couch_mrview_util.erl
tonysun83: @davisp: I changed  ```reduced_external_size/1``` to return 0 for cases when an ExternalSize is not returned. Previously, the value returned when there was no ExternalSize was ```null```. So here I was accounting for the situation when it was null. Now that I changed the value 0, it seems the only possible values here are 0 or some other number. It can't be be nil or null?

tonysun83:
hmmm, pre compaction and post compaction is slightly off, need to figure out why

tonysun83:
looks like ```{ok, Result} = couch_query_servers:reduce(Lang, [FunSrc], KVs),``` returns a slightly different after compaction. That's why post-compaction, when we do ```?term_size(Reduction)```, it returns a slightly different value.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -1038,22 +1047,32 @@ get_user_reds(Reduction) ->
     element(2, Reduction).
 
 
+get_external_reds(Reduction) when tuple_size(Reduction) == 2 ->
```
davisp: Minor nit, call this get_external_size_reds. external_reds could be a bit misleading here.

PR #615, opened 2017-06-24
Annotate important processes

iilyak:
Opening for discussion. Let me know if you think it is a bad idea. I'll annotate more processes next week.

Once in a while for busy clusters there is a need to figure out a list of processes which has a big number of messages in their mailbox. Currently it is quite hard to infer the kind of process which has backups. We have to rely on inspecting an initial calls, monitors, links and so on.
The idea is to annotate important processes by storing kind key in the process dictionary.

### DIFF
--- a/src/couch/src/couch_debug.erl
+++ b/src/couch/src/couch_debug.erl
@@ -18,6 +18,15 @@
     opened_files_contains/1
 ]).
 
+-export([
+    tired_by_kind/2,
+    tired_by_kind/3,
+    process_by_kind/1,
+    process_by_kind/2
+]).
+
+
+
 -spec opened_files() ->
     [{port(), CouchFilePid :: pid(), Fd :: pid() | tuple(), FilePath :: string()}].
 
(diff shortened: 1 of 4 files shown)


davisp:
I would say that there's some work we can do here, but for the most part figuring out thigns based on initial_call and registered names is pretty easy. Granted there are a bunch that have an initial_call of erlang:apply/3. For those ones we should look into replacing erlang:spawn* with proc_lib:spawn* though for spawn_monitor we'll have to fake it cause proc_lib doesn't have a monitor related things (and prevents the option from being used for some unknown reason).

src/couch/src/couch_debug.erl
```diff
@@ -50,3 +59,28 @@ opened_files_contains(FileNameFragment) ->
     lists:filter(fun({_Port, _Pid, _Fd, Path}) ->
         string:str(Path, FileNameFragment) > 0
     end, couch_debug:opened_files()).
+
+tired_by_kind(Kind, Threshold) ->
```
wohali: Do you mean tiered? I don't think Erlang processes get tired. :)

iilyak:
I am glad I opened the PR for discussion prior to any significant work

PR #624, opened 2017-06-29
Enable compaction daemon by default

janl:
This is the #1 potential user footgun we ship by default. Especially with `_global_changes` coming on the scene, this becomes quite the annoying issue. Pro-users can easily disable it, most users will want to have it enabled.


wohali:
zOMG +1.

janl:
Heh, the tests might assume the feature to be off, I'll check in the morning :)

The feature works fine.

wohali:
@janl +1, do it!

PR #628, opened 2017-07-03
Add stable and update support to Mango

### LINKED ISSUE #621: Support stable and update options with mango queries
`stable` and `update` options give Mango queries more flexibility in terms of when the results set is updated and whether a fixed replica set is used.

garrensmith:
This brings mango inline with views by supporting the new options
`stable` and `update`.

Fixes #621

## Testing recommendations

See the test for an example on how to test. I'm not really sure of a way to test the `stable` option.

## GitHub issue number

#621

## Checklist

- [x] Documentation reflects the changes; (https://github.com/apache/couchdb-documentation/pull/145)

### DIFF
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -138,10 +138,10 @@ composite_prefix([Col | Rest], Ranges) ->
 
 
 % The query planner
-% First choose the index with the lowest difference between its 
+% First choose the index with the lowest difference between its
 % Prefix and the FieldRanges. If that is equal, then
-% choose the index with the least number of 
-% fields in the index. If we still cannot break the tie, 
+% choose the index with the least number of
+% fields in the index. If we still cannot break the tie,
 % then choose alphabetically based on ddocId.
 % Return the first element's Index and IndexRanges.
 %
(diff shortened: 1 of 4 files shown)


tonysun83:
stable and update are replacing stale right? will the same options work for text indexes?

garrensmith:
Yes, those are replacing stale. I'm not sure about text indexes. I don't think so

All misspelling thanks to my iPhone.

ditta95aR:
#621

janl:
Please see https://github.com/apache/couchdb/issues/621#issuecomment-315613763 for comment.

janl:
added stale, rebased with master and squashed my commits.

tonysun83:
@garrensmith @janl I like the addition of stale! We'll need to add in support for text indexes for stale as well. This way we are at least trying to be consistent across index options. We may also some logic if the user specifies stable or update for text indexes (which doesn't support that option, but we may already just ignore invalid options). We could move those changes into another PR if this a pressing option we want.

davisp:
+1

@tonysun83 I'd make any changes to the text indexes as new work. I checked and mango_cursor_text will ignore any unknown options so there's no breakage there.

tonysun83:
Sounds good. +1

iilyak:
Merging it in on Garren's behalf since he is on vacation at the moment.

PR #635, opened 2017-07-04
Stop indexing activity on ddoc update

mayya-sharipova:
Currently when ddoc is updated, couch_index and couch_index_updater processes
corresponding to the previous version of ddoc will still exist until
all indexing processing initiated by them is done.
When ddoc of a big database is rapidly modified, this puts a lot
of unnecessary strain on database.

With this change, when ddoc is updated:
* all couch_index processes for the previous version of ddoc will be shutdown
* all linked to them couch_index_updater processes will die as well
* all processes waiting for indexing activity to be finished (waiters
    for couch_index:get_status) will receive an immediate reply:
    ddoc_updated. Interactive user requests (view queries) will get response:
    {404, <<"not_found">>, <<"Design document was updated or deleted.">>}

## Overview
When design document is updated, stop couch_index processes for the previous version
of the design doc.

## Testing recommendations
`make check apps=couch_mrview tests=ddoc_update_test_`

## Checklist

### DIFF
--- a/src/couch_index/src/couch_index_server.erl
+++ b/src/couch_index/src/couch_index_server.erl
@@ -106,6 +106,13 @@ get_index(Module, IdxState) ->
     Sig = Module:get(signature, IdxState),
     case ets:lookup(?BY_SIG, {DbName, Sig}) of
         [{_, Pid}] when is_pid(Pid) ->
+            DDocId = Module:get(idx_name, IdxState),
+            case ets:match_object(?BY_DB, {DbName, {DDocId, Sig}}) of
+                [] ->
+                    Args = [Pid, DbName, DDocId, Sig],
+                    gen_server:cast(?MODULE, {add_to_ets, Args});
+                _ -> ok
+            end,
             {ok, Pid};
         _ ->
             Args = {Module, IdxState, DbName, Sig},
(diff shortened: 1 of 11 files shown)


src/couch_index/src/couch_index_server.erl
```diff
@@ -257,6 +257,12 @@ handle_db_event(<<"shards/", _/binary>> = DbName, {ddoc_updated,
         couch_db:open_doc(Db, DDocId, [ejson_body, ?ADMIN_CTX])
     end),
     DbShards = [mem3:name(Sh) || Sh <- mem3:local_shards(mem3:dbname(DbName))],
+    % Send ddoc_updated message to all couch_index processes for this DDoc,
+    % with the intention to close them.
+    % There can be other DDocs with the same Signature, served by the
+    % same couch_index processes, but the possibility for it is low,
+    % the negative effect is minor, and the check itself is expensive,
+    % => this check is skipped.
```
davisp: This is slightly worrisome as we have an established pattern of creating identical design docs to create views in the background without disrupting production. When we go to swap the design docs (ie, update old design doc to contents of new version, and then delete the old version) this would cause a non trivial amount of interruption to requests in progress.

I assume when you say that the check is expensive you're referring to the cost of looking up every design document and calculating signatures to see if we have duplicates. I wonder if a slight modification to just keep track of alternate design document names when opening the index would be useful here. That sort of "alias list" would have been useful a number of times that I can remember when debugging various customer issues. Then that makes the check something more along the lines of "if length(alias_list) == 1" which would be sufficiently cheap.

mayya-sharipova: @davisp thanks Paul for the review, will work on this change

src/couch_index/src/couch_index_server.erl
```diff
@@ -106,6 +106,13 @@ get_index(Module, IdxState) ->
     Sig = Module:get(signature, IdxState),
     case ets:lookup(?BY_SIG, {DbName, Sig}) of
         [{_, Pid}] when is_pid(Pid) ->
+            DDocId = Module:get(idx_name, IdxState),
+            case ets:match_object(?BY_DB, {DbName, {DDocId, Sig}}) of
+                [] ->
+                    Args = [DbName, DDocId, Sig],
+                    gen_server:cast(?MODULE, {add_to_ets, Args});
```
davisp: You should pass the pid through the cast here so that when handling the add_to_ets message we can assert that the entry in ?BY_PID still exists and if not then ignore them message. Otherwise this might lead to race conditions where we go to insert the entry just after the index was removed from the ets tables.

PR #645, opened 2017-07-05
Pass db open options to fabric_view_map

eiri:
## Overview

We don't set db open options to `fabric_view_map` and missing `user_ctx` when opening databases. For admin-only database we inject `user_ctx` for after doc check, pulling it from passed accumulator with assumption it is always `#vacc` record. This is incorrect in case of `list` or `mango` queries and breaks fabric abstraction, where it is spec'ed for acc to be type `any()`

This patch allows to pass db open options  and `user_ctx` along with it to `fabric:query_view/7` and modifies chttpd show and view handlers accordingly.

## Testing recommendations

Javascript test suite should pass:
```
make javascript
```

For manual check on admin-only database the following should work:

```
echo '{"name": "demo", "password": "apple", "roles": [], "type": "user"}' | curl -u $CRED http://localhost:15984/_users/org.couchdb.user:demo

...

echo '{"views": {"names": {"map": "function(doc) { emit(doc.name); }"}}, "lists": {"names": "function(head, req) { var row; while (row = getRow()) { send(\"name: \" + row.key + \"\\n\"); } }"}}}' | curl -u $CRED http://localhost:15984/_users/_design/users 
```

Then
```
$ curl -u $CRED http://localhost:15984/_users/_design/users/_view/names?include_docs=true
$ curl  -u $CRED "http://localhost:15984/_users/_design/users/_list/names/names"
```

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_show.erl
+++ b/src/chttpd/src/chttpd_show.erl
@@ -211,7 +211,8 @@ handle_view_list(Req, Db, DDoc, LName, {ViewDesignName, ViewName}, Keys) ->
             <<"_all_docs">> ->
                 fabric:all_docs(Db, Options, CB, Acc, QueryArgs);
             _ ->
-                fabric:query_view(Db, VDoc, ViewName, CB, Acc, QueryArgs)
+                fabric:query_view(Db, Options, VDoc, ViewName,
+                    CB, Acc, QueryArgs)
         end
     end).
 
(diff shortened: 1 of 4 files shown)


eiri:
End-points `_list` and `_show` no working on `_users` as well. `_show` because of the missing user context and `_list` because it is using not `#vacc` record as an acc for `query_view`, so this needs a bit different approach. Changing PR on WIP for now.

tonysun83:
Tested _show and that lead to 500 
```
[error] 2017-07-07T17:59:08.236067Z node1@127.0.0.1 <0.7836.0> 354b71b309 req_err(1517023101) case_clause : {forbidden,<<"Only administrators can view design docs in the users database.">>}
    [<<"chttpd_show:maybe_open_doc/2 L26">>,<<"chttpd_show:handle_doc_show_req/3 L52">>,<<"chttpd:process_request/1 L295">>,<<"chttpd:handle_request_int/1 L231">>,<<"mochiweb_http:headers/6 L122">>,<<"proc_lib:init_p_do_apply/3 L237">>]
```
This is a different bug though.

tonysun83:
+1

PR #646, opened 2017-07-05
Make map functions distinct with each invocation

sebastianrothbucher:
## Overview
[Issue 548](https://github.com/apache/couchdb/issues/548) points out that tests fail randomly when using ad-hoc queries (i.e. create a design doc, query it, remove it again). Though I can't reproduce it (not even by doing this cycle three times in a row without wait in-between), there is a solution by simply adding comments to the map function (hence changing the signature, allowing two otherwise similar views to co-exist. As it's hardly production relevant (design docs likely won't and probably shouldn't be created and removed all the time - it's a performance no-go in most cases), we actually get *better* tests as they really indicate if things go wrong (instead of failing randomly and being ignored after some time).

In short: improves the quality of javascript tests

## Testing recommendations
This IS a test

## GitHub issue number
[Issue 548](https://github.com/apache/couchdb/issues/548)

## Related Pull Requests
(none)

## Checklist

- n/a Documentation reflects the changes;

### DIFF
--- a/test/javascript/couch.js
+++ b/test/javascript/couch.js
@@ -147,6 +147,9 @@ function CouchDB(name, httpHeaders, globalRequestOptions) {
     if (typeof(mapFun) != "string") {
       mapFun = mapFun.toSource ? mapFun.toSource() : "(" + mapFun.toString() + ")";
     }
+    if ((!language) || language.toLowerCase() == "javascript") {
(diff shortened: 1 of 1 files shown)


wohali:
Once Travis passes in about 20 minutes, a big +1. Thanks for this!

sebastianrothbucher:
great!

kocolosk:
Yes, thanks. This is exactly the sort of thing I had in mind for #548

Issue #643, opened 2017-07-05
EUnit couchdb_views_tests failure: suspend_process after termination


wohali:
https://travis-ci.org/apache/couchdb/jobs/250135101#L3825

```
  View group shutdown
    couchdb_views_tests:315: couchdb_1283...*failed*
in function erlang:suspend_process/1
  called as suspend_process(<0.27290.0>)
in call from couchdb_views_tests:'-couchdb_1283/0-fun-21-'/0 (test/couchdb_views_tests.erl, line 381)
**error:badarg
  output:<<"">>
```

Analysis of the logfile shows the compaction completes in just 200ms:

```
[info] 2017-07-04T21:35:33.260737Z nonode@nohost <0.27290.0> -------- Compaction started for db: eunit-test-db-1499204132844268 idx: _design/foo
[info] 2017-07-04T21:35:33.262754Z nonode@nohost <0.27290.0> -------- Compaction finished for db: eunit-test-db-1499204132844268 idx: _design/foo
```

so my hunch is that it is already gone by the time we call erlang:suspend_process. This test is racey.

## Possible Solution
@rnewson says on IRC we could spawn it with `{hibernate_after,0}` but that only works for `gen_server:start_link|start|enter_loop`, and we'd have to get deep inside `couch_index_server` to do that, I think.

We could bump up the populate_db to 1000 or 10000 documents as a workaround.

wohali:
Another instance:
* https://travis-ci.org/apache/couchdb/jobs/250437020

wohali:
I believe these are being caused by #652.

wohali:
Closing because PR#654 landed. Will reopen if it recurs.

wohali:
Closing because PR#654 landed. Will reopen if it recurs.

PR #654, opened 2017-07-08
Disable compaction daemon on eunit run couch startups

### LINKED ISSUE #652: Compaction daemon is interfering with eunit tests (all_dbs_active errors)
The recent enabling of the compaction daemon also has it now running during the eunit tests, which is why we're suddenly getting these `all_dbs_active` errors (for which I haven't filed a ticket yet - this is it.)

If you had a `tmp/` directory inside of your couchdb checkout, you wouldn't notice this until you cleaned it out. Try removing your `tmp/` directory and you'll see the problems.

Grep your `*/.eunit/couch.log` files for "Compaction" and you'll see it compacting `_nodes`, `_dbs`, etc......and with each eunit test not deleting its dbs (in many cases) we end up with it cycling over a lot of dbs over time, which ends up filling the LRU and sometimes exhausting fds. Tests closer to the end of the eunit test suite fail more often with `all_dbs_active`, especially `couchdb_views_tests:couchdb_1283`.

I tried to set `[compactions] _default = []` (or the null string) in `rel/files/eunit.ini` but it doesn't work. And [the trick I used](https://github.com/apache/couchdb/pull/651/files#diff-d86563699bde00df2cce2613ecbed531R193) to stabilise the JS tests in my PR doesn't work, because some eunit tests actually need the compaction daemon, and that change completely removes its definition. Grumble.

## Expected Behavior
Compaction daemon should be disabled completely during eunit test runs.

## Current Behavior
Compaction daemon is running during eunit test runs, holding databases open, filling the LRU and wreaking havoc in general.

## Possible Solution
We could set the `[compactions] _default` to something like:

```
[compactions]
_default = [{db_fragmentation, "99%"}, {view_fragmentation, "99%"}, {from, "00:00"}, {to, "00:01"}, {strict_window, true}]
```

but there's still a window during which it could be interfering with databases, and during which it's going to scan through every single database it can get its hands on, which messes up the LRU. I don't know if specifying the same `from` and `to` time would work.

`make eunit` and grep your ~/.eunit/couch.log files for the string `"Compaction"`.

wohali:
Commit 21f9544 enabled the compaction daemon by default. And commit
3afe3ad disabled the compaction daemon at the start of the compaction
daemon tests. Unfortunately, the compaction daemon remains active during
all the other EUnit tests. This has resulted in some strange behaviour,
especially filling the LRU up and `all_dbs_active` errors during tests
that shouludn't be hitting it otherwise (like
`couchdb_views_tests:couchdb_1283` which monkeys with max_dbs_open ).

I attempted to override the [compactions] _default line in the file
rel/files/eunit.ini but specifying `_default=` or `_default=[]` did not
provide the desired behaviour. (This in and of itself is worrying; how would
someone disable the compaction daemon in `local.ini` now?)

This change disables the _default config as part of
`test_util:start_couch` after startup. This means there is a very brief
period during which the daemon is running, but in empirical testing the
only thing I've seen it manage to do is compact `_dbs` before being
disabled.

Tested with `make soak-eunit` for 3 hours; this seems to eliminate the
`all_dbs_active` error we've been seeing in CI runs since the compaction
daemon was enabled by default.

Now, I realize that this PR is at odds with making sure that all of the
usual things we do in CouchDB work correctly while the compaction daemon
is running in the background, so if you feel strongly about this, don't
hesitate to -1 it.

Noting here that the alternative is that we can try and fix all of the
suddenly failing tests - but with all of the DBs that EUnit creates and
deletes rapidly, the daemon will be working overtime to try and compact
them all. And many EUnit tests don't delete their databases after
they're done (which itself is probably another bug). The LRU is
adversely affected as well, as we see in
`couchdb_views_tests:couchdb_1283`, but
`couch_db_tests:should_create_multiple_dbs` would also be affected.

Other test failures from enabling the compaction daemon: 
* #649 
* #643 

Closes #652

### DIFF
--- a/src/couch/src/test_util.erl
+++ b/src/couch/src/test_util.erl
@@ -72,6 +72,7 @@ start_couch(IniFiles, ExtraApps) ->
     load_applications_with_stats(),
     ok = application:set_env(config, ini_files, IniFiles),
     Apps = start_applications(?DEFAULT_APPS ++ ExtraApps),
+    ok = config:delete("compactions", "_default", false),
     #test_context{started = Apps}.
 
 stop_couch() ->


davisp:
+1

I don't think we should worry about disabling the daemon for tests. Adding randomness to the tests when we're making assertions about global state isn't gonna work so hot when we've got something doing things in the background without coordination. If we're worried about testing things with the daemon running we should just wrote new tests for it rather than screw with all the existing tests.

Issue #656, opened 2017-07-08
Remove OAuth implementation


wohali:
CouchDB's OAuth implementation is OAuth 1.0 (not 2.0), supports 2-legged authentication only, and does not work with the CouchDB 2.x clustered interface.

As the implementation needs to be completely reworked, I am recommending we remove the current OAuth functionality for the 2.1 release. We already marked it as deprecated in 2.0.

@rnewson gave this a +1 today; if you have any strong objections, please say so before I complete the work to remove the (broken) functionality.

wohali:
Done!

PR #657, opened 2017-07-08
Remove deprecated OAuth 1.0 implementation

wohali:
Helps resolve issue #656. Implementation broken since bigcouch merge.

Replicator oauth hooks are left in place for future work towards
adding cookie-based authentication support.

## Related Pull Requests

https://github.com/apache/couchdb-documentation/pull/142

## Checklist


rnewson:
think there's more to do in replicator, chttpd, etc. I think it best to review as a whole set than piecemeal.

wohali:
@rnewson My goal was to ensure the feature was not enable-able in 2.1 at all, which this accomplishes. Readying chttpd/replicator for a different solution is beyond my ability without intense study, and unless you're up to the task, wouldn't be done in time for 2.1. What should we do?

Issue #674, opened 2017-07-13
Flaky compaction daemon test


davisp:
Originally reported by @wohali here: https://github.com/apache/couchdb/issues/552#issuecomment-312722372

https://travis-ci.org/apache/couchdb/jobs/243780116

```
module 'couchdb_compaction_daemon_tests'
  Compaction daemon tests
    couchdb_compaction_daemon_tests:74: should_compact_by_default_rule...*failed*
in function couchdb_compaction_daemon_tests:'-should_compact_by_default_rule/1-fun-6-'/1 (test/couchdb_compaction_daemon_tests.erl, line 104)
in call from couchdb_compaction_daemon_tests:'-should_compact_by_default_rule/1-fun-7-'/1 (test/couchdb_compaction_daemon_tests.erl, line 104)
**error:{assertion_failed,[{module,couchdb_compaction_daemon_tests},
                   {line,104},
                   {expression,"is_idle ( DbName )"},
                   {expected,true},
                   {value,false}]}
```

And again in https://travis-ci.org/apache/couchdb/jobs/243544740
And again in https://travis-ci.org/apache/couchdb/jobs/243303511

davisp:
This is probably related to #638

wohali:
OK looked at a couch.log. We have a race where the compaction daemon finishes compacting the test DB, then immediately starts compacting it **again**, so our check for the db to be idle fails.

```
[info] 2017-06-16T18:49:25.122688Z nonode@nohost <0.13883.1> -------- Starting compaction for db "eunit-test-db-1497638926674066"
[notice] 2017-06-16T18:49:25.383106Z nonode@nohost <0.13883.1> -------- Compaction swap for db: /home/travis/build/apache/couchdb/tmp/data/eunit-test-db-1497638926674066.couch 1646789 53445
[info] 2017-06-16T18:49:25.386118Z nonode@nohost <0.13883.1> -------- Compaction for db "eunit-test-db-1497638926674066" completed.
[info] 2017-06-16T18:49:25.390952Z nonode@nohost <0.16501.1> -------- Compaction started for db: eunit-test-db-1497638926674066 idx: _design/foo
[notice] 2017-06-16T18:49:25.434478Z nonode@nohost <0.13891.1> -------- Compaction swap for view /home/travis/build/apache/couchdb/tmp/data/.eunit-test-db-1497638926674066_design/mrview/ee2c720a13443ab1d50210242730ca31.view 391267 28800
[...]

PR #678, opened 2017-07-14
TMP: Add debug logging for failed assertion

davisp:
Opening this as a PR so that it gets built by CI to try and diagnose #674. This isn't intended to actually be merged (unless I can't get a failure then maybe I will so it'll be around when that test eventually does fail again).

### DIFF
--- a/src/couch/test/couchdb_compaction_daemon_tests.erl
+++ b/src/couch/test/couchdb_compaction_daemon_tests.erl
@@ -293,6 +293,13 @@ is_idle(DbName) ->
     {ok, Db} = couch_db:open_int(DbName, [?ADMIN_CTX]),
     Monitors = couch_db:monitored_by(Db),
     ok = couch_db:close(Db),
+    Others = [M || M <- Monitors, M /= self()],
+    if Others == [] -> ok; true ->
+        lists:foreach(fun(Other) ->
+            Args = [Other, process_info(Other)],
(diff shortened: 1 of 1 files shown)


src/couch/test/couchdb_compaction_daemon_tests.erl
```diff
@@ -293,6 +293,13 @@ is_idle(DbName) ->
     {ok, Db} = couch_db:open_int(DbName, [?ADMIN_CTX]),
     Monitors = couch_db:monitored_by(Db),
     ok = couch_db:close(Db),
+    Others = [M || M <- Monitors, M /= self()],
+    if Others == [] -> ok; true ->
+        lists:foreach(fun(Other) ->
+            Args = [Other, process_info(Other)],
```
iilyak: `process_info/1` is dangerous in production. Since it also includes all messages in the mailbox of a process.

davisp: That's quite true. But this is a test case change to try and track down why an assertion is failing. It'll be reverted once we figure that out.

davisp:
I can't get this to fail on the PR. Anyone mind if I just commit this to master until we see a reproduction? /cc @wohali @iilyak @nickva

wohali:
+1 for XKCD error ;)

davisp:
Makes it easy to grep for. :D

PR #699, opened 2017-07-19
Explicitly mention Facebook "BSD+Patents" license in NOTICE per LEGAL-303

### LINKED ISSUE #697: Make a note of the React PATENTS file
The ASF Legal team said that any project that has a release before Aug 31 using a Facebook library with the PATENTS file needs to license note. We should figure out what the language is for that and where to put it (I assume its in the LICENSE file, but just want to double check).

wohali:
Closes #697


davisp:
+1

PR #710, opened 2017-07-22
Fix/peruser test

### LINKED ISSUE #708: couch_peruser_tests using wrong asserts
From IRC:

```irc
17:22 < chewbranca> Wohali: jan____: yeap confirmed: https://gist.github.com/chewbranca/e97917468d32bedd198c1f370be7915e
17:22 < chewbranca> jan____: well it basically means that only the last test in the function is tested
17:22 < chewbranca> you need to return a list of functions for generators
17:25 < chewbranca> Wohali: jan____: for that module, basically just need to change things like this
                    https://github.com/apache/couchdb/blob/e5f7de648c19b79ae665800aa36d657a02b682b6/src/couch_peruser/test/couch_peruser_test.erl#L230-L232 to
                    return a list of those three asserts, rather than only returning the last one
17:26 < chewbranca> ?_* is basically equivalent to fun() -> ?assert(...) end and provides delayed evaluation of the test
17:26 <+jan____> chewbranca: or we turn them into directly executed asserts?
17:28 < chewbranca> jan____: that could work too, but you'll need to switch more things around, because *_test_ instead of *_test makes a generator test
```

janl:
closes #708

### DIFF
--- a/src/couch_peruser/test/couch_peruser_test.erl
+++ b/src/couch_peruser/test/couch_peruser_test.erl
@@ -136,7 +136,6 @@ should_create_user_db(TestAuthDb) ->
 
 should_create_anon_user_db(TestAuthDb) ->
     create_anon_user(TestAuthDb, "fooo"),
-    io:fwrite("HELLO: " ++ ?b2l(TestAuthDb) ++ "~n", []),
     ?_assert(lists:member(<<"userdb-666f6f6f">>, all_dbs())).
 
 should_not_delete_user_db(TestAuthDb) ->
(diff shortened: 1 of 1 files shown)


janl:
cc @chewbranca 

@wohali had to revert your commit since tests depend on each other, so same-db names are required.

wohali:
I'd still ripple the changes I had made through the assertions so that we remove the create/delete/create pattern.

wohali:
Speak of the devil:

https://travis-ci.org/apache/couchdb/jobs/256766131

```
    couch_peruser_test:188: should_add_user_to_db_admins...*failed*
in function couch_peruser_test:'-should_add_user_to_db_admins/1-fun-0-'/1 (test/couch_peruser_test.erl, line 190)
**error:{assertEqual_failed,[{module,couch_peruser_test},
                     {line,190},
                     {expression,"proplists : get_value ( << \"admins\" >> , get_security ( UserDbName ) )"},
                     {expected,{[{<<"names">>,[<<"qux">>]}]}},
                     {value,undefined}]}
```

janl:
> Still worried about the create/delete/create pattern. Is there no way to make this work without that pattern? 

not entirely, part of the whole idea is that databases should be created/deleted on user create/delete.

wohali:
We don't have time to wait on this one. Let's merge the fixes as they stand, and if the test proves unreliable in the future, let's rework then.

+1

PR #740, opened 2017-08-09
Add Bookmark support for mango json queries

garrensmith:
This adds a bookmark that is sent with each query and can be used to
continue a query from a specific key. This will allow users to paginate
mango queries.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -315,7 +315,6 @@ doc_member(Db, RowProps, Opts) ->
             end
     end.
 
-
 is_design_doc(RowProps) ->
     case couch_util:get_value(id, RowProps) of
         <<"_design/", _/binary>> -> true;
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -39,6 +39,7 @@ create(Db, Indexes, Selector, Opts) ->
     Limit = couch_util:get_value(limit, Opts, mango_opts:default_limit()),
     Skip = couch_util:get_value(skip, Opts, 0),
     Fields = couch_util:get_value(fields, Opts, all_fields),
+    Bookmark = couch_util:get_value(bookmark, Opts), 
 
     {ok, #cursor{
         db = Db,
@@ -48,7 +49,8 @@ create(Db, Indexes, Selector, Opts) ->
         opts = Opts,
         limit = Limit,
         skip = Skip,
-        fields = Fields
+        fields = Fields,
+        bookmark = Bookmark
     }}.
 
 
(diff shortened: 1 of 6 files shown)


willholley:
testing a basic case where no results are returned, I get a response:
```
{"warning":"no matching index found, create an index to optimize query time",
"docs":[
],
"bookmark": "nil"}
```

However, passing "nil" back as a bookmark returns an error:
```
$ curl 'http://127.0.0.1:15984/test/_find' -H 'Content-Type:application/json' -d '{"selector":{"id":"willnotmatch"},"bookmark":"nil"}'
{"error":"invalid_bookmark","reason":"Invalid bookmark value: \"nil\""}
```

willholley:
suggest adding a pagination test where each document emits the same value into the index to ensure progress is made using `startkey_docid`.

garrensmith:
Thanks @willholley I've fixed the issue and added another test.

src/mango/src/mango_json_bookmark.erl
```diff
@@ -0,0 +1,68 @@
...
+    couch_util:encodeBase64Url(Bin);
+create(#cursor{bookmark = Bookmark}) ->
+    Bookmark.
+
+
+unpack(nil) ->
+    nil;
+unpack(Packed) ->
+    try
+        binary_to_term(couch_util:decodeBase64Url(Packed))
```
tonysun83: This part got me thinking a bit. What is a valid vs invalid bookmark? So when unpacked, a json bookmark is a start_key and end_key range, and a text bookmark is a node+shard range. So the composition is slightly different. So a text index bookmark should be considered invalid for a json index bookmark right? When I tried that particular scenario, the bookmark passed correctly ``` Here's a sample text bookmark: "g2wAAAACaAJkAA5zdGFydGtleV9kb2NpZG0AAAAkZTkwMDAwMWQtYmM0OC00OGE2LTliMWEtYWM5YTFmNWQxYTAzaAJkAAhzdGFydGtleWsAAStq"``` . 

If you did it the other way around...substitute a json index bookmark into a text index bookmark, it fails with invalid_bookmark.

Wondering if we should be more consistent/strict and and verify both types of bookmarks?

src/mango/src/mango_json_bookmark.erl
```diff
@@ -0,0 +1,68 @@
...
+    catch _:_ ->
+        ?MANGO_ERROR({invalid_bookmark, Packed})
+    end.
+
```
tonysun83: missing new line

src/mango/src/mango_cursor_view.erl
```diff
@@ -305,8 +315,20 @@ doc_member(Db, RowProps, Opts) ->
             end
     end.
 
+
```
tonysun83: whitespace change

src/mango/src/mango_json_bookmark.erl
garrensmith: I don't quite follow? Do you want me to remove that new line or add another?

garrensmith:
@tonysun83 thanks for the review. I've make some fixes based on your feedback

src/mango/src/mango_json_bookmark.erl
tonysun83: weird, it showed no new end of line. just normal end of file new line is fine

tonysun83:
+1 after that fixing that weird end of line issue

anuragvohraec:
> This adds a bookmark that is sent with each query and can be used to
> continue a query from a specific key. This will allow users to paginate
> mango queries.

How exactly is bookmark implmented? [I have no knowledge of Erlang, but have written a couchdb client in dart]. Is it base64 enocded of json of last index[or doc] sent to user ?

garrensmith:
@anuragvohraec the code is here https://github.com/apache/couchdb/blob/master/src/mango/src/mango_json_bookmark.erl
We basically take the startkey and startkey_docid and convert to erlang binary and then to base64

PR #742, opened 2017-08-09
couch_epi depends on crypto app

rnewson:
couch_epi calls crypto functions since 9e56cf334c33d4cdb96acbfca9cc9461e2b658aa so should ensure the app is started first.


wohali:
+1 once Travis passes.

PR #746, opened 2017-08-10
Log additional information on crashes from rexi

iilyak:
## Overview

This improve the log messages from rexi. It adds flowing info:

- nonce
- node originator of the request
- mfa of the function called
- client Pid

## Checklist

### DIFF
--- a/src/rexi/src/rexi_server.erl
+++ b/src/rexi/src/rexi_server.erl
@@ -138,7 +138,11 @@ init_p(From, {M,F,A}, Nonce) ->
     put(nonce, Nonce),
     try apply(M, F, A) catch exit:normal -> ok; Class:Reason ->
         Stack = clean_stack(),
-        couch_log:error("rexi_server ~p:~p ~100p", [Class, Reason, Stack]),
+        {ClientPid, _ClientRef} = From,
+        couch_log:error(
+            "rexi_server: from: ~s(~p) mfa: ~s:~s/~p ~p:~p ~100p", [
+            node(ClientPid), ClientPid, M, F, length(A),
+            Class, Reason, Stack]),
         exit(#error{
             timestamp = now(),
             reason = {Class, Reason},


chewbranca:
Overall I think this looks good, but I'm a bit confused as to why the nonce isn't getting logged already. We write the log entry in [1], which includes the `MsgId` in [2], and we set that value in [3] by way of [4]. This log statement is happening in `rexi_server:init_p/3` which is already recording the nonce [5], so I'm a bit confused as to why the nonce is not showing up as expected, or even the fall back msg_id [6].


[1] https://github.com/apache/couchdb/blob/e6ecb1030b1971a49e69a4d7f2042ed7eb9e6653/src/couch_log/src/couch_log_writer_syslog.erl#L76-L96
[2] https://github.com/apache/couchdb/blob/e6ecb1030b1971a49e69a4d7f2042ed7eb9e6653/src/couch_log/src/couch_log_writer_syslog.erl#L92
[3] https://github.com/apache/couchdb/blob/e6ecb1030b1971a49e69a4d7f2042ed7eb9e6653/src/couch_log/src/couch_log_formatter.erl#L44
[4] https://github.com/apache/couchdb/blob/e6ecb1030b1971a49e69a4d7f2042ed7eb9e6653/src/couch_log/src/couch_log_util.erl#L50
[5] https://github.com/cloudant/couchdb/blob/e6ecb1030b1971a49e69a4d7f2042ed7eb9e6653/src/rexi/src/rexi_server.erl#L138
[6] https://github.com/cloudant/couchdb/blob/e6ecb1030b1971a49e69a4d7f2042ed7eb9e6653/src/couch_log/src/couch_log_util.erl#L51

iilyak:
@chewbranca: You were right the Nonce shouldn't be part of a message. I just pushed the change which removes it. The nonce is added by `couch_log_writer_syslog.erl` to the syslog message. It can be accessed as `%msgid` from format pattern.

chewbranca:
+1

PR #751, opened 2017-08-14
Add [couchdb, httpd_status_codes, 503] metric

iilyak:
## Overview

The 503 response was added recently. We need to update the `stats_description.cfg` to include this new return code in order to avoid `"unknown metric: [couchdb,httpd_status_codes,503]"` errors from [here](https://github.com/apache/couchdb/blob/master/src/chttpd/src/chttpd.erl#L1127).

## Checklist


chewbranca:
+1

PR #753, opened 2017-08-15
run mango tests on travis

garrensmith:
Update our travis file to run the mango tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ cache:
 
 # logfile uploader credentials
 env:
-  matric:
+  matrix:
     - COMMAND=check
     - COMMAND=mango-on-travis
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,9 +36,6 @@ cache:
 
 # logfile uploader credentials
 env:
-  matrix:
-    - COMMAND=check
-    - COMMAND=mango-on-travis
   global:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -35,6 +36,9 @@ cache:
 
 # logfile uploader credentials
 env:
+  matrix:
+    - COMMAND=check
+    - COMMAND=mango-on-travis
```
wohali: This is going to explode our tests from currently 4 runs to 8 runs. We're already throttled with the number of simultaneous runs we can have, this is going to increase runtimes for Travis overall to at least double, if not worse.

I don't think we need the full crossbar approach. Wouldn't it be better to simply run `make check mango-on-travis` instead?

Makefile
```diff
@@ -176,6 +176,17 @@ build-test:
 	@test/build/test-configure.sh
 
 
+.PHONY: mango-on-travis
```
wohali: Why not just call it `test-mango`? I can see using this not just in Travis.

If we're not going to migrate the mango tests to etap or javascript we might as well just add this as a new target for `make check`. Then, the `.travis.yml` file won't change at all and we simply extend our normal test run slightly.

wohali: If you take the latter approach, you'll want to kill off the `./dev/run` process. If you collapsed all of the steps to a single line, you could add a `kill %1` at the end; otherwise you'll have to go fishing for the pid somehow.

garrensmith: Thanks @wohali I'll give that a try

garrensmith:
@wohali could you take another look. My makefile skills are pretty lacking. I found this much easier to just create a script file to setup couchdb before running the tests.

wohali:
Much cleaner! Won't run on Windows but we can tackle that later.

Can you look into the Mango failures on Travis before we merge this? I don't want us to get back into the habit of breaking the build. @tonysun83 are you able to help here?

wohali:
Thank you @garrensmith ! 💃

PR #754, opened 2017-08-16
Compactor optimization - PR FOR FEEDBACK ONLY

davisp:
It occurred to me the other day that we already have the docid tree stored in the original database so the first pass at compaction is wasting a lot of time and IO writing out the emsort structure. This optimizes the compactor to avoid that first use of emsort and instead reads document IDs from the original db's id_tree.

There are zero tests for correct compactor behavior right now (outside of the PSE branch) so I'm going to start writing a bunch of tests that will do mean things to the compactor to make sure that this doesn't have any issues around breaking things. However, given the change I figured I'd at least open a feedback PR while I write the test suite for the compactor behavior.

### DIFF
--- a/src/couch/src/couch_db_updater.erl
+++ b/src/couch/src/couch_db_updater.erl
@@ -26,6 +26,7 @@
 -record(comp_st, {
     old_db,
     new_db,
+    new_id_tree,
     meta_fd,
     retry
 }).
@@ -1049,10 +1050,9 @@ start_copy_compact(#db{}=Db) ->
         fun copy_purge_info/1,
         fun copy_compact/1,
         fun commit_compaction_data/1,
-        fun sort_meta_data/1,
-        fun commit_compaction_data/1,
-        fun copy_meta_data/1,
-        fun compact_final_sync/1
+        fun copy_doc_ids/1,
+        fun compact_final_sync/1,
+        fun verify_compaction/1
     ],
 
     FinalCompSt = lists:foldl(fun(Stage, CompSt) ->
@@ -1092,8 +1092,9 @@ open_compaction_files(OldDb) ->
             #comp_st{
                 old_db = OldDb,
                 new_db = Db1,
+                new_id_tree = Db0#db.id_tree,
                 meta_fd = MetaFd,
-                retry = Db0#db.id_tree
+                retry = (couch_btree:size(Db0#db.id_tree) > 0)
             };
         _ when DataHdrIsDbHdr ->
             % We tried to swap out the compaction but there were
@@ -1105,8 +1106,9 @@ open_compaction_files(OldDb) ->
             #comp_st{
                 old_db = OldDb,
                 new_db = Db1,
+                new_id_tree = Db0#db.id_tree,
                 meta_fd = MetaFd,
-                retry = Db0#db.id_tree
+                retry = true
             };
         _ ->
             % We're starting a compaction from scratch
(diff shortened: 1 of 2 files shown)


src/couch/src/couch_db_updater.erl
```diff
@@ -1034,56 +1041,90 @@ check_md5(_, _) -> throw(md5_mismatch).
...
+            Db1 = bind_emsort(Db0, MetaFd, A#comp_header.meta_st),
+            #comp_st{
+                old_db = OldDb,
+                new_db = Db1,
+                meta_fd = MetaFd,
+                retry = Db0#db.id_tree
+            };
         _ when DataHdrIsDbHdr ->
+            % We tried to swap out the compaction but there were
+            % writes to the database during compction. Start
```
ksnavely: compaction typo

src/couch/src/couch_db_updater.erl
```diff
@@ -1034,56 +1041,90 @@ check_md5(_, _) -> throw(md5_mismatch).
...
-        open_compaction_files(Name, Header, Filepath, Options),
-    erlang:monitor(process, MFd),
+    FinalCompSt = lists:foldl(fun(Stage, CompSt) ->
+        Stage(CompSt)
+    end, InitCompSt, Stages),
 
-    % This is a bit worrisome. init_db/4 will monitor the data fd
-    % but it doesn't know about the meta fd. For now I'll maintain
-    % that the data fd is the old normal fd and meta fd is special
-    % and hope everything works out for the best.
```
ksnavely: Is the concern over the lack of a monitor on the meta file descriptor still valid? Or is it moot with the `couch_file:close(MetaFd)` call now?

ksnavely: Or is it this? https://github.com/apache/couchdb/pull/754/commits/c4969f3916f52554179b69dda245f14feb1e19db#diff-c0abd721a4433ae0399b0c2cbdea8d18R1126

src/couch/src/couch_db_updater.erl
```diff
@@ -1216,7 +1218,7 @@ copy_compact(#comp_st{} = CompSt) ->
         {changes_done, 0},
         {total_changes, TotalChanges}
     ],
-    case (Retry /= nil) and couch_task_status:is_task_added() of
+    case (Retry) and couch_task_status:is_task_added() of
```
ksnavely: unnecessary parens

src/couch/src/couch_db_updater.erl
```diff
@@ -1258,6 +1313,7 @@ copy_docs(Db, #db{fd = DestFd} = NewDb, MixedInfos, Retry) ->
         % first pass. This means we may have docs that already exist
         % in the seq_tree in the .data file. Here we lookup any old
         % update_seqs so that they can be removed.
+        OldDocIdTree =
```
kocolosk: I'm a little rusty but this looks like a refactoring mistake ...

src/couch/src/couch_db_updater.erl
davisp: Yep, it got moved to clean up start_compact. That comment is way old me. I know how couch_file works better now so the comment can go as its misleading.

src/couch/src/couch_db_updater.erl
davisp: Totes was. Got mixed in when separating commits.

davisp:
Closing in favor of #778

PR #759, opened 2017-08-18
add erlang 20 for test

getong:


wohali:
We know that CouchDB fails on Erlang 20. And we no longer allow merges to master that would cause our builds to fail.

As such I am rejecting this PR until proper fixes can be made to CouchDB to make it work on 20 (as well as still work on 16B03!)

PR #762, opened 2017-08-18
Use fully qualified call to allow code upgrade

iilyak:
## Overview

`rexi_monitor:wait_monitor/1` function is called recursively. This means
that hot code upgrade will not work when we are inside that
function. Use fully qualified call to allow the update of a module.

## Testing recommendations

1. Add `-export([wait_monitors/1]).`
2. Run following snippet
    ```
    Self = self(), spawn(fun() -> rexi_monitor:wait_monitors(Self) end).
    code:soft_purge(rexi_monitor).
    ```
3. Without this change the test ^^ would always return false.
## Checklist

### DIFF
--- a/src/rexi/src/rexi_monitor.erl
+++ b/src/rexi/src/rexi_monitor.erl
@@ -12,6 +12,7 @@
 
 -module(rexi_monitor).
 -export([start/1, stop/1]).
+-export([wait_monitors/1]).
 
 
 %% @doc spawn_links a process which monitors the supplied list of items and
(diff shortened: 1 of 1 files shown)


jaydoane:
Curiously, I'm not seeing the expected behavior described under **Testing recommendations**.

I modified rexi_monitor.erl to export `wait_monitors/1`, and then ran the test:
```erlang
(node1@127.0.0.1)17> Self = self(), spawn(fun() -> rexi_monitor:wait_monitors(Self) end).
<0.6576.228>
(node1@127.0.0.1)18> code:soft_purge(rexi_monitor).
true
```
Nevertheless, this PR LGTM
+1

Issue #757, opened 2017-08-17
Race condition in couch_mrview_ddoc_updated_tets


davisp:
I've seen this assertion fail locally:

https://github.com/apache/couchdb/blob/master/src/couch_mrview/test/couch_mrview_ddoc_updated_tests.erl#L106

The reason is that it appears the query succeeds before the design doc update message kills the indexer. We should add a meck synchronization point so that we can ensure that the process is always alive when the message is sent.

mayya-sharipova:
PR for this issue was +1ed, but one mango test fails in Travis. This test will be fixed separately by @tonysun83

eiri:
Fixed by #764

eiri:
Fixed by #764

PR #764, opened 2017-08-21
Avoid race condition in couch_mrview_ddoc_updated_tests

mayya-sharipova:
issue #757

## Testing recommendations
`make check apps=couch_mrview tests=ddoc_update_test_`

## GitHub issue number
757

## Checklist

### DIFF
--- a/src/couch_mrview/test/couch_mrview_ddoc_updated_tests.erl
+++ b/src/couch_mrview/test/couch_mrview_ddoc_updated_tests.erl
@@ -28,7 +28,6 @@ setup() ->
             {<<"baz">>, {[
                 {<<"map">>, <<
                     "function(doc) {\n"
-                    "    var i = 0; while(i<1000){ i++ };\n"
                     "    emit(doc.val, doc.val);\n"
                     "}"
                 >>}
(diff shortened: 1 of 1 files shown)


src/couch_mrview/test/couch_mrview_ddoc_updated_tests.erl
```diff
@@ -47,12 +46,17 @@ setup() ->
     {ok, _} =
         couch_mrview:query_view(Db2, <<"_design/bar">>, <<"baz">>, [], CB, 0),
 
+    meck:expect(couch_index_updater, update, fun(Idx, Mod, IdxSt) ->
```
davisp: I *think* you need to add a line meck:new(couch_index_updater, [passthrough]) so that this expectation can use the meck:passthrough/1 function.

src/couch_mrview/test/couch_mrview_ddoc_updated_tests.erl
```diff
@@ -47,12 +46,17 @@ setup() ->
...
+        timer:sleep(5000), meck:passthrough([Idx, Mod, IdxSt])
```
davisp: Put meck:passthrough on its own line.

src/couch_mrview/test/couch_mrview_ddoc_updated_tests.erl
davisp: Granted in this particular case the point is to never hit the passthrough call but we should try and be correct in our usage of meck or the bad patterns might get copied to other tests if this is used as a reference.

davisp:
+1

PR #765, opened 2017-08-21
include mrview options in _explain result

willholley:
## Overview

`_explain` previously returned the options passed in by the user but
not those modified at execution time by Mango. Now we include
index-specific options (mrargs for map/reduce indexes) in the output,
allowing us to see e.g. when include_docs was used.

## Testing recommendations

Manually test by making a request to the `_explain` endpoint from the command line / Fauxton. When a query is fulfilled by a `json` index, you should see an `mrargs` field in the output.

Unit tests should provide good coverage.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -56,18 +56,31 @@ create(Db, Indexes, Selector, Opts) ->
 
 explain(Cursor) ->
     #cursor{
-        index = Idx,
-        ranges = Ranges
+        opts = Opts
     } = Cursor,
-    case Ranges of
-        [empty] ->
-            [{range, empty}];
-        _ ->
-            [{range, {[
-            {start_key, mango_idx:start_key(Idx, Ranges)},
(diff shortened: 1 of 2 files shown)


src/mango/src/mango_cursor_view.erl
```diff
@@ -56,18 +56,31 @@ create(Db, Indexes, Selector, Opts) ->
...
-    case Ranges of
-        [empty] ->
-            [{range, empty}];
-        _ ->
-            [{range, {[
-            {start_key, mango_idx:start_key(Idx, Ranges)},
-            {end_key, mango_idx:end_key(Idx, Ranges)}
-        ]}}]
-    end.
+
```
iilyak: Previous implementation returned `range` property as well. We don't return it anymore. Would it be a problem for fauxton?

willholley: fauxton just displays the raw JSON result from `_explain`, so it won't be a problem. The `range` property is now part of the `mrargs` object in the result.

PR #778, opened 2017-08-25
Compactor id tree optimization

davisp:
## Overview

This adds a new optimization to the compactor to read document ids from the `OldDb#db.id_tree` during the first pass of compaction. This saves us a large amount of I/O and CPU since we aren't writing out the emsort structure. However, after the first attempt at swapping the compaction file we have to fall back to using emsort to maintain our ability to do incremental compactions if there were writes during compaction.

## Testing recommendations

make eunit apps=couch suites=couch_db_updater_tests

## Related Pull Requests

Original PR #754 superseded by this PR cause this one includes a fairly comprehensive test suite.

## Checklist

### DIFF
--- a/src/couch/src/couch_db_updater.erl
+++ b/src/couch/src/couch_db_updater.erl
@@ -26,14 +26,16 @@
 -record(comp_st, {
     old_db,
     new_db,
-    new_id_tree,
-    meta_fd,
+    docid_fd,
+    docid_mod,
+    docid_st,
     retry
 }).
 
 -record(comp_header, {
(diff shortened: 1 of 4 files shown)


davisp:
Closing this in favor of https://github.com/apache/couchdb/pull/806 which is significantly faster than this approach. Algorithmically they're equal as O(N log N) (if my analysis is correct) but the extra calls through couch_file for this approach (via btree lookups against seq_tree) means that we end up swamping any gains we made by not copying full_doc_info records in the meta file.

PR #786, opened 2017-09-01
Run mango tests with make check

wohali:
This will force everyone to run the Mango tests with the default
`make check` command, which will also enable it to be run as part
of the Jenkins tests as well. (We were already running these test
in Travis.)

Be forewarned: some of the Mango tests are not reliable and will
need fixup and support soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ before_script:
   - cd ../..
 
 script:
-   - make check mango-test
+   - make check
 
 after_failure:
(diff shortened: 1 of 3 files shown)


iilyak:
This would introduce a mandatory dependency on nosetest. Would you mind to copy the information about dependencies from `https://github.com/apache/couchdb/blob/master/src/mango/test/README.md` to `https://github.com/apache/couchdb/blob/master/README-DEV.rst`?

PR #791, opened 2017-09-05
Check all indexed fields are in the selector

garrensmith:
Each mango json index is a filtered index, meaning that it only adds
documents to the index that have all the fields that are defined in the
index. So when selecing an index for a query we need to make sure that
all the fields defined in the index are also in the selector. Otherwise the
query could return incorrect results.

An example:
For documents:
```
{
    name: "Garren",
    location: "South Africa"
},
{
    name: "Ed",
}
```
and with an index: `["name, "location"]. If we have a query like this:

```
selector: {name: {$gte: "E"}}
```
If it uses the above defined index it will only return the first document. Not both of them.
This PR changes it so that it will not use that index and rather use another index if there is or use _all_docs.

## Checklist

### DIFF
--- a/src/mango/src/mango_idx_view.erl
+++ b/src/mango/src/mango_idx_view.erl
@@ -115,10 +115,24 @@ columns(Idx) ->
 
 is_usable(Idx, Selector) ->
     % This index is usable if at least the first column is
-    % a member of the indexable fields of the selector.
+    % a member of the indexable fields of the selector and all
+    % the fields in the index are in the selector
     Columns = columns(Idx),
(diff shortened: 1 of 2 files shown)


willholley:
JSON indexes only index documents where all the fields are non-null. I think this PR needs to account for `null` comparisons when matching the fields in the selector against the index. For example, if I have an index on:

```
["user_id", "age"]
```

and a selector `{"user_id": {"$lt":8}, "age": { "$gte": null }}` then the index would not be valid, even though the fields in the selector match the index, because using the index would skip any documents where `age` is null.

PR #797, opened 2017-09-07
Return reduce overflow errors to the client

davisp:
This changes the reduce overflow error to return an error to the client
rather than blowing up the view build. This allows views that have a
single bad reduce to build while not crushing the server's RAM usage.

## Testing recommendations

```make javascript suites=view_errors```

## Checklist


tonysun83:
I noticed in ```fabric_util``` we have an error condition for this error, and also in ```view_errors``` we're testing for this situation. Was this error not propagated correctly before and thus the view builds blew up?

davisp:
@tonysun83 My bad! Forgot to push after I found the view_errors.js test failure. I hadn't noticed the fabric_util clause but I've removed it since it'd be dead code now.

tonysun83:
+1

janl:
rebased on latest master, can be merged after travis

PR #804, opened 2017-09-11
Fix Mango error handling

willholley:
## Overview

Fixes a regression where a 500 status code was returned when  no index is available to service a _find query because the  sort order does not match any available indexes. This was actually a general issue with _find errors, introduced when index selection moved from `mango_cursor.erl` to `mango_idx.erl` in https://github.com/apache/couchdb-mango/commit/01252f971bef0c8da1d78bf5a7b506b71926ce1b.

In a separate commit, I converted the index selection tests to use `unittest` assertion functions and addressed an issue with a duplicate test name.

## Testing recommendations

Changes covered by unit tests

## Checklist

### DIFF
--- a/src/mango/src/mango_error.erl
+++ b/src/mango/src/mango_error.erl
@@ -21,19 +21,19 @@
 ]).
 
 
-info(mango_cursor, {no_usable_index, no_indexes_defined}) ->
+info(mango_idx, {no_usable_index, no_indexes_defined}) ->
     {
         400,
         <<"no_usable_index">>,
         <<"There are no indexes defined in this database.">>
     };
-info(mango_cursor, {no_usable_index, no_index_matching_name}) ->
+info(mango_idx, {no_usable_index, no_index_matching_name}) ->
     {
         400,
         <<"no_usable_index">>,
         <<"No index matches the index specified with \"use_index\"">>
     };
-info(mango_cursor, {no_usable_index, missing_sort_index}) ->
+info(mango_idx, {no_usable_index, missing_sort_index}) ->
     {
         400,
         <<"no_usable_index">>,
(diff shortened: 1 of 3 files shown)


eiri:
What's the reasoning for complete removal of module name match? As far as I can see `no_usable_index` errors still thrown from a single and not multiple modules and as a matter of fact `{no_usable_index, selector_unsupported}` kept in `mango_cursor` and doesn't require changes in error module

willholley:
@eiri I didn't see the benefit of scoping those error message conversions to a single module; we're not likely to change the status code or message based on the module that raised the error and it makes the error conversion more brittle (as illustrated by the change that led to this regression). Arguably this logic applies to all of mango_error.erl - perhaps this PR should stick to the existing convention and use a separate PR to discuss whether to remove the module scope from error conversion?

eiri:
I don't know for sure, but my guess that was a kind of [a rudimentary traceback](https://github.com/willholley/couchdb/commit/bbe0f0d87c6e08760a214b0b9d8d1a033a41c269#diff-f7c867d23c179319495d65d8f6203df7R20) that got lost in a refactoring.

Yes, as I see it you here mixing fixes for index selection refactoring with mango_error's signature changes, which I'm not against of, but would prefer to be done in more organized fashion, rather than the scope creep. If that's all right with you let's do that in separate PR for sake of accountability and simplified revert, if that'd be necessary.

willholley:
@eiri Interesting - I wonder why the logging was removed. In any case, I've reverted to the existing pattern for now and we can consider the error signature changes as a separate PR.

eiri:
Judging by minimalistic description of [commit where it was done](https://github.com/apache/couchdb/commit/422fdcdbeedb818e1e69478b7f1f492cdc71c15c#diff-f7c867d23c179319495d65d8f6203df7L19) - in heat of refactoring :)

Cool, thank you for the change! +1

PR #807, opened 2017-09-12
Avoid duplicate index selection in Mango

willholley:
## Overview

Previously, index selection for a given query was run twice for each request - once to add
a warning in case a full database scan would be performed and then again when the query was executed.

This moves the warning generation so that it occurs at the end of the query processing and we can use the existing index context to decide whether to add a warning or not.

Whilst only a minor optimisation (which also assumes we don't have cached query plans etc), it
at least moves index selection to where one might expect it to happen i.e. during query planning.

## Testing recommendations

Added a unit test to cover that a warning is appropriately generated. The text index tests should also be run manually.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -17,13 +17,15 @@
     create/3,
     explain/1,
     execute/3,
-    maybe_filter_indexes/2
+    maybe_filter_indexes/2,
+    maybe_add_warning/3
 ]).
 
 
 -include_lib("couch/include/couch_db.hrl").
 -include("mango.hrl").
 -include("mango_cursor.hrl").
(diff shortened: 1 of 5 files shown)


garrensmith:
I like the idea of this. I don't really like the fact that mango_cursor is calling the http module to send the warning. I dont think the mango_cursor module should know about the http layer. It might be better to do this via the `handle_doc({row})` callback.

willholley:
@garrensmith I kinda agree. However, it's no different to the pattern used by bookmarks or execution stats (i.e. JSON serialisation happens in the `maybe_` function rather than in httpd). Would you be more comfortable if the `maybe_add_warning` lived in a different module e.g. `mango_idx`?

PR #808, opened 2017-09-12
Add selector support for json indexes

garrensmith:
Adds selector support to json indexes. The selector can be used to
filter what documents are added to the index. When executing a query
the index will only be used if the index is specified in the use_index
field.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -90,7 +90,7 @@ execute(#cursor{index=Idx}=Cursor, UserFun, UserAcc) ->
 maybe_filter_indexes(Indexes, Opts) ->
     case lists:keyfind(use_index, 1, Opts) of
         {use_index, []} ->
-            Indexes;
+            remove_indexes_with_selector(Indexes);
         {use_index, [DesignId]} ->
             filter_indexes(Indexes, DesignId);
         {use_index, [DesignId, ViewName]} ->
@@ -112,6 +112,16 @@ filter_indexes(Indexes, DesignId0) ->
 filter_indexes(Indexes0, DesignId, ViewName) ->
     Indexes = filter_indexes(Indexes0, DesignId),
     FiltFun = fun(I) -> mango_idx:name(I) == ViewName end,
+    lists:filter(FiltFun, Indexes).
+
+
+remove_indexes_with_selector(Indexes) ->
+    FiltFun = fun(Idx) -> 
+        case mango_idx:get_idx_selector(Idx) of
+            undefined -> true;
+            _ -> false
+        end
+    end,
     lists:filter(FiltFun, Indexes).
 
 
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -17,7 +17,7 @@
     create/3,
     explain/1,
     execute/3,
-    maybe_filter_indexes/2,
+    maybe_filter_indexes_by_ddoc/2,
     maybe_add_warning/3
 ]).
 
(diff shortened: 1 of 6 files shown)


src/mango/src/mango_native_proc.erl
```diff
@@ -150,6 +148,22 @@ get_index_entries({IdxProps}, Doc) ->
...
 
+get_index_values(Fields, Selector, Doc) ->
+    lists:map(fun({Field, _Dir}) ->
+        case mango_doc:get_field(Doc, Field) of
+            not_found -> 
+                not_found;
+            bad_path -> 
+                not_found;
+            Value -> 
+                case should_index(Selector, Doc) of
```
willholley: this looks like it could be more efficient - we only need to check whether the doc should be indexed once and, if not, return immediately.

src/mango/test/16-index-selectors.py
```diff
@@ -0,0 +1,115 @@
...
+        docs = self.db.find(selector2, use_index='Selected')
+        self.assertEqual(len(docs), 3)
+
+    def test_doesnot_use_selector_when_not_specified(self):
+        selector = {"location": {"$gte": "FRA"}}
+        self.db.create_index(["location"], selector=selector, ddoc="Selected", name="Selected")
+        resp = self.db.find(selector, explain=True)
+        self.assertEqual(resp["index"]["name"], "_all_docs")
+
+    def test_doesnot_use_selector_when_not_specified_with_index(self):
```
willholley: I think we need to test that specifying an invalid partial index fails appropriately. This should follow the same rules as global indexes i.e. an index cannot be used if it does not map to the selector and sort order specified in the query.

willholley: Final thought it that we should run these same tests for partial text indexes to verify consistent behaviour - perhaps that's a separate PR though.

src/mango/test/mango.py
```diff
@@ -96,8 +96,11 @@ def create_index(self, fields, idx_type="json", name=None, ddoc=None):
             body["name"] = name
         if ddoc is not None:
             body["ddoc"] = ddoc
+        if selector is not None:
+            body["index"]["selector"] = selector
         body = json.dumps(body)
         r = self.sess.post(self.path("_index"), data=body)
+        print r.json()
```
willholley: stray debugging statement?

src/mango/test/mango.py
```diff
@@ -180,6 +183,7 @@ def find(self, selector, limit=25, skip=0, sort=None, fields=None,
         else:
             path = self.path("_find")
         r = self.sess.post(path, data=body)
+        print r.json()
```
willholley: stray debugging statement..

src/mango/src/mango_cursor.erl
```diff
@@ -90,7 +90,7 @@ execute(#cursor{index=Idx}=Cursor, UserFun, UserAcc) ->
 maybe_filter_indexes(Indexes, Opts) ->
```
garrensmith: change to `maybe_filter_on_use_index`

src/mango/src/mango_cursor.erl
```diff
@@ -115,6 +115,16 @@ filter_indexes(Indexes0, DesignId, ViewName) ->
     lists:filter(FiltFun, Indexes).
 
 
+remove_indexes_with_selector(Indexes) ->
```
garrensmith: add comment here around why we removing these indexes

src/mango/src/mango_native_proc.erl
```diff
@@ -135,19 +135,30 @@ index_doc(#st{indexes=Indexes}, Doc) ->
 
 get_index_entries({IdxProps}, Doc) ->
     {Fields} = couch_util:get_value(<<"fields">>, IdxProps),
-    Values = lists:map(fun({Field, _Dir}) ->
+    Selector = case couch_util:get_value(<<"selector">>, IdxProps) of
+        [] -> {[]};
+        Else -> Else
+    end,
+    case should_index(Selector, Doc) of
```
willholley: perhaps consider factoring out the duplication here with `get_text_entries`

PR #814, opened 2017-09-13
Don't reset logs when JS tests restart the server

davisp:
When a JS test requested a restart server we would wip the current log
file. This makes it hard to debug failing tests occasionally when they
happen just after a restart. This change prevents just opens log files
in read/write mode specifically when a test requests a server restart.

The current behavior for interactive use of `dev/run` will continue to
truncate log files on startup.


wohali:
nice, +1

PR #816, opened 2017-09-14
Fix mango json index selection

willholley:
## Overview

JSON index selection in Mango previously deemed an index to be usable if a range scan on the first component of its compound key could be generated from the query selector.

For instance, if I had an index:

[A, B]

is_usable would return true for the selector:

{"A": "foo"}

This is incorrect because JSON indexes only index documents that contain all the fields in the index; null values are ok, but the field must be present. That means that for the above selector, the index would implicitly include only documents where B exists, missing documents where {"A":5} matched but field B was not present.

This commit changes is_usable so that it only returns true if all the keys in the index are required to exist by the selector. This means that in the worst case e.g. none of the predicates can be used to generate a range query, we should end up performing a full index scan, but this is still more efficient
than a full database scan.

We leave the generation of the optimal range for a given index as a separate exercise - currently this happens after index selection.

Potentially we'd want to score indexes during index selection based on their ability to restrict the result set, etc.

## Testing recommendations

## Checklist

### DIFF
--- a/src/mango/src/mango_idx_view.erl
+++ b/src/mango/src/mango_idx_view.erl
@@ -114,11 +114,12 @@ columns(Idx) ->
 
 
 is_usable(Idx, Selector) ->
-    % This index is usable if at least the first column is
-    % a member of the indexable fields of the selector.
-    Columns = columns(Idx),
-    Fields = indexable_fields(Selector),
-    lists:member(hd(Columns), Fields) andalso not is_text_search(Selector).
+    % This index is usable if all of the columns are 
+    % restricted by the selector such that they are required to exist
+    % and the selector is not a text search (so requires a text index)
+    RequiredFields = columns(Idx),
+    mango_selector:has_required_fields(Selector, RequiredFields)
+        andalso not is_text_search(Selector).
 
 
 is_text_search({[]}) ->
--- a/src/mango/src/mango_selector.erl
+++ b/src/mango/src/mango_selector.erl
@@ -15,7 +15,8 @@
 
 -export([
     normalize/1,
-    match/2
+    match/2,
+    has_required_fields/2
 ]).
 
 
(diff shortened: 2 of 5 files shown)


willholley:
@davisp thanks for the review! I'll have a think about the naming and overloaded function issues.

To your last point, my thought there is that all predicates except, by definition, `"$exists":false`, "operate" on the index because selection of the index restricts the result set to documents where the indexed fields exist.

willholley:
@davisp I went with your second suggestion and implemented a `has_required_fields` function

src/mango/src/mango_idx_view.erl
```diff
@@ -114,11 +114,11 @@ columns(Idx) ->
...
-    % This index is usable if at least the first column is
-    % a member of the indexable fields of the selector.
-    Columns = columns(Idx),
-    Fields = indexable_fields(Selector),
-    lists:member(hd(Columns), Fields) andalso not is_text_search(Selector).
+    % This index is usable if all of the columns are 
+    % restricted by the selector such that they are required to exist
+    % and the selector is not a text search (so requires a text index)
+    RequiredFields = columns(Idx),
+    mango_selector:has_required_fields(Selector, RequiredFields) andalso not is_text_search(Selector).
```
davisp: Line too long. Break before andalso and indent it twice.

src/mango/src/mango_selector.erl
```diff
@@ -566,3 +567,111 @@ match({[{Field, Cond}]}, Value, Cmp) ->
 
 match({[_, _ | _] = _Props} = Sel, _Value, _Cmp) ->
     erlang:error({unnormalized_selector, Sel}).
+
+
+
```
davisp: One too many blank lines.

src/mango/src/mango_selector.erl
```diff
@@ -566,3 +567,111 @@ match({[{Field, Cond}]}, Value, Cmp) ->
...
+% Returns true if Selector requires all  
+% fields in RequiredFields to exist in any matching documents.
+
+% For each condition in the selector, check
+% whether the field is in RequiredFields.
+% If it is, remove it from RequiredFields and continue
+% until we match then all or run out of selector to
+% match against.
+
+% empty selector
```
davisp: Capitalize empty

src/mango/src/mango_selector.erl
```diff
@@ -566,3 +567,111 @@ match({[{Field, Cond}]}, Value, Cmp) ->
...
+has_required_fields({[]}, _) ->
+    false;
+
+has_required_fields(Selector, RequiredFields) when not is_list(Selector) ->
+    has_required_fields([Selector], RequiredFields);
+
+% "see" through $and operator. We ignore other
```
davisp: The first sentence isn't a sentence. "We can" as a prefix would be good enough.

src/mango/src/mango_selector.erl
```diff
@@ -566,3 +567,111 @@ match({[{Field, Cond}]}, Value, Cmp) ->
...
+% combination operators because they can't be used to restrict
+% an index.
+has_required_fields([{[{<<"$and">>, Args}]}], RequiredFields) when is_list(Args) ->
```
davisp: Line too long. Break before when and indent it twice.

src/mango/src/mango_selector.erl
```diff
@@ -566,3 +567,111 @@ match({[{Field, Cond}]}, Value, Cmp) ->
...
+    case Cond of
+        % $exists:false is a special case - this is the only operator
+        % that explicitly does not require a field to exist
+        {[{<<"$exists">>, false}]} ->
+            has_required_fields(Rest, RequiredFields);
+        _ ->
+            has_required_fields(Rest, lists:delete(Field, RequiredFields))
+    end;
+
+% no more required fields
```
davisp: Capitalize No

src/mango/src/mango_selector.erl
```diff
@@ -566,3 +567,111 @@ match({[{Field, Cond}]}, Value, Cmp) ->
...
+has_required_fields(_, []) ->
```
davisp: Its general practice to place your recursion terminating conditions as the first function clauses.

src/mango/src/mango_selector.erl
```diff
@@ -566,3 +567,111 @@ match({[{Field, Cond}]}, Value, Cmp) ->
...
+    true;
+
+% no more selector
```
davisp: Capitalize No and move to before the recursive call.

src/mango/src/mango_selector.erl
```diff
@@ -566,3 +567,111 @@ match({[{Field, Cond}]}, Value, Cmp) ->
...
+has_required_fields([], _) ->
+    false.
+
+    
+
+%%%%%%%% module tests below %%%%%%%%
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
```
davisp: Good test cases!

src/mango/test/05-index-selection-test.py
```diff
@@ -14,7 +14,6 @@
 import user_docs
 import unittest
 
-
```
davisp: Don't include white space only changes in PRs that aren't stylistic changes.

src/mango/test/05-index-selection-test.py
```diff
@@ -92,6 +91,80 @@ def test_invalid_use_index(self):
...
+        }
+        docs = self.db.find(selector)
+        self.assertEqual(len(docs), 1)
+        self.assertEqual(docs[0]["company"], "Pharmex")
+        self.assertNotIn("manager", docs[0])
+        
+        resp_explain = self.db.find(selector, explain=True)
+        self.assertEqual(resp_explain["index"]["type"], "special")
+
+
```
davisp: Too many blank lines between class methods here and in a few places below.

PR #818, opened 2017-09-14
Rename selector to partialfilterselector in indexes

garrensmith:
To make it easier to distinguish between a selector in _find and a
selector in _index. Rename the selector in the _index to
partialfilterselector. It also gives a bit more of an explanation of
what this selector does.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -90,9 +90,9 @@ execute(#cursor{index=Idx}=Cursor, UserFun, UserAcc) ->
 maybe_filter_indexes_by_ddoc(Indexes, Opts) ->
     case lists:keyfind(use_index, 1, Opts) of
         {use_index, []} ->
-            %We remove any indexes that have a selector 
+            % We remove any indexes that have a selector 
             % since they are only used when specified via use_index
-            remove_indexes_with_selector(Indexes);
+            remove_indexes_with_partial_filter_selector(Indexes);
         {use_index, [DesignId]} ->
             filter_indexes(Indexes, DesignId);
         {use_index, [DesignId, ViewName]} ->
@@ -117,9 +117,9 @@ filter_indexes(Indexes0, DesignId, ViewName) ->
     lists:filter(FiltFun, Indexes).
 
 
-remove_indexes_with_selector(Indexes) ->
+remove_indexes_with_partial_filter_selector(Indexes) ->
     FiltFun = fun(Idx) -> 
-        case mango_idx:get_idx_selector(Idx) of
+        case mango_idx:get_partial_filter_selector(Idx) of
             undefined -> true;
             _ -> false
         end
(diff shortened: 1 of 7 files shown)


willholley:
It needs rebasing on master but basically looks good to me. What happens if both `selector` and `partialfilterselector` are specified?

garrensmith:
@willholley I've just rebased

willholley:
@garrensmith I think we should follow the convention for other fields and use `_` as a delimiter - i.e. `partial_filter_selector` instead of `partialfilterselector`.

src/mango/src/mango_cursor.erl
```diff
@@ -92,7 +92,7 @@ maybe_filter_indexes_by_ddoc(Indexes, Opts) ->
         {use_index, []} ->
             %We remove any indexes that have a selector
```
willholley: probably add a space here

src/mango/src/mango_idx.erl
```diff
@@ -44,7 +44,7 @@
     to_json/1,
     delete/4,
     get_usable_indexes/3,
-    get_idx_selector/1
+    get_idx_partial_filter_selector/1
```
willholley: rename to `get_partial_filter_selector` - that it is in the `mango_idx` module implies that it relates to an index

src/mango/src/mango_idx.erl
```diff
@@ -139,7 +139,25 @@ new(Db, Opts) ->
 
 validate_new(Idx, Db) ->
     Mod = idx_mod(Idx),
-    Mod:validate_new(Idx, Db).
+    Def1 = maybe_rename_idx_fields(Idx#idx.def),
+    Idx1 = Idx#idx{def=Def1},
+    Mod:validate_new(Idx1, Db).
+
+
+maybe_rename_idx_fields({Props}) ->
```
willholley: rename to `maybe_rename_fields`

src/mango/test/03-operator-test.py
```diff
@@ -185,7 +185,7 @@ def test_lt_includes_null_but_not_missing(self):
 
     def test_lte_includes_null_but_not_missing(self):
         docs = self.db.find({
-                "twitter": {"$lt": 1}
+                "twitter": {"$lte": 1}
```
willholley: move this into a separate PR as it's unrelated

willholley:
after squashing, looks ok to me

garrensmith:
@davisp I've made all the required changes. Could you take a look when you get a chance.

src/mango/src/mango_idx_view.erl
```diff
@@ -199,8 +199,8 @@ opts() ->
             {tag, fields},
             {validator, fun mango_opts:validate_sort/1}
         ]},
-        {<<"selector">>, [
-            {tag, selector},
```
davisp: I see the get_legacy_selector code above, but reading through code I can't convince myself that removing this option definition won't break old indexes that used the selector name. I wonder if we shouldn't continue to support this for awhile and upgrade it internally for new indexes?

davisp: And/or perhaps just adding a test specifically for the upgrade would be enough here. And for that I'd like to see us write an old style ddoc directly and then read the index to make sure it behaves as expected.

PR #827, opened 2017-09-19
Support setting cookie domain for auth

almightyju:
Implementation for https://issues.apache.org/jira/browse/COUCHDB-1959 

It's a very simple use where you can set cookie_domain in the config which gets put on the set-cookie header.

I couldn't work out where the documentation is to update that and I'm not familiar enough with Erlang to write a test for it but a manual test with and without the config value works fine. If anyone wants to point out how to make a test for it I'm happy enough to put it in if desired.


wohali:
Before merging this we'll definitely want to see a test for this.

Can you confirm that this does the right thing on the clustered port (5984)?

almightyju:
I've been using this build myself for dev since making the pull request and it all works as expected on port 5984. 

With the tests I can see how the login is handled easy enough but I can't work out how the server is started since I guess it would have to be restarted with different a config to test the domain gets set on the cookie correctly.

wohali:
Hi @almightyju , I'd like to push forward with this. I also want you to get the credit for the change!

You don't have to restart the server with a different config, just change it as part of your `setup()` function, and set it back to the default value in your `teardown()` function. Here's an example:

https://github.com/apache/couchdb/blob/master/src/couch/test/couchdb_mrview_cors_tests.erl#L32-L57

If you need more guidance, I'm happy to work together with you on this, just let me know.

janl:
Customer update:

> I just want to give a small update.

> So we changed now the service URLs to have the CouchDB and the Auth-service on the same domain. (cloud.int.pscc.local/database -> couchdb, cloud.int.pscc.local/cas -> Auth-service)
We woud still prefer a solution where this services are behind different sub domains. But if we do not find a better solution soon, we will stay with the current one.

So they are unblocked on this for now, I told them about the timeout issue and the PR.

almightyju:
@wohali, finally got the tests done, I really do dislike erlang :/

Anyway, new tests are passing for me, not confident about the need for clustered and backdoor for this so I've only tested against the clustered port.

wohali:
Only clustered matters here, the node-local port is deprecated for all communications other than essential system admin tasks.

Looks good! Assuming the Travis changes pass, we can merge this PR.

wohali:
Thanks again @almightyju - great job, and congrats on your first-time contribution to Apache CouchDB!

PR #837, opened 2017-09-25
Add cluster info to database information

### LINKED ISSUE #812: Return `q` value a database was created with when accessing `/dbname`
`<+Wohali>	i really wish GET /dbname would return the db's q value `

## Expected Behavior

`GET /dbname` should include the `q` value the database `dbname` was created with.

## Current Behavior

`q` is not part of the db info.

## Context

This would help with debugging, operations, performance analysis.

## Your Environment

All.

eiri:
## Overview

The patch adds a "cluster" block into a database info with values for database's `q`, `n` and default `w` and `r`. This enhancement can help with debugging, operations and performance analysis.

## Testing recommendations

Changes specific tests:
`make eunit apps=fabric suites=fabric_db_info`

## GitHub issue number

Fixes #812 

## Checklist

### DIFF
--- a/src/fabric/src/fabric_db_info.erl
+++ b/src/fabric/src/fabric_db_info.erl
@@ -22,7 +22,8 @@ go(DbName) ->
     Workers = fabric_util:submit_jobs(Shards, get_db_info, []),
     RexiMon = fabric_util:create_monitors(Shards),
     Fun = fun handle_message/3,
-    Acc0 = {fabric_dict:init(Workers, nil), []},
+    {ok, ClusterInfo} = get_cluster_info(Shards),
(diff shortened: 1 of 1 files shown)


eiri:
Note that there are already `/{db}/_shards` and `/{db}/_shards/{id}` that provide more detailed information on sharding and quorum, so information for `q` and `n` could be gathered from there.

davisp:
+1

Though do we need to update docs as well?

eiri:
@davisp Good idea. I've opened apache/couchdb-documentation#164 for update on a cluster config and I guess a section about `_db_info` will need to be updated after this change end up in a release?

Issue #834, opened 2017-09-22
Mango python test suite doesn't run under Python 3


wohali:
https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/82/pipeline

Our CentOS 6 build runs under Python 3, because it's annoying to get an up-to-date Python 2 on CentOS 6 but not annoying to get a Python 3.

There are a number of syntax errors that need to be resolved pronto. master will continue to fail on Jenkins until they do.

@tonysun83 @garrensmith can you take a look here? The error log for the centos 6 build is pretty explicit as to where the errors are. `2to3` will also highlight these for you.

tonysun83:
@wohali: I'm on it.

tonysun83:
https://github.com/apache/couchdb/pull/839/

PR #839, opened 2017-09-26
Fix tests to be compatible for both python2 and python3

tonysun83:
## Overview

We make mango tests python2 and  python3 compatible.

## Testing recommendations

Tests were run successfully in python2 and python3 environments. To run nosetests with python3:

1) pyvenv myenv
2) source myenv/bin/activate
3) pip install nose
4) python3 -m "nose"

## GitHub issue number

https://github.com/apache/couchdb/issues/834

## Checklist


src/mango/test/mango.py
```diff
@@ -44,7 +44,7 @@ def url(self):
         return "http://{}:{}/{}".format(self.host, self.port, self.dbname)
 
     def path(self, parts):
-        if isinstance(parts, (str, unicode)):
+        if isinstance(parts, ("".__class__, u"".__class__)):
```
tonysun83: only part worth of note. Another way to do this is : 
```
try:
    basestring  # attempt to evaluate basestring
    def isstr(s):
        return isinstance(s, basestring)
except NameError:
    def isstr(s):
        return isinstance(s, str)
```

wohali: Cute shortcut, makes perfect sense.

wohali:
@tonysun83 Jenkins desperately needs these changes so I'm pushing the merge button. Hope you're OK with that ;)

PR #849, opened 2017-09-28
Fix external size bug when seq btree exists

tonysun83:
## Overview

When users specify "options":  {"seq_indexed": true}, we will run
reduced_external_size/1 on seq btrees. The reduce will return an
integer because the reduce function is:
couch_db_updater:btree_by_seq_reduce/2.
This leads to a bad match error. Thanks @jiangphcn for finding this.

## Testing recommendations

Unit testcase included.

## GitHub issue number

This PR is the issue.

## Checklist


src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -800,7 +800,10 @@ reduced_external_size(Tree) ->
     case couch_btree:full_reduce(Tree) of
         {ok, {_, _, Size}} -> Size;
         % return 0 for versions of the reduce function without Size
-        {ok, {_, _}} -> 0
+        {ok, {_, _}} -> 0;
+        % This is for the case when seq_btree or key_byseq_btree is not
+        % null. Reduces will return an integer.
+        {ok, Size} when is_integer(Size) -> Size
```
jiangphcn: After adding this line, we can get matched clause. However, it looks that the returned size is the number of indexed documents instead of external size for seq_indexed tree and keyseq_indexed tree.

tonysun83: Ooops, you're right. I think we may need to modify our seq_tree reduce function:
https://github.com/apache/couchdb/blob/seqt-external-size-bug/src/couch/src/couch_db_updater.erl#L566-L570

I wonder if we can get away with just using couch_btree:size for the seq_indexed trees.
Thoughts @davisp?

davisp: Depends on what you want the size of. couch_btree:size gives you the number of bytes to store the btree. Depending on how you're storing KV pairs you may have to have a custom reduce to include that info. For instance, in the id_tree we have a custom reduce that includes size information for the document bodies (since they aren't part of the KV structure of the btree) but for the seq_tree there's nothing written beyond what's in the KVs so we don't need a custom reduce.

tonysun83: @davisp:  So I think what you're saying is : for seq_btrees and keyseq_btrees, there is no additional data outside the KVs. If external size means uncompressed data that the user is going to store, then the total number of bytes to store seq_btrees is exactly that because we generated it for the user (who specified to include it). Therefore, couch_btree:size should be sufficient. 

Would want to to include the actual reduce value? so something like:
```couch_btree:size(SeqBtree) + ?term_size(reduce_external_size(SeqBtree))```

Comment (davisp) [reply src/couch_mrview/src/couch_mrview_util.erl]: Ah, no, couch_btree:size is compressed on disk size, not external size. If you're doing this for external size calculations you'll have to add that as a custom reduce function. You can take a look at the reduce function on existing views to get an idea of how it'd work:

https://github.com/apache/couchdb/blob/master/src/couch_mrview/src/couch_mrview_util.erl#L1054-L1073

**Comment by tonysun83** [comment]
@davisp @jiangphcn : does the new reduce function look right to you guys? It's almost the exact same approach as the user btree, but instead of running the user reduce and grabbing the ```?term_size``` for that value, we just do ```?term_size``` for the integer representing the number of docs. I am including the value for now, but on second thought I don't think we need it.

**Comment by jiangphcn** [comment]
@tonysun83 LGTM. I also run newly introduced eunit test case and it passed.
```
Map views
  couch_mrview_map_views_tests:68: should_map...ok
  couch_mrview_map_views_tests:82: should_map_with_range...ok
  couch_mrview_map_views_tests:96: should_map_with_limit_and_skip...ok
  couch_mrview_map_views_tests:113: should_map_with_include_docs...ok
  couch_mrview_map_views_tests:120: should_map_empty_views...ok
  couch_mrview_map_views_tests:141: should_give_ext_size_seq_indexed_test...ok
  [done in 0.669 s]
  All 6 tests passed.
```
I know that it is hard to compare actual size in eunit test case and ends up checking whether the returned size is reasonable number.

File: src/couch_mrview/src/couch_mrview_util.erl [hunk 2]
```diff
@@ -796,24 +796,30 @@ changes_ekey_opts(_StartSeq, #mrargs{end_key=EKey,
     end.
 
 
-reduced_external_size(Tree) ->
+reduced_external_size(btree, Tree) ->
```
davisp: Matching on the arg is a bit odd here. I'd make this two different functions that you call directly in calculate_eternal_size/1.

src/couch_mrview/src/couch_mrview.erl
```diff
@@ -315,14 +315,16 @@ count_view_changes_since(Db, DDoc, VName, SinceSeq, Options) ->
...
-                            {ok, N} = couch_btree:fold_reduce(
-                                    Btree, fun(_SeqStart, PartialReds, 0) ->
-                                        {ok, couch_btree:final_reduce(
-                                                    Btree, PartialReds)}
-                                    end,
-                                0, Opts),
-                            Acc0 + N
-                    end, 0, OptList);
+                case couch_btree:fold_reduce(Btree,
+                    fun(_SeqStart, PartialReds, 0) ->
```
davisp: Big narp on this formatting. Assign your anonymous function to a variable and pass that.

wohali:
Would you mind documenting what `"options": {"seq_indexed": true},` does? ;) There's no reference to it in the docs anywhere.

eiri:
@wohali `seq_indexed` turns on seq btree generation in the views used in `_view_changes`. The feature itself not exposed on API at the moment and the current implementation desire some work and fixing, so it's definitely not ready for documentation or production usage.

PR #856, opened 2017-10-01
Bypass authentication check for /_up

### LINKED ISSUE #823: Healthcheck endpoint authentication should be optional
## Expected Behavior
When setting `chttpd.require_valid_user = true`the healthcheck endpoint `/_up` should not be protected.

## Current Behavior
Setting `chttpd.require_valid_user = true` also makes `/_up` endpoint to be authenticated.

## Possible Solution
Either unprotect the `/_up` endpoint always, which doesn't sound like very dangerous as nothing but the health of the system is exposed there or make it configurable.

## Context
Currently some butt providers (Google in this case) don't allow credentials to be given to healthcheck configurations. Setting `chttpd.request_valid_user = true` is something very desirable as you enforce authentication all over the place, but you can't combine both in this case as healthchecks will fail if auth is enforced.

## Your Environment
* Version used: 2.1.0~trusty
* Browser Name and version: Doesn't apply
* Operating System and version (desktop or mobile): Ubuntu 14.04

wohali:
This will allow various automated health check systems to hit /_up
without having to provide a username/password pair when the
chttpd.require_valid_user config setting is true. Apparently, many
of these health check providers do not even allow supplying creds
for such a purpose...

Closes #823

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.pyc
 *.snap
 *.so
+*.swp
 .DS_Store
 .rebar/
 .eunit/
--- a/src/couch/src/couch_httpd_auth.erl
+++ b/src/couch/src/couch_httpd_auth.erl
@@ -85,6 +85,8 @@ basic_name_pw(Req) ->
 default_authentication_handler(Req) ->
     default_authentication_handler(Req, couch_auth_cache).
 
+default_authentication_handler(#httpd{path_parts=[<<"_up">>]}=Req, _) ->
+    Req#httpd{user_ctx=?ADMIN_USER};
 default_authentication_handler(Req, AuthModule) ->
     case basic_name_pw(Req) of
     {User, Pass} ->


wohali:
Note that I'm worried about the precedent that this sets, but I do agree that `_up` is an innocuous hole to punch through the authentication layer.

rnewson:
require_valid_user=true means every request needs auth no matter how innocuous so I’m -1 on 
principle. 

there’s a specific module that handles authorisation logic, so the patch is not quite right either. 

🙂

rnewson:
I’d remove my veto if the check is done in the right module and is separately configurable, defaulting to match require_valid_user.

wohali:
@rnewson Not sure which module you're referring to here. I _think_ the flow looks something like:

`mochiweb` → `chttpd:process_request` → `chttpd:authenticate_request` → (list of handlers, which starts with) `chttpd_auth:default_authentication_handler` → `couch_httpd_auth:default_authentication_handler`

which is why the patch ended up where it did. I can't put it in `chttpd_auth_request` because authorization happens after authentication, right?

Happy to add the extra config check, that makes perfect sense. I could even see listing the endpoints to unprotect, though that's not strictly necessary.

wohali:
from IRC:

```irc
11:41 <+rnewson> hm, yes, I see your point.
11:41 <+Wohali> if i don't punch a hole through
                couch_httpd_auth:default_authentication_handler, there's no way
                to actually get to chttpd_auth_request
11:41 <+Wohali> and chttpd_auth_request already has the right settings for _up
11:42 <+rnewson> well, that hole won't be sufficient but I understand
11:42 <+rnewson> a request with a cookie would not go through there (etc)
11:43 <+Wohali> yeah
11:43 <+Wohali> but the assumption is this is just for healthcheck services
                that won't supply any creds
11:44 <+rnewson> yes, understood.
11:44 <+rnewson> but require_valid_user=true is defined as requiring
                 authentication for every request (even _session, which is at
                 least as silly as _up)
11:44 <+Wohali> i guess i could add an entirely new {couch_httpd_auth,
                massive_security_hole_authentication_handler}
11:44 <+rnewson> so we'd need another config to punch a hole in that, rather
                 than change what require_valid_user=true means.
11:45 <+rnewson> so long as there's a config setting (default to current
                 behaviour) I'm ok with the proposed change (though not
                 necessarily the impl)
11:45 <+Wohali> ok, will look at later
```

janl:
@rnewson how do you feel about this with an off-by-default config variable? (see latest push)

PR #863, opened 2017-10-04
Fix exists false when field is explicitly defined

tonysun83:
## Overview

When users explicitly defined a text index's fields, and used $exists
with false, we tag false when constructing the query. This led to a
function clause for indexable_fields since we did not account for it.
This fix patches that up, but note that we don't care about the false
value itself since we only care about fields.

## Testing recommendations

New unit tests added to test for this particular operator

## GitHub issue number

This PR is the issue.

## Checklist

### DIFF
--- a/src/mango/test/07-text-custom-field-list-test.py
+++ b/src/mango/test/07-text-custom-field-list-test.py
@@ -12,6 +12,7 @@
 
 import mango
 import unittest
+import user_docs
 
 
 @unittest.skipUnless(mango.has_text_service(), "requires text service")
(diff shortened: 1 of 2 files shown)


src/mango/test/07-text-custom-field-list-test.py
```diff
@@ -160,3 +164,35 @@ def test_all_match(self):
...
+
+    def test_exists_object(self):
+        docs = self.db.find({"exists_object.should": {"$exists": True}})
+        assert len(docs) == 1
+        assert docs[0]["user_id"] == 11
+
+        docs = self.db.find({"exists_object.should": {"$exists": False}})
+        assert len(docs) == len(user_docs.DOCS) - 1
+        for d in docs:
+            assert d["user_id"] !=11
```
tonysun83: whitespace missing

tonysun83:
@willholley : It's worth noting that the behavior is going to be different for because of how we perceive $exists "should" work. If we indexed everything, and we queried ```{"foo": {"$exists": false}}```, then we would return  all the documents in the db that did not have "foo" as a field in the document. However, if you explicitly indexed "foo", then you would get 0 results back because you only indexed the subset of documents that contained "foo" as a field. The large (all_docs)superset is not part of the index, and therefore would not be returned. So _all_docs, explicit json index, index everything, and explicit text index would all return different results.

willholley:
@tonysun83 if a user specifies `{"$exists": false}` explicitly, we can never use an index to fulfil that predicate (you're correct that this was addressed in https://github.com/apache/couchdb/commit/503f3e34afa0a68e91258ca864e70ab18e6584e2). Can you provide an example where you think we'll trip up (or ideally, a failing test)?

As a general rule, I think we should consider it a bug whenever there is variation in results due to indexes being present, the exception being index-specific operators such as `$text`. The expectation is that a user will start with no indexes (so queries will default to _all_docs) and add indexes when necessary to improve performance or when sorting is required. Adding an index to an existing application should never change the results of queries already being used.

Another thing to note, for awareness, is that an explicit [partial index](https://github.com/apache/couchdb/pull/808) could be used to optimise an `{"$exists": false}` selector but these require users to opt-in - the query planner will never automatically select them.

tonysun83:
@willholley : I've always been champion of making the results the same across json and text indexes. If $exists: false is a special case, perhaps we default to using _all_docs as well even with a text index? No other  way to do that unless the user already indexed everything with another text index (but we have no way of knowing that).

tonysun83:
@willholley : Can you remind me of the corner cases where index-everything text indexes would return different results than _all_docs? If I recall it had to do with a field having a null value and querying for ```$ne```. If discrepancies are only query specific, then I would advocate using index-everything for ```"exists": false``` simply for the performance gain. I'll think a bit more, but index-everything should index every single document.

willholley:
@tonysun83 "index everything" does sound like a special case where `"$exists":false` could be fulfilled by the text index. Text indexes in general will behave differently to `_all_docs` or JSON indexes when range queries are used. 

For example `{ "name": {"$gt": 0}}` against a text index would only match documents where the value of `"name"` is an integer, whereas against `_all_docs` it will match documents where `"name"` is a string / object / array as well. This extends to `null` handling - text indexes do not include `null` values in range query results whereas other index types do.

I'd really like to address the difference in range handling between the index types (adding an index should never change the results of a query) but I think that needs a separate discussion / PR as it will be a breaking change to align them.

tonysun83:
@willholley : should be the same across all index types now. This is for $exists: false. we might have to check out $ne operators as well.

src/mango/test/07-text-custom-field-list-test.py
```diff
@@ -160,3 +161,52 @@ def test_all_match(self):
...
+        docs = self.db.find({"exists_object.should": {"$exists": True}})
+        assert len(docs) == 1
+        assert docs[0]["user_id"] == 11
+
+        docs = self.db.find({"exists_object.should": {"$exists": False}})
+        assert len(docs) == len(user_docs.DOCS) - 1
+        for d in docs:
+            assert d["user_id"] != 11
+
+    def test_exists_false_same_as_views(self):
```
willholley: I'm not quite clear how this test works - is there an assumption that this matches another test?

src/mango/test/07-text-custom-field-list-test.py
```diff
@@ -160,3 +161,52 @@ def test_all_match(self):
...
+    FIELDS = [
+        {"name": "exists_field", "type": "string"},
+        {"name": "exists_array.[]", "type": "string"},
+        {"name": "exists_object.should", "type": "string"},
+        {"name": "twitter", "type": "string"}
+    ]
+
+    def test_exists_field(self):
+        docs = self.db.find({"exists_field": {"$exists": True}})
+        assert len(docs) == 2
```
willholley: can you use the `self.assertEqual` function instead (e.g. https://github.com/apache/couchdb/pull/971/files#diff-ce3549b9dcbffaa7483d1f289451c786R453) as this provides much better error messages when assertions fail.

tonysun83:
Fixed the asserts.

The test you referenced is meant to return the same results as:
https://github.com/apache/couchdb/blob/master/src/mango/test/03-operator-test.py#L148

PR #866, opened 2017-10-05
Fix incorrect index selection when sort specified

willholley:
## Overview

Mango has an apparently long-standing bug whereby if an index is deemed valid for sorting a query but is not valid according to mango_idx:is_usable, the query planner would fall back to _all_docs silently, therefore ignoring the sort order specified in the query.

This reverses the index selection logic so that the list of usable indexes is generated prior to filtering them based on the sort order specified in the query.

Similar logic is applied to use_index, allowing us to generate a more specific error message when a user specifies an index which isn't valid for the current
selector.

This fix exposes that the change introduced in https://github.com/apache/couchdb/pull/816 may cause existing queries with sort fields to fail (indeed, several queries in our test suite needed fixing), so we need to consider whether this is a severe enough breaking change to warrant a major version bump.

## Testing recommendations

Run the test suite. Test that you can run Mango queries with sort criteria.

## Related Pull Requests

https://github.com/apache/couchdb/pull/816

## Checklist

### DIFF
--- a/src/mango/src/mango_idx.erl
+++ b/src/mango/src/mango_idx.erl
@@ -70,13 +70,31 @@ get_usable_indexes(Db, Selector0, Opts) ->
         ?MANGO_ERROR({no_usable_index, no_index_matching_name})
     end,
 
-    SortIndexes = mango_idx:for_sort(FilteredIndexes, Opts),
-    if SortIndexes /= [] -> ok; true ->
-        ?MANGO_ERROR({no_usable_index, missing_sort_index})
+    UsableFilter = fun(I) -> mango_idx:is_usable(I, Selector) end,
+    UsableIndexes0 = lists:filter(UsableFilter, FilteredIndexes),
+
+    UsableIndexes1 = case has_use_index(Opts) of
+        true ->
+            case mango_cursor:maybe_filter_indexes_by_ddoc(UsableIndexes0, Opts) of
+                [] -> ?MANGO_ERROR({no_usable_index, no_usable_index_matching_name});
+                UsableFilteredIndexes -> UsableFilteredIndexes
+            end;
+        false ->
+            UsableIndexes0
     end,
 
-    UsableFilter = fun(I) -> mango_idx:is_usable(I, Selector) end,
-    lists:filter(UsableFilter, SortIndexes).
+    % If a sort was specified we have to find an index that
+    % can satisfy the request.
+    case has_sort(Opts) of
+        true ->
+            case mango_idx:for_sort(UsableIndexes1, Opts) of
+                [] -> ?MANGO_ERROR({no_usable_index, missing_sort_index}); 
+                SortIndexes -> SortIndexes
+            end;
+        false ->
+            UsableIndexes1
+    end.
+
 
 recover(Db) ->
     {ok, DDocs0} = mango_util:open_ddocs(Db),
@@ -93,15 +111,29 @@ recover(Db) ->
     end, DDocs)}.
 
 
-for_sort(Indexes, Opts) ->
-    % If a sort was specified we have to find an index that
-    % can satisfy the request.
+has_use_index(Opts) ->
     case lists:keyfind(sort, 1, Opts) of
-        {sort, {SProps}} when is_list(SProps) ->
-            for_sort_int(Indexes, {SProps});
+        {use_index, _} ->
+            true;
         _ ->
-            Indexes
+            false
     end.
+
+
+has_sort(Opts) ->
+    case lists:keyfind(sort, 1, Opts) of
+        {sort, {[]}} ->
+            false;
+        {sort, {SProps}} when is_list(SProps) ->
+            true;
+        _ ->
+            false
+    end.
+
+
+for_sort(Indexes, Opts) ->
+    {sort, {SProps}} = lists:keyfind(sort, 1, Opts),
+    for_sort_int(Indexes, {SProps}).
 
 
 for_sort_int(Indexes, Sort) ->
--- a/src/mango/src/mango_idx.erl
+++ b/src/mango/src/mango_idx.erl
@@ -20,7 +20,6 @@
 -export([
     list/1,
     recover/1,
-    for_sort/2,
 
     new/2,
     validate_new/2,
@@ -36,7 +35,7 @@
     def/1,
     opts/1,
     columns/1,
-    is_usable/2,
+    is_usable/3,
     start_key/2,
     end_key/2,
     cursor_mod/1,
@@ -57,9 +56,8 @@ list(Db) ->
     {ok, Indexes} = ddoc_cache:open(db_to_name(Db), ?MODULE),
     Indexes.
 
-get_usable_indexes(Db, Selector0, Opts) ->
-    Selector = mango_selector:normalize(Selector0),
 
+get_usable_indexes(Db, Selector, Opts) ->
     ExistingIndexes = mango_idx:list(Db),
     if ExistingIndexes /= [] -> ok; true ->
         ?MANGO_ERROR({no_usable_index, no_indexes_defined})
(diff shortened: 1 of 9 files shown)


src/mango/src/mango_idx.erl
```diff
@@ -93,17 +113,33 @@ recover(Db) ->
     end, DDocs)}.
 
 
-for_sort(Indexes, Opts) ->
-    % If a sort was specified we have to find an index that
-    % can satisfy the request.
+has_use_index(Opts) ->
+    erlang:display(<<"sort opts">>),
```
garrensmith: debugging

src/mango/test/02-basic-find-test.py
```diff
@@ -222,6 +222,20 @@ def test_sort(self):
         docs2 = list(reversed(sorted(docs1, key=lambda d: d["age"])))
         assert docs1 is not docs2 and docs1 == docs2
 
+    def test_sort_desc_complex(self):
```
garrensmith: Could you add a test without the `manager: $exists: true` and make sure it returns an error

src/mango/src/mango_idx.erl
```diff
@@ -70,13 +70,33 @@ get_usable_indexes(Db, Selector0, Opts) ->
...
 
-    SortIndexes = mango_idx:for_sort(FilteredIndexes, Opts),
-    if SortIndexes /= [] -> ok; true ->
-        ?MANGO_ERROR({no_usable_index, missing_sort_index})
+    UsableFilter = fun(I) -> mango_idx:is_usable(I, Selector) end,
+    UsableIndexes0 = lists:filter(UsableFilter, FilteredIndexes),
+
+    UsableIndexes1 = case has_use_index(Opts) of
+        true ->
+            UsableFilteredIndexes = mango_cursor:maybe_filter_indexes_by_ddoc(UsableIndexes0, Opts),
```
tonysun83: maybe this can be written as:
```
case mango_cursor:maybe_filter_indexes_by_ddoc(UsableIndexes0, Opts) of
    [] -> ?MANGO_ERROR({no_usable_index, no_usable_index_matching_name});
    UsableFilteredIndexes -> UsableFilteredIndexes
end;
```

src/mango/src/mango_idx.erl
```diff
@@ -70,13 +70,33 @@ get_usable_indexes(Db, Selector0, Opts) ->
...
+            UsableIndexes0
     end,
 
-    UsableFilter = fun(I) -> mango_idx:is_usable(I, Selector) end,
-    lists:filter(UsableFilter, SortIndexes).
+    % If a sort was specified we have to find an index that
+    % can satisfy the request.
+    case has_sort(Opts) of
+        true ->
+            SortIndexes = mango_idx:for_sort(UsableIndexes1, Opts),
```
tonysun83: similar stylist change like above:
```
case mango_idx:for_sort(UsableIndexes1, Opts) of
    [] -> ?MANGO_ERROR({no_usable_index, missing_sort_index}); 
    SortIndexes -> SortIndexes
end;
```

tonysun83:
This is a tricky spot. I think the situation where the user would complain about backwards compatibility is as follows:

1) User had a sort specified on one field ```a```
2) User had an index created for fields : ```[a, b]```
3) User Searched on ```a``` and incorrectly  got back documents which only included docs with both a and b. 
4) User gets a list of sorted docs returned, but missing some documents.

In the scenario where we ignore sort order and used _all_docs, I would assume the user would have already complained that their documents are not sorted.

Perhaps we should improve the error messages returned?

willholley:
@tonysun83 this PR should improve the error message in the case where a user specifies an index with `use_index` that is no longer deemed valid. We should have a corresponding documentation PR which explains the error messages and how to address them.

willholley:
@tonysun83 If a `sort` is specified, perhaps it's reasonable to assume that the user expects the field to exist in all documents that are returned. To preserve existing behaviour, we could implicitly add an `{"$exists": true}` predicate for each sort field to the selector. 

This would allow us to safely validate the selector/index combination against `is_usable` whilst preserving existing behaviour as much as possible.

tonysun83:
@willholley : I considered appending $exists:true for backwards compatibility. We've done it in the past when we appended $gt: null to do the _all_docs trick for users. If we do this though, users would not see an error message, but they could still be getting incorrect results back as we're forcing the predicate fields to exist.  I.e

Index: [a, b, c]
Query: ```{"c": "foo"}```
Sort:[a, b]
New modified becomes: {"c": "foo", {"a":{"$exists": true}}, {"b":{"$exists": true}}}

It's a tradeoff: let their apps continue to run with possible incorrect data or let it break and force them to possibly re-index and re-write their app. I'd like more opinions @garrensmith @davisp ?

garrensmith:
Adding $exists: true is a nice idea. +1

src/mango/src/mango_idx.erl
```diff
@@ -93,17 +109,31 @@ recover(Db) ->
     end, DDocs)}.
 
 
-for_sort(Indexes, Opts) ->
-    % If a sort was specified we have to find an index that
-    % can satisfy the request.
+has_use_index(Opts) ->
+    case lists:keyfind(sort, 1, Opts) of
+        {use_index, _} ->
```
davisp: This check will never return true as sort is not equal to use_index.

src/mango/src/mango_selector.erl
```diff
@@ -398,6 +400,14 @@ negate({[{Field, Cond}]}) ->
     {[{Field, negate(Cond)}]}.
 
 
+norm_deduplicate({[{Op, [H|T]}]}) ->
```
davisp: I'm not seeing where this is called.

src/mango/src/mango_selector.erl
```diff
@@ -398,6 +400,14 @@ negate({[{Field, Cond}]}) ->
...
+    Arg = [H | [X || X <- norm_deduplicate(T), X /= H]],
```
davisp: This really hurts my brain to read. And not knowing what's supposed to call it I can't figure out if its correct.

src/mango/src/mango_idx.erl
davisp: Which means we're likely missing a test for this since travis returned green.

davisp:
Couple things to unpack in the discussion:

First, for #816 I think that's actually the opposite, its not whether it needs a new version bump but how many versions should be be back porting this fix to. A database that gives bad answers to queries is more a bug that needs to be patched backwards, not something we cut the cord on and just bump version numbers.

There is of course the issue around notifying users that a pending change may break their application (technically fix, but users have a different perspective :), so this is definitely something we should talk about. We'll also want to update the documentation as we've long said that using column prefixes was a suggestion for sharing indexes so that'll need to be fixed.

For @tonysun83's example of this changing a query from something that uses _all_docs to something that uses an existing index I don't think that's an issue. Originally I would have ixnayed it because of the lack of range restriction on A and B but we've already relaxed that by having an _all_docs default. Also the behavior here is different than the issue introduced by #816 in that previously Tony's example would have been an incorrect sort returned where the new version changes that (ie, same set of docs would be returned both places, just now its sorted). Which I think is quite a bit different than subtley different sets of documents returned. Also, if a user specifies a sort and isn't detecting the wrong sort order based on the _all_docs fall back then I'm failing to see how they'd be relying on that sort, and thus fixing it seems like not an issue.

For #816 with the _all_docs fallback is that also not more of a concern about performance rather than an app that stops working? Ie, all queries will continue to work and be "fixed" but some unknown number of queries will suddenly switch to using _all_docs and become much slower? Granted suddenly slower performance that lasts until an index can be built isn't nothing either.

src/mango/src/mango_idx.erl
willholley: There is a test but this situation is handled elsewhere. I'll remove `has_use_index` and its usage.

src/mango/src/mango_selector.erl
willholley: It's a step in the selector normalization - (line 40 above). The intention is to deduplicate predicates (particularly `"$exists": true`) but this isn't necessary for correctness - it just makes the normalized selector cleaner. Thinking about it again, I'll remove it here and will consider improving selector normalization in a different PR.

willholley:
I've pushed a new fix which doesn't require rewriting the selector. Instead, we just include sort fields when considering whether a JSON index provides coverage for the selector. This seems a lot simpler and, unlike the previous implementation, doesn't impact the validity of other index types (i.e. text indexes).

src/mango/src/mango_selector.erl
```diff
@@ -16,7 +16,8 @@
 -export([
     normalize/1,
     match/2,
-    has_required_fields/2
+    has_required_fields/2,
+    append_sort_predicate/2
```
davisp: This appears to be left over from the previous commits?

PR #328, opened 2015-07-20
Validate host header

kxepal:
By default, the Host header must be either localhost or the
fully-qualified domain name of the host.

Administrators can disable this check with;

[httpd]
validate_host = false

Administrators can override the valid host list with;

[httpd]
valid_hosts = foo,bar,baz

COUCHDB-2752

PR #879, opened 2017-10-08
Kxepal 2752 validate host

janl:
closes #328


janl:
wtffff, sorry for the noise

PR #880, opened 2017-10-08
Backport validate host feature

janl:
fresh rebase, closes #328


kxepal:
Thanks for catching this, @janl !
It seems build hanged on building Fauxton. I think we better remove it completely from 1.x.x branch as it far too outdated from what master has today. And it has optimizations preventing to work with good old CouchDB 1.x

PR #881, opened 2017-10-09
Fix maximum key value when using JSON indexes

willholley:
## Overview

Mango previously constrained range queries against JSON indexes
(map/reduce views) to startkey=[]&endkey=[{}]. In Mango, JSON
index keys are always compound (i.e. always arrays), but this
restriction resulted in Mango failing to match documents where
the indexed value was an object.

For example, an index with keys:

[1],
[2],
[{"foo": 3}]

would be restricted such that only [1] and [2] were returned
if a range query was issued.

On its own, this behaviour isn't necessarily unintuitive, but
it is different from the behaviour of a non-indexed Mango
query, so the query results would change in the presence of an
index.

Additonally, it prevented operators or selectors which explicitly
depend on a full index scan (such as $exists) from returning a
complete result set.

This commit changes the maximum range boundary from [{}] to {},
so all array/compound keys will be included.

## Testing recommendations

Create some documents with a field that contains a mix of value types - integers, strings, objects, nulls.

Without adding an index, run some Mango queries using range operators (`$lt`, `$lte`, `$gt`, `$gte`) and those that depend on full index scans - `$exists`, `$regex` etc. 

If you then add a JSON index on the field, ensure you get the same results.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -29,7 +29,7 @@
 -include_lib("couch/include/couch_db.hrl").
 -include_lib("couch_mrview/include/couch_mrview.hrl").
 -include("mango_cursor.hrl").
-
+-include("mango_idx_view.hrl").
 
 create(Db, Indexes, Selector, Opts) ->
     FieldRanges = mango_idx_view:field_ranges(Selector),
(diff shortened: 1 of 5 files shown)


janl:
@willholley this seems to have broken other mango tests (see Travis for details, all builds failed).

willholley:
oops - this turned out to be a bit trickier than I first thought because `end_key` returns a list. I've changed the PR to just add a special case when we want no upper bound.

willholley:
@davisp I'd appreciate a second review of this. My niggly concern is that it dosn't distinguish between a user specified upper bound of `[{}]` and `json_max_value`, but this seems like very edgey case. In that situation, all that would happen is that the index would not be as restrictive as it could be - the correct result should still be returned, with some records potentially filtered out in memory.

PR #886, opened 2017-10-10
Fix cluster upgrades when deploying ddoc_cache

davisp:
## Overview

As it turns out I made a bit of a mistake when I forgot that the old
ddoc_cache implementation had an ets_lru process registered as
ddoc_cache_lru. These cast messages were causing that process to crash.
If a cluster had enough design document activity and enough nodes this
would cause nodes with the old ddoc_cache implementation to reboot the
entire VM. This was a cascading failure due to the ets_lru process
restarting frequently enough that it brought down the entire ddoc_cache
application.

## Testing recommendations

Try a rolling reboot upgrade of a large cluster with significant write load on design documents.

## Checklist

### DIFF
--- a/src/ddoc_cache/src/ddoc_cache_lru.erl
+++ b/src/ddoc_cache/src/ddoc_cache_lru.erl
@@ -39,6 +39,9 @@
 -include("ddoc_cache.hrl").
 
 
+-define(OPENER, ddoc_cache_opener).
+
+
 -record(st, {
     pids, % pid -> key
     dbs, % dbname -> docid -> key -> pid
@@ -143,11 +146,11 @@ handle_call(Msg, _From, St) ->
 
 
 handle_cast({evict, DbName}, St) ->
-    gen_server:abcast(mem3:nodes(), ?MODULE, {do_evict, DbName}),
+    gen_server:abcast(mem3:nodes(), ?OPENER, {do_evict, DbName}),
     {noreply, St};
 
 handle_cast({refresh, DbName, DDocIds}, St) ->
-    gen_server:abcast(mem3:nodes(), ?MODULE, {do_refresh, DbName, DDocIds}),
+    gen_server:abcast(mem3:nodes(), ?OPENER, {do_evict, DbName, DDocIds}),
     {noreply, St};
 
 handle_cast({do_evict, DbName}, St) ->


src/ddoc_cache/src/ddoc_cache_lru.erl
```diff
@@ -143,11 +146,11 @@ handle_call(Msg, _From, St) ->
 
 
 handle_cast({evict, DbName}, St) ->
-    gen_server:abcast(mem3:nodes(), ?MODULE, {do_evict, DbName}),
+    gen_server:abcast(mem3:nodes(), ?OPENER, {do_evict, DbName}),
     {noreply, St};
 
 handle_cast({refresh, DbName, DDocIds}, St) ->
-    gen_server:abcast(mem3:nodes(), ?MODULE, {do_refresh, DbName, DDocIds}),
+    gen_server:abcast(mem3:nodes(), ?OPENER, {do_evict, DbName, DDocIds}),
```
chewbranca: Is changing the destination module and the cast message to `do_evict` intentional?

davisp: Yes, that's the fix.

https://github.com/apache/couchdb/blob/master/src/ddoc_cache/src/ddoc_cache_opener.erl#L47-L55

Its reverting the message passing protcol to match the old version. For old nodes this means they continue to behave the same. For new nodes, the ddoc_cache_opener module exists solely to forward and translate the old style messages to new style. My mistake was just not realizing that ddoc_cache_lru wouldn't be silently ignored by the old version (because the ets_lru process exists).

davisp: I should note that this is a bit convoluted because of the upgrade. In the future after the new ddoc_cache has been on master for enough versions we'll remove ddoc_cache opener and revert this commit to avoid the jump through ddoc_cache_opener.

As a note for the intrepid observer, upgrades that involve changing messages passed between nodes are totes the hardest.

chewbranca: aha! Makes sense now.

chewbranca:
+1

davisp:
Will merge this once travis finishes.

PR #894, opened 2017-10-17
Mango text index test fixes

willholley:
## Overview

The text index tests are not run routinely because CouchDB doesn't natively support them. This PR corrects some bad assertions as a result of recent improvements/fixes to Mango and ensures that all of the expected tests are run.

## Testing recommendations

Run the test suite with and without text index support.

## Checklist


src/mango/test/12-use-correct-index-test.py
```diff
@@ -68,15 +68,15 @@ def test_choose_index_with_two(self):
...
+        self.db.create_index(["name"], ddoc="bbb")
         self.db.create_index(["name"], ddoc="zzz")
         explain = self.db.find({"name": "Eddie", "age": {"$gte": 12}}, explain=True)
         self.assertEqual(explain["index"]["ddoc"], '_design/aaa')
 
     def test_choose_index_most_accurate(self):
-        self.db.create_index(["name", "location", "user_id"], ddoc="aaa")
-        self.db.create_index(["name", "age", "user_id"], ddoc="bbb")
+        self.db.create_index(["name", "age", "user_id"], ddoc="aaa")
+        self.db.create_index(["name", "age"], ddoc="_design/bbb")
```
garrensmith: This test is failing. you don't need the `_design` here.

PR #907, opened 2017-10-21
Improve release/RC tarball creation, bump ver to 2.1.1

### LINKED ISSUE #908: Fix release build process tooling
Currently, there are 3 problems with the RC and release build process that I'm aware of:

1. The Makefile functionality to diff a given RC against the last one is broken for RC1 (because there is no RC0 tag).
1. There is no easy way to bump the version number - you have to manually edit both `Makefile` and `rel/reltool.config`. See: https://github.com/apache/couchdb/pull/907 Sure would be nice to pull this from an env var or tag or something.
1. There's no equivalent of "make reltarball" that does roughly the same thing as "make rc" when you check out from a release tag (e.g. 2.1.0). This could be collapsed with "make rc" if it was assumed that all RCs have to be built from a proper git tag.

wohali:
RC1 had improper version numbering, and the Makefile script fails when the RC number is 1. This isn't a full workaround but it gets us in place to build and release 2.1.1 proper.

Closes #908

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ include version.mk
 
 REBAR?=$(shell echo `pwd`/bin/rebar)
 IN_RELEASE = $(shell if [ ! -d .git ]; then echo true; fi)
-COUCHDB_VERSION_SUFFIX = $(shell if [ -d .git ]; then echo '-`git rev-parse --short --verify HEAD`'; fi)
+COUCHDB_VERSION_SUFFIX = $(shell if [ ! -z "$(COUCH_RC)" ]; then echo '-RC$(COUCH_RC)'; else if [ -d .git ]; then echo '-`git rev-parse --short --verify HEAD`'; fi; fi)
 COUCHDB_VERSION = $(vsn_major).$(vsn_minor).$(vsn_patch)$(COUCHDB_VERSION_SUFFIX)
 
 DESTDIR=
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -357,11 +357,11 @@ else
 	@rm -rf apache-couchdb-*
 	@$(MAKE) dist 2>&1 > /dev/null
 	@rm apache-couchdb-*.tar.gz
-	@mv apache-couchdb-* apache-couchdb-2.1.0-RC$(COUCH_RC)
-	@tar czf apache-couchdb-2.1.0-RC$(COUCH_RC).tar.gz apache-couchdb-2.1.0-RC$(COUCH_RC)
-	@echo "Done apache-couchdb-2.1.0-RC$(COUCH_RC).tar.gz"
+	@mv apache-couchdb-* apache-couchdb-2.1.1-RC$(COUCH_RC)
```
janl: best to use `COUCHDB_VERSION` defined in https://github.com/apache/couchdb/pull/907/files#diff-b67911656ef5d18c4ae36cb6741b7965R18

rel/reltool.config
```diff
@@ -12,7 +12,7 @@
 
 {sys, [
     {lib_dirs, ["../src"]},
-    {rel, "couchdb", "2.1.0", [
+    {rel, "couchdb", "2.1.1", [
```
janl: not sure if this can be scripted. If not, we could make this the canonical source where we change the version number by hand instead of `https://github.com/apache/couchdb/blob/master/version.mk` (which I see is still on 2.1.0 too.

wohali: Will address this later...for now I'll just fix the Makefile problem.

wohali:
@janl please review again.

To bump the version number, you'll still need to modify both `version.mk` and `rel/reltool.config`. I'll update the release procedure with this info for now.

The new command for building an RC, say RC2, is `COUCH_RC=2 make dist`. If accepted for merge I'll also merge to master and update the [wiki page](https://cwiki.apache.org/confluence/display/COUCHDB/Release+Procedure).

The nice thing is that `make rc` is completely gone, as it was mostly just redoing what `make dist` had been getting wrong. I dropped the `git diff` a) because it wouldn't work for RC1, b) it wouldn't show the changes from the last _release_, only the last RC, and c) there are ways of getting this info through GitHub anyway that are formatted more nicely for a release manager these days.

janl:
nice work

wohali:
Wiki has been updated.

PR #915, opened 2017-10-25
Disallow duplicate json keys

janl:


rnewson:
please remove whitespace changes

rnewson:
+1 once ws changes are removed

janl:
@davisp @rnewson the WS changes are a separate commit

davisp:
@janl The issue is that the new whitespace is also wrong as its using two space indents instead of our standard 4. If you update to have correct indentation I'm fine leaving it as per SOP.

src/chttpd/src/chttpd_auth_cache.erl
```diff
@@ -45,7 +45,7 @@ get_user_creds(_Req, UserName) when is_binary(UserName) ->
             Props;
         UserProps when is_list(UserProps) ->
             couch_auth_cache:add_roles(Props,
-	        couch_util:get_value(<<"roles">>, UserProps))
+            couch_util:get_value(<<"roles">>, UserProps))
```
davisp: This is missing an extra two indents since its a continuation line.

janl:
sorry this slipped, fixed in latest push

PR #921, opened 2017-10-26
Fix build for 1.x.x

kxepal:
This updates libmozjs package on Travis and drops Fauxton.


janl:
screen shot 2017-10-26 at 14 44 47 😱

kxepal:
Yeah. And `+15 −150,212` - perfect diff! (:
Actually, it's just remove all the Fauxton bits. Too much of them >_>

wohali:
+1

For the record: Fauxton has moved way past being 1.x compatible, and the version here is broken. It was only ever shipped as an experimental feature. So we're ditching it for a 1.7 release, which I'm fine with.

janl:
👍

PR #929, opened 2017-10-28
feat: port time funs, as per #883

janl:
See #883 for details


janl:
From IRC: [16:38:54]  <vatamane>	+1

PR #937, opened 2017-10-31
Fix 1.x.x docs build

janl:


janl:
looks like the sphinx version on travis still has the classic theme, I’ll see if we can get that updated, otherwise its gonna be weird building releases off of this not on CI.

kxepal:
Hm...build failure reason is weird.

.travis.yml
```diff
@@ -4,7 +4,7 @@ before_install:
    - sudo apt-get -y install libtool automake autoconf autoconf-archive
    - sudo apt-get -y install texlive-latex-base texlive-latex-recommended
    - sudo apt-get -y install texlive-latex-extra texlive-fonts-recommended texinfo
-   - sudo apt-get -y install python-pygments python-docutils python-sphinx
+   - pip install pygments docutils sphinx
```
kxepal: Much likely this will fail without sudo or `--user` argument.

janl: in master, we just do pip install…

kxepal: Good!

janl: there we are ;) trying sudo now. — also, I’m gonna split the commits once this is working

kxepal: There could be another gotcha. Your motivation for this fix was build failure due to docs warning about template name change, right? Blind `pip install sphinx` will bring you latest version which may trigger more warnings due to new changes (IIRC, there was such fight for couchdb-documentation repo). If this will happens, try to install minimal required version to get the new stylesheet - 1.3.3, for instance - or any else 1.3.x. This should be enough to bypass this issue.

janl:
This is annoying, `make distcheck` passes locally :/

kxepal:
More interesting is what have changed since https://github.com/apache/couchdb/pull/921 where build was green.

kxepal:
@janl May I ask what the original build failure you're trying to fix?

janl:
I’m giving up on this, never mind :)

kxepal:
Ok. Hope nothing critical. Locally and on travis now everything is fine. Or should be so (:

PR #938, opened 2017-10-31
Update NOTICE: remove autoconf/m4 references and revert react patents

janl:


NOTICE
```diff
@@ -6,26 +6,6 @@ The Apache Software Foundation (http://www.apache.org/).
...
-
- * ac_check_curl.m4 (http://autoconf-archive.cryp.to/ac_check_curl.html)
-
-   Copyright 2008, Akos Maroy <darkeye@tyrell.hu>
-
- * pkg.m4 (http://http://pkg-config.freedesktop.org/wiki/)
-
-   Copyright 2004, Scott James Remnant <scott@netsplit.com>
-
- * jQuery (http://jquery.org/)
```
wohali: Cannot remove, we still use bootstrap which still depends on jQuery.

NOTICE
```diff
@@ -6,26 +6,6 @@ The Apache Software Foundation (http://www.apache.org/).
...
-
-   Copyright 2012 jQuery Foundation and other contributors
- 
- * jQuery UI (http://jqueryui.com)
```
wohali: Cannot confirm we have completely removed jQuery UI - would leave it in for now

NOTICE
janl: re-added, good catch

PR #951, opened 2017-11-01
Do not always return partial_filter_selector

garrensmith:
For Get / _index only return the partial_filter_selector for an index if
it is set to something other than default

## Checklist

### DIFF
--- a/src/mango/src/mango_idx_view.erl
+++ b/src/mango/src/mango_idx_view.erl
@@ -120,7 +120,7 @@ is_usable(Idx, Selector, SortFields) ->
     % and the selector is not a text search (so requires a text index)
     RequiredFields = columns(Idx),
 
-    % sort fields are required to exist in the results so
+    % sort fields are required to exist in the results so 
     % we don't need to check the selector for these
     RequiredFields1 = ordsets:subtract(lists:usort(RequiredFields), lists:usort(SortFields)),
 
--- a/src/mango/src/mango_idx_view.erl
+++ b/src/mango/src/mango_idx_view.erl
@@ -120,7 +120,7 @@ is_usable(Idx, Selector, SortFields) ->
     % and the selector is not a text search (so requires a text index)
     RequiredFields = columns(Idx),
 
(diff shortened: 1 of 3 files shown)


src/mango/test/16-index-selectors-test.py
```diff
@@ -161,6 +161,12 @@ def test_saves_partial_filter_selector_in_index(self):
         indexes = self.db.list_indexes()
         self.assertEqual(indexes[1]["def"]["partial_filter_selector"], selector)
 
+    def test_partial_filter_only_in_return_if_not_default(self):
```
willholley: do we have a test for the negative as well (i.e. partial_filter_selector is not present in the json when it has the default value / was not specified)?

src/mango/test/16-index-selectors-test.py
```diff
@@ -264,4 +270,11 @@ def test_text_old_selector_still_supported(self):
         resp = self.db.find(selector, explain=True, use_index='oldschooltext')
         self.assertEqual(resp["index"]["name"], "oldschooltext")
         docs = self.db.find(selector, use_index='oldschooltext')
-        self.assertEqual(len(docs), 3)
\ No newline at end of file
+        self.assertEqual(len(docs), 3)
+
+    @unittest.skipUnless(mango.has_text_service(), "requires text service")
+    def test_text_partial_filter_only_in_return_if_not_default(self):
```
willholley: same as above - do we test the opposite situation?

src/mango/src/mango_idx_view.erl
```diff
@@ -120,7 +120,7 @@ is_usable(Idx, Selector, SortFields) ->
     % and the selector is not a text search (so requires a text index)
     RequiredFields = columns(Idx),
 
-    % sort fields are required to exist in the results so 
+    % sort fields are required to exist in the results so
```
willholley: unrelated whitespace change

garrensmith: This was my editor removing whitespaces.

PR #957, opened 2017-11-02
Increase default pbkdf2 iterations to 10k

wohali:
Our default was previously 10, which is dangerously low.

NIST recommends [10,000 iterations](https://pages.nist.gov/800-63-3/sp800-63b.html#sec5).

So does the [OWASP password storage cheat sheet from 2015](https://www.owasp.org/index.php/Password_Storage_Cheat_Sheet).

As of [November 2015](https://cryptosense.com/parameter-choice-for-pbkdf2/), a pbkdf2-encrypted 8 random lowercase letters (37 bits of entropy) with *1000* iterations can be brute force cracked on a GPU setup in just 12 hours. It'll only be faster now. (10k iterations for the same poor password would take 5 days.)


wohali:
Looks like the increased iterations is slowing down `dev/run` sufficiently as to break most (but not all!) mango tests on travis. I'll add a workaround.

rnewson:
Until the replicator acquires a session cookie, this seems unwise.

wohali:
@rnewson So because the replicator uses basic auth, we want to keep the # of iterations low to improve replication connection performance at the expense of general security? That seems like a perilous tradeoff.

What about enforcing a different iteration limit for the hashing of admin passwords specifically?

rnewson:
I'm not happy about it (I did write the original pbkdf2 implementation for couchdb to improve security) but it does feel like a blocker. Can we make a couchdb release where replication becomes a) much slower and b) cpu bound. I don't think so.

rnewson:
separate iteration counts for admins vs not-admins is a good idea and simple to implement. Unfortunately it's a requirement to use admin creds if you want to successfully replicate design documents, but that's another part of the couchapp story, and perhaps we can document around it, telling folks to publish their design docs separately to replication.

wohali:
OK, I'll close this PR for now and open a new ticket with the idea from above. There's no rush on this, it just occurred to me that I could probably crack most 10-iteration passwords with my GPU in an hour or less ;)

rnewson:
yes, you could, assuming the password under attack was very weak.

Sebatyne:
Hi,

I read in https://github.com/apache/couchdb/issues/959 that :

> The replicator now supports session.

Maybe I missed some discussions, but could this Merge Request be open again ?

PR #962, opened 2017-11-03
Mango: generate a warning instead of an error when an user-specified index cannot be used

willholley:
## Overview

If a user specifies a value for use_index that is not valid for the current query - i.e. it does not meet the coverage requirements of the selector or sort fields - attempt to fall back to a valid index (or database scan) rather than returning a 400 error to the client.

When a fallback occurs, populate the "warning" field in the response (as we already do when a full database scan takes place) to feed back to the user that `use_index` was ignored.

This change is partially as mitigation for #816, which may lead to some previously valid indexes being deemed invalid, and also to make `use_index` less brittle for production applications in general. 

## Testing recommendations

Run the test suite. Run some mango queries with and without `use_index`. Observe that a warning is returned in the JSON and in Fauxton.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -18,6 +18,7 @@
     explain/1,
     execute/3,
     maybe_filter_indexes_by_ddoc/2,
+    remove_indexes_with_partial_filter_selector/1,
     maybe_add_warning/3
 ]).
 
@@ -47,16 +48,18 @@
 create(Db, Selector0, Opts) ->
     Selector = mango_selector:normalize(Selector0),
     UsableIndexes = mango_idx:get_usable_indexes(Db, Selector, Opts),
-
-    {use_index, IndexSpecified} = proplists:lookup(use_index, Opts),
-    case {length(UsableIndexes), length(IndexSpecified)} of
-        {0, 0} ->
+    case length(UsableIndexes) of
+        0 ->
             AllDocs = mango_idx:special(Db),
             create_cursor(Db, AllDocs, Selector, Opts);
-        {0, _} ->
-            ?MANGO_ERROR({no_usable_index, selector_unsupported});
         _ ->
-            create_cursor(Db, UsableIndexes, Selector, Opts)
+            case mango_cursor:maybe_filter_indexes_by_ddoc(UsableIndexes, Opts) of
+                [] ->
+                    % use_index doesn't match a valid index - fall back to a valid one
+                    create_cursor(Db, UsableIndexes, Selector, Opts);
+                UserSpecifiedIndex ->
+                    create_cursor(Db, UserSpecifiedIndex, Selector, Opts)
+            end
     end.
 
 
@@ -90,9 +93,7 @@ execute(#cursor{index=Idx}=Cursor, UserFun, UserAcc) ->
 maybe_filter_indexes_by_ddoc(Indexes, Opts) ->
     case lists:keyfind(use_index, 1, Opts) of
         {use_index, []} ->
-            % We remove any indexes that have a selector 
-            % since they are only used when specified via use_index
-            remove_indexes_with_partial_filter_selector(Indexes);
+            [];
         {use_index, [DesignId]} ->
             filter_indexes(Indexes, DesignId);
         {use_index, [DesignId, ViewName]} ->
@@ -150,12 +151,54 @@ group_indexes_by_type(Indexes) ->
     end, ?CURSOR_MODULES).
 
 
-maybe_add_warning(UserFun, #idx{type = IndexType}, UserAcc) ->
-    case IndexType of
+maybe_add_warning(UserFun, #cursor{index = Index, opts = Opts}, UserAcc) ->
+    UseIndexInvalid = case lists:keyfind(use_index, 1, Opts) of
+        {use_index, []} ->
+            [];
+        {use_index, [DesignId]} ->
+            case filter_indexes([Index], DesignId) of
+                [] ->
+                    [fmt("_design/~s was not used because it does not contain a valid index for this query.", 
+                        [ddoc_name(DesignId)])];
+                _ ->
+                    []
+            end;
+        {use_index, [DesignId, ViewName]} ->
+            case filter_indexes([Index], DesignId, ViewName) of
+                [] ->
+                    [fmt("_design/~s, ~s was not used because it is not a valid index for this query.", 
+                        [ddoc_name(DesignId), ViewName])];
+                _ ->
+                    []
+            end
+    end,
+
+    NoIndex = case Index#idx.type of
         <<"special">> ->
-            Arg = {add_key, warning, <<"no matching index found, create an index to optimize query time">>},
-            {_Go, UserAcc0} = UserFun(Arg, UserAcc),
-            UserAcc0;
+            [<<"no matching index found, create an index to optimize query time">>];
         _ ->
-            UserAcc
-    end.
+            []
+    end,
+
+    maybe_add_warning_int(UseIndexInvalid ++ NoIndex, UserFun, UserAcc).
+
+
+maybe_add_warning_int([], _, UserAcc) ->
+   UserAcc;
+
+maybe_add_warning_int(Warnings, UserFun, UserAcc) ->
+    % only include the first warning
+    Arg = {add_key, warning, hd(Warnings)},
+    {_Go, UserAcc0} = UserFun(Arg, UserAcc),
+    UserAcc0.
+
+
+fmt(Format, Args) ->
+    iolist_to_binary(io_lib:format(Format, Args)).
+
+
+ddoc_name(<<"_design/", Name/binary>>) ->
+    Name;
+
+ddoc_name(Name) ->
+    Name.
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -152,43 +152,42 @@ group_indexes_by_type(Indexes) ->
 
 
 maybe_add_warning(UserFun, #cursor{index = Index, opts = Opts}, UserAcc) ->
-    UseIndexInvalid = case lists:keyfind(use_index, 1, Opts) of
+    NoIndexWarning = case Index#idx.type of
+        <<"special">> ->
+            <<"no matching index found, create an index to optimize query time">>;
+        _ ->
+            ok
+    end,
+
+    UseIndexInvalidWarning = case lists:keyfind(use_index, 1, Opts) of
         {use_index, []} ->
-            [];
+            NoIndexWarning;
         {use_index, [DesignId]} ->
             case filter_indexes([Index], DesignId) of
                 [] ->
-                    [fmt("_design/~s was not used because it does not contain a valid index for this query.", 
-                        [ddoc_name(DesignId)])];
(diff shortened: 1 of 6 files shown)


src/mango/src/mango_cursor.erl
```diff
@@ -150,12 +151,53 @@ group_indexes_by_type(Indexes) ->
     end, ?CURSOR_MODULES).
 
 
-maybe_add_warning(UserFun, #idx{type = IndexType}, UserAcc) ->
-    case IndexType of
+maybe_add_warning(UserFun, #cursor{index = Index, opts = Opts}, UserAcc) ->
+
```
tonysun83: Is this whitespace intended? I don't think I've seen this before. Might want to check styling guidelines.

src/mango/src/mango_cursor.erl
```diff
@@ -150,12 +151,53 @@ group_indexes_by_type(Indexes) ->
...
+    UseIndexInvalid = case lists:keyfind(use_index, 1, Opts) of
+        {use_index, []} ->
+            [];
+        {use_index, [DesignId]} ->
+            case filter_indexes([Index], DesignId) of
+                [] ->
+                    [fmt("_design/~s was not used because it does not contain a valid index for this query.", [ddoc_name(DesignId)])];
```
tonysun83: this line is too long

src/mango/src/mango_cursor.erl
```diff
@@ -150,12 +151,53 @@ group_indexes_by_type(Indexes) ->
...
+                _ ->
+                    []
+            end;
+        {use_index, [DesignId, ViewName]} ->
+            case filter_indexes([Index], DesignId, ViewName) of
+                [] ->
+                    [fmt("_design/~s, ~s was not used because it is not a valid index for this query.", [ddoc_name(DesignId), ViewName])];
```
tonysun83: super long line

src/mango/src/mango_cursor.erl
```diff
@@ -150,12 +151,53 @@ group_indexes_by_type(Indexes) ->
...
+
+fmt(Format, Args) ->
+    iolist_to_binary(io_lib:format(Format, Args)).
+
+
+ddoc_name(<<"_design/", Name/binary>>) ->
+    Name;
+
+ddoc_name(Name) ->
+    Name.
```
tonysun83: missing endline

src/mango/src/mango_error.erl
```diff
@@ -21,12 +21,6 @@
 ]).
 
 
-info(mango_idx, {no_usable_index, no_indexes_defined}) ->
-    {
-        400,
-        <<"no_usable_index">>,
-        <<"There are no indexes defined in this database.">>
-    };
 info(mango_idx, {no_usable_index, no_index_matching_name}) ->
```
tonysun83: do we need to remove this error as well?

src/mango/src/mango_idx.erl
```diff
@@ -59,20 +59,16 @@ list(Db) ->
...
-        ?MANGO_ERROR({no_usable_index, no_indexes_defined})
-    end,
 
-    FilteredIndexes = mango_cursor:maybe_filter_indexes_by_ddoc(ExistingIndexes, Opts),
-    if FilteredIndexes /= [] -> ok; true ->
-        ?MANGO_ERROR({no_usable_index, no_index_matching_name})
-    end,
+    GlobalIndexes = mango_cursor:remove_indexes_with_partial_filter_selector(ExistingIndexes),
+    UserSpecifiedIndex = mango_cursor:maybe_filter_indexes_by_ddoc(ExistingIndexes, Opts),
+    UsableIndexes0 = lists:usort(GlobalIndexes ++ UserSpecifiedIndex),
```
tonysun83: why are we sorting the indexes here?

src/mango/test/05-index-selection-test.py
```diff
@@ -77,34 +73,78 @@ def test_uses_index_when_no_range_or_equals(self):
         resp_explain = self.db.find(selector, explain=True)
         self.assertEqual(resp_explain["index"]["type"], "json")
 
-
```
tonysun83: whitespace

src/mango/src/mango_idx.erl
willholley: this is to remove duplicates - an index might belong to GlobalIndexes and be specified by the user explicitly.

willholley:
@tonysun83 Regarding your concerns about the fallback behaviour:
1. This is a concern in general and doesn't seem specific to this PR. Assuming users don't specify `use_index`, we depend on the query planner to select an appropriate index and it should be the case that the only indexes which can change the results are partial indexes. Indexes that are not usable for the query are excluded from the set of candidate indexes, so the fallback is safe so long as the text indexes correctly exclude themselves.

2. I think this points to a general weakness around index management for Query. Ideally it would be safe to add and remove indexes in production databases whilst only impacting performance. Unfortunately, the query planner doesn't take the index state into account, so if an index is added to a production database and it takes some time to build, the query planner would use it and time out. Again, I'm not really sure that's a concern specific to this PR, but one that should be addressed independently.

tonysun83:
Wanted to clarify a bit:

For scenario 1): I agree that fixing the difference between text and json indexes is outside the scope of the PR, but I was thinking about the specific backwards-compatibility situation where a user specifies ```use_index``` for a json index, and it fails, and the fallback is a text index. Would we rather have them possibly get different results and complain later down the road versus failing them right away with an error?

For scenario 2): Similarly to scenario 1), the issue is whether or not the fallback leads to inadvertent changes for the user. In this case, he  is modifying his indexes based on perceived notion of what is being used. The performance degradation for _all_docs is noticed and he complains about the issue. We notify him saying that an index was deleted, but he says the index wasn't being used. Now he says ok he wants to rebuild an index...and that may take a long time and negatively affect his cluster.

I wonder if it's better to explicitly send back index names as part an failure error message. But could that be a security risk?

src/mango/test/05-index-selection-test.py
```diff
@@ -46,7 +46,7 @@ def test_use_most_columns(self):
                 "name.last": "Something or other",
                 "age": {"$gt": 1}
             }, explain=True)
-        self.assertNotEqual(resp["index"]["ddoc"], "_design/" + ddocid)
+        self.assertNotEqual(resp["index"]["ddoc"], ddocid)
```
tonysun83: this looks a random testcase fix not related to the PR?

src/mango/test/05-index-selection-test.py
```diff
@@ -108,41 +103,53 @@ def test_reject_use_index_ddoc_and_name_invalid_fields(self):
...
         }
-        try:
-            self.db.find(selector, use_index=[ddocid,name])
-        except Exception as e:
-            self.assertEqual(e.response.status_code, 400)
-        else:
-            raise AssertionError("did not reject bad use_index")
+        
+        resp = self.db.find(selector, use_index=[ddocid,name], return_raw=True)
+        self.assertEqual(resp["warning"], "{0}, {1} was not used because it is not a valid index for this query.".format(ddocid, name))
```
tonysun83: in the test case above, you're checking that a value is added
```
# should still return a correct result
for d in r["docs"]:
    self.assertEqual(d["company"], "Pharmex")
```
We should be consistent?

src/mango/test/05-index-selection-test.py
```diff
@@ -108,41 +103,53 @@ def test_reject_use_index_ddoc_and_name_invalid_fields(self):
...
-        except Exception as e:
-            self.assertEqual(e.response.status_code, 400)
-        else:
-            raise AssertionError("did not reject bad use_index")
+
+        resp_explain = self.db.find(selector, sort=["foo", "bar"], use_index=ddocid_invalid, explain=True)
+        self.assertEqual(resp_explain["index"]["ddoc"], ddocid_valid)    
+
+        resp = self.db.find(selector, sort=["foo", "bar"], use_index=ddocid_invalid, return_raw=True)
+        self.assertEqual(resp["warning"], '{0} was not used because it does not contain a valid index for this query.'.format(ddocid_invalid))
```
tonysun83: same as above, we should get a value back here right? It's going to fall back on the ddoc_valid and return a result, so just confirm a result?

src/mango/src/mango_cursor.erl
```diff
@@ -150,12 +151,54 @@ group_indexes_by_type(Indexes) ->
...
-    case IndexType of
+maybe_add_warning(UserFun, #cursor{index = Index, opts = Opts}, UserAcc) ->
+    UseIndexInvalid = case lists:keyfind(use_index, 1, Opts) of
+        {use_index, []} ->
+            [];
+        {use_index, [DesignId]} ->
+            case filter_indexes([Index], DesignId) of
+                [] ->
+                    [fmt("_design/~s was not used because it does not contain a valid index for this query.", 
+                        [ddoc_name(DesignId)])];
```
tonysun83: Should make the warning message a little bit more explicit here and add "we used index <other index name> " instead?

src/mango/src/mango_cursor.erl
```diff
@@ -150,12 +151,54 @@ group_indexes_by_type(Indexes) ->
...
+                _ ->
+                    []
+            end;
+        {use_index, [DesignId, ViewName]} ->
+            case filter_indexes([Index], DesignId, ViewName) of
+                [] ->
+                    [fmt("_design/~s, ~s was not used because it is not a valid index for this query.", 
+                        [ddoc_name(DesignId), ViewName])];
```
tonysun83: same as the above comment

src/mango/src/mango_cursor.erl
```diff
@@ -150,12 +151,54 @@ group_indexes_by_type(Indexes) ->
...
-            UserAcc0;
+            [<<"no matching index found, create an index to optimize query time">>];
         _ ->
-            UserAcc
-    end.
\ No newline at end of file
+            []
+    end,
+
+    maybe_add_warning_int(UseIndexInvalid ++ NoIndex, UserFun, UserAcc).
```
tonysun83: Do we really need to append two messages and then take the head of that list later? I feel like it's either going to be a
1) "we used a diff index" warning or
2) "no indexes found, we used all docs" warning.

Instead of returning empty lists, perhaps you could just move the NoIndex value as the return value for the case clauses.  That way no need to add and then take the head. 

Ex:

```
{use_index, []} ->
    NoIndex;
````

File: src/mango/src/mango_cursor.erl [hunk 14]
```diff
@@ -150,12 +151,54 @@ group_indexes_by_type(Indexes) ->
...
+
+
+maybe_add_warning_int([], _, UserAcc) ->
+   UserAcc;
+
+maybe_add_warning_int(Warnings, UserFun, UserAcc) ->
+    % only include the first warning
```
tonysun83: see comment about about the warning list

src/mango/test/05-index-selection-test.py
willholley: agree that we should check the result in the cases above. In this test, however, I don't expect any matches (and can assert that).

src/mango/src/mango_cursor.erl
willholley: I figured the user can use _explain to see which index was used

tonysun83:
+1

PR #966, opened 2017-11-06
Version bump to 2.2.0 for future release

wohali:
Given PSE work will probably land, we need to bump the minor version number.


davisp:
+1

PR #986, opened 2017-11-12
Revert to ubuntu Jenkins label, uids now standardised

wohali:
Previously, we used the `couchdbtest` label to select a specific subset of nodes with a known uid/gid combination for the `jenkins` user. This is necessary to ensure a successful build inside of a docker container without file permissions issues.

Infra has requested to delete that label, and has agreed to standardise the uid/gid to 910/910 across all nodes labelled `ubuntu`. That's done now.

I will also need to merge this change to the 2.1.x branch in a secondary PR.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,6 +12,7 @@ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 License for the specific language governing permissions and limitations under
 the License.
 */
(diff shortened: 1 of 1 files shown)


wohali:
```
17:01 <+jan_____> Wohali: I think you’re the only one who can reasonably review
                  this, feel free to self+1
```

PR #923, opened 2017-10-27
Mango test failures

tonysun83:
## Overview
Mango tests fail due to flaky tests with index deletion. I cannot recreate locally.
@davisp and @garrensmith
mentioned that it could be due to the latest ddoc_cache changes and suggested
changing r=3 to see if that would help. However, we recently changed tests to use
n=1 as default for all dbs created. I wonder if having w=3 and n=1 could be the reason.
We get a 202 back, but the index isn't deleted in time when we go to retrieve it. So changing
w=3  to w=1 to see if that will help. If not, I may add a wait loop inside delete.

## Testing recommendations
Let's see if it passes travis.

## Related Issues or Pull Requests
https://github.com/apache/couchdb/issues/918

## Checklist

- [X] Code is written and works correctly;
- [X] Changes are covered by tests;
- [ ] Documentation reflects the changes;

PR #995, opened 2017-11-15
Improve Mango test suite performance

willholley:
## Overview

877c886d added delays into the Mango test suites to improve stability. However, the failures were never reproduced locally and significantly increase the time that it takes to run the tests. This is a PR to experiment with removing those delays to see whether we can get the test suite stable in Travis without them.

## Testing recommendations

Run the Mango test suite.

#923

## Checklist

### DIFF
--- a/src/mango/test/mango.py
+++ b/src/mango/test/mango.py
@@ -82,9 +82,7 @@ class Database(object):
 
     def recreate(self):
         self.delete()
-        delay()
         self.create()
-        delay()
 
     def save_doc(self, doc):
         self.save_docs([doc])
@@ -126,7 +124,6 @@ class Database(object):
             body["index"]["partial_filter_selector"] = partial_filter_selector
         body = json.dumps(body)
         r = self.sess.post(self.path("_index"), data=body)
-        delay()
         r.raise_for_status()
         assert r.json()["id"] is not None
         assert r.json()["name"] is not None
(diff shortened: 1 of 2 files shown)


src/mango/test/01-index-crud-test.py
```diff
@@ -193,7 +193,7 @@ def test_recreate_index(self):
         pre_indexes = self.db.list_indexes()
         for i in range(5):
             ret = self.db.create_index(["bing"], name="idx_recreate")
-            assert ret is True
+            self.assertTrue(ret)
```
kxepal: `self.assertIs` is better. This assertion will be passed if ret would be not just `True`, but `"1"`, `42`, `["ololo"]`, e.g. any non-falsy value.

src/mango/test/mango.py
```diff
@@ -126,11 +124,14 @@ def create_index(self, fields, idx_type="json", name=None, ddoc=None,
...
         r = self.sess.post(self.path("_index"), data=body)
-        delay()
         r.raise_for_status()
         assert r.json()["id"] is not None
         assert r.json()["name"] is not None
-        return r.json()["result"] == "created"
+
+        created = r.json()["result"] == "created"
+        if not created:
+            print r.json()["result"]
```
kxepal: print

willholley:
This seems pretty solid now so I think is ready for review. I've run the latest commit through Travis >10 times without a failure. 697345f needs to be removed before merging to master.

willholley:
also worth adding that this brings the time to run the Mango tests down from ~15 minutes to 20 seconds on my local machine.

tonysun83:
+1, smart use of list_indexes to see if the index is available to use and to make sure our index is deleted. let's give this a try

Issue #1001, opened 2017-11-16
[1.6.x][Compaction] Random compaction failures


starting:
## Expected Behavior
1. Compaction completes without any error. 

## Current Behavior
1. Temporary file is not deleted.  
2. couch.log has following stack trace 
`[Thu, 17 Aug 2017 11:14:30 GMT] [error] [emulator] Error in process <0.12887.1> with exit value: {function_clause,[{couch_db_updater,'-copy_docs/4-fun-3-',[not_found],[{file,"couch_db_updater.erl"},{line,857}]},{lists,map,2,[{file,"lists.erl"},{line,1224}]},{lists,map,2,[{file,"lists.erl"},{line,1224}]},{couch_db_updater,copy_docs... 

[Thu, 17 Aug 2017 11:14:30 GMT] [error] [<0.5045.1>] ** Generic server <0.5045.1> terminating 
** Last message in was {'EXIT',<0.12887.1>,
                           {function_clause,
                               [{couch_db_updater,'-copy_docs/4-fun-3-',
                                    [not_found],
                                    [{file,"couch_db_updater.erl"},
                                     {line,857}]},
                                {lists,map,2,[{file,"lists.erl"},{line,1224}]},
                                {lists,map,2,[{file,"lists.erl"},{line,1224}]},
                                {couch_db_updater,copy_docs,4,
                                    [{file,"couch_db_updater.erl"},
                                     {line,856}]},
                                {couch_db_updater,'-copy_compact/3-fun-0-',7,
                                    [{file,"couch_db_updater.erl"},
                                     {line,917}]},
                                {couch_btree,stream_kv_node2,8,
                                    [{file,"couch_btree.erl"},{line,708}]},
                                {couch_btree,stream_kp_node,7,
                                    [{file,"couch_btree.erl"},{line,637}]},
                                {couch_btree,fold,4,
                                    [{file,"couch_btree.erl"},{line,159}]}]}}
** When Server state == {db,<0.5044.1>,<0.5045.1>,<0.12887.1>,
                            <<"1502965878196162">>,<0.5046.1>,<0.5042.1>,
                            <0.5048.1>,
                            {db_header,6,885493,0,
                                {6119000697,{39549,257,136740145},7541350},
                                {6119002975,42754,3498231},
                                {6119018635,[],8490},
                                286,6117892339,5154762888,1000},
                            885493,
                            {btree,<0.5042.1>,
                                {6119000697,{39549,257,136740145},7541350},
[...]

PR #1003, opened 2017-11-16
Fix: Added a temporary fix to avoid compaction failures #1001

starting:
## Overview
Fix for #1001 
Discarded any document not available in couchdb but only in main index.  It seems couch db main index is not getting properly updated after purging documents. Compactor fails with error 'not_found' when it ties to get the full document from the couchdb.

## Testing recommendations
Compaction fails continuously when a document is actually purged but not cleaned up in main index.  This issue is random.  I tested couple of databases that had the issue and compaction is working fine after the fix. 

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 before_install:
    - sudo apt-get -y update
-   - sudo apt-get -y install libicu-dev libmozjs-dev pkg-config help2man
+   - sudo apt-get -y install libicu-dev libmozjs185-dev pkg-config help2man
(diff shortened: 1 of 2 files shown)


starting:
Seems CI fails continuously, anyway to get through this?

wohali:
You want the 1.7.x branch. 1.6 is no longer under development nor is it supported, see https://blog.couchdb.org/2017/11/14/apache-couchdb-cve-2017-12635-and-cve-2017-12636/

Please resubmit.

PR #1011, opened 2017-11-20
Allow to use $or with a mango JSON index

nono:
## Overview

Since #816, mango JSON index on compound fields can be selected only if the selector make sure that all the fields listed in the index are always present. This commit add a special case where all clauses of an `$or` can ensure that a field is present.

For instance, if I had an index:

```
[A, B]
```

is_usable would now return true for the selector:

```json
{
  "A": "foo",
  "$or": {
    "B": "bar",
    "B": "baz"
  }
}
```

## Testing recommendations

- #816

## Checklist

### DIFF
--- a/src/mango/src/mango_selector.erl
+++ b/src/mango/src/mango_selector.erl
@@ -596,9 +596,19 @@ has_required_fields(Selector, RequiredFields) when not is_list(Selector) ->
 % We can "see" through $and operator. We ignore other
 % combination operators because they can't be used to restrict
 % an index.
-has_required_fields([{[{<<"$and">>, Args}]}], RequiredFields) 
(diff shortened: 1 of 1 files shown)


willholley:
Thanks for this @nono. Would you be able to write an integration test (e.g. https://github.com/apache/couchdb/blob/master/src/mango/test/05-index-selection-test.py#L24)? 

It's probably also looking at https://github.com/apache/couchdb/pull/1014 as that PR changes the implementation of `has_required_fields` a little and `$or` would be susceptible to the related bug.

willholley:
fixed in #1038

nono:
Thanks @willholley!

PR #1014, opened 2017-11-22
Fix index validation for nested $and

willholley:
## Overview

mango_selector:has_required_fields checks that a list of
indexed fields is covered by a given selector. The implementation
recurses through the selector, tracking fields that encounters.

Unfortunately, this skipped peers of combination operators. For
example,

```
"selector": {
	"$and":[
		"$and":[
			"A": "foo"
		],
		"$and":[
			"B": "bar"
		]
	]
}
```

would skip the first nested "$and" operator and only return "B"
as a covered field.

This commit explicitly handles this situation (the only combination
operator we care about is $and), so for the above selector we
would correctly identify "A" and "B" as covered fields.

## Testing recommendations

Run the test suite. Test that Query selectors with nested `$and` operators correctly use `json` indexes that should cover them.

## Checklist

### DIFF
--- a/src/mango/src/mango_selector.erl
+++ b/src/mango/src/mango_selector.erl
@@ -578,36 +578,46 @@ match({[_, _ | _] = _Props} = Sel, _Value, _Cmp) ->
 % until we match then all or run out of selector to
 % match against.
 
+has_required_fields(Selector, RequiredFields) ->
+    Remainder = has_required_fields_int(Selector, RequiredFields),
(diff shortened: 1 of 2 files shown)


willholley:
Thanks to @afinne for identifying the issue. The original bug report showed that the index / query combination at https://pastebin.com/WiQ5Edgu failed.

tonysun83:
+1

PR #1029, opened 2017-11-29
Fix: Added a temporary fix to avoid compaction failures #1001

starting:
## Overview

Fix for #1001
Discarded any document not available in couchdb but only in main index. It seems couch db main index is not getting properly updated after purging documents. Compactor fails with error 'not_found' when it ties to get the full document from the couchdb.

This fix is cherry-picked from the fix done on top of 1.6.1 in #1003

## Testing recommendations

Compaction fails continuously when a document is actually purged but not cleaned up in main index. This issue is random. I tested couple of databases that had the issue and compaction is working fine after the fix.

## Checklist

### DIFF
--- a/src/couchdb/couch_db_updater.erl
+++ b/src/couchdb/couch_db_updater.erl
@@ -884,7 +884,10 @@ copy_docs(Db, #db{updater_fd = DestFd} = NewDb, InfoBySeq0, Retry) ->
                         fun({_, _, _, AttLen, _, _, _, _}, S) -> S + AttLen end,
                         SummarySize, AttsInfo),
                     {IsDel, Pos, Seq, TotalLeafSize}
-                end, RevTree)}
+                end, RevTree)};
+        (U) ->
+             ?LOG_INFO("Drop a document from compaction: not found in db : ~p", [U]),
+            []
         end, LookupResults),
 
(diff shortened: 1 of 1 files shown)


davisp:
-1 on this change. The bug is whatever is happening in purge that's broken. Dropping documents during compaction is a big no-go for me as that's kind of the crux of our consistency during compaction swaps.

starting:
I too share your view @davisp.   

What should we do if a document not present in the db during compaction? I feel it is better to be flexible and notify the error, rather failing complete compaction process. This help for the availability of the database. 

In Summary, We should find and fix the root of this issue, but still a single document failure should not cause the whole DB to be unavailable. (I couldn't still find an exact location that generates this issue.

davisp:
At the very least this should be a required flag or config setting. Preferably I'd even just never commit it and document it as a temp fix that can be compiled and dropped in separately to fix broken compactions. Having code on master that drops data during compactions scares the daylights out of me.

janl:
Closing this issues because there won‘t [be any more CouchDB 1.x releases](https://lists.apache.org/thread.html/6fa798e96686b7b0013ec2088140d00aeb7d34487d3f5ad032af6934@%3Cdev.couchdb.apache.org%3E). If you experience this issue on current 2.x releases, please re-open it with additional information.

PR #1038, opened 2017-12-01
Allow to use index with or

willholley:
## Overview

Implementation of #1011 taking into account recent selector fixes.

## Testing recommendations

#1011

## Checklist

### DIFF
--- a/src/mango/src/mango_selector.erl
+++ b/src/mango/src/mango_selector.erl
@@ -693,7 +693,7 @@ has_required_fields_and_false_test() ->
     Normalized = normalize(Selector),
     ?assertEqual(false, has_required_fields(Normalized, RequiredFields)).
 
-has_required_fields_or_test() ->
+has_required_fields_or_false_test() ->
     RequiredFields = [<<"A">>],
     Selector = {[{<<"$or">>,
           [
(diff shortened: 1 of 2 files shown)


src/mango/src/mango_selector.erl
```diff
@@ -680,7 +697,8 @@ has_required_fields_nested_and_true_test() ->
           ]
     }]},
 
-    ?assertEqual(true, has_required_fields(Selector, RequiredFields)).
+    Normalized = normalize(Selector),
```
tonysun83: wups, did I miss this in the last review?

src/mango/src/mango_selector.erl
```diff
@@ -704,4 +722,103 @@ has_required_fields_or_test() ->
...
+          [
+              Selector1,
+              Selector2
+          ]
+    }]},
+    Normalized = normalize(Selector),
+    ?assertEqual(true, has_required_fields(Normalized, RequiredFields)).
+
+
+-endif.
```
tonysun83: I know the remainder mechanism was added for $and and it should work with $or's peers as well, but should we 
have a redundant nested $or with $or test just to feel safe?

garrensmith: It might be over kill. But what about a $and in an $or

src/mango/src/mango_selector.erl
willholley: Normalization wasn't necessary for the test to pass but I think this makes it clearer that we expect the input to be Normalized. This is because normalized selectors always have an outer `$and` operator so we assume that only `$and` needs to handle the case where it's children are a list and it has no siblings i.e. [here](https://github.com/willholley/couchdb/blob/66510494c8fbadfa86021f82d6f0804d2a92db3a/src/mango/src/mango_selector.erl#L602).

I'm tempted to go further and have a specific record type for normalized selectors (which would also prevent them being normalized multiple times) but that's one for another PR, I think.

src/mango/src/mango_selector.erl
willholley: @garrensmith we test `$and` in an `$or` at https://github.com/apache/couchdb/pull/1038/files#diff-484bd99289bb1293f65aedcd7bc029f7R770

willholley: @garrensmith we test `$and` in an `$or` at https://github.com/apache/couchdb/pull/1038/files#diff-484bd99289bb1293f65aedcd7bc029f7R770

garrensmith: Awesome. Sorry I missed that test.

PR #1039, opened 2017-12-02
Fix haproxy stats

redgeoff:
- `stats scope .` ends up blocking all backends
- Renaming endpoint from `_stats` to `haproxy_stats` so that it doesn't collide with CouchDB's `_stats` endpoint
- Added a commented out stats auth line so that you can easily password protect

## Testing recommendations

Visit `http://example.com:5984/_stats` now and you won't see any stats for any backends. With my changes you will see the backend specified when you visit `http://example.com:5984/haproxy_stats`

### DIFF
--- a/rel/haproxy.cfg
+++ b/rel/haproxy.cfg
@@ -30,7 +30,7 @@ defaults
         timeout connect 500
 
         stats enable
-        stats uri /haproxy_stats
+        stats uri /_haproxy_stats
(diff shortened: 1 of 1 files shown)


kocolosk:
Good call. It's a nit but maybe `_haproxy_stats` is better so it's guaranteed not to clash with any CouchDB database name?

redgeoff:
Changed to `_haproxy_stats` as per @kocolosk's recommendation

PR #1067, opened 2017-12-14
Fix mango native proc crash

eiri:
## Overview

When soft limit is reached couch_proc_manager evicts idle processes by casting on them `stop`
message. Since mango_native_proc doesn't handle this message it results to its crash with
`invalid_cast` reason.

## Testing recommendations

`make eunit apps=mango` or `make eunit apps=mango suites=mango_native_proc` for isolated test.

`make mango-test` for integration test.

## Checklist

### DIFF
--- a/src/mango/src/mango_native_proc.erl
+++ b/src/mango/src/mango_native_proc.erl
@@ -113,6 +113,9 @@ handle_cast(garbage_collect, St) ->
     erlang:garbage_collect(),
     {noreply, St};
 
+handle_cast(stop, St) ->
+    {stop, normal, St};
+
 handle_cast(Msg, St) ->
     {stop, {invalid_cast, Msg}, St}.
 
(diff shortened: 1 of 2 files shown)


src/mango/src/mango_native_proc.erl
```diff
@@ -363,4 +366,35 @@ validate_index_info(IndexInfo) ->
...
+        erlang:trace(Pid, true, [garbage_collection]),
+        receive {trace, Pid, gc_start, _} ->
+            erlang:trace(Pid, false, [garbage_collection]),
+            exit(gc_start)
+        end
+    end),
+    erlang:yield(),
+    ?assertEqual({noreply, []}, handle_cast(garbage_collect, [])),
+    receive
+        {'DOWN', TracerRef, _, _, Msg} -> ?assertEqual(gc_start, Msg)
```
tonysun83: So iiuc, when the spawned process starts garbage collection, it'll exit with ```gc_start`` which would then cause our mango_native_proc to exit with an ```{invalid_cast, gc_start}```

tonysun83:
The fix lgtm. The garbage collection test is a little tricky, just want to understand it a bit better.

src/mango/src/mango_native_proc.erl
eiri: No, no, it's different.

We spawn and monitor a new process that sets a trace on the test process, because process can't trace itself. Then in the test process we call a garbage collection by executing `handle_cast(garbage_collect, [])` and asserting that we are getting proper `noreply` response. The tracer receives `gc_start` trace message, indicating that its target (the test process) started a garbage collection and exits with reason `gc_start`. Because we are monitoring it in our test process, we are receiving its exit reason and can assert that it was indeed for a garbage collection.

Interesting bit here is `erlang:yield()` - we need to give a scheduler a chance to run the tracer process or gc will be finished before the trace set.

Conceptually it's is not different from when we are mocking module with `meck` and then waiting for some of its functions to be called, but I had to use trace here because you can't mock `garbage_collection`. Plus, unlike mock, this is a real deal, we are checking actual, not simulated functionality.

eiri:
Hey @tonysun83 , was my explanation sufficient? Do you want me to change anything in the tests?

tonysun83:
+1

PR #1069, opened 2017-12-15
Mango: change catch-all field range priority

willholley:
## Overview

01252f97 introduced a "catch-all" feature to Mango that allowed
queries to fall back on a full database scan (_all_docs) when
no valid index was available.

This worked by creating a special index range representing
the full database scan.

For example, a selector:

`{ "_id": "foo" }`

would be translated into a field range of:

`[{ "startkey": "foo", "endkey": "foo"}]`

then prepending the catch-all field range, we would have:

```
[
{ "startkey": null, "endkey": max_json_value},
{ "startkey": "foo", "endkey": "foo"}
]
```

This set gets passed into
mango_cursor_view:choose_best_index to determine most selective
index and field range combination to use. Unfortunately, in
the event that we have one possible index (all_docs) and multiple
valid ranges, it just chooses the first range it finds -
the full index scan in this case.

This commit makes the catch-all field range the last
available option, ensuring we use the more selective
range where available.

## Testing recommendations

Run queries against a large database with no indexes. Check execution_stats and _explain output to see that the startkey and endkey used are appropriate.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_special.erl
+++ b/src/mango/src/mango_cursor_special.erl
@@ -31,10 +31,13 @@
 create(Db, Indexes, Selector, Opts) ->
     InitialRange = mango_idx_view:field_ranges(Selector),
     CatchAll = [{<<"_id">>, {'$gt', null, '$lt', mango_json_max}}],
-    FieldRanges = lists:append(CatchAll, InitialRange),
+    % order matters here - we only want to use the catchall index
+    % if no other range can fulfill the query (because we know)
+    % catchall is the most expensive range
+    FieldRanges = InitialRange ++ CatchAll,
     Composited = mango_cursor_view:composite_indexes(Indexes, FieldRanges),
     {Index, IndexRanges} = mango_cursor_view:choose_best_index(Db, Composited),
-
+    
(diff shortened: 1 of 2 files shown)


src/mango/src/mango_cursor_special.erl
```diff
@@ -31,10 +31,12 @@
...
     InitialRange = mango_idx_view:field_ranges(Selector),
     CatchAll = [{<<"_id">>, {'$gt', null, '$lt', mango_json_max}}],
-    FieldRanges = lists:append(CatchAll, InitialRange),
+    % order matters here - we only want to use the catchall index
+    % if no other range can fulfill the query (because we know)
+    % catchall is the most expensive range
+    FieldRanges = InitialRange ++ CatchAll,
     Composited = mango_cursor_view:composite_indexes(Indexes, FieldRanges),
     {Index, IndexRanges} = mango_cursor_view:choose_best_index(Db, Composited),
-
```
tonysun83: whitespace change

tonysun83:
+1 after whitespace issue, nice catch.

Issue #977, opened 2017-11-10
COPY doc doesn't decode the destination header


garrensmith:
We have an issue when a user tries to clone/copy a doc in Fauxton and the new doc id has Unicode characters. The Fauxton URL request to CouchDB encodes the user input and sets the Destination header with the encoded value of the new ID when sending the COPY request. CouchDB does not decode this, and just creates the ID including escaped values literally.

#Context
Cloning/copying a document with a new ID that includes Unicode chars (e.g. xxx啊xxx) fails because CouchDB takes the encoded value literally, creating a doc with ID xxx%E5%95%8Axxx instead of xxx啊xxx. E.g.:
curl 'https://antonio-cloudant.cloudant.com/appts1/original1' -X COPY -H 'Content-Type: application/json' -H 'Destination: xxx%E5%95%8Axxx'

Note that using xxx啊xxx directly in curl works. The new doc will have ID xxx啊xxx:
curl 'https://antonio-cloudant.cloudant.com/appts1/original1' -X COPY -H 'Content-Type: application/json' -H 'Destination: xxx啊xxx'.
Which means the API works although it doesn't follow 100% the Destination header's spec which states the header should contain an URI, hence its value should be encoded if it includes characters outside the valid range.

CouchDB COPY doc: http://docs.couchdb.org/en/2.0.0/api/document/common.html#copy--db-docid

## Possible Solution
From the Fauxton side we might bypass the COPY api and just create a new doc via the PUT request. But this doesn't solve the CouchDB issue. Ideally CouchDB needs to decode the destination header.

## Steps to Reproduce (for bugs)
1. Go to a database in Fauxton
2. Click on a document.
3. Click on Clone Document
4. Add a doc id like `this/is/an/example` or `xxx啊xxx`
5. Click on Clone document
6. The newly created document will be incorrect as it will be the url encoded doc.

garrensmith:
Just to note that most of the bug report research was done by @Antonio-Maranhao.

PR #1072, opened 2017-12-18
Decode destination header for doc copy

jiangphcn:
## Overview

 The Fauxton URL request to CouchDB encodes the user input and sets the Destination header with the encoded value of the new ID when sending the COPY request. This PR is aimed to make CouchDB to decode Destination header, and creates the ID without escaped values.

```
jiangphs-mbp-2:geotest jiangph$ curl -u foo:bar http://127.0.0.1:15984/testdb1/point11 -X COPY -H 'Content-Type: application/json'  -H 'Destination: xxx%E5%95%8Axxx'
{"ok":true,"ok":true,"id":"xxx啊xxx","rev":"1-ffb6c2ae737918f8f911e1fce668333e"}
jiangphs-mbp-2:geotest jiangph$ curl -u foo:bar http://127.0.0.1:15984/testdb1/point11 -X COPY -H 'Content-Type: application/json'  -H 'Destination: this/is/an/example'
{"ok":true,"ok":true,"id":"this/is/an/example","rev":"1-ffb6c2ae737918f8f911e1fce668333e"}
```

## Testing recommendations

```
make check skip_deps+=couch_epi apps=chttpd tests=all_test_
chttpd db tests
  chttpd_db_test:84: should_return_ok_true_on_bulk_update...[0.089 s] ok
  chttpd_db_test:111: should_accept_live_as_an_alias_for_continuous...[0.043 s] ok
  chttpd_db_test:126: should_return_404_for_delete_att_on_notadoc...[0.004 s] ok
  chttpd_db_test:148: should_return_409_for_del_att_without_rev...[0.038 s] ok
  chttpd_db_test:166: should_return_200_for_del_att_with_rev...[0.069 s] ok
  chttpd_db_test:187: should_return_409_for_put_att_nonexistent_rev...[0.011 s] ok
  chttpd_db_test:282: should_return_correct_id_on_doc_copy...[0.083 s] ok
  [done in 3.898 s]
  All 7 tests passed.
==> rel (eunit)
==> couchdb (eunit)
```

issue #977

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -851,7 +851,8 @@ db_doc_req(#httpd{method='COPY', user_ctx=Ctx}=Req, Db, SourceDocId) ->
         missing_rev -> nil;
         Rev -> Rev
     end,
-    {TargetDocId, TargetRevs} = couch_httpd_db:parse_copy_destination_header(Req),
+    {TargetDocId0, TargetRevs} = couch_httpd_db:parse_copy_destination_header(Req),
+    TargetDocId = list_to_binary(mochiweb_util:unquote(TargetDocId0)),
     % open old doc
     Doc = couch_doc_open(Db, SourceDocId, SourceRev, []),
     % save new doc
--- a/src/chttpd/test/chttpd_db_test.erl
+++ b/src/chttpd/test/chttpd_db_test.erl
@@ -19,6 +19,10 @@
 -define(PASS, "pass").
 -define(AUTH, {basic_auth, {?USER, ?PASS}}).
 -define(CONTENT_JSON, {"Content-Type", "application/json"}).
+-define(DESTHEADER1, {"Destination", "foo%E5%95%8Abar"}).
+-define(DESTHEADER2, {"Destination", "foo%2Fbar%23baz%3Fpow%3Afiz"}).
+
+
 -define(FIXTURE_TXT, ?ABS_PATH(?FILE)).
 -define(i2l(I), integer_to_list(I)).
 
@@ -70,7 +74,8 @@ all_test_() ->
                     fun should_succeed_on_all_docs_with_queries_limit_skip/1,
                     fun should_succeed_on_all_docs_with_multiple_queries/1,
                     fun should_succeed_on_design_docs_with_multiple_queries/1,
-                    fun should_fail_on_multiple_queries_with_keys_and_queries/1
+                    fun should_fail_on_multiple_queries_with_keys_and_queries/1,
+                    fun should_return_correct_id_on_doc_copy/1
                 ]
             }
         }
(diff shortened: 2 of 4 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -825,7 +825,8 @@ db_doc_req(#httpd{method='COPY', user_ctx=Ctx}=Req, Db, SourceDocId) ->
         missing_rev -> nil;
         Rev -> Rev
     end,
-    {TargetDocId, TargetRevs} = couch_httpd_db:parse_copy_destination_header(Req),
+    {TargetDocId0, TargetRevs} = couch_httpd_db:parse_copy_destination_header(Req),
+    TargetDocId = list_to_binary(mochiweb_util:unquote(TargetDocId0)),
```
rnewson: I think the problem here is that the doc id isn't always escaped (that change is fairly recent).

rnewson:
There's several places where we url encode and url decode, and we've changed where we do that over multiple versions of couchdb. I'm struggling to see if we can call unquote unilaterally here. Do we need to care about this across versions or is it sufficient to be internally consistent?

src/chttpd/src/chttpd_db.erl
jiangphcn: Thanks, Bob. According to statement from Garren in issue #977

```
Which means the API works although it doesn't follow 100% the Destination header's spec which states the header should contain an URI, hence its value should be encoded if it includes characters outside the valid range.
```

If spec implies that destination header for couchdb should be encoded, we need to decode here. If there is no escaped character in docid for target doc, calling `mochiweb_util:unquote/1` will not change the content of docid.

jiangphcn:
https://tools.ietf.org/html/rfc4918#section-10.3
https://tools.ietf.org/html/rfc3986

From above links, we know that
```
10.3.  Destination Header

   The Destination request header specifies the URI that identifies a
   destination resource for methods such as COPY and MOVE, which take
   two URIs as parameters.

      Destination = "Destination" ":" Simple-ref


   If the Destination value is an absolute-URI (Section 4.3 of
   [RFC3986]), it may name a different server (or different port or
   scheme).  If the source server cannot attempt a copy to the remote
   server, it MUST fail the request.  Note that copying and moving
   resources to remote servers is not fully defined in this
   specification (e.g., specific error conditions).
```

```
Abstract

   A Uniform Resource Identifier (URI) is a compact sequence of
   characters that identifies an abstract or physical resource.  This
   specification defines the generic URI syntax and a process for
   resolving URI references that might be in relative form, along with
   guidelines and security considerations for the use of URIs on the
   Internet.  The URI syntax defines a grammar that is a superset of all
   valid URIs, allowing an implementation to parse the common components
   of a URI reference without knowing the scheme-specific requirements
   of every possible identifier.  This specification does not define a
   generative grammar for URIs; that task is performed by the individual
   specifications of each URI scheme.
```

jiangphcn:
My understanding is that it is not restricted to encode or un-encode URI. For with url encoding, it is safe from security's point of view.

When discussing with Bob and Paul, Bob mentioned "we should encode the destination header value, and decode it wherever we might need to (if anywhere)".

Here is one place we need to decode URI.

Issue #822, opened 2017-09-14
Provide info=true and keys=[] arguments for /_all_dbs


wohali:
## Expected Behavior
Right now, if I want the database info for all databases on a server, I must individually access /{db} for each database. It'd be swell to make a single request to retrieve all of that.

While I'm at it, it'd be extra nice if I could also specify keys=[....] to select a subset of all databases on a server to monitor.

## Current Behavior
N requests must be made for N databases if info is to be retrieved on all databases.

jiangphcn:
welcome back @wohali ! I created PR #1082 for this issue. Would you please like to review them? Thanks

iilyak:
My 2¢

Accessing all databases on the server is expensive operation. Since get_db_info goes through `couch_server` process. For clusters with big number of databases opening all of them at the same time would cause a series of evictions from `couch_lru`. This in its turn causes lots of update_lru messages to couch_server. 

We could try to modify `couch_db:get_db_info` to open database in read only mode without using `couch_server`. However with this approach we wouldn't be able to return the same information as we currently do. Therefore we would have to introduce new endpoint. Which is a good thing since it would make it expressible in OpenAPI spec which we can benefit from latter on. 

Following keys are problematic:
  * update_seq
  * purge_seq
  * compact_running
  * instance_start_time
  * committed_update_seq
  * compacted_seq

davisp:
@iilyak What's the issue with update_seq and purge_seq? Those are both in the header. I understand why the others are an issue.

Also I do understand the read-only approach, though I wonder if a "do-not-cache" option to couch_server:open might not also be a possibility here? Just spitballing on that front though.

iilyak:
You are right on both fronts `update_seq` and `purge_seq` could be retrieved and `do-not-cache` approach might work.

PR #1082, opened 2017-12-28
Introduce new _dbs_info endpoint to get info of a list of databases

jiangphcn:
## Overview

Before this PR, if you want the database info for all databases on a server, you must individually access /{db} for each database. This PR makes a single request to retrieve all of that.

You can specify keys=[....] to select a subset of all databases on a server to monitor.

```
curl -X POST 'localhost:15984/_dbs_info' -u foo:bar -H "Content-Type: application/json"  -d @dblist.json -g|jq .
100   971    0   938  100    33   104k   3770 --:--:-- --:--:-- --:--:--  114k
[
  {
    "key": "dbname1",
    "info": {
      "db_name": "dbname1",
      "update_seq": "0-g1AAAAFDeJzLYWBg4MhgTmHgz8tPSTV0MDQy1zMAQsMcoARTIkOS_P___7MSGXAqSVIAkkn2hFQ5gFTFE1KVAFJVT0BVHguQZGgAUkCF85FVGmNVuQCicj9hMw9AVN4nrPIBRCXInVkAHz1WeA",
      "sizes": {
        "file": 33936,
        "external": 0,
        "active": 0
      },
      "purge_seq": 0,
      "other": {
        "data_size": 0
      },
      "doc_del_count": 0,
      "doc_count": 0,
      "disk_size": 33936,
      "disk_format_version": 6,
      "data_size": 0,
      "compact_running": false,
      "cluster": {
        "q": 8,
        "n": 3,
        "w": 2,
        "r": 2
      },
      "instance_start_time": "0"
    }
  },
  {
    "key": "dbname2",
    "info": {
      "db_name": "dbname2",
      "update_seq": "0-g1AAAAFDeJzLYWBg4MhgTmHgz8tPSTV0MDQy1zMAQsMcoARTIkOS_P___7MSGXAqSVIAkkn2hFQ5gFTFE1KVAFJVT0BVHguQZGgAUkCF8wmrXABRuZ-wygMQlfcJq3wAUQlyZxYAHlFWdg",
      "sizes": {
        "file": 33936,
        "external": 0,
        "active": 0
      },
      "purge_seq": 0,
      "other": {
        "data_size": 0
      },
      "doc_del_count": 0,
      "doc_count": 0,
      "disk_size": 33936,
      "disk_format_version": 6,
      "data_size": 0,
      "compact_running": false,
      "cluster": {
        "q": 8,
        "n": 3,
        "w": 2,
        "r": 2
      },
      "instance_start_time": "0"
    }
  }
]
```
```
more dblist.json
{"keys" : ["dbname1", "dbname2"]}
```

```
curl -X GET 'localhost:15984/_dbs_info?keys=["db1"]' -u foo:bar -g|jq .
100    60  100    60    0     0  10284      0 --:--:-- --:--:-- --:--:-- 12000
{
  "error": "method_not_allowed",
  "reason": "Only POST allowed"
}
```

for database which doesn't exist, there is the following behavior.
```
 curl -X POST 'localhost:15984/_dbs_info' -u foo:bar -H "Content-Type: application/json"  -d @dblist.json -g|jq .
100   555    0   519  100    36  13046    904 --:--:-- --:--:-- --:--:-- 13307
[
  {
    "key": "dbname1",
    "info": {
      "db_name": "dbname1",
      "update_seq": "0-g1AAAAFDeJzLYWBg4MhgTmHgz8tPSTV0MDQy1zMAQsMcoARTIkOS_P___7MSGeBKjNCUJCkAySR7NFXoBiU5gFTFE1KVAFJVT0BVHguQZGgAUkCF8wmrXABRuZ-wygMQlffx-Rei8gFEJcidWQAfhVZ4",
      "sizes": {
        "file": 33936,
        "external": 0,
        "active": 0
      },
      "purge_seq": 0,
      "other": {
        "data_size": 0
      },
      "doc_del_count": 0,
      "doc_count": 0,
      "disk_size": 33936,
      "disk_format_version": 6,
      "data_size": 0,
      "compact_running": false,
      "cluster": {
        "q": 8,
        "n": 3,
        "w": 2,
        "r": 2
      },
      "instance_start_time": "0"
    }
  },
  {
    "key": "unknown_db",
    "error": "not_found"
  }
]
```
Note:
1. Only POST is supported
2. one configuration item is introduced to limit maximum number of databases 
```
; Limit maximum number of databases when tying to get detailed information using
; _dbs_info in a request
max_db_number_for_dbs_info_req = 100
```
## Testing recommendations

make check skip_deps+=couch_epi apps=chttpd tests=dbs_info_test_
```

Compiled test/chttpd_dbs_info_test.erl
    Running test function(s):
      chttpd_dbs_info_test:dbs_info_test_/0
chttpd dbs info tests
  chttpd_dbs_info_test:73: should_return_error_for_get_db_info...[0.013 s] ok
  chttpd_dbs_info_test:86: should_return_dbs_info_for_single_db...[0.045 s] ok
  chttpd_dbs_info_test:102: should_return_dbs_info_for_multiple_dbs...[0.006 s] ok
  chttpd_dbs_info_test:123: should_return_error_for_exceeded_keys...[0.014 s] ok
  chttpd_dbs_info_test:139: should_return_error_for_missing_keys...[0.004 s] ok
  chttpd_dbs_info_test:153: should_return_dbs_info_for_dbs_with_mixed_state...[0.006 s] ok
  [done in 2.399 s]
  All 6 tests passed.
==> rel (eunit)
```
issue #822

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_misc.erl
+++ b/src/chttpd/src/chttpd_misc.erl
@@ -14,6 +14,7 @@
 
 -export([
     handle_all_dbs_req/1,
+    handle_dbs_info_req/1,
     handle_node_req/1,
     handle_favicon_req/1,
     handle_favicon_req/2,
@@ -36,6 +37,8 @@
 -import(chttpd,
     [send_json/2,send_json/3,send_method_not_allowed/2,
     send_chunk/2,start_chunked_response/3]).
+
+-define(MAX_DB_NUM_FOR_DBS_INFO, 100).
 
 % httpd global handlers
 
@@ -140,6 +143,38 @@ all_dbs_callback(complete, #vacc{resp=Resp0}=Acc) ->
 all_dbs_callback({error, Reason}, #vacc{resp=Resp0}=Acc) ->
     {ok, Resp1} = chttpd:send_delayed_error(Resp0, Reason),
     {ok, Acc#vacc{resp=Resp1}}.
+
+handle_dbs_info_req(#httpd{method='POST'}=Req) ->
+    chttpd:validate_ctype(Req, "application/json"),
+    Props = chttpd:json_body_obj(Req),
+    Keys = couch_mrview_util:get_view_keys(Props),
+    case Keys of
+        undefined -> throw({bad_request, "`keys` member must exist."});
+        _ -> ok
+    end,
+    MaxNumber = config:get_integer("chttpd",
+        "max_db_number_for_dbs_info_req", ?MAX_DB_NUM_FOR_DBS_INFO),
+    case length(Keys) =< MaxNumber of
+        true -> ok;
+        false -> throw({bad_request, too_many_keys})
+    end,
+    {ok, Resp} = chttpd:start_json_response(Req, 200),
+    send_chunk(Resp, "["),
+    lists:foldl(fun(DbName, AccSeparator) ->
+        case catch fabric:get_db_info(DbName) of
+            {ok, Result} ->
+                Json = ?JSON_ENCODE({[{key, DbName}, {info, {Result}}]}),
+                send_chunk(Resp, AccSeparator ++ Json);
+            _ ->
+                Json = ?JSON_ENCODE({[{key, DbName}, {error, not_found}]}),
+                send_chunk(Resp, AccSeparator ++ Json)
+        end,
+        "," % AccSeparator now has a comma
+    end, "", Keys),
+    send_chunk(Resp, "]"),
+    chttpd:end_json_response(Resp);
+handle_dbs_info_req(Req) ->
+    send_method_not_allowed(Req, "POST").
 
 handle_task_status_req(#httpd{method='GET'}=Req) ->
     {Replies, _BadNodes} = gen_server:multi_call(couch_task_status, all),
--- a/src/chttpd/test/chttpd_dbs_info_test.erl
+++ b/src/chttpd/test/chttpd_dbs_info_test.erl
@@ -1,0 +1,169 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(chttpd_dbs_info_test).
+
+-include_lib("couch/include/couch_eunit.hrl").
+-include_lib("couch/include/couch_db.hrl").
+
+-define(USER, "chttpd_db_test_admin").
+-define(PASS, "pass").
+-define(AUTH, {basic_auth, {?USER, ?PASS}}).
+-define(CONTENT_JSON, {"Content-Type", "application/json"}).
+
+
+setup() ->
+    Hashed = couch_passwords:hash_admin_password(?PASS),
+    ok = config:set("admins", ?USER, ?b2l(Hashed), _Persist=false),
+    Addr = config:get("chttpd", "bind_address", "127.0.0.1"),
+    Port = mochiweb_socket_server:get(chttpd, port),
+    Url = lists:concat(["http://", Addr, ":", Port, "/"]),
+    Db1Url = lists:concat([Url, "db1"]),
+    create_db(Db1Url),
+    Db2Url = lists:concat([Url, "db2"]),
+    create_db(Db2Url),
+    Url.
+
+teardown(Url) ->
+    Db1Url = lists:concat([Url, "db1"]),
+    Db2Url = lists:concat([Url, "db2"]),
+    delete_db(Db1Url),
+    delete_db(Db2Url),
+    ok = config:delete("admins", ?USER, _Persist=false).
+
+create_db(Url) ->
+    {ok, Status, _, _} = test_request:put(Url, [?CONTENT_JSON, ?AUTH], "{}"),
+    ?assert(Status =:= 201 orelse Status =:= 202).
+
+delete_db(Url) ->
+    {ok, 200, _, _} = test_request:delete(Url, [?AUTH]).
+
+dbs_info_test_() ->
+    {
+        "chttpd dbs info tests",
+        {
+            setup,
+            fun chttpd_test_util:start_couch/0, fun chttpd_test_util:stop_couch/1,
+            {
+                foreach,
+                fun setup/0, fun teardown/1,
+                [
+                    fun should_return_error_for_get_db_info/1,
+                    fun should_return_dbs_info_for_single_db/1,
+                    fun should_return_dbs_info_for_multiple_dbs/1,
+                    fun should_return_error_for_exceeded_keys/1,
+                    fun should_return_error_for_missing_keys/1,
+                    fun should_return_dbs_info_for_dbs_with_mixed_state/1
+                ]
+            }
+        }
+    }.
+
+
+should_return_error_for_get_db_info(Url) ->
+    ?_test(begin
+        {ok, Code, _, ResultBody} = test_request:get(Url ++ "/_dbs_info?"
+            ++ "keys=[\"db1\"]", [?CONTENT_JSON, ?AUTH]),
+        {Body} = jiffy:decode(ResultBody),
+        [
+            ?assertEqual(<<"method_not_allowed">>,
+                couch_util:get_value(<<"error">>, Body)),
+            ?assertEqual(405, Code)
+        ]
+    end).
+
+
+should_return_dbs_info_for_single_db(Url) ->
+    ?_test(begin
+        NewDoc = "{\"keys\": [\"db1\"]}",
+        {ok, _, _, ResultBody} = test_request:post(Url ++ "/_dbs_info/",
+            [?CONTENT_JSON, ?AUTH], NewDoc),
+        BodyJson = jiffy:decode(ResultBody),
+        {Db1Data} = lists:nth(1, BodyJson),
+        [
+            ?assertEqual(<<"db1">>,
+                couch_util:get_value(<<"key">>, Db1Data)),
+            ?assertNotEqual(undefined,
+                couch_util:get_value(<<"info">>, Db1Data))
+        ]
+    end).
+
+
+should_return_dbs_info_for_multiple_dbs(Url) ->
+    ?_test(begin
+        NewDoc = "{\"keys\": [\"db1\", \"db2\"]}",
+        {ok, _, _, ResultBody} = test_request:post(Url ++ "/_dbs_info/",
+            [?CONTENT_JSON, ?AUTH], NewDoc),
+        BodyJson = jiffy:decode(ResultBody),
+        {Db1Data} = lists:nth(1, BodyJson),
+        {Db2Data} = lists:nth(2, BodyJson),
+        [
+            ?assertEqual(<<"db1">>,
+                couch_util:get_value(<<"key">>, Db1Data)),
+            ?assertNotEqual(undefined,
+                couch_util:get_value(<<"info">>, Db1Data)),
+            ?assertEqual(<<"db2">>,
+                couch_util:get_value(<<"key">>, Db2Data)),
+            ?assertNotEqual(undefined,
+                couch_util:get_value(<<"info">>, Db2Data))
+        ]
+    end).
+
+
+should_return_error_for_exceeded_keys(Url) ->
+    ?_test(begin
+        NewDoc = "{\"keys\": [\"db1\", \"db2\"]}",
+        ok = config:set("chttpd", "max_db_number_for_dbs_info_req", "1"),
+        {ok, Code, _, ResultBody} = test_request:post(Url ++ "/_dbs_info/",
+                   [?CONTENT_JSON, ?AUTH], NewDoc),
+        {Body} = jiffy:decode(ResultBody),
+        ok = config:delete("chttpd", "max_db_number_for_dbs_info_req"),
+        [
+            ?assertEqual(<<"bad_request">>,
+                couch_util:get_value(<<"error">>, Body)),
+            ?assertEqual(400, Code)
+        ]
+    end).
+
+
+should_return_error_for_missing_keys(Url) ->
+    ?_test(begin
+        NewDoc = "{\"missingkeys\": [\"db1\", \"db2\"]}",
+        {ok, Code, _, ResultBody} = test_request:post(Url ++ "/_dbs_info/",
+            [?CONTENT_JSON, ?AUTH], NewDoc),
+        {Body} = jiffy:decode(ResultBody),
+        [
+            ?assertEqual(<<"bad_request">>,
+                couch_util:get_value(<<"error">>, Body)),
+            ?assertEqual(400, Code)
+        ]
+    end).
+
+
+should_return_dbs_info_for_dbs_with_mixed_state(Url) ->
+    ?_test(begin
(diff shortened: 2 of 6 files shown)


src/chttpd/src/chttpd_misc.erl
```diff
@@ -103,6 +103,52 @@ maybe_add_csp_headers(Headers, _) ->
     Headers.
 
 handle_all_dbs_req(#httpd{method='GET'}=Req) ->
+    Params = lists:flatmap(fun({K, V}) -> parse_all_dbs_param(K, V) end,
```
iilyak: Inconsistent formatting of funs. I think most of our codebase uses following formatting.
 ```
lists:flatmap(fun(Element) -> 
   <function body>
end, List),
```

Comment (iilyak) [reply src/chttpd/src/chttpd_misc.erl]: Unless whole expression fits the max line length (which is 80 characters).

**Comment by iilyak** [comment]
This PR introduces new APIs. Therefore libraries would need an update. @ricellis could you take a look?

File: src/chttpd/src/chttpd_misc.erl [hunk 2]
```diff
@@ -442,3 +486,25 @@ message_queues(Registered) ->
...
+    throw({query_parse_error, ?l2b(Msg)}).
+
+parse_all_dbs_param("info", Value) ->
+    [{info, parse_bool_param("info", Value)}];
+parse_all_dbs_param("keys", Value) ->
+    [{keys, ?JSON_DECODE(Value)}];
+parse_all_dbs_param(Key, Value) ->
+    [{extra, {Key, Value}}].
+
+validate_alldbs_query(info, Value, Args) ->
```
iilyak: should it be named `validate_all_dbs_query`? To be consistent with:
 -  `parse_all_dbs_param`
 - `handle_all_dbs_req`

src/chttpd/src/chttpd_misc.erl
iilyak: If you change `parse_all_dbs_param(K, V)` to accept tuple you would be able to write:

```
lists:flatmap(fun parse_all_dbs_param/1, chttpd:qs(Req)),
```

src/chttpd/src/chttpd_misc.erl
```diff
@@ -103,6 +103,52 @@ maybe_add_csp_headers(Headers, _) ->
     Headers.
 
 handle_all_dbs_req(#httpd{method='GET'}=Req) ->
+    Params = lists:flatmap(fun({K, V}) -> parse_all_dbs_param(K, V) end,
+        chttpd:qs(Req)),
+    Args = lists:foldl(fun({K, V}, Arg) -> validate_alldbs_query(K, V, Arg)
```
iilyak: Same here. If you change validate_alldbs_query to accept tuple and accumulator you could write:

```
Args = lists:foldl(fun validate_alldbs_query/2, Params),
```

src/chttpd/src/chttpd_misc.erl
```diff
@@ -103,6 +103,52 @@ maybe_add_csp_headers(Headers, _) ->
...
+    maybe_handle_all_dbs_info(Req, Args#alldbs_query_args.info,
+        Args#alldbs_query_args.keys);
+handle_all_dbs_req(Req) ->
+    send_method_not_allowed(Req, "GET,HEAD").
+
+maybe_handle_all_dbs_info(Req, true, []) ->
+    Params = couch_mrview_http:parse_params(Req, undefined),
+    ShardDbName = config:get("mem3", "shards_db", "_dbs"),
+    Options = [{user_ctx, Req#httpd.user_ctx}],
+    Callback = fun({meta, _}, []) ->
```
iilyak: weird indentation. I would write it as:
```
Callback = fun
    ({meta, _}, []) ->
        {ok, []};
    ({row, Props}, Acc) ->
        something
end, 
```
If you prefer to keep your style please add 4 more spaces before `{ok, []};`

src/chttpd/src/chttpd_misc.erl
```diff
@@ -442,3 +486,25 @@ message_queues(Registered) ->
         {Type, Length} = process_info(whereis(Name), Type),
         {Name, Length}
     end, Registered).
+
+parse_bool_param(_, Val) when is_boolean(Val) ->
+    Val;
+parse_bool_param(_, "true") -> true;
+parse_bool_param(_, "false") -> false;
+parse_bool_param(Name, Val) ->
+    Msg = io_lib:format("Invalid value for ~s: ~p", [Name, Val]),
```
iilyak: The messaging doesn't suggest that boolean is expected.

src/chttpd/src/chttpd_misc.erl
```diff
@@ -103,6 +103,52 @@ maybe_add_csp_headers(Headers, _) ->
...
+    Params = lists:flatmap(fun({K, V}) -> parse_all_dbs_param(K, V) end,
+        chttpd:qs(Req)),
+    Args = lists:foldl(fun({K, V}, Arg) -> validate_alldbs_query(K, V, Arg)
+        end, #alldbs_query_args{}, Params),
+    maybe_handle_all_dbs_info(Req, Args#alldbs_query_args.info,
+        Args#alldbs_query_args.keys);
+handle_all_dbs_req(Req) ->
+    send_method_not_allowed(Req, "GET,HEAD").
+
+maybe_handle_all_dbs_info(Req, true, []) ->
```
iilyak: I don't think we should allow getting info about all databases. It is too expensive and it is a perfect way to make a couch_server unresponsive due to message queue size.

src/chttpd/src/chttpd_misc.erl
```diff
@@ -103,6 +103,52 @@ maybe_add_csp_headers(Headers, _) ->
...
+            {error, Reason}
+    end,
+    {ok, Keys} = fabric:all_docs(ShardDbName, Options, Callback, [], Params),
+    case Keys of
+        [] ->
+            chttpd:send_response(Req, 200, [], <<"[]">>);
+        _ ->
+            maybe_handle_all_dbs_info(Req, true, Keys)
+    end;
+maybe_handle_all_dbs_info(Req, true, Keys) ->
```
iilyak: We should have a config which limits the number of databases in Keys to prevent DOS from malicious users.

ricellis:
I wonder if accepting `keys` should only be part of a `POST` body. I think this has two advantages:
i) It avoids the URL length limitations that we've seen previously with including long lists of keys in a URL.
ii) It also avoids URL encoding work for the caller (although admittedly that is less of an issue with a DB name than with, for example, view keys).

This would be consistent with `_all_docs` and `_view` which allow `POST` of a list of keys. Whilst those endpoints also support `keys` as a query parameter I think that is only for backwards compatibility - so for a new API accepting `POST` only for `keys` could make the most sense.

IMO one of the pain points in using the CouchDB API in the past has been changes in the response schema for the same endpoint (based on e.g. a query parameter). It seems like this is proposing to introduce yet another example of this similar problem:
`_all_dbs` -> `["db1", "db2"]` (list of strings)
`_all_dbs?info=true` -> [{"key": "db1", "info": {...}}, {"key": "db2", "info": {...}}]` (list of objects)

I think using a new endpoint e.g. `_all_dbs_info` (or whatever) makes a more usable API because it can be made clear that the returned schema is different than that used for `_all_dbs`. Further, given that it appears this operation isn't even going to be allowed for cases where there are more than 100 DBs, then something like `_dbs_info` makes more sense as a name than `_all_dbs` (since it isn't necessarily going to be all the DBs).

Combining the above with the concern that this should probably be limited to avoid retrieving metadata for all the DBs at once under most circumstances then perhaps a more intuitive API would be one that only accepts a list of keys e.g.:

`POST` `_dbs_info` `{"keys" : ["dbName1", "dbName2"]}`

jiangphcn:
@ricellis good suggestions from API's point of view.
```
POST _dbs_info {"keys" : ["dbName1", "dbName2"]}
```
Ping @davisp @rnewson @mikerhodes for further comments.

If not, I can continue to change to provide above new API :-)

davisp:
+1 to _dbs_info

jiangphcn:
Introduced new _dbs_info endpoint to get info of a list of databases using 4a84a1e.

wohali:
Is _all_dbs?info=true retained, with the new _dbs_info endpoint added instead of specifying a keys=[] query argument?

If so I am completely +1, LGTM!

jiangphcn:
hi @wohali, with introduction of `_dbs_info` endpoint, the current intention is not to introduce _all_dbs?info=true anymore. I think the reason includes
- `_dbs_info` can have same function as _all_dbs?info=true and clearer
- using `_dbs_info` instead of _all_dbs?info=true  can avoid to have implication of query of information of a large set of databases.

src/chttpd/src/chttpd_misc.erl
```diff
@@ -141,6 +144,38 @@ all_dbs_callback({error, Reason}, #vacc{resp=Resp0}=Acc) ->
     {ok, Resp1} = chttpd:send_delayed_error(Resp0, Reason),
     {ok, Acc#vacc{resp=Resp1}}.
 
+handle_dbs_info_req(#httpd{method='POST'}=Req) ->
+    chttpd:validate_ctype(Req, "application/json"),
+    Props = couch_httpd:json_body_obj(Req),
```
iilyak: I think it would be better if you would use functions which are defined in the current application. The `chttpd:json_body_obj/1` provides the same functionality.

src/chttpd/src/chttpd_misc.erl
```diff
@@ -141,6 +144,38 @@ all_dbs_callback({error, Reason}, #vacc{resp=Resp0}=Acc) ->
...
+    Keys = couch_mrview_util:get_view_keys(Props),
```
iilyak: I am not sure about it. On one side the `get_view_keys` does the validation we need. On the other side this has nothing to do with views.
Feel free to ignore.

src/chttpd/src/chttpd_misc.erl
```diff
@@ -141,6 +144,38 @@ all_dbs_callback({error, Reason}, #vacc{resp=Resp0}=Acc) ->
...
+    case Keys of
+        undefined -> throw({bad_request, "`keys` member must exist."});
+        _ -> ok
+    end,
+    MaxNumber = config:get_integer("chttpd",
+        "max_db_number_for_dbs_info_req", ?MAX_DB_NUM_FOR_DBS_INFO),
+    case length(Keys) =< MaxNumber of
+        true -> ok;
+        false -> throw({bad_request, request_keys_too_large})
```
iilyak: I might be wrong, but I parse `large` as `big`. Other non native speakers might have the same confusion. Would `{bad_request, too_many_keys}` be better?

src/chttpd/src/chttpd_misc.erl
```diff
@@ -141,6 +144,38 @@ all_dbs_callback({error, Reason}, #vacc{resp=Resp0}=Acc) ->
...
+    end,
+    {ok, Resp} = chttpd:start_json_response(Req, 200),
+    send_chunk(Resp, "["),
+    lists:foldl(fun(DbName, AccSeparator) ->
+        case catch fabric:get_db_info(DbName) of
+            {ok, Result} ->
+                Json = ?JSON_ENCODE({[{key, DbName}, {info, {Result}}]}),
+                send_chunk(Resp, AccSeparator ++ Json);
+            _ ->
```
iilyak: Please make sure this API endpoint do not allow non authenticated users to check the existence of the database with specified name (Ideally add test for this). You might want to add a rule in `chttpd_auth_request.erl`.

src/chttpd/src/chttpd_misc.erl
jiangphcn: changed to chttpd:json_body_obj/1

src/chttpd/src/chttpd_misc.erl
jiangphcn: Above is just trying to get `keys` but we need to make sure that `keys` exists in current case. So added additional validation below.

src/chttpd/src/chttpd_misc.erl
jiangphcn: changed to  `too_many_keys`

src/chttpd/src/chttpd_misc.erl
jiangphcn: Added a new rule in `chttpd_auth_request.erl`

jiangphcn: working on test right now

jiangphcn: I performed below test
1. create two users: one is admin (`foo/bar`) and another is normal user (`normal_user/member`). 
```
curl -X POST http://localhost:15984/_users      -H "Accept: application/json"      -H "Content-Type: application/json"      -d @member.json
{"ok":true,"id":"org.couchdb.user:normal_user","rev":"1-b39703a91aa01702c53d0297e33bb5f9"}
```
2. create one database (`db1-for-foo`) using admin user (`foo/bar`). 
```
curl -u foo:bar http://localhost:15984/db1-for-foo -X PUT
{"ok":true}
```
3. use normal user (`normal_user/member`) to check db1-for-foo, and can get information of this database. 
```
curl -u normal_user:member http://localhost:15984/db1-for-foo
{"db_name":"db1-for-foo","update_seq":"0-g1AAAAFDeJzLYWBg4MhgTmHgz8tPSTV0MDQy1zMAQsMcoARTIkOS_P___7MSGXAqSVIAkkn2hFQ5gFTFE1KVAFJVT0BVHguQZGgAUkCF8wmrXABRuZ-wygMQlfcJq3wAUQlyZxYAHlFWdg","sizes":{"file":33936,"external":0,"active":0},"purge_seq":0,"other":{"data_size":0},"doc_del_count":0,"doc_count":0,"disk_size":33936,"disk_format_version":6,"data_size":0,"compact_running":false,"cluster":{"q":8,"n":3,"w":2,"r":2},"instance_start_time":"0"}
```
4. use normal user to check `_all_dbs`, and get list with db1-for-foo
```
curl -u normal_user:member http://localhost:15984/_all_dbs
["_global_changes","_replicator","_users","db1-for-foo"]
```

Following above behavior, I make `_dbs_info` to be visible in couchdb.

If there is need to restrict access to `_dbs_info`, we will add logic in upper level.

jiangphcn:
@iilyak thanks a lot for your review. I addressed your comments in couchdb. Could you please review again? Thanks

src/chttpd/src/chttpd_misc.erl
iilyak: Please add `should_deny_non_authenticated` test into `chttpd_dbs_info_test.erl`. The test would be something like (WARNING pseudo code):
```
    ?_test(begin
        {ok, Code, _, ResultBody} = test_request:get(Url ++ "/_dbs_info?"
            ++ "keys=[\"db1\"]", [?CONTENT_JSON]), %% no ?AUTH
        {Body} = jiffy:decode(ResultBody),
        [
            ?assertEqual(<<"unauthorized">>,
                couch_util:get_value(<<"error">>, Body)),
            ?assertEqual(401, Code)
        ]
    end).
```

src/chttpd/test/chttpd_dbs_info_test.erl
```diff
@@ -0,0 +1,169 @@
...
+    }.
+
+
+should_return_error_for_get_db_info(Url) ->
+    ?_test(begin
+        {ok, Code, _, ResultBody} = test_request:get(Url ++ "/_dbs_info?"
+            ++ "keys=[\"db1\"]", [?CONTENT_JSON, ?AUTH]),
+        {Body} = jiffy:decode(ResultBody),
+        [
+            ?assertEqual(405, Code),
```
iilyak: Please swap the order of assertions for every test in a test suite.
You would have:
 1. assert on error 
 2. assert on return code

If `couch_util:get_value(<<"error">>, Body))` assertion goes first we would see a reason in test failure. Currently we would see return code don't match but not a meaningful error.

Issue #969, opened 2017-11-07
GET /db/_all_docs?update_seq=true missing 'offset' and 'update_seq' when 'keys' also provided


dharders:
## Expected Behavior

A `GET /db/_all_docs?update_seq=true&keys=["aValidKey"]` should return 
```javascript
{
  total_rows: X,
  rows: [{id: "aValidKey", ...}],
  offset: Y,
  update_seq: "Z-ABC..."
}
```
## Current Behavior

Instead it returns
```javascript
{
  total_rows: X,
  rows: [{id: "aValidKey", ...}]
}
```
i.e. it is **missing 'offset' and 'update_seq'** when 'keys' also provided.

**NOTE:** this is also true for `POST`

## Possible Solution

Specifying `key` instead of `keys` produces the desired behavior so possibly check that code path

i.e.  `GET /db/_all_docs?update_seq=true&key="aValidKey"` works as expected

## Steps to Reproduce (for bugs)

1. 
```javascript
 curl -X PUT "localhost:5984/testdb"
```
2. 
```javascript
curl -X POST -H "Content-Type: application/json" -d '{"docs": [{"_id": "doc1"}, {"_id": "doc2"}]}' "localhost:5984/testdb/_bulk_docs"
```
3.
```javascript
curl -X GET 'localhost:5984/testdb/_all_docs?update_seq=true&keys=\["doc1"\]'
```
4. Response is missing `offset` and `update_seq`
5. Below response is as expected when using `key` instead of `keys`
```javascript
curl -X GET 'localhost:5984/testdb/_all_docs?update_seq=true&key="doc1"'
```
6. Response should contain `offset` and `update_seq` as expected
7. POST is also not returning `offset` and `update_seq`
```javascript
curl -X POST -H "Content-Type: application/json" -d '{"keys": ["doc1"]}' 'localhost:5984/testdb/_all_docs?update_seq=true'
```
## Context

Noticed this bug when putting together a PR to add a feature to PouchDB (see [PouchDB issue 6230](https://github.com/pouchdb/pouchdb/issues/6230), i.e. missing `update_seq` option in allDocs API)

## Your Environment

* Version used: CouchDB v2.1.0
* Browser Name and version: curl, Chrome 61.0.3163.100 (Official Build) (64-bit)
* Operating System and version (desktop or mobile): Windows 10 64-bit Desktop

jiangphcn:
@dharders I have issued one PR  #1085 to address your issue. Currently, it is under review process.

[...]

PR #1085, opened 2018-01-03
Return update_seq and offset when update_seq is true and keys is set

jiangphcn:
## Overview

Before this PR, if `keys` parameter is set for `_all_docs` endpoint, there is no `update_seq` and `offset` returned in response even if `update_seq` parameter is set to true. This is not consistent with description in http://docs.couchdb.org/en/2.1.1/api/database/bulk-api.html

```
update_seq (boolean) – Response includes an update_seq value indicating which sequence id of the underlying database the view reflects. Default is false.
```

This PR is aimed to return update_seq and offset with null if `update_seq` is set to true and `keys` is specified as well.

```
 curl -X GET 'localhost:15984/db3/_all_docs?update_seq=true&keys=["bar"]&skip=0' -u foo:bar -g|jq .
100   157    0   157    0     0  14275      0 --:--:-- --:--:-- --:--:-- 15700
{
  "total_rows": 14,
  "offset": null,
  "update_seq": null,
  "rows": [
    {
      "id": "bar",
      "key": "bar",
      "value": {
        "rev": "1-9393bddaf4aecebfe0cd04a6558d8731"
      }
    }
  ]
}
```

## Testing recommendations

```
make check skip_deps+=couch_epi apps=chttpd tests=all_test_
chttpd db tests
  chttpd_db_test:78: should_return_ok_true_on_bulk_update...[0.104 s] ok
  chttpd_db_test:105: should_accept_live_as_an_alias_for_continuous...[0.050 s] ok
  chttpd_db_test:120: should_return_404_for_delete_att_on_notadoc...[0.005 s] ok
  chttpd_db_test:142: should_return_409_for_del_att_without_rev...[0.040 s] ok
  chttpd_db_test:160: should_return_200_for_del_att_with_rev...[0.053 s] ok
  chttpd_db_test:181: should_return_409_for_put_att_nonexistent_rev...[0.008 s] ok
  chttpd_db_test:196: should_return_update_seq_when_set_on_all_docs...[0.257 s] ok
  chttpd_db_test:212: should_not_return_update_seq_when_unset_on_all_docs...[0.251 s] ok
  [done in 2.258 s]
  All 8 tests passed.
==> rel (eunit)
```

issue #969

## Checklist

### DIFF
--- a/src/chttpd/test/chttpd_db_test.erl
+++ b/src/chttpd/test/chttpd_db_test.erl
@@ -20,6 +20,7 @@
 -define(AUTH, {basic_auth, {?USER, ?PASS}}).
 -define(CONTENT_JSON, {"Content-Type", "application/json"}).
 -define(FIXTURE_TXT, ?ABS_PATH(?FILE)).
+-define(i2l(I), integer_to_list(I)).
 
 setup() ->
     Hashed = couch_passwords:hash_admin_password(?PASS),
@@ -62,7 +63,9 @@ all_test_() ->
                     fun should_return_404_for_delete_att_on_notadoc/1,
                     fun should_return_409_for_del_att_without_rev/1,
                     fun should_return_200_for_del_att_with_rev/1,
-                    fun should_return_409_for_put_att_nonexistent_rev/1
+                    fun should_return_409_for_put_att_nonexistent_rev/1,
+                    fun should_return_update_seq_when_set_on_all_docs/1,
+                    fun should_not_return_update_seq_when_unset_on_all_docs/1
                 ]
             }
         }
@@ -184,6 +187,34 @@ should_return_409_for_put_att_nonexistent_rev(Url) ->
             {<<"error">>,<<"not_found">>},
             {<<"reason">>,<<"missing_rev">>}]},
             ?JSON_DECODE(RespBody))
+    end).
+
+
+should_return_update_seq_when_set_on_all_docs(Url) ->
+    ?_test(begin
+        [create_doc(Url, "testdoc" ++ ?i2l(I)) || I <- lists:seq(1, 3)],
+        {ok, RC, _, RespBody} = test_request:get(Url ++ "/_all_docs/"
+            ++ "?update_seq=true&keys=[\"testdoc1\"]",[?CONTENT_JSON, ?AUTH]),
+        ?assertEqual(200, RC),
+        {ResultJson} = ?JSON_DECODE(RespBody),
+        ?assertNotEqual(undefined,
+            couch_util:get_value(<<"update_seq">>, ResultJson)),
+        ?assertNotEqual(undefined,
+            couch_util:get_value(<<"offset">>, ResultJson))
+    end).
+
+
+should_not_return_update_seq_when_unset_on_all_docs(Url) ->
+    ?_test(begin
+        [create_doc(Url, "testdoc" ++ ?i2l(I)) || I <- lists:seq(1, 3)],
+        {ok, RC, _, RespBody} = test_request:get(Url ++ "/_all_docs/"
+            ++ "?update_seq=false&keys=[\"testdoc1\"]",[?CONTENT_JSON, ?AUTH]),
+        ?assertEqual(200, RC),
+        {ResultJson} = ?JSON_DECODE(RespBody),
+        ?assertEqual(undefined,
+            couch_util:get_value(<<"update_seq">>, ResultJson)),
+        ?assertNotEqual(undefined,
+            couch_util:get_value(<<"offset">>, ResultJson))
     end).
 
 
--- a/src/fabric/src/fabric_view_all_docs.erl
+++ b/src/fabric/src/fabric_view_all_docs.erl
@@ -59,7 +59,8 @@ go(DbName, Options, QueryArgs, Callback, Acc0) ->
         conflicts = Conflicts,
         skip = Skip,
         keys = Keys0,
-        extra = Extra
+        extra = Extra,
+        update_seq = UpdateSeq
     } = QueryArgs,
     DocOptions1 = case Conflicts of
         true -> [conflicts|DocOptions0];
@@ -97,7 +98,12 @@ go(DbName, Options, QueryArgs, Callback, Acc0) ->
     end,
     case Resp of
         {ok, TotalRows} ->
-            {ok, Acc1} = Callback({meta, [{total, TotalRows}]}, Acc0),
+            Meta = case UpdateSeq of
+                false -> [{total, TotalRows}, {offset, null}];
+                true ->
+                    [{total, TotalRows}, {offset, null}, {update_seq, null}]
+            end,
+            {ok, Acc1} = Callback({meta, Meta}, Acc0),
             {ok, Acc2} = doc_receive_loop(
                 Keys3, queue:new(), SpawnFun, MaxJobs, Callback, Acc1
             ),


src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -59,7 +59,8 @@ go(DbName, Options, QueryArgs, Callback, Acc0) ->
         conflicts = Conflicts,
         skip = Skip,
         keys = Keys0,
-        extra = Extra
+        extra = Extra,
+        update_seq = UpdateSeq
```
iilyak: We should split the update of the record into two commits. 
1. add compatibility clause so with this commit we would be able to handle both kinds of records. This commit would convert new record to old one.
2. add new field and change conversion clause so it converts the old one into new.

The reason for the split is handling of mixed cluster case when some nodes are upgraded and some of them don't. 

This would allow us to do two stage upgrade of the cluster. First we would deploy version with first commit to all nodes. Then we wait a little bit and deploy the version with second commit.

This dance is usually needed for the cases when we upgrade records we send to other nodes. Which is the case for `#mrargs{}` .

eiri:
Are `update_seq` and `offset` even make sense with `keys` in sharded environment? We probably should consider just changing documentation instead of adding network overhead for ~useless attributes.

Ping @davisp on a competent opinion.

jiangphcn:
@eiri I read the following from issue #969 
```
Noticed this bug when putting together a PR to add a feature to PouchDB (see PouchDB issue 6230, i.e. missing update_seq option in allDocs API)
```

src/fabric/src/fabric_view_all_docs.erl
jiangphcn: Thanks @iilyak. However, wants to double check with you whether we need two commits. Actually, the definition for `mrargs` was not changed. https://github.com/apache/couchdb/blob/master/src/couch_mrview/include/couch_mrview.hrl#L86.  The `update_seq` filed was already present in any nodes in mixed cluster. In my newly introduced codes,  the value of `update_seq` field is fetched here and used to check whether we need to return update_seq value or not. If request is sent to nodes which are not upgraded, update_seq information is not returned. Otherwise, for upgraded nodes, the update_seq information will be returned.

davisp:
Well, offset certainly doesn't make any sense. Though to match things we could make that return offset:null or similar.

Secondly, this is gating the inclusion of these keys on whether update_seq=true is passed, it seems like if we're going to add them they should be included regardless.

For the network overhead I'm not sure it'd be a huge deal? I guess we could add a more specific endpoint rather than shipping entire db info blobs but I'm not super concerned about it.

Although, this update_seq is kind of a lie because we open each document individually so a) the update_seq may or may not reflect the shards used to open documents and b) it could change for each invidual doc open call depending on the combination of shards that responded.

I read the PouchDB issue and I'm not seeing the motivation for adding this other than its not the same as the other behavior. I wonder if we shouldn't either a) update the documentation, or b) just set offset and update_seq to null if they're requested. However without knowing the motivation for wanting these keys I'm not sure if that breaks the downstream use or not.

dharders:
Hi, PouchDB PR author here.

PouchDB was missing `{ update_seq: true }` option, so we added it due to user request. One of our tests caught the `keys` failing case when `update_seq=true`. I agree that the `offset` and `update_seq` information does not make a great deal of sense in a sharded environment.

Our main motivation is to match CouchDB API, so whatever you decide is best we will follow.

If you decide that it is easier to update the documentation (for the `keys` case) then we are fine with that. Returning `null` is also fine. As long as the API matches the documentation we will make it work!

jiangphcn:
hey @dharders thanks for your information. Let me try to re-work to return null for this case.

jiangphcn:
Return null for update_seq and offset if update_seq is true using commit af8d9f3. Can anyone help review them? Thanks

src/chttpd/test/chttpd_db_test.erl
```diff
@@ -187,6 +190,36 @@ should_return_409_for_put_att_nonexistent_rev(Url) ->
     end).
 
 
+should_return_update_seq_when_set_on_all_docs(Url) ->
+    ?_test(begin
+        [create_doc(Url, "testdoc" ++ ?i2l(I)) || I <- lists:seq(1, 3)],
+        {ok, RC, _, RespBody} = test_request:get(Url ++ "/_all_docs/"
+            ++ "?update_seq=true&keys=[\"testdoc1\"]",[?CONTENT_JSON, ?AUTH]),
+        {ResultJson} = ?JSON_DECODE(RespBody),
+        [
```
eiri: You don't need a list _here_ actually, you are probably confusing this with eunit's requirement for a list of _test objects_.

Macro `?_test` not executed in place, it returns a function that treated as a single test object. This function executed after `setup` and marks the test as a failure if it's crashes in any of its parts, so it doesn't matter if `?assert...` macro in list or sequential, the result is the same, you don't need to overwrap them.

src/chttpd/test/chttpd_db_test.erl
```diff
@@ -187,6 +190,36 @@ should_return_409_for_put_att_nonexistent_rev(Url) ->
...
+    end).
+
+
+should_not_return_update_seq_when_unset_on_all_docs(Url) ->
+    ?_test(begin
+        [create_doc(Url, "testdoc" ++ ?i2l(I)) || I <- lists:seq(1, 3)],
+        {ok, RC, _, RespBody} = test_request:get(Url ++ "/_all_docs/"
+            ++ "?update_seq=false&keys=[\"testdoc1\"]",[?CONTENT_JSON, ?AUTH]),
+        {ResultJson} = ?JSON_DECODE(RespBody),
+        [
```
eiri: Same as above and a side effect is that you might fit macro in a single line.

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -97,7 +98,14 @@ go(DbName, Options, QueryArgs, Callback, Acc0) ->
     end,
     case Resp of
         {ok, TotalRows} ->
-            {ok, Acc1} = Callback({meta, [{total, TotalRows}]}, Acc0),
+            case UpdateSeq of
```
eiri: Two things:

1. If we really want back compatibility we need `offset` in meta in both cases, this is how it was in CouchDB 1.x. Query attribute `update_seq` only controls show/hide of `update_seq` in meta, not both.

2. You are exporting variable `Meta` from case here. This is bad erlang and compiler would warn you if we'd have `+warn_export_vars` turned on with something like
```bash
~/couchdb/src/fabric/src/fabric_view_all_docs.erl:108: Warning: variable 'Meta' exported from 'case' (line 101)
```

The idiomatic way to do this is to assign `Meta` the return value of `case`, i.e.
```erlang
Meta = case UpdateSeq of
  false -> [{total, TotalRows}, {offset, null}];
  true -> [{total, TotalRows}, {offset, null}, {update_seq, null}]
end.
```

src/chttpd/test/chttpd_db_test.erl
```diff
@@ -187,6 +190,32 @@ should_return_409_for_put_att_nonexistent_rev(Url) ->
     end).
 
 
+should_return_update_seq_when_set_on_all_docs(Url) ->
+    ?_test(begin
+        [create_doc(Url, "testdoc" ++ ?i2l(I)) || I <- lists:seq(1, 3)],
+        {ok, RC, _, RespBody} = test_request:get(Url ++ "/_all_docs/"
```
eiri: Getting `/home/vagrant/git/release/couchdb/src/chttpd/test/chttpd_db_test.erl:196: Warning: variable 'RC' is unused` here.
Maybe add `?assertEqual(200, RC)` here and below to catch possible bad reply from  `/_all_docs`?

wohali:
@dharders note that this will be in 2.2.0, which is releasing imminently - may be something you'll want to review for your next PouchDB release.

PR #1089, opened 2018-01-04
Extract helper to execute a function in a new process

willholley:
## Overview

Adds a helper that executes a function in a new process, with standardized error handling and logging.

## Testing recommendations

The test suite should exercise the code paths that have changed (currently document update and _find).

n/a

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -984,35 +984,11 @@ http_code_from_status(Status) ->
     end.
 
 update_doc(Db, DocId, #doc{deleted=Deleted, body=DocBody}=Doc, Options) ->
-    {_, Ref} = spawn_monitor(fun() ->
-        try fabric:update_doc(Db, Doc, Options) of
-            Resp ->
-                exit({exit_ok, Resp})
-        catch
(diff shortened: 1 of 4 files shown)


eiri:
I don't think this helper belongs to fabric, it's pretty generic, so I'd say it fits better in [couch_util](https://github.com/apache/couchdb/blob/master/src/couch/src/couch_util.erl).

As a matter a fact a quite similar `couch_util:with_proc/4` function is already there, so it might be better to fuse them together, given that `defer` here is missing explicit timeout handeling.

willholley:
thanks @eiri - good thought. I wasn't aware of `couch_util:with_proc/4` so I'll have a look at whether that can be used instead.

willholley:
huh - I can't find any uses of `couch_util:with_proc/4` - is it redundant?

eiri:
@willholley Used in [cassim](https://github.com/apache/couchdb-cassim/blob/master/src/cassim_metadata_cache.erl#L200) evidently. But can't disagree, it's not widly known, I discovered it only because I thought your helper better fits into couch_util and went to scan code there :)

janl:
Closing until progress.

PR #1098, opened 2018-01-08
Mango: fallback to "selector" on empty "partial_filter_selector"

willholley:
## Overview

Mango text indexes historically supported partial indexes
defined via a "selector" field. This was renamed to
"partial_filter_selector" in b98de40 but the fallback
code did not correctly handle the case where
a "selector" existed alongside a "partial_filter_selector".

This situation can occur when a the _index endpoint is
used to create a text index with a "selector". The resulting
design document contains an empty "partial_filter_selector"
field *and* the "selector" field that was passed in. The
previous implementation of the fallback would detect the
presence of "partial_filter_selector" and use the empty
value (match all docs) instead of faling back to the
"selector" field.

This commit changes the behaviour so that a "selector"
will be used even if an empty "partial_filter_selector"
is present. A secondary fix would be to change the index
creation so that we never use "selector" in the underlying
index design document, even if it is passed to the _index
API.

## Testing recommendations

Run the test suite (with text indexes enabled). Attempt to create text indexes using the _index endpoint, specifying a "selector" field. Test that the partial index contains only the documents that match the selector.

#818 

## Checklist

### DIFF
--- a/src/mango/src/mango_native_proc.erl
+++ b/src/mango/src/mango_native_proc.erl
@@ -172,19 +172,12 @@ get_text_entries({IdxProps}, Doc) ->
 
 
 get_index_partial_filter_selector(IdxProps) ->
-    case couch_util:get_value(<<"partial_filter_selector">>, IdxProps) of
-        undefined ->
+    case couch_util:get_value(<<"partial_filter_selector">>, IdxProps, {[]}) of
+        {[]} ->
             % this is to support legacy text indexes that had the partial_filter_selector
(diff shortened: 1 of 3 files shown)


willholley:
@davisp oops - rebased to fixed

tonysun83:
+1, change looks good

I ran a full regression with text index tests and the new test all passed. I got some errors with the users db tests, but that's probably related to my environment and not related to text indexes.

PR #1118, opened 2018-01-19
Set update_lru_on_read=false as default

wohali:
See https://github.com/apache/couchdb/issues/796#issuecomment-359060816

From IRC:

```
15:18 <+Wohali> davisp: is update_lru_on_read something that should be in default.ini?
15:18 <+davisp> Mebbe
15:18 <+davisp> We run it by default cause a good number of clusters would get crushed without it
15:19 <+davisp> But its a behavior change so never made default
```

Discussion welcome.


wohali:
@nickva can I get a perfunctory +1 from you on this? Thanks.

janl:
+1, but would like a team-cloudant +1 as well

Issue #1122, opened 2018-01-22
should_create_delete_database_continuously (100 times) fails on Jenkins CI


wohali:
The test seems to have started failing after #1062 landed (/cc @eiri @nickva).

I beg you to not ignore Jenkins :)

## Expected Behavior
Test runs should be green across the board

## Current Behavior
8 of 11 runs in Jenkins timed out on this test in the last run on `master`

https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/173/pipeline/46
https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/173/pipeline/47
https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/173/pipeline/48
https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/173/pipeline/56
https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/173/pipeline/55
https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/173/pipeline/50
https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/173/pipeline/52
https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/173/pipeline/51

## Possible solution
FYI the semantics of the test changed with the PR. See:

https://github.com/apache/couchdb/blob/e81c54e32d2a165dc762ad4217b753da723dc1fa/src/couch/test/couch_db_tests.erl

vs. 

https://github.com/apache/couchdb/blob/master/src/couch/test/couch_db_tests.erl#L129-L139

We are calling `couch_server:delete()` now; previously we were calling `couch_db:close()` which only de-monitors the DB.

eiri:
No, the test actually been calling `couch_db:delete/2` inside `cycle`, inside `loop`, inside `_assert,` inside `timeout`, inside a list comprehension - [right here](https://github.com/apache/couchdb/blob/e81c54e32d2a165dc762ad4217b753da723dc1fa/src/couch/test/couch_db_tests.erl#L127). It just hard to see because the thing been rather convoluted. 

The semantics didn't changed, we are still calling same `couch_db:create/2`, `couch_db:close/1` (inside `create_db` helper) and then `couch_server:delete/2` inside `delete_db` helper repeatedly. 

[...]

PR #1123, opened 2018-01-22
Set eunit timeout on a whole test object

eiri:
## Overview

I've accidentally overrode custom 2 mins timeout for `should_create_delete_database_continuously` tests to default 5 sec with`?_test` wrapper.  It went unnoticed in local tests and travis, since in general "should_create_delete_database_continuously (100 times)" passes within 1.096s, but Jenkins prove to be more sensitive.

## Testing recommendations

Standard `make eunit apps=couch` should pass. More importantly Jenkins now should not throw timeout on this test.

Fixes: #1122 

## Checklist

### DIFF
--- a/src/couch/test/couch_db_tests.erl
+++ b/src/couch/test/couch_db_tests.erl
@@ -128,15 +128,13 @@ should_delete_multiple_dbs(DbNames) ->
 
 should_create_delete_database_continuously(Times, DbName) ->
     {lists:flatten(io_lib:format("~b times", [Times])),
-    ?_test(begin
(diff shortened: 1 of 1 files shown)


eiri:
@wohali This should fix Jenkins, but I guess it needs to hit "master" to confirm. Sorry about this mixup, the original test was kind of hard to untangle, that's mainly the reason I took on refactoring it.

Alternatively we can just remote this "100 times" run, honestly it doesn't really prove or demonstrate anything "10 times" run doesn't show.

wohali:
Nice, looks good. +1

If you want to also remove the 100x test, go for it.

Issue #603, opened 2017-06-20
Error 500 when creating a db below quorum


calonso:
## Expected Behavior
On a 4 nodes cluster, having two down, if I try to create a new database the server should either reject the request or accept it returning a 202 status code. Once the down nodes come back, the new db should be replicated to them.

## Current Behavior
Currently an error 500 is returned but the db is indeed created (not sure if it replicated when the down nodes came back).

## Possible Solution
Return a more friendlier status code as `202 Accepted` if the db was actually created or completely reject it with a `412 Precondition Failed` instead.

## Steps to Reproduce (for bugs)
1. Setup a 4 nodes cluster
2. Bring 2 of them down
3. `curl -X PUT "http://xxx.xxx.xxxx.xxx:5984/testdb"`
4. An error 500 is returned.
5. `curl -X GET "http://xxx.xxx.xxx.xxx:5984/_all_dbs"` will return `testdb`

## Context
I'm building a [CouchDB automatic administration toolset](https://github.com/cabify/couchdb-admin) and I was trying out different scenarios to see it working, in this case I was simulating an operation where on a cluster without quorum (4 nodes, 2 of them down), one tries to create a new database.

## Your Environment
It is a Kubernetes managed cluster where the Docker image in use is [https://hub.docker.com/r/klaemo/couchdb/](https://hub.docker.com/r/klaemo/couchdb/)
* Version used: 2.0.0
* Link to your project: [https://github.com/cabify/couchdb-admin](https://github.com/cabify/couchdb-admin)

wohali:
Hi there, can you please include an excerpt from `couch.log` showing what happens when the 500 is returned? Also, if possible, can you include the exact message returned to your client along with the 500, if any? Thanks.

calonso:
Sure!!

```
$ curl -X PUT "http://admin:password@127.0.0.1:5984/newtestdb2"
{"error":"error","reason":"internal_server_error"}
```
and the log, in debug mode
```
[debug] 2017-06-20T13:42:58.420676Z couchdb@couch-0.couchdb2 <0.18754.0> ac122995ef cache miss for admin
[debug] 2017-06-20T13:42:58.421663Z couchdb@couch-0.couchdb2 <0.18754.0> ac122995ef no record of user admin
[...]

PR #1127, opened 2018-01-25
Fix for issue #603 - Error 500 when creating a db below quorum

jjrodrig:
## Overview
Current behaviour of DB creation in a cluster degraded situation is not consistent with the general behaviour described for document creation in the same situation.

> The number of copies of a document with the same revision that have to be read before CouchDB returns with a 200 is equal to a half of total copies of the document plus one. It is the same for the number of nodes that need to save a document before a write is returned with 201. If there are less nodes than that number, then 202 is returned. Both read and write numbers can be specified with a request as r and w parameters accordingly.

The current behaviour for database creation in a cluster is:
 - Database creation returns 201 - Created if all nodes responds ok
 - Database creation returns 202 - Accepted if the quorum is met
 - Database creation returns 500 - Error if the responses are bellow quorum

The quorum is the default: Number of nodes/2 +1

This PR changes the database creation result with the following behaviour:
- Database creation returns 201 - Creation if the quorum is met
- Database creation returns 202 - Accepted if at least one node responds ok
- Database creation returns 500 - Error if there is no correct response from any node

## Testing recommendations

- All previous tests are ok
- I've focused on chttpd and javascript tests 
- I've skiped reduce_builtin.js test as it is failing even in master branch
`test/javascript/tests/reduce_builtin.js                        
    Error: {gen_server,call,[<0.2426.1>,{get_state,49},infinity]}`

`make check apps=chttpd ignore_js_suites=reduce_builtin`

I didn't identify testing infrastructure for testing cluster degradation issues. So I decided to add some support for testing in different cluster conditions. Now make has two more task:
- test-cluster-with-quorum, which launch a three node cluster and stops one node then executes the tests
- test-cluster-without-quorum which launch a three node cluster and stops two nodes then executes the tests

This PR can be tested in this way
```
 make test-cluster-with-quorum
 make test-cluster-without-quorum
```

Issue #603 

## Checklist

- [x] Changes are covered by tests; (DB creation yes, Cluster degradation no)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,8 @@ fauxton: share/www
 .PHONY: check
 # target: check - Test everything
 check: all
+	@$(MAKE) test-cluster-with-quorum
+	@$(MAKE) test-cluster-without-quorum
 	@$(MAKE) eunit
 	@$(MAKE) javascript
 	@$(MAKE) mango-test
@@ -129,6 +131,7 @@ endif
             'test/javascript/run --suites "$(suites)" \
             --ignore "$(ignore_js_suites)"'
 
+# TODO: port to Makefile.win
 .PHONY: test-cluster-with-quorum
 test-cluster-with-quorum:
 	@mkdir -p share/www/script/test
@@ -146,6 +149,7 @@ endif
             --ignore "$(ignore_js_suites)" \
 	    --path test/javascript/tests-cluster/with-quorum'
 
+# TODO: port to Makefile.win
 .PHONY: test-cluster-without-quorum
 test-cluster-without-quorum:
 	@mkdir -p share/www/script/test
--- a/Makefile
+++ b/Makefile
@@ -129,6 +129,40 @@ endif
             'test/javascript/run --suites "$(suites)" \
             --ignore "$(ignore_js_suites)"'
 
+.PHONY: test-cluster-with-quorum
+test-cluster-with-quorum:
+	@mkdir -p share/www/script/test
+ifeq ($(IN_RELEASE), true)
+	@cp test/javascript/tests/lorem*.txt share/www/script/test/
+else
+	@mkdir -p src/fauxton/dist/release/test
+	@cp test/javascript/tests/lorem*.txt src/fauxton/dist/release/test/
+endif
+	@rm -rf dev/lib
+	@dev/run -n 3 -q --with-admin-party-please \
+            --enable-erlang-views --degrade-cluster 1 \
+            -c 'startup_jitter=0' \
+            'test/javascript/run --suites "$(suites)" \
+            --ignore "$(ignore_js_suites)" \
+	    --path test/javascript/tests-cluster/with-quorum'
+
+.PHONY: test-cluster-without-quorum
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -129,6 +129,40 @@ endif
...
+	@mkdir -p share/www/script/test
+ifeq ($(IN_RELEASE), true)
+	@cp test/javascript/tests/lorem*.txt share/www/script/test/
+else
+	@mkdir -p src/fauxton/dist/release/test
+	@cp test/javascript/tests/lorem*.txt src/fauxton/dist/release/test/
+endif
+	@rm -rf dev/lib
+	@dev/run -n 3 -q --with-admin-party-please \
+						--enable-erlang-views --degrade-cluster 2 \
```
rnewson: indentation

Makefile
```diff
@@ -129,6 +129,40 @@ endif
...
+	@mkdir -p src/fauxton/dist/release/test
+	@cp test/javascript/tests/lorem*.txt src/fauxton/dist/release/test/
+endif
+	@rm -rf dev/lib
+	@dev/run -n 3 -q --with-admin-party-please \
+            --enable-erlang-views --degrade-cluster 1 \
+            -c 'startup_jitter=0' \
+            'test/javascript/run --suites "$(suites)" \
+            --ignore "$(ignore_js_suites)" \
+						--path test/javascript/tests-cluster/with-quorum'
```
rnewson: indentation

src/fabric/src/fabric_db_create.erl
```diff
@@ -146,9 +146,9 @@ maybe_stop(W, Counters) ->
         {ok, {W, Counters}};
     false ->
         case lists:sum([1 || {_, ok} <- Counters]) of
-        W ->
+        NumOk when NumOk >= (W div 2 +1) ->
             {stop, ok};
-        NumOk when NumOk >= (W div 2 + 1) ->
+        NumOk when NumOk > 0 ->
```
rnewson: this is the right change.

rnewson:
+1

janl:
@jjrodrig we wanted to make it so that `make test-cluster-with[out]-quorum` is run as part of `make check`, but this doesn’t seem to work on Travis, would you be able to assist with debugging this?

jjrodrig:
Ok, I'll check it.

janl:
Thank you! :)

janl:
I’m trying a new variant where the new make targets run first in `make check`

jjrodrig:
@janl I've reproduced the problem in my environment. It seems that the stop after the mango test is collisioning with the start of the cluster test. 
If you see this commit:

https://github.com/jjrodrig/couchdb/commit/f45b404de930b7da81af7e01d2071451d21acde7

I've introduced an sleep after mango tests and the check is working

Also, I've noticed that the mango test do not clean up dev/lib so it can be affected by previous test executions.

janl:
merged this, and filed an issue for cleaning up mango: https://github.com/apache/couchdb/issues/1134

Thanks a lot for this contribution!

PR #1135, opened 2018-01-31
Fix for issue #1134 clean up dev/lib before run mango tests

### LINKED ISSUE #1134: mango tests don’t clean up after themselves
See https://github.com/apache/couchdb/pull/1127#issuecomment-361648968 and https://github.com/jjrodrig/couchdb/commit/f45b404de930b7da81af7e01d2071451d21acde7

jjrodrig:
## Overview

**make mango-test** is not cleaing previous state in dev/lib so it can be affected by previous mango or javascript test execution

## Testing recommendations
Following execution sequence produces testing errors beffore this PR and not after PR
```
make test-cluster-with-quorum
make mango-test
```` 

Fixes #1134 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ soak-eunit: couch
 
 .PHONY: javascript
 # target: javascript - Run JavaScript test suites or specific ones defined by suites option
-javascript:
+javascript: devclean
 	@mkdir -p share/www/script/test
 ifeq ($(IN_RELEASE), true)
 	@cp test/javascript/tests/lorem*.txt share/www/script/test/
@@ -124,7 +124,6 @@ else
 	@mkdir -p src/fauxton/dist/release/test
 	@cp test/javascript/tests/lorem*.txt src/fauxton/dist/release/test/
 endif
-	@rm -rf dev/lib
 	@dev/run -n 1 -q --with-admin-party-please \
             --enable-erlang-views \
             -c 'startup_jitter=0' \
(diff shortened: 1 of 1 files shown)


eiri:
I think it'l be cleaner to just make `mango-test` target to depend on `devclean` target before `all`.

jjrodrig:
Thanks @eiri

devclean is removing data directories `@rm -rf dev/lib/*/data`  but javascript test is remove all the dev/lib content `@rm -rf dev/lib`
I think that the local.ini is modified by the dev/run script and it is not cleaned by devclean target. It is enough to clean just data?

rm -rf dev/lib is done also by javascript, test-cluster-with-quorum and test-cluster-without-quorum targets. Should these targets also depend on devclean?

eiri:
As far as I can tell `dev/run` overwrites `[admin]` and `[cluster]` sections on each run, so it should be enough just to remove leftover databases to get a clean start state, but you can modify `devclean` to remove whole `dev/lib` if you have any concerns.

For `javascript` and the new targets: yes, I think it's better to reduce cut'n'paste when makefile allows that. It's bigger than original scope of this PR, so you can make it for `mango-test` only if you don't feel like extending your changes and we can always return to this later.

jjrodrig:
I've modified the javascript, test-cluster* and test-mango targets.

soak-javascript target seems to be not correct, a syntax error is raised. I didn’t modify it.

eiri:
@jjrodrig : Looks good. Thank you for the fix!

PR #1146, opened 2018-02-06
Validate user auth from Erlang

abc3:
## Overview

PR consist translation to Erlang from AUTH_DB_DOC_VALIDATE_FUNCTION javascript file. 

## Testing recommendations

I tested throught adding this fun to "_users > _design/_auth" doc. It seems to work :)

## Checklist


wohali:
This would be a nice one for 2.2.0, if we can swing it, though it's not a trivial change.

We'll need to go the extra mile to add it to `_users` by default, and presumably remove the pre-existing one.

rnewson:
I don't see where this gets hooked in, but I don't like that the erlang is expressed in a macro. I assume the author is following the existing method we use for JS, but that's done out of necessity. Let's pull that erlang into a proper module please. will review again after.

wohali:
Hi @abc3, is there any chance of doing what @rnewson says above and putting the Erlang into a proper module rather than as a macro?

abc3:
Yes, but right now I'm a little bit busy.

PR #1161, opened 2018-02-13
Add ./configure --dev as alias for -c --disable-{docs,fauxton}

janl:
As a CouchDB developer, I find myself typing `./configure -c --disable-docs --disable-fauxton` a lot, and I long for a shortcut, so I’m introducing `./configure --dev` as an alias of the above.


wohali:
+1, me too

PR #1164, opened 2018-02-14
Fix: Proper error handling for file:open() call in couch_file.erl

AlexanderKaraberov:
## Overview

According to Erlang's [file:open](http://erlang.org/doc/man/file.html#open-2) docs this function returns `{error, Reason}` in case of failure but inside `couch_file:sync` and `init` functions there are no checks for this in two places. Consequently when I received `{error, emfile}` kind of error, process crashed with `badmatch` instead of a proper error.

## Testing recommendations

Simulate `{error, emfile}` conditions (e.g. decrease `ulimit` on your OS) and observe an error.

## Checklist

### DIFF
--- a/src/couch/src/couch_file.erl
+++ b/src/couch/src/couch_file.erl
@@ -216,9 +216,9 @@ sync(Filepath) when is_list(Filepath) ->
     case file:open(Filepath, [append, raw]) of
         {ok, Fd} ->
             try ok = file:sync(Fd) after ok = file:close(Fd) end;
-        Error ->
-            throw(Error)
-        end;
+        {error, Error} ->
+            erlang:error(Error)
+    end;
 sync(Fd) ->
     gen_server:call(Fd, sync, infinity).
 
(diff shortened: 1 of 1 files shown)


rnewson:
-1.

We assert that we get the `ok` we expect. If we don't, we want the process to crash, which it does.

Adding error handling code for errors we don't intend to handle (in the sync case, you immediately throw it anyway) is an anti-pattern in erlang, though common (or mandatory) in other languages.

rnewson:
@davisp if you would like to comment on the init portions.

src/couch/src/couch_file.erl
```diff
@@ -213,8 +213,12 @@ truncate(Fd, Pos) ->
 %%----------------------------------------------------------------------
 
 sync(Filepath) when is_list(Filepath) ->
-    {ok, Fd} = file:open(Filepath, [append, raw]),
-    try ok = file:sync(Fd) after ok = file:close(Fd) end;
+    case file:open(Filepath, [append, raw]) of
+      {ok, Fd} ->
```
davisp: Use four space indents everywhere.

src/couch/src/couch_file.erl
```diff
@@ -213,8 +213,12 @@ truncate(Fd, Pos) ->
...
+         try ok = file:sync(Fd) after ok = file:close(Fd) end;
+      Error ->
+         throw(Error)
```
davisp: Match {error, Error} and then erlang:error(Error) here.

davisp:
@rnewson yes and no. The issue is where we want that badmatch to happen and with what as the badmatch. For the init section this is a good change as it matches the initial open. What's being added is the handling when the second of two fails to behave the same as the first. This is fine as we already handle the error. And it cleans up logs a bit in that the emfile error happens in the process doing the open rather than a EXIT/DOWN message when the couch_file pid dies on a badmatch.

As for the init process this just reuses the existing logic which seems cleaner than going back and rewriting it to use proc_lib properly.

+1 after style issues are addressed

rnewson:
if @davisp is happy, I'm happy.

AlexanderKaraberov:
@davisp Thanks for review. Will address this in a few days when I have access to my laptop.

src/couch/src/couch_file.erl
```diff
@@ -213,8 +213,12 @@ truncate(Fd, Pos) ->
...
 
 sync(Filepath) when is_list(Filepath) ->
-    {ok, Fd} = file:open(Filepath, [append, raw]),
-    try ok = file:sync(Fd) after ok = file:close(Fd) end;
+    case file:open(Filepath, [append, raw]) of
+        {ok, Fd} ->
+            try ok = file:sync(Fd) after ok = file:close(Fd) end;
+        {error, Error} ->
+            erlang:error(Error)
+        end;
```
davisp: This `end` should be dedented one level.

src/couch/src/couch_file.erl
```diff
@@ -389,14 +393,18 @@ init({Filepath, Options, ReturnPid, Ref}) ->
...
         {ok, Fd_Read} ->
-            {ok, Fd} = file:open(Filepath, OpenOptions),
-            %% Save Fd in process dictionary for debugging purposes
-            put(couch_file_fd, {Fd, Filepath}),
-            ok = file:close(Fd_Read),
-            maybe_track_open_os_files(Options),
-            {ok, Eof} = file:position(Fd, eof),
-            erlang:send_after(?INITIAL_WAIT, self(), maybe_close),
-            {ok, #file{fd=Fd, eof=Eof, is_sys=IsSys, pread_limit=Limit}};
+             case file:open(Filepath, OpenOptions) of
```
davisp: You've got an extra space here.

src/couch/src/couch_file.erl
```diff
@@ -389,14 +393,18 @@ init({Filepath, Options, ReturnPid, Ref}) ->
...
+                      %% Save Fd in process dictionary for debugging purposes
+                      put(couch_file_fd, {Fd, Filepath}),
+                      ok = file:close(Fd_Read),
+                      maybe_track_open_os_files(Options),
+                      {ok, Eof} = file:position(Fd, eof),
+                      erlang:send_after(?INITIAL_WAIT, self(), maybe_close),
+                      {ok, #file{fd=Fd, eof=Eof, is_sys=IsSys, pread_limit=Limit}};
+                  Error ->
+                      init_status_error(ReturnPid, Ref, Error)
+              end;
```
davisp: And two extra spaces here.

PR #1165, opened 2018-02-14
Fix: validate_doc_update behaviour for Erlang functions

AlexanderKaraberov:
## Overview

This issue hit me when I was implementing support for custom `validate_doc_read` functions (written in both native Erlang and JS), because `validate_doc_update` is not sufficient for our production use case.

When you implement `validate_doc_update` as an Erlang function it doesn't work because of a missing case clause, namely: when `couch_query_servers:try_compile()` calls `proc_prompt_raw()` which in turn calls `erlang:apply(Fun)` for a native function the return value of it is an atom `ok` or `true` but in case of JS function we have response code `1`. I added an additional guard check for all these situations in the `validate_doc_update` and now it works for Erlang functions as well. This was crucial in our case because JS functions are much slower and in our production system we are going to use validation functions implemented in Erlang.

## Testing recommendations

Implement `validate_doc_update` as Erlang function in the design doc (`"language": "erlang"`), run CouchDB with `--enable-erlang-views` and update some docs in the db in order to trigger this call.

## Checklist


janl:
Thanks!

PR #1178, opened 2018-02-22
Prevent chttpd multipart zombie processes

davisp:
Occasionally it's possible to lose track of our RPC workers in the main
multipart parsing code. This change monitors each worker process and
then exits if all workers have exited before the parser considers itself
finished.

Fixes part of #745

## Testing recommendations

This was tested under the conditions for 745 which are non-trivial but can be duplicated by reading that PR.

Issue #745 was the original report for this issue

## Checklist


wohali:
+1

:postal_horn: :imp: :unicorn:

PR #1185, opened 2018-02-28
TMP - Debug logging for Jenkins failures

davisp:
Debugging compaction daemon test failure


davisp:
Don't merge this yet. Just running some travis runs to try and trigger the failure in compaction daemons tests.

wohali:
@davisp Did we fix this one? I think this might be safe to close out.

PR #1186, opened 2018-03-01
[WIP] Clustered purge mixed cluster upgrade

jiangphcn:
## Overview

This is preparatory work for adding clustered purge which is based on pluggable storage engines. The motivation for this ground work is NOT to break internal replication between nodes. 

For clustered purge during internal replication, additional step was added to synchronize purges. This work involves creating a stub function for synchronizing purges that returns an empty list of purges, as well as, creating duplicates of some other functions.

In addition, we added logic to support downgrade of couch database to the previous format in case of the need of rollback of clustered purge.

## Testing recommendations

There's no behavior changes so the existing test suite covers the existing logic.

## JIRA issue number
https://issues.apache.org/jira/browse/COUCHDB-3326

We are going to open another PR for COUCHDB-3326 which is the main Clustered Purge work that will be asking to merge into this PR. When we go to merge these we'll merge this PR first, and then merge 3326 into master.
https://github.com/apache/couchdb/pull/1187

## Checklist


src/couch/src/couch_bt_engine.erl
```diff
@@ -759,6 +777,45 @@ set_default_security_object(Fd, Header, Compression, Options) ->
     end.
 
 
+% This function is to rollback pf clustered purge.
+% It replaces purge_tree_state and upurge_tree_state with
+% with purged_docs disk pointer that stores only the last purge request
+downgrade_purge_info(Fd, Header) ->
+    {db_header, _, UpSeq, _, IdTreeState, SeqTreeState, LocalTreeState, PSeq,
```
davisp: Can you reformat this so that it's one field per line and includes names for the _ variables that are ignored? I had to refer back and forth to the various header definitions to understand what was being dropped in each of those wild card matches.

src/couch/src/couch_bt_engine.erl
```diff
@@ -759,6 +777,45 @@ set_default_security_object(Fd, Header, Compression, Options) ->
     end.
 
 
+% This function is to rollback pf clustered purge.
+% It replaces purge_tree_state and upurge_tree_state with
+% with purged_docs disk pointer that stores only the last purge request
```
davisp: This function is wrong as its not a mirror of what happens when we upgrade the state turning the whole purge docs into multiple rows. It also occurs to me that we transform this in such a way that we'd not be able to know how to undo the upgrade since we would not be able to distinguish between an upgrade and purges applied after the upgrade.

Also if we rollback this approach would also likely invalidate all secondary indices since we're grabbing the last purge sequence as well.

davisp:
So thinking more on this, I think the best approach is to not upgrade the purge values to a tree until we get a purge request. Thus for the normal case when a customer has no purge activity before and after the upgrade/downgrade there won't be any issue.

I'm also trying to decide if we should change the disk version for this instead of just altering the purge_docs field contents.

The thing I'm most worried about at this point is what happens when we downgrade after purge requests have been issued over the new API. If we're doing this quickly then its likely that we'd have to invalidate all secondary indexes to be sure they're valid? Or else make some sort of check against all of the _local purge docs?

src/couch/src/couch_bt_engine.erl
jiangphcn: Re-formatted usage of db_header no matter whether downgrade_purge_info/2 will be used in pre-Clustered Purge build or not.

jiangphcn:
Thanks @davisp for pointing out about possibility of index being rebuilt. It looks to me that it is good idea of not upgrading the purge values to a tree until we get a purge request. I need some time to implement them.

davisp:
Yeah, but this is also gonna play badly with internal replication which will also try pushing and pulling those things across. Also trying to figure out index resets when we upgrade and have an old index that's not queried very often. Am trying to figure out if there's a way to force all indexes to write out their tracking doc without forcing indexing to start.

davisp:
However that seems like it'd get awfully complicated quite quickly. Am trying to avoid having a large impact on clusters that have lots of unused indexes.

jiangphcn:
Still trying to think about downgrade and upgrade from pre-CP to Clustered Purge.

There is another situation where database are created after we upgrade to Clustered Purge. The database header will be created with purge tree instead of purge value, like
```
(dbcore@db1.testy011.cloudant.net)3> couch_db:open(<<"shards/c0000000-dfffffff/testy011-user1/db2.1520242912">>, []).
{ok,{db,1,
        <<"shards/c0000000-dfffffff/testy011-user1/db2.1520242912">>,
        "/srv/db/shards/c0000000-dfffffff/testy011-user1/db2.1520242912.couch",
        {couch_bt_engine,{st,"/srv/db/shards/c0000000-dfffffff/testy011-user1/db2.1520242912.couch",
                             <0.7078.55>,#Ref<0.0.69.200720>,
                             [before_header,after_header,on_file_open],
                             {db_header,7,0,0,nil,nil,nil,0,nil,184,1000,
                                        <<"6139f441"...>>,
                                        [{...}],
                                        0,...},
                             false,
                             {btree,<0.7078.55>,nil,
                                    #Fun<couch_bt_engine.id_tree_split.1>,
                                    #Fun<couch_bt_engine.id_tree_join.2>,undefined,
                                    #Fun<couch_bt_engine.id_tree_reduce.2>,
                                    {deflate,6}},
                             {btree,<0.7078.55>,nil,
                                    #Fun<couch_bt_engine.seq_tree_split.1>,
                                    #Fun<couch_bt_engine.seq_tree_join.2>,undefined,
                                    #Fun<couch_bt_engine.seq_tree_reduce.2>,
                                    {deflate,6}},
                             {btree,<0.7078.55>,nil,
                                    #Fun<couch_bt_engine.local_tree_split.1>,
                                    #Fun<couch_bt_engine.local_tree_join.2>,undefined,nil,
                                    {deflate,6}},
                             {deflate,6},
                             {btree,<0.7078.55>,nil,undefined,undefined,undefined,
                                    #Fun<couch_bt_engine.purge_tree_reduce.2>,snappy},
                             {btree,<0.7078.55>,nil,undefined,undefined,undefined,
                                    #Fun<couch_bt_engine.purge_tree_reduce.2>,...}}},
        <0.7077.55>,nil,0,<<"1520242912784404">>,
        {user_ctx,null,[],undefined},
        [],undefined,nil,nil,nil,
        [{default_security_object,[]},
         {n,"3"},
         {q,"8"},
         {placement,undefined}],
        undefined}} 
```

If we need to downgrade for such database, we might have to start without any purge history. Or other thoughts? @davisp

davisp:
I've not got anything good no. Other than I'm starting to really really wanting to avoid the possibility of downgrading. The one thing that does make things a bit easier is that most databases won't already have purge information since its not available through the clustered API and doesn't appear to be an API most people use normally.

However the thing that's mostly worrying me now is what to do about unused/idle indexes after upgrade in terms of both a) can we auto-create _local docs for them or b) can we force them all to run an initial step. The fear being that if we upgrade and allow for purging when an index exists but is unused it seems possible to have an invalid index laying around (i.e., if we compact purge requests such that a index never sees them).

jiangphcn:
Thanks Paul @davisp for your reply. I understand. For completeness, we need to consider downgrade when there is purge request between pre-CP and CP. However, only with CP build, the clustered purge API is available. Before that, there is less possibility to have purge request. So downgrade may not be effort we should focus on.

Regarding the second point "unused/idle indexes after upgrade", as you know, we have two methods to trigger init/update of secondary index, search index and geo index.
- issue search request and trigger update of index
- use `ken` or similar approach to automatically trigger init/update of index

If there is unused/idle index, what about leveraging `ken` or similar approach to trigger index o run an initial step? Thus, if there is update on database document, ken will trigger init/update of index, thus the local purge document will be created so that purge requests will not be compacted before a index see them.

wohali:
@jiangphcn remember that `ken` does not exist in CouchDB itself. Any solution that relies on it would need an open-source counterpart (or, yanno, just open source `ken` :) )

davisp:
@wohali We must've forgotten to circle back on the PR discussions to here but the ken thing is a no go cause its effectively no different than just relying on customers to query all indexes which is a no-go. The solution ended up being to add a couch_epi pre-compact hook to allow indexers to pre-populate the _local/purge- docs as required.

jiangphcn:
Thanks for checking @wohali and response @davisp. Yes, we go with EPI based approach to inject local purge document before actual compaction. The current implementation are https://github.com/apache/couchdb/pull/1187/commits/6351cfc3cd1a7b0a3302456efd423e16b919a156.

wohali:
@jiangphcn FYI you can land this and the other PRs anytime, 2.2.x has been forked off of master.

jiangphcn:
Thanks @wohali for your information. It is cool that 2.2. is final. I will double check with Paul and then merge clustered purge PRs.

PR #1193, opened 2018-03-02
Use `chttpd:send_error/2` in mango_httpd

iilyak:
## Overview

We used to call non exported function chttpd:send_error/6. Switch to
public `chttpd:send_error/2` for sending error.

## Testing recommendations

induce error in mango or update code as follows 
```
diff --git a/src/mango/src/mango_httpd.erl b/src/mango/src/mango_httpd.erl
index 2e87771..e69001a 100644
--- a/src/mango/src/mango_httpd.erl
+++ b/src/mango/src/mango_httpd.erl
@@ -33,6 +33,8 @@
 }).

 handle_req(#httpd{} = Req, Db0) ->
+    Stack = erlang:get_stacktrace(),
+    chttpd:send_error(Req, {<<"ErrorStr">>, <<"ReasonStr">>, Stack}),
```

Issue mango query request
```
http --auth=adm:pass http://127.0.0.1:15984/mango_test_ffdbb10509f046cea3883d144586c02d/_index
Content-Length: 59
Date: Fri, 02 Mar 2018 15:16:21 GMT
Server: CouchDB/2.1.1 (Erlang OTP/17)
X-Couch-Request-ID: f942b096d7
X-Couch-Stack-Hash: 1466922193
X-CouchDB-Body-Time: 0

{
    "error": "ErrorStr",
    "reason": "ReasonStr",
    "ref": 1466922193
}
```

Check that 
- response has `error` and `reason`.
- response code is 500
- logs contain stacktrace
## Checklist

### DIFF
--- a/src/mango/src/mango_httpd.erl
+++ b/src/mango/src/mango_httpd.erl
@@ -41,7 +41,7 @@ handle_req(#httpd{} = Req, Db0) ->
             case mango_error:info(Module, Reason) of
             {500, ErrorStr, ReasonStr} ->
                 Stack = erlang:get_stacktrace(),
-                chttpd:send_error(Req, 500, [], ErrorStr, ReasonStr, Stack);
+                chttpd:send_error(Req, {ErrorStr, ReasonStr, Stack});
             {Code, ErrorStr, ReasonStr} ->
                 chttpd:send_error(Req, Code, ErrorStr, ReasonStr)
             end


eiri:
Using `chttpd:send_error/4` seems to be closer to the original intent? It's public func as well.

janl:
*bump*

iilyak:
> Using chttpd:send_error/4 seems to be closer to the original intent? It's public func as well.

`chttpd:send_error/4` doesn't support passing of stack trace. The stack trace is converted [here](https://github.com/apache/couchdb/blob/master/src/chttpd/src/chttpd.erl#L1003). There are no calls to `json_stack/1` from  `chttpd:send_error/4`. `chttpd:send_error/4` calls `chttpd:send_error/5` with `Stack = []`.

PR #1194, opened 2018-03-02
Eliminate "default" Erlang platform from Jenkins CI

wohali:
Currently, we run Jenkins CI builds on 11 platforms:

Erlang 18.3 + [centos6|centos7|debian8|debian9|ubuntu1404|ubuntu1604]
"Default" (ships-with-the-distro) Erlang + [same list but not centos6]

The intent was ensure that CouchDB ran fine both with our preferred
Erlang version (right now, 18.3) as well as the Erlang that ships with
various common distributions. This would help ease consumption of
CouchDB by end users, we thought.

With CouchDB 2.1.0 we started providing convenience binary packages,
built with Erlang 18.3. This (along with Docker) has become the most
common way people consume CouchDB today. Those binaries are generated
directly from the Jenkins builds.

Further, Travis CI already runs our test suite against a variety of
Erlang distributions, from 16B03 right through 20.1 (as of today).

There is little value in the extra Jenkins runs at this point, beyond
catching any vagaries of distro-specific Erlang weirdness that we
would probably wash our hands of anyway.

This change removes 5 of our 11 platform builds. This will be valuable
as we move towards regression testing other platforms (ppc64le, arm).
It also lightens our load on the ASF Jenkins infrastructure.

NOTE that we will still need to maintain the Docker image for
ubuntu1404erlangdefault, as this contains 16B03, the earliest version
of Erlang that we support. The initial "Build" step in our Jenkins
run uses this to build the couchdb.tar.gz dist tarball, which is then
compiled and tested on each platform during the "Test" step.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -118,31 +118,6 @@ pipeline {
             deleteDir()
           } // node
         },
-        centos7erlangdefault: {
-          node(label: 'ubuntu') {
-            timeout(time: 45, unit: "MINUTES") {
-              sh 'docker pull couchdbdev/centos-7-erlang-default'
-              withDockerContainer(image: 'couchdbdev/centos-7-erlang-default', args: '-e LD_LIBRARY_PATH=/usr/local/bin') {
-                sh 'rm -f apache-couchdb-*.tar.gz'
-                unstash 'tarball'
-                sh '''
-                  cwd=$(pwd)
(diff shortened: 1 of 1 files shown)


wohali:
For reference, Jenkins is running this branch's PR here: https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/jenkins-faster/1/pipeline

PR #1195, opened 2018-03-02
Add support for bulk get with Accept Content Type: "multipart/mixed", "multipart/related"

AlexanderKaraberov:
## Overview

CouchDB currently has support for `_bulk_get` requests which are intended to improve performance of client pull replications, by allowing the client to request multiple documents in one request. But current implementation has some limitations namely: attachment bodies can only be encoded inline (base64) not as MIME multipart bodies. Another issue which was much more important in our case: current _bulk_get implementation in CouchDB is not compatible with the [Couchbase Lite 1.4.x](https://github.com/couchbase/couchbase-lite-ios/tree/release/1.4.1) for iOS and Android because their `BulkDownloader` accepts [multipart/related content type](https://github.com/couchbase/couchbase-lite-ios/blob/release/1.4.1/Source/CBLBulkDownloader.m#L85) and not `application/json`. 
This PR adds support for `multipart/related`/`multipart/mixed` _bulk_get response which is compatible with [Couchbase Lite/Sync Gateway](https://github.com/couchbase/sync_gateway/wiki/Bulk-GET). Code is pretty much based on the [rcouch](https://github.com/rcouch) implementation with some changes to make it compatible with current CouchDB 2 architecture and also doesn't use any external libraries. 
Logic is pretty simple: depending on the `case MochiReq:accepts_content_type("multipart/mixed")` check I pick either a default CouchDB's _bulk_get implementation or a multipart one. I specifically added a `send_doc_multipart` function so that not to break an existing `send_docs_multipart` one which wasn't compatible with the logic.
We already tested this with some of the big production CouchDB databases and speedup was very noticeable. I think this feature will be very useful because there are a lot of Couchbase Lite users who can't benefit from the _bulk_get because of `application/json` only response. Plus this is a completely additive change which doesn't break/change existing functionality. Test suite is based on the `chttpd_db_bulk_get_test`. If some of the tests are missing or there are some issues with the implementation - feel free to point and I will fix them.

## Testing recommendations
For testing create a `db1`, put some docs there: 
```
curl -u admin:rm -X PUT http://127.0.0.1:5983/db1
curl -u admin:rm -H "Content-Type: application/json" -X PUT http://127.0.0.1:5983/db1/doc1 -d '{"type":"a"}'
curl -u admin:rm -H "Content-Type: application/json" -X PUT http://127.0.0.1:5983/db1/doc2 -d '{"type":"a"}'
curl -u admin:rm -H "Content-Type: application/json" -X PUT http://127.0.0.1:5983/db1/doc3 -d '{"type":"b"}'
```
Also add attachments for some of the docs (.png images or whatnot). After this you can execute similar curl command:
```
curl -u "admin:rm" -X POST -H "Content-Type: application/json" -H "Accept: multipart/related" -H "X-Accept-Part-Encoding: gzip"  -o "response.json" -d '{"docs": [{"id":"doc1", "rev": "1-8d0539dc5c5fe98d4069d2d0b06d9fa6"},{"id": "doc2", "rev":"1-b59ac7ccf4c21b26f6a4e1d72c509f23"},{"id": "doc3", "rev": "1-fc9b2660a790d8d98589102162ccb63d"}]}' http://127.0.0.1:5983/db1/_bulk_get?revs=true&attachments=true
```
Then in the response.json you can observe something similar to this:
```json
-----------------------------abe834719cdf618f78719d2a8f3edd06

{"_id":"doc1","_rev":"1-8d0539dc5c5fe98d4069d2d0b06d9fa6","fp_owner":"global"
-----------------------------abe834719cdf618f78719d2a8f3edd06
X-Doc-Id: _design/briefcase
X-Rev-Id: 3-9df87eb32d30ce1e01fe4ca121ac725e
Content-Type: multipart/related; boundary=---------------------------edc31bb71a98be453c3ea40d4f96019e

-----------------------------edc31bb71a98be453c3ea40d4f96019e

{"_id":"doc2","_rev":"3-9df87eb32d30ce1e01fe4ca121ac725e","fp_owner":"global"}
-----------------------------edc31bb71a98be453c3ea40d4f96019e
Content-Disposition: attachment; filename="IMG_0361.PNG"
Content-Type: image/png
Content-Length: 2466384

�PNG


IHDRe	����aiCCPkCGColorSpaceDisplayP3(�c``RI,(�aa``��+)
rwR���R`����b
����>@%0|����/���:%5�I�^��b��Ջ�D�0գ����d ��S��JS�l�����):
Ȟb�C�@�$�XMH�3�}�VH�H�����IBOGbC�n��ₜ�J�c�%��V��h���ʢ���G`(�*x�%��(�30����s 8,�� Ě�30������n���~��@�\;b��'v$%�����)-����r�H�@=��i�F`yF'�{��Vc``����w�������w1P��y!e�5����iTXtXML:com.adobe.xmp<x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="XMP Core 5.4.0">
   <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
      <rdf:Description rdf:about=""
            xmlns:photoshop="http://ns.adobe.com/photoshop/1.0/"
            xmlns:exif="http://ns.adobe.com/exif/1.0/">
         <photoshop:DateCreated>2018-01-25T11:00:11</photoshop:DateCreated>
         <exif:UserComment>Screenshot</exif:UserComment>
      </rdf:Description>
   </rdf:RDF>
</x:xmpmeta>
�8�@IDATx����ga'~����~�o����#I+��I�Y��|�8��d0�9� &���c R
g8rp��C�Q"Ğ�H#Y�u?�GR?�ܭR���=�n�^.uMu�����OWmy�z��㶄`�����8�p��8�p��8�p��8�����
-----------------------------edc31bb71a98be453c3ea40d4f96019e--

-----------------------------abe834719cdf618f78719d2a8f3edd06--
```
This format is completely compatible with Couchbase Lite and also Sync Gateway implementation.

## Checklist


janl:
Hi @AlexanderKaraberov,

this is pretty cool. Before I do a more thorough review, I’m wondering why you didn’t opt to re-use `couch_doc:doc_to_multi_part_stream()` / `couch_httpd_db:encode_multipart_stream()` for the multipart streaming?

AlexanderKaraberov:
Hi @janl ,
Thanks for a good question. Long story short: I tried to re-use couch_doc:doc_to_multi_part_stream() but it wasn’t working, because content headers wasn’t encoded properly. Hence I decided to write a custom one based on rcouch instead of modifying/breaking existing functionality. Now I understand that this was because I didn’t use (didn’t know) about couch_httpd_db:encode_multipart_stream(). Function name looks like exactly what I needed at that point of time... Unfortunately I still have a superficial knowledge of full CouchDB codebase and its hidden gems. If you don’t mind I can try to reimplement this functionality using mentioned functions instead of mine and test whether this will work with Sync Gateway.

janl:
Yeah, I really don't feel comfortable having two multipart implementations in CouchDB. We already don't like the one we have. ;)

AlexanderKaraberov:
Hi @janl , thank you for your advice to stick to the `couch_doc:doc_to_multi_part_stream()` and `couch_httpd_db:encode_multipart_stream()`. I realised that I wrote a lot of duplicated code :)
I pushed a revised implementation.

AlexanderKaraberov:
Hi,
is there a small chance this PR might be reviewed before the 2.2.0 release? It would nice to have this functionality in the CouchDB 2.2

wohali:
@AlexanderKaraberov Sorry about not being able to get this included in 2.2.0, but it might land for 2.2.1 or 2.3.0.

@mikerhodes @davisp another one that could use your reviews. Would also love any comments from the PouchDB crew e.g. @garrensmith @daleharvey

wohali:
@AlexanderKaraberov Can you resolve the conflicts on this PR please?

@garrensmith do you think you could review this PR please?

src/chttpd/src/chttpd_db.erl
```diff
@@ -476,19 +477,59 @@ db_req(#httpd{method='POST', path_parts=[_, <<"_bulk_get">>]}=Req, Db) ->
...
+            AcceptMixedMp = MochiReq:accepts_content_type("multipart/mixed"),
+            AcceptRelatedMp = MochiReq:accepts_content_type("multipart/related"),
+            AcceptMp = AcceptMixedMp orelse AcceptRelatedMp,
+            case AcceptMp of
+               false ->
+                  {ok, Resp} = start_json_response(Req, 200),
+                  send_chunk(Resp, <<"{\"results\": [">>),
+                  lists:foldl(fun(Doc, Sep) ->
+                      {DocId, Results, Options1} = bulk_get_open_doc_revs(Db, Doc,
+                                                                          Options),
```
nickva: Indent this 4 space to the right `{DocId..` on previous line:

```
{DocId, Results, ...}
    Options)
```

src/chttpd/src/chttpd_db.erl
```diff
@@ -476,19 +477,59 @@ db_req(#httpd{method='POST', path_parts=[_, <<"_bulk_get">>]}=Req, Db) ->
...
+                  CType = {"Content-Type", MpType ++ "; boundary=\"" ++
+                                        ?b2l(OuterBoundary) ++  "\""},
+                  {ok, Resp} = start_chunked_response(Req, 200, [CType]),
+                  lists:foldl(fun(Doc, _Pre) ->
+                      case bulk_get_open_doc_revs(Db, Doc, Options) of
+                          {_, {ok, []}, _Options1} ->
+                              ok;
+                          {_, {ok, Results}, Options1} ->
+                              send_docs_multipart(bulk_get, Req, Results, Options1,
+                                                  OuterBoundary, Resp);
```
nickva: Indent 4 spaces to the right of send_doc_multipart like in the previous `{DocId, Results,...}` case

src/chttpd/src/chttpd_db.erl
```diff
@@ -476,19 +477,59 @@ db_req(#httpd{method='POST', path_parts=[_, <<"_bulk_get">>]}=Req, Db) ->
...
+                          {DocId, {error, {RevId, Error, Reason}}, _Options1} ->
+                              Json = ?JSON_ENCODE({[{<<"id">>, DocId},
+                                                    {<<"rev">>, RevId},
```
nickva: Keep `[` on `?JSON_ENCODE` line but indent all the {..., ...} tuples 4 spaces to the right of `Json`

Like:
```
Json = ?JSON_ENCODE({[
    {<<"id">>, DocId},
    {<<"rev">>, RevId},
    ...
]}),
```

src/chttpd/src/chttpd_db.erl
```diff
@@ -476,19 +477,59 @@ db_req(#httpd{method='POST', path_parts=[_, <<"_bulk_get">>]}=Req, Db) ->
...
+                                                    {<<"error">>, Error},
+                                                    {<<"reason">>, Reason}]}),
+                              couch_httpd:send_chunk(Resp,
+                                  [<<"\r\n--", OuterBoundary/binary>>,
```
nickva: Keep the `[` on the `couch_httpd:` line

src/chttpd/src/chttpd_db.erl
```diff
@@ -476,19 +477,59 @@ db_req(#httpd{method='POST', path_parts=[_, <<"_bulk_get">>]}=Req, Db) ->
...
+                                  <<"\r\nContent-Type: application/json; error=\"true\"\r\n\r\n">>,
+                                  Json])
```
nickva: Maybe put `])` on their own line lining up with `couch_httpd:send_chunk`

src/chttpd/src/chttpd_db.erl
```diff
@@ -952,21 +1025,34 @@ send_docs_multipart(Req, Results, Options1) ->
             couch_doc:doc_to_multi_part_stream(InnerBoundary, JsonBytes, Atts,
                     fun(Data) -> couch_httpd:send_chunk(Resp, Data)
                     end, true),
-             couch_httpd:send_chunk(Resp, <<"\r\n--", OuterBoundary/binary>>)
```
nickva: This wasn't indented correctly to start with, but let's keep reformatting for a separate PR.

src/chttpd/src/chttpd_db.erl
```diff
@@ -952,21 +1025,34 @@ send_docs_multipart(Req, Results, Options1) ->
...
+            couch_httpd:send_chunk(Resp,
                 [<<"\r\nContent-Type: application/json; error=\"true\"\r\n\r\n">>,
                 Json,
                 <<"\r\n--", OuterBoundary/binary>>])
          end, Results),
     couch_httpd:send_chunk(Resp, <<"--">>),
     couch_httpd:last_chunk(Resp).
 
+bulk_get_multipart_headers({0, []}, Id, Boundary) ->
+    [<<"\r\nX-Doc-Id: ", Id/binary>>,
```
nickva: Maybe a bit pedantic but let's keep `[` on its own line and `<<...>>>`  on separate lines indented by 4 spaces. Then final `]` matching with the original `[`. Like so:

```
[
    <<"\r\n...."/binary>>,
    <<....>>>
];
```

src/chttpd/src/chttpd_db.erl
```diff
@@ -932,7 +973,39 @@ send_doc_efficiently(#httpd{mochi_req=MochiReq}=Req, #doc{atts=Atts}=Doc, Header
         send_json(Req, 200, Headers, couch_doc:to_json_obj(Doc, Options))
     end.
 
-send_docs_multipart(Req, Results, Options1) ->
+send_docs_multipart(bulk_get, _Req, Results, Options1, OuterBoundary, Resp) ->
```
nickva: Start at `Options0` then after the first update make `Options`

nickva: I think you were working at the same time :-). But yeah I agree, make a separate function as there is not much of a benefit to make it the same function.

Also a single squashed commit rebased on master would be ok.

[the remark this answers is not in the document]
AlexanderKaraberov: Sorry overlooked this comment when I had written my questions above. Sure will do.

src/chttpd/test/chttpd_db_bulk_get_multipart_test.erl
```diff
@@ -0,0 +1,309 @@
...
+    mock(couch_stats),
+    mock(fabric),
+    mock(mochireq),
+    Pid = spawn_accumulator(),
+    Pid.
+
+
+teardown(Pid) ->
+    ok = stop_accumulator(Pid),
+    meck:unload(config),
```
nickva: I think you can just do one meck:unload() and it will unload all mocked modules.

nickva:
Look better. Nice work cleaning it up.

I see eunit test failures locally and in Travis. Not sure exactly why, maybe because of the name change of the function or maybe something on master changed?

Also see a minor comment about simplifying meck unloading.

src/chttpd/test/chttpd_db_bulk_get_multipart_test.erl
AlexanderKaraberov: This is my blunder. I overlooked that `meck` has this convenient function.

AlexanderKaraberov:
@nickva I checked Travis logs and seemingly only one job `Job #6237.5` for OTP 17.5 failed for some odd reason. All other ones are successful. I've checked the logs and I do believe this error has nothing to do with my functionality. Perhaps some configs are wrong:
```
ERROR: One or more eunit tests failed.
ERROR: eunit failed while processing /home/travis/build/apache/couchdb/src/couch: rebar_abort
make[1]: *** [eunit] Error 1
make[1]: Leaving directory `/home/travis/build/apache/couchdb'
make: *** [check] Error 2
The command "make check" exited with 2.
```

nickva:
Ah it's probably flaky, I'll try to restart it

AlexanderKaraberov:
@nickva Thanks! Indeed it was some sporadic fail. All is good now. Let me know if I need to address something else before this can be merged into `master`.

src/chttpd/src/chttpd_db.erl
```diff
@@ -481,18 +482,62 @@ db_req(#httpd{method='POST', path_parts=[_, <<"_bulk_get">>]}=Req, Db) ->
...
-                bulk_get_send_docs_json(Resp, DocId, Results, Options1, Sep),
-                <<",">>
-            end, <<"">>, Docs),
-
-            send_chunk(Resp, <<"]}">>),
-            end_json_response(Resp)
+            % Decide whether to use multipart response or application/json
+            AcceptMixedMp = MochiReq:accepts_content_type("multipart/mixed"),
+            AcceptRelatedMp = MochiReq:accepts_content_type("multipart/related"),
+            AcceptMp = AcceptMixedMp orelse AcceptRelatedMp,
```
nickva: Sorry one more thing I realized. I think this logic would make multipart response become the default if there is no accept header or it accepts '*/*'. 

In general the semantics would be correct, but given that this is a new behavior, it would be better if in such cases we returned json, and only if json is not accepted but multipart is then we send multipart.

So I think we'd want something like:

```
AcceptJson =  MochiReq:accepts_content_type("application/json"),
...
SendMp = not AcceptJson andalso (AcceptMixedMp orelse AcceptRelatedMp)

case SendMp of ...
```

So we are saying we only send mp if json wasn't requested (either because of a */* accept header or explicitly) and one of the mp flavors was. Otherwise we send json as before.

src/chttpd/src/chttpd_db.erl
```diff
@@ -997,6 +1075,23 @@ send_docs_multipart(Req, Results, Options1) ->
...
+    RevStr = couch_doc:rev_to_str({Start, FirstRevId}),
+    [
+        <<"\r\nX-Doc-Id: ", Id/binary>>,
+        <<"\r\nX-Rev-Id: ", RevStr/binary>>,
+        <<"\r\nContent-Type: multipart/related; boundary=", Boundary/binary, "\r\n\r\n">>
+    ].
+
+bulk_get_multipart_boundary() ->
+    Unique = couch_uuids:random(),
+    <<"---------------------------", Unique/binary>>.
```
nickva: Another question. Why does it have to be `<<"---------------------------"` and not `<<"--"`? Was it for testing mainly or another reason?

src/chttpd/src/chttpd_db.erl
AlexanderKaraberov: @nickva Valid point even though I consider this kind of scenario a rare edge case. I'll make appropriate changes.

src/chttpd/src/chttpd_db.erl
AlexanderKaraberov: > Was it for testing mainly

Exactly. Will clean it up.

src/chttpd/src/chttpd_db.erl
AlexanderKaraberov: @nickva Actually this change broke my tests with `got invalid Content-Type 'application/json'`
Have to investigate why.

AlexanderKaraberov: For some reason `MochiWeb`'s `accepts_content_type("application/json")` returns `true` even though the client sent these headers:
```
Accept = "multipart/related";
"Content-Type" = "application/json";
```
With previous logic it works perfectly because we have an explicit check for multipart `Accept`.

src/chttpd/src/chttpd_db.erl
AlexanderKaraberov: > Was it for testing mainly

Exactly. I've just forgotten to clean this up. Committed a fix.

src/chttpd/src/chttpd_db.erl
AlexanderKaraberov: @nickva Also I see the same logic in the normal db/doc request [here](https://github.com/apache/couchdb/blob/master/src/chttpd/src/chttpd_db.erl#L756) where we first just check for `multipart` `Accept` header. When i wrote this implementation for `_bulk_get` I mostly copied `db_doc_req` logic.

nickva: I think the difference is that with existing open_revs logic it's ok if it is inconsistent because it is already release, but I am worried about changing the default behavior of what _bulk_gets if there is no `Accept` header or `Accept` header contains `*/*` (that is accept anything).

In those cases I think `_bulk_get` will start returning mp content instead of json as before. Try for example sending a request before the PR to _bulk_get without an Accept header and after the PR. For me, it looked like the default changed and that might be unexpected (even if technically correct from the HTTP request / response point).

AlexanderKaraberov: @nickva Yeah you are right here. Sorry for my unjustified stubbornness. I pushed a fix and squashed all the commits. Also the branch is rebased again.

nickva:
Added a rebased on master PR https://github.com/apache/couchdb/pull/1195

Not sure why "update branch" button in GH doesn't allow rebasing on master :-)

nickva:
@AlexanderKaraberov 

Thank you again for you contribution. I rebased and merged your PR. 

Please make a documentation PR to along with it.

AlexanderKaraberov:
@nickva Thanks! Sure, I shall allocate some time in order to update related documentation and submit a PR for it.

wohali:
@AlexanderKaraberov Thanks, please be advised I am moving swiftly into 2.3.0 release time, so a little urgency would be much appreciated. :)

PR #1199, opened 2018-03-03
Prevent access to Fauxton on node-local port (5986)

### LINKED ISSUE #1198: Fauxton should not be available on port 5986
People mistakenly stumble across Fauxton on port 5986 and think that this is "where CouchDB is." I've seen it multiple times on IRC, and professionally, where people stand up "single node CouchDB 2.1.1" and put all their databases on port 5986. It even sometimes encourages people to muck with the node-local system databases, including `_dbs` and `_nodes`, cluelessly.

# AUGH. No.

Removing Fauxton from the node-local port should help stop some of this mess.

Our documentation even mentions it: https://github.com/apache/couchdb-documentation/blob/master/src/cluster/sharding.rst#scaling-out

This has to go.

wohali:
Will help stop people shooting themselves in the foot and/or using
node-local CouchDB as their "main" CouchDB port.

I'll file a separate docs repo PR on this once this merges.

Test results:

```
$ curl -v http://localhost:15986/_utils
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 15986 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 15986 (#0)
> GET /_utils HTTP/1.1
> Host: localhost:15986
> User-Agent: curl/7.52.1
> Accept: */*
>
< HTTP/1.1 410 Gone
< Cache-Control: must-revalidate
< Content-Length: 110
< Content-Type: application/json
< Date: Sat, 03 Mar 2018 00:45:22 GMT
< Server: CouchDB/2.2.0-51cb6aecc (Erlang OTP/19)
<
{"error":"no_node_local_fauxton","reason":"The web interface is no longer available on the node-local port."}
```

Closes #1198

### DIFF
--- a/src/couch/src/couch_httpd_misc_handlers.erl
+++ b/src/couch/src/couch_httpd_misc_handlers.erl
@@ -61,30 +61,9 @@ handle_file_req(#httpd{method='GET'}=Req, Document) ->
 handle_file_req(Req, _) ->
     send_method_not_allowed(Req, "GET,HEAD").
 
-handle_utils_dir_req(#httpd{method='GET'}=Req, DocumentRoot) ->
-    "/" ++ UrlPath = couch_httpd:path(Req),
-    case couch_httpd:partition(UrlPath) of
-    {_ActionKey, "/", RelativePath} ->
-        % GET /_utils/path or GET /_utils/
-        CachingHeaders = [{"Cache-Control", "private, must-revalidate"}],
-        EnableCsp = config:get("csp", "enable", "false"),
-        Headers = maybe_add_csp_headers(CachingHeaders, EnableCsp),
-        couch_httpd:serve_file(Req, RelativePath, DocumentRoot, Headers);
-    {_ActionKey, "", _RelativePath} ->
-        % GET /_utils
-        RedirectPath = couch_httpd:path(Req) ++ "/",
-        couch_httpd:send_redirect(Req, RedirectPath)
-    end;
 handle_utils_dir_req(Req, _) ->
-    send_method_not_allowed(Req, "GET,HEAD").
-
-maybe_add_csp_headers(Headers, "true") ->
-    DefaultValues = "default-src 'self'; img-src 'self' data:; font-src 'self'; "
-                    "script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline';",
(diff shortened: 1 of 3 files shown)


src/couch/src/couch_httpd_misc_handlers.erl
```diff
@@ -61,22 +61,8 @@ handle_file_req(#httpd{method='GET'}=Req, Document) ->
...
-        Headers = maybe_add_csp_headers(CachingHeaders, EnableCsp),
-        couch_httpd:serve_file(Req, RelativePath, DocumentRoot, Headers);
-    {_ActionKey, "", _RelativePath} ->
-        % GET /_utils
-        RedirectPath = couch_httpd:path(Req) ++ "/",
-        couch_httpd:send_redirect(Req, RedirectPath)
-    end;
 handle_utils_dir_req(Req, _) ->
-    send_method_not_allowed(Req, "GET,HEAD").
+   throw({forbidden, <<"The Fauxton interface is no longer available on the node-local port.">>}).
```
wohali: If people don't like returning a `403` here, we could consider returning `410` (Gone) instead.

chewbranca: This should be indented one space further in, eg 4 spaces instead of 3.

chewbranca:
+0.5 from me (after the WS nit). The approach seems fine, but I don't have strong enough opinions on removal of Fauxton from `:5986` to give it a full +1, so it would be good to get some more feedback.

rnewson:
losing the CSP tests seems a bit unfortunate.

wohali:
These were only testing CSP on the node-local port anyway. I can port them to chttpd, give me a day or two.

wohali:
@janl Hm, if we bind chttpd to 0.0.0.0 it could be tricky to determine the right IP address to choose, 127.0.0.1 might not be visible if they are coming from outside (and mistakenly have bound httpd to 0.0.0.0 as well.) It'd sure be a nice to have, though.

wohali:
@rnewson I've moved the CSP tests over to chttpd, where they should be anyway, and of course they pass with flying colours because we are awesome. :)

janl:
@wohali good call, with a load balancer it gets even worse.

PR #1201, opened 2018-03-03
This fixes couch_bt_engine:fold_local_docs/4

davisp:
Somehow managed to fix without getting it into the PSE PR that was
merged the other day.

## Testing recommendations

curl _local_docs endpoint

## Checklist

### DIFF
--- a/src/couch/src/couch_bt_engine.erl
+++ b/src/couch/src/couch_bt_engine.erl
@@ -464,7 +464,10 @@ fold_docs(St, UserFun, UserAcc, Options) ->
 
 
 fold_local_docs(St, UserFun, UserAcc, Options) ->
-    fold_docs_int(St, St#st.local_tree, UserFun, UserAcc, Options).
+    case fold_docs_int(St, St#st.local_tree, UserFun, UserAcc, Options) of
+        {ok, _Reds, FinalAcc} -> {ok, null, FinalAcc};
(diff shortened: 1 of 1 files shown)


src/couch/src/couch_bt_engine.erl
```diff
@@ -464,7 +464,10 @@ fold_docs(St, UserFun, UserAcc, Options) ->
 
 
 fold_local_docs(St, UserFun, UserAcc, Options) ->
-    fold_docs_int(St, St#st.local_tree, UserFun, UserAcc, Options).
+    case fold_docs_int(St, St#st.local_tree, UserFun, UserAcc, Options) of
+        {ok, _Reds, FinalAcc} -> {ok, null, FinalAcc};
```
tonysun83: In fold_docs_int, when it's a ```local_tree``` we return ```{ok, 0, FinalUserAcc};```. What's the reasoning behind changing it to 0?

davisp: Good call. Its actually higher up when its expected to be null in fabric. What happens is that depending on what shards have actual docs you get a combination of null and 0 which leads to badarith exceptions when trying to do `0 + null` basically.

This is definitely a bit of a sharp corner on PSE vs features that assume legacy engine behavior. But for v1 its mostly a game of trying to define existing behavior so that we can refine what's expected/acceptable for future engines.

tonysun83:
+1

PR #1215, opened 2018-03-09
Introduce bin_opt_info erl_opts compilation option

jaydoane:
## Overview

As discussed in
http://erlang.org/doc/efficiency_guide/binaryhandling.html, the
bin_opt_info option can have the compiler print information about
binary optimizations. Typically, it issues warnings about code that
can not be optimized. For example, with the option enabled, it warns
about this line of code (among many others) like this:
```
src/chttpd/src/chttpd_auth_cache.erl:183: Warning: NOT OPTIMIZED: sub binary is used or returned
```
Not only should having this option enabled draw awareness to existing
sub-optimal binary constructs, it should help developers avoid
creating new ones.

## Testing recommendations

This is a compiler configuration change, so test it by running `make clean && make`. You should now see numerous warnings about failed binary optimizations such as
```
/Users/jay/proj/couchdb/src/couch_replicator/src/couch_replicator_api_wrap.erl:401: Warning: INFO: matching anything else but a plain variable to the left of binary pattern will prevent delayed sub binary optimization; SUGGEST changing argument order
/Users/jay/proj/couchdb/src/couch_replicator/src/couch_replicator_api_wrap.erl:1019: Warning: NOT OPTIMIZED: sub binary is used or returned
```

## Checklist

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -92,7 +92,7 @@ AddConfig = [
     {deps, lists:map(MakeDep, DepDescs)},
     {sub_dirs, SubDirs},
     {lib_dirs, ["src"]},
-    {erl_opts, [debug_info, {i, "../"}]},
+    {erl_opts, [bin_opt_info, debug_info, {i, "../"}]},
     {eunit_opts, [verbose]},
     {plugins, [eunit_plugin]},
     {dialyzer, [


eiri:
@jaydoane @janl @wohali 

I propose to revert this commit for the for time being. It doesn't seems that anybody doing anything to fix all those warnings and as I've tried a couple of times on mrview modules I can tell it's often not so trivial thing to do.

Meanwhile it generates a wall of the warnings making hunting for an actual warnings and compilation errors during dev process quite a chore.

jaydoane:
While I agree the wall of warnings is annoying (and also not trivial to fix), I personally find it helpful to prevent new warnings when working on new code. However, if it makes it even more difficult to work on the code base, I'm ok with reverting (although I'm also aware my vote doesn't count).

eiri:
@jaydoane But does it prevents the new warnings? We don't count the diff, just dump a bunch of noise, so new warnings just get lost in the old ones. I assume the intent was to annoy people into fixing those warnings, but two month later I don't see that this is working

wohali:
fine by me

PR #1228, opened 2018-03-21
Make bcrypt optional

iilyak:
## Overview

Support additional dependencies for couch app and make bcrypt optional. This changes a build system to allow exclusion of bcrypt dependency. There should be no functionality changes.

## Testing recommendations

1. do a fresh clone
2. run `./configure`
3. run `make && dev/run --admin=adm:pass`
4. run `dev/remsh`
5. In `remsh` make sure you still have `bcrypt` by running the following 

```
lists:usort([A || {A, _, _} <- application:loaded_applications()]).
```

None

## Checklist

### DIFF
--- a/rel/apps/couch.config
+++ b/rel/apps/couch.config
@@ -1,2 +1,3 @@
 {applications, [
+    bcrypt
 ]}.
--- a/rel/apps/couch.config
+++ b/rel/apps/couch.config
@@ -1,0 +1,2 @@
+{applications, [
+]}.
--- a/src/couch/src/couch.app.src.script
+++ b/src/couch/src/couch.app.src.script
@@ -9,6 +9,8 @@
 % WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 % License for the specific language governing permissions and limitations under
 % the License.
+ConfigFile = filename:join([os:getenv("COUCHDB_APPS_CONFIG_DIR"), "couch.config"]).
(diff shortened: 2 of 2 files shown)


rnewson:
this only addresses the dependency, I'd like to review it context of (presumably) the code that returns a clean error message when configuring a user or admin with a bcrypt hashed password.

I also don't think this needs to be a generic mechanism, a "--disable-bcrypt" option at build time, say. If we add further such tweaks, we could invent a more general thing.

wohali:
The "more general thing" would be useful for things like optionally including geo/search at runtime. That could easily be extended to replace a `--disable-bcrypt` option. But for now a `./configure` flag would be sufficient, I think.

And I agree that knowing what to do if only bcrypt-ed passwords are in `.ini` files and bcrypt is disabled would be important. :)

iilyak:
70% done
 ```
git diff --stat
 configure                            |  8 ++++++++
 rebar.config.script                  | 18 +++++++++++++-----
 src/chttpd/src/chttpd_auth_cache.erl | 19 +++++++++++++++++--
 src/couch/rebar.config.script        |  3 +++
 src/couch/src/couch.app.src.script   |  4 +++-
 src/couch/src/couch_users_db.erl     |  5 ++++-
 6 files changed, 48 insertions(+), 9 deletions(-)
```

Too complex and would require tests. Dropping it.

PR #1231, opened 2018-03-22
Replace resource expensive bcrypt test with shorter version

eiri:
Since we are using 3rd party app for `bcrypt` and `couch_passwords: bcrypt/2` is just an interface for it there are no need to test correctness of the lib itself, just the fact that we are using it properly.

Test `bcrypt_logRounds_18` is kind of a resource hog, so I suggest to remove it to easy a  burden on Travis and Jenkins. This reduces the test runtime on my laptop from 75s to 15s and I suspect the difference for CI will be even greater.

### DIFF
--- a/src/couch/test/couch_passwords_tests.erl
+++ b/src/couch/test/couch_passwords_tests.erl
@@ -53,19 +53,15 @@ pbkdf2_test_()->
            )}}]}.
 
 
-
-setup() ->
-    test_util:start(?MODULE, [bcrypt]).
-
-teardown(Ctx)->
-    test_util:stop(Ctx).
-
 bcrypt_test_() ->
     {
         "Bcrypt",
         {
-            foreach,
-            fun setup/0, fun teardown/1,
+            setup,
+            fun() ->
+                test_util:start_applications([bcrypt])
+            end,
+            fun test_util:stop_applications/1,
             [
                 {timeout, 1, fun bcrypt_logRounds_4/0},
                 {timeout, 1, fun bcrypt_logRounds_5/0},
--- a/src/couch/test/couch_passwords_tests.erl
+++ b/src/couch/test/couch_passwords_tests.erl
@@ -68,6 +68,7 @@ bcrypt_test_() ->
             fun setup/0, fun teardown/1,
             [
                 {timeout, 1, fun bcrypt_logRounds_4/0},
+                {timeout, 1, fun bcrypt_logRounds_5/0},
                 {timeout, 5, fun bcrypt_logRounds_12/0},
                 {timeout, 5, fun bcrypt_null_byte/0}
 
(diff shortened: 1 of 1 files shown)


iilyak:
good idea.

wohali:
We don't need to validate that the algorithm works - just that it's able to function correctly when linked and used.

wohali:
Oh, thought - maybe make it `bcrypt_logRounds_5` instead, or some other (lower) number, as a way to validate that you can adjust the figure and the library still does the right thing?

eiri:
@wohali I've added `bcrypt_logRounds_5` and (surprise!) found that the tests were actually  broken - generators been returning test objects instead of running assertions, infamous and hugely common error.

I fixed that, simplified setup (we've been starting and stopping bunch of unused apps and for some reason did it for each test individually) and added the tests descriptions. How does it look for you now?

wohali:
Good catch! +1 and go for it.

wohali:
@eiri we got a random failure in Jenkins related to encryption after this hit master, can you investigate please?

https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/223/pipeline

Logs at https://couchdb-vm2.apache.org/ci_errorlogs/jenkins-couchdb-223-2018-03-23T23:15:34.663810/couchlog.tar.gz

```
test/javascript/tests/users_db_security.js                     
    Error: undefined
Trace back (most recent call first):
    
  52: test/javascript/test_setup.js
      T((void 0))
  35: test/javascript/tests/users_db_security.js
      ("jchris1")
  39: test/javascript/tests/users_db_security.js
      ([object Object],"org.couchdb.user:fdmanana","jchris1")
 168: test/javascript/tests/users_db_security.js
      ("pbkdf2",(function (derived_key) {TEquals(40, derived_key.length, "de
 413: test/javascript/tests/users_db_security.js
      ()
 395: test/javascript/couch_test_runner.js
      run_on_modified_server([object Array],(function () {try {testFun(schem
 411: test/javascript/tests/users_db_security.js
      ("pbkdf2",0,[object Array])
 399: test/javascript/tests/users_db_security.js
      ()
  37: test/javascript/cli_runner.js
      runTest()
  48: test/javascript/cli_runner.js
```

eiri:
@wohali It's hard to imagine that fixing eunit tests would break API tests.

I looked into it and as far as I can see with `bcrypt` PR API test `users_db_security.js` was changed to be prune to racing condition - user database [got deleted and re-created](https://github.com/apache/couchdb/blob/master/test/javascript/tests/users_db_security.js#L414), which is known anti-pattern.

wohali:
Thanks @eiri, sorry for leaving this request here. I wasn't accusing you of making the mistake :) I'll open a new ticket. Thank you again!

PR #1245, opened 2018-03-27
feat: add debug log output for shard open errors

janl:
We have a number of issue reports that show “No DB shards could be opened” which we are stuck in resolving, because we don’t know *why* db shards can’t be opened. This PR aims to add debug logging for when shards can’t be opened.


wohali:
Moving this to `debug` level will bury these errors at normal log levels, won't it?

wohali:
This is extra logging at `debug` level, the `No DB shards could be opened` error stays, if I'm reading this right. If we don't have any deeper insight, this is reasonable. Would like to hear from someone else before +1.

PR #1249, opened 2018-03-28
add test covering loading admins from config

janl:


janl:
Travis suggests this new test is flakey.

Is `config:set()` async? e.g. can there be no admin configured if trying to `config:get()` right after?

janl:
new patch with proper wait

eiri:
This should do it, I think. For a reason why this needed - `config:set` is sync, but config change broadcasting and consequently pass hashing isn't, so half of time `get_admin/1` in couch_auth cache will be getting "password" instead of prefixed hash and default to `nil`

Issue #1221, opened 2018-03-14
`_changes` filtered on `_doc_ids` returns unexpected `last_seq`


kdoran:
## Expected Behavior
From: http://docs.couchdb.org/en/2.1.1/api/database/changes.html#post--db-_changes:

> "last_seq is the update sequence of the last update returned (Equivalent to the last item in the results)."

When filtering on `_doc_ids`, I was expecting `last_seq` to be the same thing as the last `seq` in the list of changes returned when filtering on _doc_ids.

## Current Behavior
When filtering on `_doc_ids`, `_changes` responds with the latest sequence in the database or, if there's a `limit` argument, the sequence number after the limit number.

## Steps to Reproduce (for bugs)
1. `POST /{db}/_changes?filter=_doc_ids&since=0` with body `{doc_ids: ['one-example-id']}`
2. See that `last_seq` in response is not the last `seq` in the list of changes.

## Context
Grabbing the `last_seq` from the `_changes` response and using it as the `since` parameter on the next request to `_changes` is PouchDB's approach to a replication filtered on IDs. Using that later sequence number means documents are (sometimes) missed in the replication. 

https://github.com/pouchdb/pouchdb/issues/6809

## Your Environment
Saw this when running a single node CouchDB 2.1 & on CouchDB 1.6 locally on a macbook pro.

wohali:
@kdoran Could this be a duplicate of #748?

kdoran:
@wohali it seemed like it could be, but then I saw the same behavior happening on a 1.7x... so would that mean it's a separate issue?

wohali:
@kdoran Definitely worth investigating, I missed that it was also an issue in 1.x.

kdoran:
👍  for our case it ended up being just fine to use the last `seq` in the change results list instead of `last_seq` in case others stumble across this.

wohali:
@kdoran We need a `last_seq` and the final row's `seq` value to investigate further, of a response that has caused the issue. Can you provide? Thanks!

wohali:
I wonder if this is related to #984 and @rnewson 's investigation there.

kdoran:
@wohali yup, sorry for the delay! Here's a quick js example I ran in the chrome devtools console while on the fauxton page on my local CouchDB, which is 2.1.0 as a single node. The same discrepancy between `seq` and `last_seq` also happens when running this on a remote EC2 CouchDB running 2.1.1 also as a single node:

```
[...]

PR #1251, opened 2018-03-28
Fix shard substitution in changes feeds

davisp:
PSE changed the local_tree's join function so we need to update this. Previously we read _local docs out of the btree that ended up just giving us {DocId, {RevId, {BodyProps}}} to our callback. PSE changed things such that we now use couch_db:fold_local_docs which provides `#doc` records.

Shard substitution happens when a changes feed is provided with a since_seq that references a ndoe that does not or has errored out some how. The substitution then goes and looks for an internal replication document so that we don't have to rewrind the changes feed entirely (i.e., we look at the internal replication checkpoint and can rewind to the place where the node-being-replaced last sent us documents).


wohali:
@davisp is this at all related to #1221 ? i'm not sure it does since it looks like the reported bug applies even when shards have not yet been substituted. if so ignore this comment.

davisp:
@wohali Narp. That's unrelated.

chewbranca:
+1

PR #1252, opened 2018-03-28
Fix killing of OS processes

davisp:
This was a latent bad merge that failed to remove the duplicate receive
statement. This ended up discarding the monitor's 'DOWN' message which
leads to an infinite loop in couch_os_proces:killer/1.


chewbranca:
+1

PR #1255, opened 2018-03-29
Fix compaction resumption for the BT engine

davisp:
This bug prevents the proper resumption of compactions that died during
the meta copy phase. The issue is that we were setting the update_seq
but not copying over the id and seq tree states. Thus when compaction
resumed from the bad files we'd end up skipping the part where we copy
docs over and then think everything was finished. Thus completely
clearing a database of its contents.

Luckily this isn't release code and as such should have fairly minimal
impact other than those who might be running off master.

Note: Once I finish emergency patching everything affected by this bug I'm gonna add a unit test that shows this behavior.

### DIFF
--- a/src/couch/src/couch_bt_engine.erl
+++ b/src/couch/src/couch_bt_engine.erl
@@ -92,6 +92,7 @@
 % Used by the compactor
 -export([
     set_update_seq/2,
+    update_header/2,
     copy_security/2
 ]).
 
--- a/src/couch/src/couch_bt_engine_compactor.erl
+++ b/src/couch/src/couch_bt_engine_compactor.erl
@@ -386,7 +386,7 @@ commit_compaction_data(#st{header = OldHeader} = St0, Fd) ->
     MetaFd = couch_emsort:get_fd(St0#st.id_tree),
     MetaState = couch_emsort:get_state(St0#st.id_tree),
     St1 = bind_id_tree(St0, St0#st.fd, DataState),
-    Header = St1#st.header,
+    Header = couch_bt_engine:update_header(St1, St1#st.header),
     CompHeader = #comp_header{
         db_header = Header,
         meta_state = MetaState


rnewson:
+1

src/couch/test/couch_bt_engine_compactor_tests.erl
```diff
@@ -0,0 +1,130 @@
...
+        {ok, _} = couch_db:update_docs(Db, [Doc1, Doc2, Doc3]),
+        couch_db:ensure_full_commit(Db)
+    end).
+
+
+compact_db(DbName) ->
+    couch_util:with_db(DbName, fun(Db) ->
+        {ok, _} = couch_db:start_compact(Db)
+    end),
+    wait_db_compact_done(DbName, ?WAIT_DELAY_COUNT).
```
iilyak: There is a helper function for this in test_util. It can be easily used here as:
```
    test_util:wait(fun() ->
        IsDone = couch_util:with_db(DbName, fun(Db) ->
            not is_pid(couch_db:get_compactor_pid(Db))
        end),
        if IsDone -> ok; true -> wait end
    end).
```

src/couch/test/couch_bt_engine_compactor_tests.erl
```diff
@@ -0,0 +1,130 @@
...
+
+check_db_validity(DbName) ->
+    couch_util:with_db(DbName, fun(Db) ->
+        ?assertEqual({ok, 3}, couch_db:get_doc_count(Db)),
+        ?assertEqual(3, couch_db:count_changes_since(Db, 0))
+    end).
+
+
+with_mecked_emsort(Fun) ->
+    meck:new(couch_emsort, [passthrough]),
```
iilyak: I am a bit worried about use of `meck:new` and `meck:expect` in the body of the test. It is known to cause problems with tests which run after the one which uses the mock. However in this case I don't see any easy ways to change that. Besides you implemented try/catch. I think it should be fine. Just logging the concern for posterity.

src/couch/test/couch_bt_engine_compactor_tests.erl
iilyak: Correction:
```
    ?assertEqual(ok, test_util:wait(fun() ->
        IsDone = couch_util:with_db(DbName, fun(Db) ->
            not is_pid(couch_db:get_compactor_pid(Db))
        end),
        if IsDone -> ok; true -> wait end
    end)).
```

davisp: Ah good call. I just modified a different test that used compaction. Will replace this bit.

src/couch/test/couch_bt_engine_compactor_tests.erl
davisp: Yar, as long as its got try catch I'm pretty sure that's about all we can do. And yeah, unfortunately I need to call both the mocked and unmocked versions to assert the behavior given that we're explicitly testing a retry.

PR #1272, opened 2018-04-09
Record and expose last compaction time

rnewson:
For a clustered database, report the earliest compaction time from any
shard. If any shard does not have a recorded compaction time, show
this as null to avoid confusion.

## Checklist


davisp:
What's the motivation here? We're already logging when compaction finishes so seems kinda fishy to have it recorded twice. Also not sure on adding it to the API as it'd seem out of place for non-compacting storage engines.

rnewson:
@davisp the motivation is to allow users to know when old revisions are finally gone from the live database files, handy if you need/want to know that a delete has finally happened.

davisp:
Ah makes sense. Only thing that concerns me is that clustered purge work is happening around this as well so we'll want to figure out how best to order things for that.

src/couch/src/couch_bt_engine_header.erl
```diff
@@ -229,6 +236,7 @@ upgrade_disk_version(#db_header{}=Header) ->
         3 -> throw({database_disk_version_error, ?OLD_DISK_VERSION_ERROR});
         4 -> Header#db_header{security_ptr = nil}; % [0.10 - 0.11)
         5 -> Header; % pre 1.2
+        6 -> Header;
```
rnewson: @davisp on reflection, I don't think I need to bump disk version as we're only adding a new field.

davisp:
I'm also mulling over how we might make this a bit more general rather than specifically inserting a last_compaction_time or w/e. Seems like having a storage-engine-defined set of extra info would be generally useful. Only sticking point is how we combine that for the cluster endpoint.

Though on a semi-related note I was wanting an option the other day to show me the db info blobs for all shards rather than the combined version. Perhaps if we do that we can make an arbitrary "engine_info" key that allows storage engines to expose any data they might feel like having there?

rnewson:
Hm, does that block this, though? If we make it more generic later, we can re-implement 'last compaction' within that framework. I don't intend to write that framework, do you?

davisp:
@chewbranca Had a PR or diff at least for the separate db info blobs. The PSE level bit of work would be relatively straight forward. Though for the most part on this I'm still worried about the clustered purge interaction and which order we should be thinking of things there.

janl:
@rnewson do you wanna bring this feature forward? I’m asking for a 2.3.0 milestone :)

wohali:
@rnewson ping. If you're not able to get to this in the next week, I'm removing this from the 2.3.0 milestone.

wohali:
This looks pretty gnarly to get back up to speed @rnewson , any chance for 3.x?

wohali:
@rnewson Any plans to get this in? I am preparing to mass-close old PRs that never got merged.

wohali:
```
15:25 <+rnewson> happy to ditch it.
```

PR #1273, opened 2018-04-09
Adopt fake_db to PSE changes

eiri:
## Overview

With db headers moved into engine's state, any `fake_db` call that is trying to setup sequences for tests (e.g. in mem3_shards) crashing with context setup failed.

It's not trivial to compose a proper `engine` field outside of couch app, so instead this fix makes `fake_db` to set engine transparently, unless it provided explicitly in a payload.

## Testing recommendations

Without this change running `make eunit apps=mem3 suites=mem3_shards` would crash with 

```
...
    *** context setup failed ***
**in function erlang:setelement/3
  called as setelement(undefined,{db,1,<<"dbs">>,undefined,
    {couch_bt_engine,undefined},
    nil,nil,undefined,undefined,
    {user_ctx,null,[],undefined},
    [],undefined,nil,nil,...},0)
in call from test_util:'-fake_db/1-fun-0-'/3 (src/test_util.erl, line 244)
in call from lists:foldl/3 (lists.erl, line 1261)
in call from mem3_shards:setup_changes/0 (src/mem3_shards.erl, line 738)
**error:badarg
  [done in 1.194 s]
  Failed: 0.  Skipped: 0.  Passed: 10.
```

This was not exposed by travis because in eunit crash in context setup doesn't count as a failure.

With this change the same command should pass successfully with `All 11 tests passed.`

## Checklist

### DIFF
--- a/src/couch/src/test_util.erl
+++ b/src/couch/src/test_util.erl
@@ -15,6 +15,7 @@
 -include_lib("couch/include/couch_eunit.hrl").
 -include("couch_db.hrl").
 -include("couch_db_int.hrl").
+-include("couch_bt_engine.hrl").
 
 -export([init_code_path/0]).
 -export([source_file/1, build_file/1]).
@@ -234,7 +235,8 @@ stop(#test_context{mocked = Mocked, started = Apps}) ->
     meck:unload(Mocked),
     stop_applications(Apps).
 
-fake_db(Fields) ->
+fake_db(Fields0) ->
+    {ok, Db, Fields} = maybe_set_engine(Fields0),
     Indexes = lists:zip(
             record_info(fields, db),
             lists:seq(2, record_info(size, db))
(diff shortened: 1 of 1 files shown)


src/couch/src/test_util.erl
```diff
@@ -234,15 +235,40 @@ stop(#test_context{mocked = Mocked, started = Apps}) ->
...
+            {ok, Header, Fields} = get_engine_header(Fields0),
+            Db = #db{engine = {couch_bt_engine, #st{header = Header}}},
+            {ok, Db, Fields}
+    end.
+
+get_engine_header(Fields) ->
+    Keys = [disk_version, update_seq, unused, id_tree_state,
+        seq_tree_state, local_tree_state, purge_seq, purged_docs,
+        security_ptr, revs_limit, uuid, epochs, compacted_seq],
+    {HeadFields, RestFields} = lists:foldl(fun({K, _} = E, {HAcc, RAcc}) ->
```
iilyak: It seems the following would work as well:
```
    {HeadFields, RestFields} = lists:partition(
        fun({K, _}) -> lists:member(K, Keys) end, Fields),
```

eiri:
@iilyak Amusingly, now when `mem3_shards` tests are actually running this exposed what appears to be a race in `mem3` test suite. I'm getting the suite failing on me about once in five runs with:
```
 mem3_shards:679: t_cache_insert_ignores_stale_update_and_kills_worker...ok
  Test mem3_shards changes listener
    mem3_shards:754: should_kill_changes_listener_on_shutdown...*failed*
in function mem3_shards:'-should_kill_changes_listener_on_shutdown/1-fun-1-'/1 (src/mem3_shards.erl, line 757)
in call from mem3_shards:'-should_kill_changes_listener_on_shutdown/1-fun-4-'/1 (src/mem3_shards.erl, line 757)
**error:{assertion_failed,[{module,mem3_shards},
                   {line,757},
                   {expression,"is_process_alive ( ChangesPid )"},
                   {expected,true},
                   {value,false}]}


    [done in 0.003 s]
  [done in 1.471 s]
```

Going to fix that before merging, don't want to make our pipeline more finicky, than it is already.

eiri:
...and just like this I can't reproduce it anymore. I'm going to open a separate issue for mem3_shards test and rebase and merge this PR.

PR #1299, opened 2018-04-29
Force use of SMP enabled BEAM VM, fixes #1296

### LINKED ISSUE #1296: bcrypt support breaks couchdb on single proc platforms
In analysis for #1293 I ran across this issue while testing in a VM that has only a single processor.

We enable bcrypt support in `test/javascript/tests/users_db_security.js` for testing purposes. However, on a single-processor machine, this test fails.

Looking at `dev/logs/node1.log` I see the following, then a crash:

```
enif_send: env==NULL on non-SMP VM
```

First hit on a web search was [this bug](https://github.com/basho/eleveldb/issues/120) which says that `enif_send` does not work on the non-SMP BEAM VM.

Confirmed on the erlang mailing list:

```
> On Tuesday, March 21, 2017 10:28 AM, Sverker Eriksson <[hidden email]> wrote:
> enif_send can be called from a non scheduler thread in an SMP enabled VM
> (beam.smp)
> but not in a non-SMP VM (beam). 
```

Our choices are:

1. Eliminate our single call to `enif_send` [here](https://github.com/apache/couchdb-erlang-bcrypt/blob/master/c_src/bcrypt_nif.c#L126) somehow
1. Put `-smp enable` in `vm.args` everywhere

/cc @janl @davisp 

As this breaks us on a lot of machines, I'm adding this to the 2.2.0 milestone to ensure we don't release without fixing this problem.

wohali:
enif_send (in bcrypt support) requires beam.smp. So, force its use everywhere.

Fixes #1296

### DIFF
--- a/rel/overlay/etc/vm.args
+++ b/rel/overlay/etc/vm.args
@@ -45,3 +45,6 @@
 
 # Comment this line out to enable the interactive Erlang shell on startup
 +Bd -noinput
+
+# Force use of the smp scheduler, fixes #1296
+-smp enable


wohali:
I have +1s from @chewbranca and @dch on IRC.

PR #1308, opened 2018-05-02
Bump fauxton to fix CI builds

wohali:
Pursuant to apache/couchdb-fauxton#1079 we need to bump our fauxton dependency to fix Fauxton builds. (Right now, Jenkins is completely broken because of a renamed dependency in npm-land preventing Fauxton from building.)


wohali:
As this is a Jenkins CI-blocking issue, I'll be merging this as soon as Travis CI goes green.

@janl Remember how we were talking about swapping Jenkins in for our CI builds? This sets an awkward precedent...if our build can break simply by someone renaming npm packages, we can end up in a state where a working master suddenly is no longer working, and the absolute ONLY PR that can land will be to fix that build breakage first. (That is, without manually switching off the CI gate on commits to master via an urgent Infra ticket...)

PR #1336, opened 2018-05-22
Mango: implicit coverage for _id and _rev during index selection

willholley:
## Overview

Mango requires that a JSON index can only be used to fulfil
a query if the "selector" field covers all of the fields the
index. For example, if an index contains ["a", "b"] but the
selector only requires field ["a"] to exist in the matching documents,
the index would not be valid for the query (because it only includes
documents containing both "a" and "b").

There is a special case here around built-in fields; _id and _rev
specifically, because they are guaranteed to exist in any matching
documents.

If a user declares an index ["a", "_id"], we can safely exclude "_id"
from the index coverage check, so a selector of {"a": "foo"} should be
able to use this index.

Prior to this commit, a user would have to alter the selector so that
it covered the "_id" field, e.g. {"a": "foo",
"_id": {"$exists": true}}. The commit removes the need to explicitly
cover _id or _rev fields in the query selector.

## Testing recommendations

Tests have been added to the Mango test suite. To manually test, create an index which includes `"_id"` and/or `"_rev"` and another field. Check that calling `_find` (or `_explain`) with a selector that does not include `"_id"` or `"_rev"` but does reference the other indexed fields will use the index.

## Checklist

### DIFF
--- a/src/mango/src/mango_idx_view.erl
+++ b/src/mango/src/mango_idx_view.erl
@@ -127,7 +127,7 @@ is_usable(Idx, Selector, SortFields) ->
     % _id and _rev are implicitly in every document so
     % we don't need to check the selector for these either
     RequiredFields2 = ordsets:subtract(
-        lists:usort(RequiredFields1), 
+        RequiredFields1,
         [<<"_id">>, <<"_rev">>]),
 
     mango_selector:has_required_fields(Selector, RequiredFields2)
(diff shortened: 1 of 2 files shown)


src/mango/src/mango_idx_view.erl
```diff
@@ -124,7 +124,13 @@ is_usable(Idx, Selector, SortFields) ->
     % we don't need to check the selector for these
     RequiredFields1 = ordsets:subtract(lists:usort(RequiredFields), lists:usort(SortFields)),
 
-    mango_selector:has_required_fields(Selector, RequiredFields1)
+    % _id and _rev are implicitly in every document so
+    % we don't need to check the selector for these either
+    RequiredFields2 = ordsets:subtract(
+        lists:usort(RequiredFields1),
```
eiri: `RequiredFields1` just came out of `ordsets:subtract/2` up a line, it's an ordset itself, so you don't need `usort` it.

willholley:
thanks @eiri - updated to remove the superfluous sort

PR #1337, opened 2018-05-23
 Enable discovery of cluster nodes through DNS SRV 

### LINKED ISSUE #1108: Query DNS to automatically discover cluster peers
We can streamline cluster setup by querying DNS for an SRV record containing the hostnames of other cluster members. Service records are defined in [RFC 2782](https://www.ietf.org/rfc/rfc2782.txt) for this purpose but this is not a strictly academic exercise. As a practical example Kubernetes will create SRV records for members of a StatefulSet. I created a sidecar container that uses those SRV records to automatically join up the members of a StatefulSet here:

https://github.com/kocolosk/couchdb-statefulset-assembler

I think it would be interesting to push that logic down into mem3 directly. I'm envisioning a system that can be configured like so

```
[mem3]
use_dns_service_discovery = true
service_record_name = _couchdb._tcp.example.net
```

If a user omits the `service_record_name` CouchDB would attempt to generate the appropriate record for the query by prepending `_couchdb._tcp.` to the DNS domain name portion of the host / VM / container FQDN.

kocolosk:
## Overview

This patch enables CouchDB nodes to check SRV records in DNS for cluster peers and automatically add any discovered peers to the nodes DB. This behavior is disabled by default and can be enabled as follows:

[mem3]
use_dns_service_discovery = true
service_record_name = _couchdb._tcp.example.net

If the `service_record_name` is omitted we attempt to generate the appropriate record for the query by prepending _couchdb._tcp. to the DNS domain name portion of the host / VM / container FQDN.

I wrote this code a few months back and it passed some smoke tests in Kube but I never opened a PR. Opening it now so I don't forget about it.

## Testing recommendations

1) Start one CouchDB node
2) Create an SRV record pointing to it
3) Start a second node with the [mem3] configuration above
4) Check to see if they automatically join into a cluster

Fixes #1108 

## Checklist


src/mem3/src/mem3_dns.erl
```diff
@@ -0,0 +1,109 @@
...
+        end, [], inet_dns:msg(DnsMsg, anlist)),
+        {ok, Hostnames};
+    Error ->
+        Error
+    end.
+
+construct_service_record() ->
+    case config:get("mem3", "service_record_name") of
+    undefined ->
+        [_ | Domain] = string:tokens(net_adm:localhost(), "."),
```
wohali: Tested this on Windows and it fails:

```erlang
Eshell V8.3  (abort with ^G)
1> [_ | Domain] = string:tokens(net_adm:localhost(), ".").
["DESKTOP-1AA2L3B"]
2> {ok, string:join(["_couchdb", "_tcp" | Domain], ".")}.
{ok,"_couchdb._tcp"}
```

I know this isn't your ideal deployment scenario, but someone's going to try it...and we need to have a fallback when it fails to get a proper domain name.

wohali: I get that you can just override it with `mem3:service_record_name`, but in case you don't, we should fail semi-gracefully.

kocolosk: Good catch, thanks.

wohali:
@kocolosk are we abandoning this work in favour of the current -helm approach?

kocolosk:
If someone wanted to bring it over the finish line I'd be happy to review and merge, but it's a low priority for me given the seedlist capability now adopted by the Helm chart. We can close this PR to clean things up.

PR #1369, opened 2018-06-01
[4/5] Clustered Purge: Add on_compact EPI hook

davisp:
This adds an on_compact hook to the couch_db_plugin. This is so that secondary indices can create their local purge sequence docs when a database is upgraded to use clustered purge.

This depends on PRs #1366, #1367, and #1368.

### DIFF
--- a/src/couch/src/couch_db_engine.erl
+++ b/src/couch/src/couch_db_engine.erl
@@ -627,7 +627,8 @@
     count_changes_since/2,
 
     start_compaction/1,
-    finish_compaction/2
+    finish_compaction/2,
+    trigger_on_compact/1
 ]).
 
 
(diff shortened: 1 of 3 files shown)


src/couch/src/couch_db_engine.erl
```diff
@@ -891,3 +892,31 @@ finish_compaction(Db, CompactInfo) ->
     end,
     ok = gen_server:call(couch_server, {db_updated, NewDb}, infinity),
     {ok, NewDb}.
+
+
+trigger_on_compact(DbName) ->
+    {ok, DDocs} = get_ddocs(DbName),
```
iilyak: Here we are calling high level API from low level API. Which is a sign of architectural problems. I am ok with that as long as @davisp is aware of the issue.

As an alternative solution we might consider passing ddoc enumerator function (one way or another) down the stack (which I am sure would complicate things).

Issue #1352, opened 2018-05-29
Restore /{db}/purge endpoint


wohali:
Work is ongoing to restore purge functionality in a clustered database: https://github.com/apache/couchdb/commit/5230cf01d847705171ab28d8219dbef1c2d6f2a6

This ticket is to track these PRs landing.

It'd be ideal if these can land in 2.2.0, but if the work is going to take a while, let's push out to 2.2.1 or 2.3.0.

/cc @davisp @jiangphcn

janl:
@wohali I’ve been closing issues that have corresponding PRs, feel free to open, if you want a unified place for multiple PRs

jiangphcn:
Added reference to https://github.com/apache/couchdb/pull/1370

PR #1370, opened 2018-06-01
[5/5] Clustered Purge Implementation

davisp:
## Overview

This PR implements clustered purge. Its big.

There are roughly five large chunks for clustered purge:

1. Single shard APIs
2. Updates to couch_mrview
3. Internal replication
4. Read-repair
5. Clustered API

At the shard level the major changes are to add two new indexes for purge requests. These indexes store the history of purge requests to a shard (rather than the old method of just storing the most recent purge). This allows for secondary indices and internal replication to manage their eventual consistency. This is the bulk of the PR in that its adding both the implementation and a whole bunch of tests for clustered purge.

One note on naming internally and in commit notes is that a "purge request" is what's sent in a single HTTP request to the purge end point. A "purge info" is a single `{DocId, Revs}` entry. The new indexes in the shard deal with purge infos as they lose any notion of bundling from which request they came in after they're stored.

The changes to couch_mrview are fairly straightforward after the single node API changes. The only thing that will be interesting in this commit is how we use a _local doc to track where the secondary index has processed in the purge sequence. This is necessary so that compaction knows when it can discard purge infos.

Internal replication is fairly straightforward. This update just ensures that we synchronize our purge infos between each shard so that we don't inadvertently undo a purge request when synchronizing shards.

Read-repair is somewhat tricky. The thing to note here is that when we open a document we track which revisions came from which nodes. Then if read-repair is required we pass that information along with the update request. This way a node can filter out requests from any node that is either a) not up to date with its synchronization or b) we have explicitly purged a revision that has not been sent to the node from whence the revision came back. That last bit of logic may sound kind of odd but the thing to remember is that we could completely remove a revision from a cluster and then re-introduce it later on. This means that we have to reject re-application of a revision for a closed period of time (rather than rejecting it forever).

The clustered API is rather straightforward for anyone familiar any of the existing fabric coordinator and HTTPD handler code.

## Testing recommendations

`make check`

Clustered purge comes with some fairly extensive testing though I'm very much open for suggestions on new tests.

This PR depends on:

#1366 
#1367 
#1368 
#1369
https://github.com/apache/couchdb/issues/1352

## Checklist

- [x] Code is written and works correctly (hopefully ;)

For docs this doesn't change the existing purge APIs but I will need to open a PR against the docs to write up some of the caveats and footgun warnings that we want to add for this.


davisp:
Also of note, I highly recommend moving through this PR a commit at a time as its quite big and the commits are specifically written to try and separate out all of the major changes to ease review.

src/couch/src/couch_bt_engine.erl
```diff
@@ -707,6 +814,23 @@ init_state(FilePath, Fd, Header0, Options) ->
             {compression, Compression}
         ]),
 
+    PurgeTreeState = case couch_bt_engine_header:purge_tree_state(Header) of
+        0 -> nil;
+        PTS -> PTS
+    end,
```
kocolosk: Is this header upgrade logic? Should that be moved to `upgrade_purge_info/2` for clarity?

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,145 @@ get_full_doc_info(Db, Id) ->
...
+    UUId :: binary(),
+    PurgeInfo :: {PurgeSeq, UUId, Id, [Rev]} | not_found,
+    PurgeSeq :: non_neg_integer(),
+    Id :: binary(),
+    Rev :: {non_neg_integer(), binary()}.
+get_purge_infos(Db, UUIDs) ->
+    couch_db_engine:load_purge_infos(Db, UUIDs).
+
+
+get_minimum_purge_seq(#db{} = Db) ->
```
kocolosk: Reading through commits 06 and 07 taught me that this little function is incredibly important and subtle. Every consumer of the database who wants to honor purge requests needs to create one of these local documents and keep it updated.

jiangphcn:
Thanks @kocolosk for your review. I tried to address most of your comments on commits 01, 02 and 03 in new commit 7e0dabb. Will work on other comments. @davisp hey Paul, feel free to add your response :-)

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,145 @@ get_full_doc_info(Db, Id) ->
...
+                        couch_log:error(Fmt, [DocId, ClientSeq]),
+                        {ok, erlang:min(OldestPurgeSeq, SeqAcc)}
+                end;
+            _ ->
+                {stop, SeqAcc}
+        end
+    end,
+    InitMinSeq = PurgeSeq - PurgeInfosLimit,
+    Opts = [
+        {start_key, list_to_binary(?LOCAL_DOC_PREFIX ++ "purge-")}
```
jiangphcn: Good suggestion. I changed code to use `?LOCAL_DOC_PREFIX` in `FoldFun` as well.

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,145 @@ get_full_doc_info(Db, Id) ->
...
+    catch error:badarg ->
+        Fmt1 = "Missing index module '~p' for purge checkpoint '~s'",
+        couch_log:error(Fmt1, [M0, DocId]),
+        throw(failed)
+    end,
+
+    F0 = couch_util:get_value(<<"verify_function">>, Props),
+    try
+        F = binary_to_existing_atom(F0, latin1),
+        fun M:F/2
```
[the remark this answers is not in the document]
jiangphcn: yes, I used `erlang:function_exported/3` before calling `fun M:F/2`.

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,145 @@ get_full_doc_info(Db, Id) ->
...
+    OldestPurgeSeq = couch_db_engine:get_oldest_purge_seq(Db),
+    PurgeInfosLimit = couch_db_engine:get_purge_infos_limit(Db),
+
+    FoldFun = fun(#doc{id = DocId, body = {Props}}, SeqAcc) ->
+        case DocId of
+            <<"_local/purge-", _/binary>> ->
+                ClientSeq = couch_util:get_value(<<"purge_seq">>, Props),
+                case ClientSeq of
+                    CS when is_integer(CS), CS >= PurgeSeq - PurgeInfosLimit ->
+                        {ok, SeqAcc};
```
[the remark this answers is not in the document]
jiangphcn: Yes, the first clause is used to directly return `{ok, SeqAcc}` without calling `purge_client_exists`. I am not quite sure why you mentioned that the `purge_client_exists` check is not necessary when `CS < PurgeSeq - PurgeInfosLimit` occurs.  Under `CS < PurgeSeq - PurgeInfosLimit` situation, we have to call `purge_client_exists` to decide which client seq needs to be used.

src/couch/src/couch_httpd_db.erl
```diff
@@ -376,17 +376,22 @@ db_req(#httpd{path_parts=[_,<<"_bulk_docs">>]}=Req, _Db) ->
...
-    {ok, PurgeSeq, PurgedIdsRevs} ->
-        PurgedIdsRevs2 = [{Id, couch_doc:revs_to_strs(Revs)} || {Id, Revs} <- PurgedIdsRevs],
-        send_json(Req, 200, {[{<<"purge_seq">>, PurgeSeq}, {<<"purged">>, {PurgedIdsRevs2}}]});
-    Error ->
-        throw(Error)
-    end;
+    {ok, Replies} = couch_db:purge_docs(Db, PurgeReqs),
+
+    Results = lists:zipwith(fun({{Id, _}, Reply}) ->
+        {Id, couch_doc:revs_to_strs(Reply)}
```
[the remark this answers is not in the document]
jiangphcn: Good catch, and fixed via commit 276d4ed

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -41,6 +43,27 @@
...
+        undefined ->
+            Reason = binary_to_list(Key) ++ " must exist in Options.",
+            throw({bad_request, Reason});
+        Value -> Value
+    end.
+
+
+get_signature_from_filename(FileName) ->
+    FilePathList = filename:split(FileName),
+    [PureFN] = lists:nthtail(length(FilePathList) - 1, FilePathList),
```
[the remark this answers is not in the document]
jiangphcn: Good suggestion, and changed in commit 276d4ed

src/couch_mrview/src/couch_mrview_cleanup.erl
```diff
@@ -41,7 +41,23 @@ run(Db) ->
 
     lists:foreach(fun(FN) ->
         couch_log:debug("Deleting stale view file: ~s", [FN]),
-        couch_file:delete(RootDir, FN, [sync])
+        couch_file:delete(RootDir, FN, [sync]),
+        Sig = couch_mrview_util:get_signature_from_filename(FN),
+        if length(Sig) < 16 -> ok; true ->
+            case re:run(Sig,"^[a-fA-F0-9]+$",[{capture, none}]) of
```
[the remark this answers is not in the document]
jiangphcn: Just double check the signature extracted from view file path is valid, i.e. `^[a-fA-F0-9]+$`.

src/couch/src/couch_db.erl
jiangphcn: Yes, this function is quite important. Not only for mrview index, but also for search index and geospatial index, etc.

src/couch_index/src/couch_index_updater.erl
```diff
@@ -208,12 +211,27 @@ update(Idx, Mod, IdxState) ->
...
-            {ok, IdxState};
-        DbPurgeSeq == IdxPurgeSeq + 1 ->
-            {ok, PurgedIdRevs} = couch_db:get_last_purged(Db),
-            Mod:purge(Db, DbPurgeSeq, PurgedIdRevs, IdxState);
-        true ->
-            reset
+    if IdxPurgeSeq == DbPurgeSeq -> {ok, IdxState}; true ->
+        FoldFun = fun({PurgeSeq, _UUId, Id, Revs}, Acc) ->
+            Mod:purge(Db, PurgeSeq, [{Id, Revs}], Acc)
+        end,
```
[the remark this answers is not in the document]
jiangphcn: Understand that `couch_mrview_index:purge/4` or `couch_mrview_updater:purge/4` can deal with PurgedIdRevs (i.e. multiple Ids with Revs) in one call. However, when calling `couch_db:fold_purge_infos/5`, only one Id and Revs are available for one PurgeSeq in the callback.

src/couch_mrview/src/couch_mrview_cleanup.erl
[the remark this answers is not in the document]
davisp: The idea is that we don't want to blow away any non-view files that someone may have stuck in that directory. We should probably move all validation to the `couch_mrview_util:get_signature_from_filename/1` function though.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -41,6 +43,27 @@
...
+    PureFNExt = filename:extension(PureFN),
```
davisp: We should also remove this and hard code the extension as ".view". And then also add a check that the filename is 16 hex characters. Probably doesn't require a regexp either as we can just cycle over the binary and assert fairly easily that a byte is acceptable.

src/couch_index/src/couch_index_updater.erl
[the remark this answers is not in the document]
davisp: @jiangphcn We could always batch up in an accumulator and only invoke Mod:purge/4 occasionally rather than for individual rows.

@kocolosk The only reason its not batching was mulling over simple vs complex for questionable benefits. Given that purge is basically a "Break Glass in Case of Emergency" feature and shouldn't be part of an application's core workflow it seemed not overly useful to concern ourselves with performance.

src/mem3/src/mem3_rep.erl
```diff
@@ -169,20 +200,131 @@ find_source_seq_int(#doc{body={Props}}, SrcNode0, TgtNode0, TgtUUID, TgtSeq) ->
...
+        batch_size = Count,
+        seq = UpdateSeq,
+        target = Target
+    } = Acc0,
+    #shard{
+        node = TgtNode,
+        name = TgtDbName
+    } = Target,
+
+    with_src_db(Acc0, fun(Db) ->
```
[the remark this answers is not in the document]
davisp: Its definitely subtle and I wasn't a fan either but it was the least bad I could think of at the time. For the couch_server, the `#acc.db` instance guarantees that the database is held open in the `couch_db` ets cache so the with_src_db call is guaranteed to be cheap from that respect.

The second bit about not re-using `#acc.db` was that it makes this bit of logic harder/impossible:

https://github.com/apache/couchdb/blob/d1b73e75891753aef6366bafbfaa1cbcbcbb4bb5/src/mem3/src/mem3_rep.erl#L81-L87

Specifically, that after clause can't reference the return of `repl/1` because its not guaranteed to have returned. Thus threading db instances through means you have to not close the first instance the first time you re-open or else you'll break that outside close. That seemed a lot more complicated than just using the with_src_db approach.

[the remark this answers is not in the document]
davisp: Oooh fancy! GH inlined my code reference.

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,145 @@ get_full_doc_info(Db, Id) ->
...
+                throw({badreq, Reason})
+            end
+        end, get_purge_infos(Db, UUIDs))
+    end,
+    increment_stat(Db, [couchdb, database_purges]),
+    gen_server:call(Pid, {purge_docs, UUIDsIdsRevs, Options}).
+
+-spec get_purge_infos(#db{}, [UUId]) -> [PurgeInfo] when
+    UUId :: binary(),
+    PurgeInfo :: {PurgeSeq, UUId, Id, [Rev]} | not_found,
```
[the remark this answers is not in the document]
davisp: Yeah, I agree that's a bit weird but it was written to be consistent with the existing APIs that do the same thing.

src/couch_index/src/couch_index_updater.erl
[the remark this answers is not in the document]
davisp: To clarify, not concerning ourselves with performance is more of a "No premature optimization" comment. We did of course make think a little so we didn't walk ourselves into some sort of `O(M^N)` exponential blow up thing. 😀

src/couch_mrview/src/couch_mrview_cleanup.erl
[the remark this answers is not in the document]
kocolosk: @davisp this patch is still deleting every file; the regex only guards against trying to remove an associated local doc if the file was not a view file.

src/couch_index/src/couch_index_updater.erl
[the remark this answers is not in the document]
kocolosk: Yeah, sure, I'm fine either way. I noticed that all the infrastructure seemed to be in place for it, and you were already doing some batching in internal replication (following the pattern that already existed).

src/mem3/src/mem3_rep.erl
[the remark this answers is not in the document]
kocolosk: OK got it.

src/fabric/src/fabric_doc_purge.erl
```diff
@@ -0,0 +1,572 @@
...
+        true -> error
+    end.
+
+
+has_quorum(Resps, Count, W) ->
+    OkResps = [R || {ok, _} = R <- Resps],
+    OkCounts = lists:foldl(fun(R, Acc) ->
+        dict:update_counter(R, 1, Acc)
+    end, dict:new(), OkResps),
+    MaxOk = lists:max([0 | element(2, lists:unzip(dict:to_list(OkCounts)))]),
```
kocolosk: For a dataset this small an `orddict` is faster and more lightweight than a `dict`, and you can also skip the `to_list` conversion (it's a noop). I think you're creating O(N^2) of these dictionaries for a purge request with N entries in it so this efficiency may actually matter

kocolosk:
Very nice work all.

Historically one could think about external consumers of `_changes` and internal consumers of the sequence index as being on equal footing, and there's something appealing about that. The latter group has some efficiency advantages but fundamentally an external process could go and do everything that the internal consumers do using the node-local HTTP interface. The lack of HTTP access to the purge index breaks that. An external indexer, for example, has no way of knowing revisions have been purged from the main database short of checking the individual shard's purge sequence and rebuilding the entire index if it changes, whereas the internal consumers can be smarter.

All that said, I wouldn't change anything about this PR, as it contains everything needed to surface an external API to the purge structures if we should conclude there's a need for it at a later date.

I'm +1 pending resolution of my (minor) review comments.

src/fabric/src/fabric_rpc.erl
```diff
@@ -298,6 +314,104 @@ with_db(DbName, Options, {M,F,A}) ->
...
+                    false ->
+                        {ok, Acc}
+                end;
+            _ ->
+                % We've processed all _local mem3 purge docs
+                {stop, Acc}
+        end
+    end,
+    InitAcc = [{list_to_binary(atom_to_list(Node)), 0} || Node <- Nodes],
+    Opts = [{start_key, <<?LOCAL_DOC_PREFIX, "/purge-mem3-">>}],
```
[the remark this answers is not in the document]
jiangphcn: Good catch, and fixed in 404ce54

src/fabric/src/fabric_doc_purge.erl
jiangphcn: addressed in commit 9038557

src/couch_mrview/src/couch_mrview_util.erl
jiangphcn: addressed in commit 9038557

src/couch_mrview/src/couch_mrview_cleanup.erl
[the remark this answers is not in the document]
jiangphcn: Hey @kocolosk I added check in commit 9038557 and do nothing if the file is not ending with `".view"`.

jiangphcn:
Thanks again for your review @kocolosk 
Hey @kocolosk @davisp, I tried to address comments related to code change in commit 9038557 and other previous commits. Any additional comments are highly welcome.

src/fabric/src/fabric_rpc.erl
```diff
@@ -21,6 +21,7 @@
     delete_shard_db_doc/2]).
 -export([get_all_security/2, open_shard/2]).
 -export([compact/1, compact/2]).
+-export([get_purge_seq/2, purge_docs/3, set_purge_infos_limit/3]).
```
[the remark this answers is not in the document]
davisp: Blargh. They totes do. Must've missed it when re-ordering commits to make reviewing easier.

src/couch/src/couch_db_engine.erl
```diff
@@ -1015,7 +1015,7 @@ get_ddocs(<<"shards/", _/binary>> = DbName) ->
 get_ddocs(DbName) ->
     couch_util:with_db(DbName, fun(Db) ->
         FoldFun = fun(FDI, Acc) ->
-            Doc = couch_db:open_doc_int(Db, FDI, []),
```
jiangphcn: Small nit here. Should be `{ok, Doc}` instead of `Doc`. Fixed in commit 2132681

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,153 @@ get_full_doc_info(Db, Id) ->
...
+    LagWindow = config:get_integer(
+            "purge", "index_lag_warn_seconds", 86400), % Default 24 hours
+
+    {Mega, Secs, _} = os:timestamp(),
+    NowSecs = Mega * 1000000 + Secs,
+    LagThreshold = NowSecs - LagWindow,
+
+    try
+        CheckFun = get_purge_client_fun(DocId, Props),
+        Exists = CheckFun(DbName, DocId, Props),
```
iilyak: I don't quite understand it. We are calling a function with arity 3 here. 
- `get_purge_client_fun/2` returns the function with arity 2 [see here](https://github.com/apache/couchdb/pull/1370/files#diff-66129b4f374c3cde3fc9a49fa8a9452eR509).
- mem3_rep:verify_purge_checkpoint has arity 3
- couch_mrview_index:verify_index_exists has arity 1

I am sure I missing something.

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,153 @@ get_full_doc_info(Db, Id) ->
...
+        Exists
+    catch _:_ ->
+        % If we fail to check for a client we have to assume that
+        % it exists.
+        true
+    end.
+
+
+get_purge_client_fun(DocId, Props) ->
+    M0 = couch_util:get_value(<<"verify_module">>, Props),
```
iilyak: What are implications on security? Users can write `_local/` documents. That means that they can craft the doc which has `verify_module` and `verify_function` keys. Which means they would be able to invoke any function which matches the expected signature. What if there is some dangerous function in otp or couchdb. The function which receives the proplist and modifies beam behavior would do. I didn't figured out the exact signature yet. But it seems the attack surface is big.

src/couch_mrview/src/couch_mrview_index.erl
```diff
@@ -207,3 +217,104 @@ index_file_exists(State) ->
     } = State,
     IndexFName = couch_mrview_util:index_file(DbName, Sig),
     filelib:is_file(IndexFName).
+
+
+verify_index_exists(Props) ->
+    ShardDbName = couch_mrview_util:get_value_from_options(
```
iilyak: Users can write to `_local/` docs. What prevents malicious user from specifying someone else `dbname`? What would happen if they would do it?

src/mem3/src/mem3_rep.erl
```diff
@@ -118,6 +121,34 @@ make_local_id(SourceThing, TargetThing, Filter) ->
     <<"_local/shard-sync-", S/binary, "-", T/binary, F/binary>>.
 
 
+make_purge_id(SourceUUID, TargetUUID) ->
+    <<"_local/purge-mem3-", SourceUUID/binary, "-", TargetUUID/binary>>.
```
iilyak: The purge docid for mrview has version of purge implementation in it [see here](https://github.com/apache/couchdb/pull/1370/commits/c67815e8272ef2e90aa6a2a07522a9590bbbc49a#diff-2bd31f00dcfba780c6c65bc69c256854R46). Should we embed version here as well?

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,153 @@ get_full_doc_info(Db, Id) ->
...
+            if is_integer(Updated) and Updated > LagThreshold -> ok; true ->
+                Diff = NowSecs - Updated,
+                Fmt = "Purge checkpoint '~s' not updated in ~p seconds",
+                couch_log:error(Fmt, [DocId, Diff])
+            end
+        end,
+        Exists
+    catch _:_ ->
+        % If we fail to check for a client we have to assume that
+        % it exists.
```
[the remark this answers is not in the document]
jiangphcn: Added an error log in 94fcf71

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,153 @@ get_full_doc_info(Db, Id) ->
...
+                    CS when is_integer(CS) ->
+                        case purge_client_exists(Db, DocId, Props) of
+                            true -> {ok, erlang:min(CS, SeqAcc)};
+                            false -> {ok, SeqAcc}
+                        end;
+                    _ ->
+                        % If there's a broken doc we have to keep every
+                        % purge info until the doc is fixed or removed.
+                        Fmt = "Invalid purge doc '~s' with purge_seq '~w'",
+                        couch_log:error(Fmt, [DocId, ClientSeq]),
```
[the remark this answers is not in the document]
jiangphcn: Added db info in 94fcf71

src/mem3/src/mem3_rep.erl
jiangphcn: I added versioning in 94fcf71 for mem3/replication case.

src/couch/src/couch_db.erl
jiangphcn: Thanks @iilyak for bringing this up. I can understand that users can create `_local/` document on their own, and specify `verify_module` and `verify_function`. The specified `M:F/1` will be called during database compaction. I wonder how the customized logic for `M:F/1` is injected into couchdb. If users only specifies module name and function name, and failed to tamper with called logic, we are kind of safe in some extent. Of course, we can see comments from @davisp

src/couch_mrview/src/couch_mrview_index.erl
jiangphcn: Can malicious user have right to work on other's database?

src/couch/src/couch_db.erl
iilyak: I don't think we are safe. Any function in OTP which doesn't validate it's input can be used to mount an attack. For example let's assume any of the following functions don't handle the case when they receive proplist:
- `zip:extract/1` --- zip bomb attack
- `os:cmd/1` --- execute any command on the host
- `erlang:throw/1` --- crash the current process with attacker provided reason.
- `erlang:halt/1` -- stop node

jiangphcn: I agreed that there is still potential risk even if only module name and function name are specified by users. Perhaps there are mature methodology to solve such issue. For me, I thought to use blacklist to check whether specified M:F/1 is attack-able. However, this may not be enumerable. As you suggested, we may consider whitelist or something like this.

davisp: Good call Ilya. Should be easy enough to add a module whitelist via EPI no? Then its just a matter of moving our validation logic to an otherwise empty module so that its trivial to verify we're not calling anything possibly dangerous.

davisp: Ooh, or alternatively, just make an EPI hook for the verification rather than having each purge handler from specifying which function in the doc itself. That seems even better than the whitelist.

src/couch/src/couch_db_updater.erl
```diff
@@ -692,6 +641,87 @@ update_local_doc_revs(Docs) ->
...
+    couch_event:notify(Db2#db.name, updated),
+    {ok, Db2, Replies}.
+
+
+apply_purge_reqs([], IdFDIs, _USeq, Replies) ->
+    {IdFDIs, lists:reverse(Replies)};
+
+apply_purge_reqs([Req | RestReqs], IdFDIs, USeq, Replies) ->
+    {_UUID, DocId, Revs} = Req,
+    {value, {_, FDI0}, RestIdFDIs} = lists:keytake(DocId, 1, IdFDIs),
```
iilyak: Is it a possibility that `PurgeReqs` has multiple entries with the same `DocId`? If so we would crash here when we encounter the `DocId` we already processed.

iilyak: Never mind we add taken element back [here](https://github.com/apache/couchdb/pull/1370/commits/edd6e7162004e3c29453b938af88ea75d54157ae#diff-c0abd721a4433ae0399b0c2cbdea8d18R720)

src/couch/src/couch_db_updater.erl
```diff
@@ -692,6 +641,87 @@ update_local_doc_revs(Docs) ->
...
+purge_docs(Db, PurgeReqs) ->
+    Ids = lists:usort(lists:map(fun({_UUID, Id, _Revs}) -> Id end, PurgeReqs)),
+    FDIs = couch_db_engine:open_docs(Db, Ids),
+    USeq = couch_db_engine:get_update_seq(Db),
+
+    IdFDIs = lists:zip(Ids, FDIs),
+    {NewIdFDIs, Replies} = apply_purge_reqs(PurgeReqs, IdFDIs, USeq, []),
+
+    Pairs = lists:flatmap(fun({DocId, OldFDI}) ->
+        {DocId, NewFDI} = lists:keyfind(DocId, 1, NewIdFDIs),
```
iilyak: It seems like this line would be executed very often. Every call is O(n). We would be traversing the list over O(2n+2*n^2) times (purge_docs and apply_purge_reqs combined) (I am sure my calculation is not precise). 
Did you consider using `gb_sets` to hold `IdFDIs`?

src/couch/src/couch_db.erl
kocolosk: Ah, good catch Ilya. Sorry I missed that!

src/couch/src/couch_db_updater.erl
davisp: Hopefully its not executed too often or that suggests we have a user that didn't understand the messaging that this feature is a last resort API and not intended to be part of normal application workflow.

Also HTTP should be limiting N here to less than 100 which means the linear searching should be cheaper than building a dict or some other data structure for lookups. There's a similar case in the implementation of `couch_util:reorder_results/1` for reference on the size issue.

src/couch/test/couch_bt_engine_upgrade_tests.erl
```diff
@@ -0,0 +1,177 @@
...
+        NewDbFilePath
+    end, DbFileNames),
+    {Ctx, NewPaths}.
+
+
+teardown({Ctx, Paths}) ->
+    lists:foreach(fun(Path) ->
+        file:delete(Path)
+    end, Paths),
+    test_util:stop_couch(Ctx).
```
iilyak: should we swap the order? 
1. stop_couch
2. remove files

src/couch/test/couch_bt_engine_upgrade_tests.erl
```diff
@@ -0,0 +1,177 @@
...
+
+
+upgrade_test_() ->
```
iilyak: Do we have tests for case when db is in old format without `purge_tree` and `purge_seq_tree`.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -41,6 +43,27 @@
...
+get_local_purge_doc_id(Sig) ->
+    Version = "v" ++ config:get("purge", "version", "1") ++ "-",
+    ?l2b(?LOCAL_DOC_PREFIX ++ "purge-" ++ Version ++ "mrview-" ++ Sig).
+
+
+get_value_from_options(Key, Options) ->
+    case couch_util:get_value(Key, Options) of
+        undefined ->
+            Reason = binary_to_list(Key) ++ " must exist in Options.",
+            throw({bad_request, Reason});
```
iilyak: If you search code base we usually return `{bad_request, binary()}`. This can be done as `Reason = <<"'", Key/binary, "' must exists in options.">>`

src/couch_mrview/test/couch_mrview_purge_docs_fabric_tests.erl
```diff
@@ -0,0 +1,272 @@
...
+    ok = fabric:create_db(DbName, [?ADMIN_CTX, {q, 1}]),
+    meck:new(couch_mrview_index, [passthrough]),
+    meck:expect(couch_mrview_index, ensure_local_purge_docs, fun(A, B) ->
+        meck:passthrough([A, B])
+    end),
+    DbName.
+
+
+teardown(DbName) ->
+    meck:unload(couch_mrview_index),
```
iilyak: `#3` from https://github.com/cloudant/db/wiki/PR-review-checklist#meck

src/couch/test/couch_bt_engine_upgrade_tests.erl
jiangphcn: See `t_upgrade_without_purge_req/0` test case with `db_without_purge_req.couch` database

src/mem3/src/mem3_rep.erl
jiangphcn: Based on discussion with @davisp, removing all versioning for local purge document and change is reflected in 407b1dc for mem3 and mrview.

src/couch/test/couch_bt_engine_upgrade_tests.erl
jiangphcn: swapped in 407b1dc

src/couch_mrview/src/couch_mrview_util.erl
jiangphcn: changed in 407b1dc

src/couch_mrview/test/couch_mrview_purge_docs_fabric_tests.erl
jiangphcn: use `meck:unload/0` in 407b1dc

src/couch/src/couch_db.erl
jiangphcn: I tried to solve this problem using commit 407b1dc. More test will be added.

src/mem3/src/mem3_rep.erl
```diff
@@ -169,20 +200,131 @@ find_source_seq_int(#doc{body={Props}}, SrcNode0, TgtNode0, TgtUUID, TgtSeq) ->
...
+    with_src_db(Acc0, fun(Db) ->
+        SrcUUID = couch_db:get_uuid(Db),
+        {LocalPurgeId, Infos, ThroughSeq, Remaining} =
+                mem3_rpc:load_purge_infos(TgtNode, TgtDbName, SrcUUID, Count),
+
+        if Infos == [] -> ok; true ->
+            {ok, _} = couch_db:purge_docs(Db, Infos, [replicated_edits]),
+            Body = purge_cp_body(Acc0, ThroughSeq),
+            mem3_rpc:save_purge_checkpoint(
+                    TgtNode, TgtDbName, LocalPurgeId, Body)
```
iilyak: nitpick: indentation looks to be more than 4 characters.

davisp: Continuation lines are indented twice to not confuse the reader on structure.

src/fabric/src/fabric_rpc.erl
```diff
@@ -236,14 +240,26 @@ get_missing_revs(DbName, IdRevsList, Options) ->
...
-    true ->
-        X = replicated_changes;
-    _ ->
-        X = interactive_edit
+    {Docs1, Type} = case couch_util:get_value(read_repair, Options) of
+        NodeRevs when is_list(NodeRevs) ->
+            Filtered = read_repair_filter(DbName, Docs0, NodeRevs, Options),
+            {Filtered, replicated_changes};
+        undefined ->
+            X = case proplists:get_value(replicated_changes, Options) of
```
iilyak: It looks weird that in the same function two very similar functions are used:
- `couch_util:get_value/2`
- `proplists:get_value/2`

src/fabric/src/fabric_rpc.erl
```diff
@@ -298,6 +314,104 @@ with_db(DbName, Options, {M,F,A}) ->
...
+% request before the remote shard copy. So to check that we
+% need to look at the purge infos that we have locally but
+% have not yet sent to the remote copy.
+%
+% NodeRevs is a list of the {node(), [rev()]} tuples passed
+% as the read_repair option to update_docs.
+read_repair_filter(Db, Docs, NodeRevs) ->
+    [#doc{id = DocId} | _] = Docs,
+    Nodes = lists:usort([Node || {Node, _} <- NodeRevs, Node /= node()]),
+    NodeSeqs = get_node_seqs(Db, Nodes),
```
iilyak: I would use `gb_sets` here since the number of elements returned by `couch_db:fold_local_docs` inside `get_node_seq` can be big. It seems we need to get `NodeSeq` for given Node quite a few times here. We do it in the loop when calculate `RecentNodeRevs` and in another loop when we calculate `AllowableRevs`.

iilyak: Feel free to ignore if you are sure that the number of elements in `NodeSeqs` list is always low (less than 1000).

davisp: Nodes from line 349 is capped at 2 (or more specifically N-1) so performance should be acceptable for any reasonable N.

davisp: Also, the reason for that is that this is a read repair is only ever called from fabric_doc_open and fabric_doc_open_revs and we're just sending the revisions that we get from each node in the response. And worst case would be every node in the cluster if we ever improved that algorithm to do parallel opens.

src/fabric/src/fabric_rpc.erl
```diff
@@ -298,6 +314,104 @@ with_db(DbName, Options, {M,F,A}) ->
...
+            end
+        end,
+        {ok, FiltRevs} = couch_db:fold_purge_infos(Db, StartSeq, FoldFun, Revs),
+        lists:usort(FiltRevs ++ RevAcc)
+    end, [], RecentNodeRevs),
+
+    % Finally, filter the doc updates to only include revisions
+    % that have not been purged locally.
+    DocFiltFun = fun(#doc{revs = {Pos, [Rev | _]}}) ->
+        lists:member({Pos, Rev}, AllowableRevs)
```
iilyak: Possible can use `gb_sets` here.

src/fabric/src/fabric_rpc.erl
```diff
@@ -298,6 +314,104 @@ with_db(DbName, Options, {M,F,A}) ->
...
+
+get_node_seqs(Db, Nodes) ->
+    % Gather the list of {Node, PurgeSeq} pairs for all nodes
+    % that are present in our read repair group
+    FoldFun = fun(#doc{id = Id, body = {Props}}, Acc) ->
+        case Id of
+            <<?LOCAL_DOC_PREFIX, "purge-mem3-", _/binary>> ->
+                TgtNode = couch_util:get_value(<<"target_node">>, Props),
+                PurgeSeq = couch_util:get_value(<<"purge_seq">>, Props),
+                case lists:keyfind(TgtNode, 1, Acc) of
```
iilyak: If you would decide against using `gb_trees`, this case can be written using `lists:keytake/3` and `[NewEntry | Acc]`.

iilyak: Never mind. As Paul mentioned earlier the Acc size is `N - 1` which is a small number.

src/couch/src/couch_db.erl
```diff
@@ -369,8 +379,145 @@ get_full_doc_info(Db, Id) ->
...
+    purge_docs(Db, IdRevs, []).
+
+-spec purge_docs(#db{}, [{UUId, Id, [Rev]}], [PurgeOption]) ->
+    {ok, [Reply]} when
+    UUId :: binary(),
+    Id :: binary(),
+    Rev :: {non_neg_integer(), binary()},
+    PurgeOption :: interactive_edit | replicated_changes,
+    Reply :: {ok, []} | {ok, [Rev]}.
+purge_docs(#db{main_pid = Pid} = Db, UUIDsIdsRevs, Options) ->
```
iilyak: I think this wouldn't work for PSE. We suppose to not have public access to `#db{}` record [see this commit](https://github.com/apache/couchdb/pull/495/commits/7b329b46ee7dccc549580e5f519023964771972c)

src/mem3/src/mem3_rep.erl
```diff
@@ -169,20 +200,131 @@ find_source_seq_int(#doc{body={Props}}, SrcNode0, TgtNode0, TgtUUID, TgtSeq) ->
...
+        end,
+
+        if Remaining =< 0 -> ok; true ->
+            OldestPurgeSeq = couch_db:get_oldest_purge_seq(Db),
+            PurgesToPush = couch_db:get_purge_seq(Db) - OldestPurgeSeq,
```
iilyak: I don't explanation yet, but dialyzer complains here with:
```
mem3_rep:pull_purges/1[221-254]:The call erlang:'-'({'ok',_},OldestPurgeSeq::{'ok',_}) will never return since it differs in the 1st and 2nd argument from the success typing arguments: (number(),number())
506c712
```

Which means in some (or all) cases the result of `couch_db:get_purge_seq` and `couch_db:get_oldest_purge_seq` is `{ok, integer()}` while we expect it to be just `integer()` (`non_neg_integer()` to be more precise).

src/couch/src/couch_db.erl
davisp: couch_db and couch_db_updater are the only two modules that know about the record definition so they can introspect it.

src/mem3/src/mem3_rep.erl
davisp: That versioning thing was something that was planned for a rollback but after more development it became clear that we can't really do a parity rollback so there's no reason to try and code it. Due to how we want to transform the current purge data structure into a tree there's no obvious way to undo a tree into the existing structure without a whole bunch of code that would be exceedingly rarely executed. Given that complexity tradeoff it seems a lot more useful to just test the upgrade thoroughly.

src/couch_index/src/couch_index_plugin_couch_db.erl
```diff
@@ -13,9 +13,19 @@
 -module(couch_index_plugin_couch_db).
 
 -export([
+    get_purge_client_fun/1,
     on_compact/2
 ]).
 
 
+get_purge_client_fun(Type) ->
```
iilyak:
```
get_purge_client_fun(<<"mrview">>) ->
    {decided, fun couch_mrview_index:verify_index_exists/1};
get_purge_client_fun(_) ->
    no_decission
end.
```

Same thing. Up to you, whatever appeal to you better.

src/chttpd/src/chttpd_db.erl
```diff
@@ -496,24 +496,20 @@ db_req(#httpd{path_parts=[_, <<"_bulk_get">>]}=Req, _Db) ->
...
-        PurgedIdsRevs2 = [{Id, couch_doc:revs_to_strs(Revs)} || {Id, Revs}
-            <- PurgedIdsRevs],
-        send_json(Req, 200, {[
-            {<<"purge_seq">>, PurgeSeq},
-            {<<"purged">>, {PurgedIdsRevs2}}
-        ]});
-    Error ->
-        throw(Error)
-    end;
+    {ok, Results} = fabric:purge_docs(Db, IdsRevs2, Options),
```
iilyak: according to the spec of `fabric:purge_docs` it can return `{accepted, Results}` are we intended to crash here in this case?

src/chttpd/src/chttpd_db.erl
```diff
@@ -623,6 +619,20 @@ db_req(#httpd{method='GET',path_parts=[_,<<"_revs_limit">>]}=Req, Db) ->
 db_req(#httpd{path_parts=[_,<<"_revs_limit">>]}=Req, _Db) ->
     send_method_not_allowed(Req, "PUT,GET");
 
+db_req(#httpd{method='PUT',path_parts=[_,<<"_purged_infos_limit">>]}=Req, Db) ->
+    Limit = chttpd:json_body(Req),
+    Options = [{user_ctx, Req#httpd.user_ctx}],
+    case chttpd:json_body(Req) of
```
iilyak: Why do we call `chttpd:json_body(Req)` twice (see two lines above)?

src/chttpd/src/chttpd_db.erl
iilyak: Never mind. The signature is 
```
{ok, [{Health, [revision()]}] | {error, any()}} when
    Health :: ok | accepted.
```

src/couch_pse_tests/src/cpse_util.erl
```diff
@@ -116,6 +119,123 @@ shutdown_db(Db) ->
...
+    after
+        couch_db:close(Db)
+    end.
+
+
+uuid() ->
+    couch_uuids:random().
+
+
+assert_db_props(DbName, Props) when is_binary(DbName) ->
```
iilyak: I think this function should receive `?LINE` argument to pinpoint to the problem better.

iilyak:
Very nice work. 💯 for a thorough test suite.

jiangphcn:
I just added commit https://github.com/apache/couchdb/pull/1370/commits/a5c1d1330c8051a9912871049ead60869745c4db to address issue of exposing `verify_module` and `verify_function` for unexpected execution. Currently `couch_epi:any/5` is used to listen whether there is client reporting a valid index. 

In addition, it looks that there is problem with `couch_db:get_design_docs/1` and I stripped the suffix (the part after dot) before calling `fabric:design_docs/1`.

src/couch_pse_tests/src/cpse_util.erl
jiangphcn: There is no `erlang:error` in subsequent call.

src/chttpd/src/chttpd_db.erl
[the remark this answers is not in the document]
davisp: The idea was that _bulk_docs is missing the limit and its only historical issues that prevent us from adding one. For this case we'll definitely want to do something around limiting the number of revisions and/or ids that can be passed in a single purge request.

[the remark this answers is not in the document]
jiangphcn: Addressed in d3d66c9

src/chttpd/src/chttpd_db.erl
jiangphcn: Removed unnecessary line

src/mem3/src/mem3_rep.erl
jiangphcn: ok-tag is not removed in some calls. Removed.

src/couch_mrview/src/couch_mrview_index.erl
```diff
@@ -219,11 +219,10 @@ index_file_exists(State) ->
     filelib:is_file(IndexFName).
 
 
-verify_index_exists(Props) ->
+verify_index_exists(DbName, Props) ->
     try
         Type = couch_util:get_value(<<"type">>, Props),
         if Type =/= <<"mrview">> -> false; true ->
-            DbName = couch_util:get_value(<<"dbname">>, Props),
```
iilyak: Nice. Two birds with one stone. Love it.

src/mem3/src/mem3_rep.erl
```diff
@@ -249,8 +248,9 @@ pull_purges(#acc{} = Acc0) ->
         end,
 
         if Remaining =< 0 -> ok; true ->
-            OldestPurgeSeq = couch_db:get_oldest_purge_seq(Db),
-            PurgesToPush = couch_db:get_purge_seq(Db) - OldestPurgeSeq,
+            {ok, PurgeSeq} = couch_db:get_purge_seq(Db),
```
iilyak: Did you verify it returns `{ok, pos_integer()}`? Even though dialyzer complains the tests are written in such a way that `ok` tag is not suppose to be there [see](https://github.com/apache/couchdb/blob/master/src/couch/src/test_engine_purge_docs.erl#L32). So either these tests are not running (which wouldn't surprise me given our previous experience with eunit) or the tag is added elsewere. Could you show me the lines where `{ok, integer()}` reply is formed? Also please verify that tests I pointed to are running.

src/mem3/src/mem3_rep.erl
jiangphcn: can you please check https://github.com/apache/couchdb/pull/1370/files#diff-66129b4f374c3cde3fc9a49fa8a9452eR522

src/chttpd/src/chttpd_db.erl
```diff
@@ -502,6 +502,19 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_purge">>]}=Req, Db) ->
     Options = [{user_ctx, Req#httpd.user_ctx}, {w, W}],
     {IdsRevs} = chttpd:json_body_obj(Req),
     IdsRevs2 = [{Id, couch_doc:parse_revs(Revs)} || {Id, Revs} <- IdsRevs],
+    MaxIds = config:get_integer("purge", "max_document_id_number", 100),
+    case length(IdsRevs2) =< MaxIds of
+        false -> throw({bad_request, "Document numbers larger than expected"});
```
iilyak: The working is unclear. How about:
- `Too many documents in a single purge request`
- `The number of documents in a single purge requests is greater than the limit`
- `Exceeded maximum number of documents`

src/chttpd/src/chttpd_db.erl
```diff
@@ -502,6 +502,19 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_purge">>]}=Req, Db) ->
...
+        true -> ok
+    end,
+    RevsLen = lists:foldl(fun({_Id, Revs}, Acc) ->
+        length(Revs) + Acc
+    end, 0, IdsRevs2),
+    MaxRevs = config:get_integer("purge", "max_revisions_number", 1000),
+    case RevsLen =< MaxRevs of
+        false -> throw({bad_request, "Document revs larger than expected"});
```
iilyak: The wording is unclear. How about:
- `Exceeded maximum number of revisions`

src/chttpd/src/chttpd_db.erl
jiangphcn: Used `Exceeded maximum number of documents.`

src/chttpd/src/chttpd_db.erl
jiangphcn: Used `Exceeded maximum number of revisions.`.

src/chttpd/src/chttpd_db.erl
```diff
@@ -504,15 +504,15 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_purge">>]}=Req, Db) ->
     IdsRevs2 = [{Id, couch_doc:parse_revs(Revs)} || {Id, Revs} <- IdsRevs],
     MaxIds = config:get_integer("purge", "max_document_id_number", 100),
     case length(IdsRevs2) =< MaxIds of
-        false -> throw({bad_request, "Document numbers larger than expected"});
+        false -> throw({bad_request, "Exceeded maximum number of documents."});
```
iilyak: 👍

src/chttpd/src/chttpd_db.erl
```diff
@@ -504,15 +504,15 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_purge">>]}=Req, Db) ->
...
         true -> ok
     end,
     RevsLen = lists:foldl(fun({_Id, Revs}, Acc) ->
         length(Revs) + Acc
     end, 0, IdsRevs2),
     MaxRevs = config:get_integer("purge", "max_revisions_number", 1000),
     case RevsLen =< MaxRevs of
-        false -> throw({bad_request, "Document revs larger than expected"});
+        false -> throw({bad_request, "Exceeded maximum number of revisions."});
```
iilyak: 👍

src/mem3/src/mem3_rep.erl
jiangphcn: There are some differences between `couch_db:get_purge_seq/1` and `couch_bt_engine:get_purge_seq/1`. The first returns `{ok, PurgeSeq}` while the second returns `PurgeSeq`. The code for `couch_db:get_purge_seq/1` is 

```
get_purge_seq(#db{}=Db) ->
    {ok, couch_db_engine:get_purge_seq(Db)}.
```

iilyak: Thank you I found it [here](https://github.com/apache/couchdb/blob/COUCHDB-3326-clustered-purge-pr5-implementation/src/couch/src/couch_db.erl#L521)

This means that return values from couch_db are inconsistent. Here is a list of functions which deal with different sequences:

- `get_update_seq` returns `integer()`
- `get_purge_seq` returns `{ok, integer}`
- `get_compacted_seq` returns `integer()`
- `get_committed_update_seq` returns `integer()`

I think that since we are updating purge feature we should take an opportunity to make it consistent, change return type and update all callers.

src/fabric/src/fabric_rpc.erl
jiangphcn: The `Options` passed on to `proplists:get_value/2` looks like
```
[replicated_changes,
 {user_ctx,{user_ctx,<<"chttpd_db_test_admin">>,[<<"_admin">>],<<"default">>}},
 {w,"1"}]
```
So need to use `proplists:get_value/2` to check value of `replicated_changes`.

jiangphcn:
Hey @iilyak I used 1dae6af to adjust returned value from `get_purge_seq/1` and `get_oldest_purge_seq/1`, and use 8a62f68 to add information about module and line. Also added explantation why we use `proplists:get_value/2` in fabric_rpc.

src/couch_pse_tests/src/cpse_util.erl
```diff
@@ -183,16 +183,24 @@ uuid() ->
     couch_uuids:random().
 
 
-assert_db_props(DbName, Props) when is_binary(DbName) ->
+assert_db_props(DbName, Props, Module, Line) when is_binary(DbName) ->
```
iilyak: Nitpick (kind of): I would put `Module` and `Line` to be first arguments. The reason is pure esthetics (very subjective). We expect `Props` to be long so it is better to have it as the very last argument. Compare:

```
assert_db_props(?MODULE, ?LINE, DbName, [
         {doc_count, 1},
         {doc_count, 1},
         {del_doc_count, 0},
         {del_doc_count, 0},
         {update_seq, 2},
         {update_seq, 2},
         {purge_seq, 0},
         {purge_seq, 0},
         {purge_infos, []}
]).
```

with 

```
assert_db_props(DbName, [
         {doc_count, 1},
         {doc_count, 1},
         {del_doc_count, 0},
         {del_doc_count, 0},
         {update_seq, 2},
         {update_seq, 2},
         {purge_seq, 0},
         {purge_seq, 0},
         {purge_infos, []}
], ?MODULE, ?LINE).

```

On the other hand most of the functions in erlang stdlib don't have this convention. For example all functions in `lists` module has long function definition as the first argument. I guess your version is less surprising to erlang developers. Therefore I propose to keep yours. Unless you think it would be better to change the order.

iilyak: Another option is to define macro (somewhere) as:
```
-define(assert_db_props(DbName, Props), 
      cpse_util:assert_db_props(DbName, Props, ?MODULE, ?LINE)).
```

Then you don't need to pass additional arguments at all:
```
?assert_db_props(DbName, [
         {doc_count, 1},
         {doc_count, 1},
         {del_doc_count, 0},
         {del_doc_count, 0},
         {update_seq, 2},
         {update_seq, 2},
         {purge_seq, 0},
         {purge_seq, 0},
         {purge_infos, []}
]).
```

iilyak: As I said not a big deal. Feel free to ignore.

src/chttpd/src/chttpd_db.erl
iilyak: 👍

iilyak:
I'm +1 pending resolution of my review comments. However I am concerned with deployment aspect of the PR. It is scary that we don't have a rollback. Could we at least have a documentation on what we would do if we would need to rollback to a version which doesn't have additional trees in the header? Ideally I would prefer to have a PR (which we would not be merging) with an implementation of a header downgrade.

src/couch_pse_tests/src/cpse_util.erl
jiangphcn: Move `?MODULE` and `?LINE` to the beginning of `assert_db_props/4` using commit 43f3183d275a4fae6f7ba5332ddca0a498315ec9

jiangphcn:
thanks again @iilyak I added one response https://github.com/apache/couchdb/pull/1370#discussion_r201238042. Hopefully there is no pending comment from your side except for downgrade database. I am working on downgrade case, but this will be in different PR.

wohali:
@davisp @jiangphcn With 2.2.0 down to just 4 open issues, and already with PSE in there, let's release that first.

Please hold merging these purge PRs until the 2.2.0 release is cut.

I anticipate that in the next 2 weeks or less.

jiangphcn:
okay @wohali thanks for letting us know.

jiangphcn:
Hi Paul, @davisp When using `git log --oneline`, the order is correct. 
```
localhost:src jiangph$ git log --oneline
bf809a8 [10/10] Clustered Purge: Clustered HTTP API
110dcdc [09/10] Clustered Purge: Fabric API
4dcfb50 [08/10] Clustered Purge: Update read-repair
973a599 [07/10] Clustered Purge: Internal replication
a947b5c [06/10] Clustered Purge: Update mrview indexes
3279c38 [05/10] Clustered Purge: Add upgrade tests
a192aef [04/10] Clustered Purge: Update couch_pse_tests
ac8d0f6 [03/10] Clustered Purge: Update couch_bt_engine
4d695f1 [02/10] Clustered Purge: Update single node APIs
0254305 [01/10] Clustered Purge: Define new purge API
830240b Create new on_compact trigger
4c87d78 Enhance PSE tests with setup/teardown functions
c998aa4 Update to use new couch_pse_tests app
0d85abe Update PSE test definitions for new util module
3c94bc4 Rename PSE test modules
db28f4c Move PSE tests to their own app
f45c036 Rewrite the PSE test suite to use couch_server
da43788 Simplify logic in mem3_rep
aea2833 Update fabric_doc_open eunit tests
bb27d5e Fix race on couch_db:reopen/1
151b9b0 Fix default security object handling
da2e22a Fix bug during purge
3852a00 Fix typos in couch_db_engine.erl
```

davisp:
Confirmed locally through not GitHub. Must be ordered on dates or something.

+1 to merge

Issue #1374, opened 2018-06-05
Retaining user_ctx after authentication


tonysun83:
After authenticating, we set the user_ctx inside httpd. We then begin authorization.
If authorization fails, we return the original HttpReq in the catch clause. This drops
the authenticated user_ctx.

chttpd.erl
```
    try
        couch_httpd:validate_host(HttpReq),
        check_request_uri_length(RawUri),
        check_url_encoding(RawUri),
        case chttpd_cors:maybe_handle_preflight_request(HttpReq) of
        not_preflight ->
            case chttpd_auth:authenticate(HttpReq, fun authenticate_request/1) of
            #httpd{} = Req ->
                HandlerFun = chttpd_handlers:url_handler(
                    HandlerKey, fun chttpd_db:handle_request/1),
                AuthorizedReq = chttpd_auth:authorize(possibly_hack(Req),
                    fun chttpd_auth_request:authorize_request/1),
                {AuthorizedReq, HandlerFun(AuthorizedReq)};
            Response ->
                {HttpReq, Response}
            end;
        Response ->
            {HttpReq, Response}
        end
    catch Tag:Error ->
        {HttpReq, catch_error(HttpReq, Tag, Error)}
    end.
```

## Expected Behavior
We should retain the user_ctx after authentication even if an error occurs.

## Current Behavior
We drop the user_ctx, so we don't know if authentication occurred.

## Possible Solution
Refactor the try catch or add a new try catch to retain the #user_ctx record.

## Steps to Reproduce (for bugs)
1) Create a user with only _reader role.
2) Try to create a database
3) Check in after_request method for #user_ctx. It should be undefined.

## Your Environment
Any couchdb instance.

big-r81:
Closing, fixed by #1375.

PR #1375, opened 2018-06-06
Refactor process_request to not drop req

tonysun83:
## Overview
Previously, when we fail authorization but pass authentication, we use
the old HttpReq which drops user_ctx (so it is undefined). This change retains the authenticated HttpReq which includes a user_ctx. This allows for after_requests to utilize a user_ctx that is not
undefined.

## Testing recommendations
Eunit test added to chttpd module.

https://github.com/apache/couchdb/issues/1374

## Checklist


iilyak:
Good catch!

Issue #1142, opened 2018-02-06
fabric:query_view/4 fails if number of docs in the DB is more than 60


AlexanderKaraberov:
## Expected Behaviour
Currently I'm implementing new view filter functionality using CouchDB 2.2 (master branch) codebase. Namely I want to filter `_changes` feed using additional query params like this:
`_changes?include_docs=true&filter=_view&view=index/type&key="d"'`, `start_key_docid` and so on. 
I managed to implement a working solution. I parse query string to `#mrargs` record (I reuse `fitler_args` field) and then pass this to this function (I wrote an additional case for a 5-tuple).
```erlang
filter(Db, DocInfo, {FilterType, Style, DDoc, VName, FilterArgs})
        when FilterType == view; FilterType == fast_view
```
And then I use `query_view/4` from `fabric.erl` proxy layer in order to execute a view and receive this kind of response `{row,[{id,<<"doc2">>},{key,<<"d">>}]}]`  and then I implemented something similar to this: `{ok, Passes} = couch_query_servers:filter_view(DDoc, VName, Docs),` functionality.
I would expect this to work on any number of docs (in my case an I have in average ~50000 docs and up to 200000 docs).

## Current Behaviour
My implementation works flawlessly with very small test databases (1-55 docs). I can apply additional filters (`type`, `keys` and so on) as a query param and all my tests succeed but when I have more than 50-60 docs this function fails. I have `results [ ` printed in logs and then just an empty space. I receive no timeout as well. Just in case I also checked in `fabric_view_map:go` that timeout value for erlang:send_after is fairly large:
```erlang
case rexi_utils:recv(Workers, #shard.ref, fun handle_message/3,
        State, infinity, 1000 * 60 * 60) of
``` 
Currently for the sake of testing I'm using only map views but i see the same `rexi_util_recv` call in `fabric_view_reduce:go2` as well.

## Possible Solution
It looks like Erlang `gen_server`'s `handle_call` doesn't have enough time to receive a response chunk and just infinitely calls `changes_callback` with no result but I might be wrong. Probably that's because of clustered configuration (I have a default 3-node setup) and now `query_view` takes additional time to query all nodes but timeout for `erlang:send_after` wasn't increased.

## Steps to Reproduce (for bugs)
You can very easily reproduce this. 
1. Just insert this call to query_view in `couch_changes:filter` with filter type == `view`:
```erlang
[_, DName] = binary:split(element(2, DDoc), <<"/">>),
[...]

PR #1382, opened 2018-06-11
Fix for a function_clause crash in couch_native_process

AlexanderKaraberov:
## Overview

In case when CouchDB has a native query server enabled and we call an Erlang `map` function for the first time (therefore there is no view index) `couch_mrview_updater:start_update` calls `run(State, [<<"map_doc">> , Doc])` from the `couch_native_process` and after this view index is built and data is read from it. Unfortunately when the same Erlang map function is used for changes feed with `_view` filter, `couch_native_process` crashes with:
```
{"error":"error","reason":"{{function_clause,\n     [{couch_native_process,ddoc,\n          [{evstate, ....
```
Changes feed with `_view` filter allows to use existing map function as the filter and it doesn’t query the view index files. Furthermore `seq_indexed` option doesn't work correctly in the mrview (more details in [this thread](https://github.com/apache/couchdb/issues/1142#issuecomment-378228093)) hence `fast_view` filter (which allows to use a dedicated `seq` and `keyseq` indexes) was disabled in the CouchDB 2. As a consequence changes feed with `_view` will always call `run(#evstate{ddocs=DDocs}=State, [<<"ddoc">>])` and this call will crash with the aforementioned error.
This crash is caused by a missing implementation of `ddoc` function for `<<"views">>` FunPath case, because for requests like `/_changes?style=all_docs&filter=_view&view=test/repl` the `FunPath` will be: `[<<"views">>,<<"repl">>,<<"map">>]`. For JS map functions this does work because in this case `couch_query_server` delegates map to an external `couch_js` OS process.
This fix effectively provides that missing `ddoc` clause and it is based on the `FilterFun` one but matches return values of the `erlang:put()` function (which is called from the native `Emit` function) and it also matches `ok` and `false` return values for the cases when docs were not emitted and Erlang view just returned either `ok` or a `false` atom. 

## Testing recommendations

Here is a simple shell script which makes curl requests in order to populate a test db with dummy docs and design docs containing Erlang and JS `map` functions:

```bash
#!/bin/sh

# creating the db
curl -u admin:admin1 -X PUT http://127.0.0.1:5983/db1

#insert ddoc with erlang view
curl -u admin:admin1 -H "Content-Type: application/json" -X PUT http://127.0.0.1:5983/db1/_design/test -d '{"_id":"_design/test","language":"erlang","options":{"include_deleted":true,"include_design":true},"views":{"repl":{"map":"fun({Doc}) ->\n    Repl = lists:keyfind(<<\"repl\">>, 1, Doc),\n    if\n         Repl =:= {<<\"repl\">>, <<\"unsafe\">>} ->\n              ok;\n         true ->\n              DocId = case lists:keyfind(<<\"_id\">>, 1, Doc) of\n                   {<<\"_id\">>, Id} -> Id;\n                   false -> undefined\n              end,\n              Emit(DocId, null)\n    end\nend."}}}'

#insert second ddoc with js view
curl -u admin:admin1 -H "Content-Type: application/json" -X PUT http://127.0.0.1:5983/db1/_design/testjs -d '{"_id":"_design/testjs", "language":"javascript","options":{"include_deleted":true,"include_design":true},"views":{"repl":{"map":"function (doc) {\n\tif (doc.repl !== \"unsafe\") emit(doc._id, null);\n}"}}}'

#populate with 50 repl safe docs
for IDX in {1..50}; do curl -u admin:admin1 -H "Content-Type: application/json" -X POST http://127.0.0.1:5983/db1 -d '{"index":"'$IDX'", "repl": "safe"}'; done

#populate with 10 repl unsafe docs
for IDX in {51..60}; do curl -u admin:admin1 -H "Content-Type: application/json" -X POST http://127.0.0.1:5983/db1 -d '{"index":"'$IDX'", "repl": "unsafe"}'; done

#populate with 10 random docs
for IDX in {61..70}; do curl -u admin:admin1 -H "Content-Type: application/json" -X POST http://127.0.0.1:5983/db1 -d '{"index":"'$IDX'", "data": "test"}'; done

```

In order to test just run changes feed for both JS map:
```
curl -u admin:admin1 'http://127.0.0.1:5983/db1/_changes?style=all_docs&filter=_view&view=testjs/repl'
```
and Erlang one `filter=_view&view=test/repl` and compare the output including the `"last_seq"` which ought to be the same.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ DESTDIR=
 
 # Rebar options
 apps=
-skip_deps=folsom,meck,mochiweb,triq,snappy
+skip_deps=folsom,meck,mochiweb,triq,snappy,bcrypt,hyper
 suites=
 tests=
 
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ help:
 
 
 .PHONY: couch
-# target: couch - Build CouchDB core
+# target: couch - Build CouchDB core, use ERL_OPTS to provide custom compiler's options
 couch: config.erl
 	@COUCHDB_VERSION=$(COUCHDB_VERSION) $(REBAR) compile $(COMPILE_OPTS)
 	@cp src/couch/priv/couchjs bin/
--- a/Makefile
+++ b/Makefile
@@ -34,16 +34,19 @@ skip_deps=folsom,meck,mochiweb,triq,snappy,bcrypt,hyper
 suites=
 tests=
 
+COMPILE_OPTS=$(shell echo "\
+	apps=$(apps) \
+	" | sed -e 's/[a-z_]\{1,\}= / /g')
 EUNIT_OPTS=$(shell echo "\
 	apps=$(apps) \
 	skip_deps=$(skip_deps) \
 	suites=$(suites) \
 	tests=$(tests) \
-	" | sed -e 's/[a-z]\+= / /g')
+	" | sed -e 's/[a-z]\{1,\}= / /g')
 DIALYZE_OPTS=$(shell echo "\
 	apps=$(apps) \
 	skip_deps=$(skip_deps) \
-	" | sed -e 's/[a-z]\+= / /g')
+	" | sed -e 's/[a-z]\{1,\}= / /g')
 
 #ignore javascript tests
 ignore_js_suites=
@@ -75,7 +78,7 @@ help:
 .PHONY: couch
 # target: couch - Build CouchDB core
 couch: config.erl
-	@COUCHDB_VERSION=$(COUCHDB_VERSION) $(REBAR) compile
+	@COUCHDB_VERSION=$(COUCHDB_VERSION) $(REBAR) compile $(COMPILE_OPTS)
 	@cp src/couch/priv/couchjs bin/
 
 
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -88,8 +88,8 @@ MakeDep = fun
 end,
 
 ErlOpts = case os:getenv("ERL_OPTS") of
-    "bin_opt_info" -> [bin_opt_info, debug_info, {i, "../"}];
-    Else -> [Else, {i, "../"}]
+    false -> [];
+    Opts -> [list_to_atom(O) || O <- string:tokens(Opts, ",")]
 end,
 
 AddConfig = [
@@ -98,7 +98,7 @@ AddConfig = [
     {deps, lists:map(MakeDep, DepDescs)},
     {sub_dirs, SubDirs},
     {lib_dirs, ["src"]},
-    {erl_opts, ErlOpts},
+    {erl_opts, [{i, "../"} | ErlOpts]},
     {eunit_opts, [verbose]},
     {plugins, [eunit_plugin]},
     {dialyzer, [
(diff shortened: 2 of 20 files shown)


wohali:
paging @eiri for PR review please

eiri:
@wohali @AlexanderKaraberov Just to let you know, I'll get to it later this week, suffering from a flu right now.

eiri:
This is rather embarrassing, but I still can't find time to look into it.

@iilyak, @nickva - if one of you have free cycles, can you revew this, please?

iilyak:
This PR needs tests. I do believe that following would be sufficient. I am going to test it tonight:
```
diff --git a/src/couch/test/couch_changes_tests.erl b/src/couch/test/couch_changes_tests.erl
index 673f2faad..f529eb312 100644
--- a/src/couch/test/couch_changes_tests.erl
+++ b/src/couch/test/couch_changes_tests.erl
@@ -153,7 +153,9 @@ filter_by_view() ->
             fun setup/0, fun teardown/1,
             [
                 fun should_filter_by_view/1,
-                fun should_filter_by_fast_view/1
+                fun should_filter_by_fast_view/1,
+                fun should_filter_by_erlang_view/1
+
             ]
         }
     }.
@@ -665,6 +667,39 @@ should_filter_by_user_ctx({DbName, _}) ->
             ?assertEqual(UpSeq, LastSeq)
         end).

+should_filter_by_erlang_view({DbName, _}) ->
+    ?_test(
+        begin
+            DDocId = <<"_design/app">>,
+            DDoc = couch_doc:from_json_obj({[
+                {<<"_id">>, DDocId},
+                {<<"language">>, <<"javascript">>},
+                {<<"views">>, {[
+                    {<<"valid">>, {[
+                        {<<"map">>, <<"fun({Doc}) ->"
+                            " case lists:keyfind(<<\"doc3\">>, 1, Doc) ->"
+                                " Emit(<<\"doc3\">>, null); "
+                                " _ -> ok"
+                            " end "
+                        "end.">>}
+                    ]}}
+                ]}}
+            ]}),
+            ChArgs = #changes_args{filter = "_view"},
+            Req = {json_req, {[{
+                <<"query">>, {[
+                    {<<"view">>, <<"app/valid">>}
+                ]}
+            }]}},
+            ok = update_ddoc(DbName, DDoc),
+            {Rows, LastSeq, UpSeq} = run_changes_query(DbName, ChArgs, Req),
+            ?assertEqual(1, length(Rows)),
+            [#row{seq = Seq, id = Id}] = Rows,
+            ?assertEqual(<<"doc3">>, Id),
+            ?assertEqual(6, Seq),
+            ?assertEqual(UpSeq, LastSeq)
+        end).
```

iilyak:
@AlexanderKaraberov: I did test the change with the test above. It fails without your change and passes with it. Could you either add the test to your PR or write another one if you think of better ways to test it?

Thank you for your contribution.

AlexanderKaraberov:
@iilyak 
Yes sure. Sorry, I forgot to write proper tests before actually submitting my PR. I will run your test suite, verify whether it covers all the use cases and will include it in the PR. Hopefully today or tomorrow.

AlexanderKaraberov:
Hi @iilyak , I've checked the test you had added and it's completely fine. Thanks!
`should_filter_by_erlang_view` is exactly what's needed and I've included this test in the PR.

AlexanderKaraberov:
@iilyak 
Excuse me for inconvenience, I've screwed up my fork. Will re-open this PR as a new one to avoid all the clutter if you don't mind.

AlexanderKaraberov:
Reopened as https://github.com/apache/couchdb/pull/1401
Shame on me

PR #1390, opened 2018-06-15
Fix active size calculations for views

davisp:
It was brought to my attention that the active size looked a bit funny
occasionally as it could be larger than the file size. Given that active
size is defined to be the number of bytes used in a view file it must be
strictly greater than or equal to zero as well as less than or equal to
file size. Thus the great hunt had begun!

While I'd love more than anything to regail you, Dear Reader, with the
tales, the joys, the highs, and yes, even the lows of this search, it
turns out that I cannot. I must not. For there were none!

Turns out this was a trivial bug where we were re-using the ExternalSize
calculation instead of applying `couch_btree:size/1` to all of the
btrees in the `#mrview` records. Simple bug comes with a correspondingly
simple fix.

I also noticed that the info tests were broken and not being run so I
spent a few minutes cleaning those up to make the various assumptions.

## Checklist


rnewson:
+1

PR #1392, opened 2018-06-18
Add set mqd off heap

davisp:
## Overview

Use `process_flag(message_queue_data, off_heap)` on critical processes on Erlang VMs that support it.

## Testing recommendations

`make eunit`

## Checklist

- [x] No user observable difference in behavior;


chewbranca:
+1 although it would be nice to see ddoc_cache in here as well.

davisp:
@chewbranca Added to ddoc_cache_lru.

PR #1397, opened 2018-06-21
[wip] Create property storage in db engine

garrensmith:
## Overview

Add the ability to properties to a database. This would allow us to define if the database supports `shardkeys` or `_access` or anything unique to it.

## Testing recommendations

The eunit tests should all pass.

## Checklist

### DIFF
--- a/src/couch/src/couch_bt_engine.erl
+++ b/src/couch/src/couch_bt_engine.erl
@@ -39,12 +39,15 @@
     get_purge_seq/1,
     get_revs_limit/1,
     get_security/1,
+    get_prop/2,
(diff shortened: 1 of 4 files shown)


wohali:
I know this is a WIP, but is there any intent to expose these at `GET /{db}`?

Out of curiosity, what new attributes are coming?

Do these attributes replicate when replicating a shard directly, or is it like the `_security` object where that does not occur?

garrensmith:
This is purely internal. Just the ability to set flags for a db. Similar to the rev_limits or security object. But we won’t expose to the user.

PR #1411, opened 2018-06-25
Use new mochiweb recbuf undefined setting

### LINKED ISSUE #1409: Network performance issue (30 secs for 8mb PUT)
## Expected Behavior

```
time curl -N -X PUT http://admin:pass@somserver.com:9999/test/doc2001/file.txt -H "Content-Type: application/octet-stream" -d@8mb_file.dat
```

Should complete in a reasonable amount of time, ie < 1 second (its only a 8mb file).

## Current Behavior

The above command if issued from a remote client (but over a very fast network) **can take up to 30 seconds.**

The same command issue from a client on the same network takes < 1 second.

Network is very fast between servers, an scp of same file takes < 0.1 seconds.

Tested on dedicated servers and multiple AWS servers. Always same result.

It seems like maybe couchdb network layer maybe sending too small a packet?

I have configured a total of 6 servers mix of AWS and dedicated, all experience the same slowness

1.  Setup a 16.04 Ubuntu server
2. Installed couchdb per documentation
3. Issue a PUT of a 8MB binary file from a remote server to couchdb server

## Your Environment

* Version used:  2.1.1
* Browser Name and version:  curl
* Operating System and version:  Ubuntu 14.04 or 16.04

just single nodes and no special configuration.

wohali:
Fixes #1409.

The new version of mochiweb that we've moved to supports `undefined` as a `recbuf` default setting when starting up `mochiweb_http:start/1`. This should help CouchDB with attachments on high-latency links, without causing regression for CouchDB over low-latency links.

Tests were clear locally; let's see how they fare in the more heavily contended Travis CI environment.

/cc @kocolosk


kxepal:
Hi @wohali !

This patch worth to check on MacOSX and FreeBSD, because these two were the reason of the last buffer change. IIRC, originally, we didn't have such tweaks at all, so OS managed them in the same way. Linux would be fine in any case.

wohali:
@kxepal Interesting! If you look at #1409 you'll see that this problem specifically cropped up on Linux running on AWS, so these defaults don't work anymore for Linux. 

If you can help run the test cases, I'd greatly appreciate it. We still don't have dedicated FreeBSD/MacOS Jenkins boxes, but they are coming.

I don't think it makes sense to bake OSX/FreeBSD defaults into the code itself, but perhaps we could add comments in `default.ini` and the documentation about setting these values explicitly on other OSes if necessary? Given the increased up-take in Docker and packaged installations, I don't think it makes sense to hobble Linux distributions with bad default settings.

wohali:
Ooh! Another option is to address this with `etc/default.d/10-socket.ini` files in the BSD ports tree (/cc @dch) and over in @janl 's OSX installer builder. That's what `etc/default.d/` is intended for - package maintainer default files for distro-specific changes.

wohali:
@kxepal in your comments on https://issues.apache.org/jira/browse/COUCHDB-1986 you state that multipart parser speedup as per https://issues.apache.org/jira/browse/COUCHDB-1953 solves the issue, yes? And that landed a while ago, so isn't your concern already addressed?

kxepal:
Hm...I fear that AWS applies some own specifics on top of. Debugging clouds is hard.

>If you can help run the test cases, I'd greatly appreciate it. We still don't have dedicated FreeBSD/MacOS Jenkins boxes, but they are coming.

I have FreeBSD at work and will check it tomorrow. The only interesting test is replication one where we check how good we are in replication of various local-remote combinations some docs with sort of ~10MB attachments. It triggers issue perfectly by timeout and just slow run.

At those time we tried to fix OSX performance at the first place. May be that problem is not actual anymore.

> Another option is to address this with etc/default.d/10-socket.ini files in the BSD ports tree (/cc @dch) and over in @janl 's OSX installer builder. That's what etc/default.d/ is intended for - package maintainer default files for distro-specific changes.

That's great idea! I like it :+1:

kxepal:
@wohali 
Wow! I don't recall 1953 merge in upstream at all.  Feels like it should solve the problem, but better check twice.

kocolosk:
For what it's worth mochiweb only handles the `recbuf` case; it explicitly looks for `undefined` and if it finds that atom it declines to pass any option to the underlying `gen_tcp` system:

https://github.com/apache/couchdb-mochiweb/commit/42d37e6

The reason they did it this way was because the backwards-compatible behavior for mochiweb was that leaving the option blank meant mochiweb would fix the initial `recbuf` to 8192.

By contrast, I would guess that the `sndbuf` setting is quietly returning `{error, einval}` when we use the atom undefined here, although it ultimately achieves the same goal of letting the OS manage the buffer.

wohali:
I dropped the sndbuf change, but the recbuf change is still failing. Need help.

kocolosk:
I wonder if other parts of the codebase are using those options directly with `gen_tcp`. Will take a look.

kocolosk:
OK I see the problem -- mochiweb only does this `recbuf` dance when opening a new socket to add to the acceptor pool. CouchDB is actually executing this code to set the socket options at the beginning of every *request*; it uses `mochiweb_socket:setopts/2` on the existing socket which simply calls `inet:setopts/2` (and fails).

Need to see if how we configure this so that our socket options overrides are set on a per-connection basis instead of a per-request basis.

wohali:
OK, with the help of @kocolosk and @rnewson it seems we should pass `{recbuf, undefined}` to `[chttpd] server_options`, and remove the `recbuf` setting from `socket_options` entirely.

PR updated.

wohali:
FYI this PR is on hold until we can review socket buffer size behaviour on all major supported OSes (Linux, FreeBSD, macOS, Windows) and determine sane defaults for all.

@rnewson is concerned that suddenly dropping our default 256k size by half, or more, could have a deleterious impact on many people. I concur. 

It may be that letting the OS manage buffer size would be better, esp. for OSes that do automatic buffer scaling like FreeBSD. What we really need is some benchmark data, across slow/high latency network connections. This will take some time.

At the very least, we will update our documentation...once we know better info about defaults.

kocolosk:
Ah good find on `server_options` - that's clearly what we wanted.

I understand the reticence about changing defaults. I also think it's rather unlikely that the CouchDB project is smarter about managing socket buffers than any reasonably modern operating system. I'd support switching to `undefined` as a default and documenting the local.ini setting necessary to restore previous behavior.

wohali:
I agree @kocolosk . I would still like to run a basic perf test before we do this. Does anyone at Cloudant have a harness already set up that would be willing to run something? I know I've seen perf stuff show up before from the likes of @davisp and @vatamane.

PR #1423, opened 2018-07-04
Mango to nodes

garrensmith:
## Overview

 Move mango selector matching to the shard level

 This moves the Mango selector matching down to the shard level.
 This would mean that the document is retrieved from the index and
 matched against the selector before being sent to the coordinator node.
 That way only the documents that match the query are sent to the coordinator node.

## Testing recommendations

Mango tests should all pass

## Checklist


src/mango/src/mango_cursor_view.erl
```diff
@@ -209,22 +213,66 @@ choose_best_index(_DbName, IndexRanges) ->
...
+
+
+view_cb({meta, Meta}, Acc) ->
+    % Map function starting
+    put(mango_docs_examined, 0),
+    ok = rexi:stream2({meta, Meta}),
+    {ok, Acc};
+view_cb({row, Row}, #mrargs{extra = Options} = Acc) ->
+    ViewRow = baseViewRow(Row),
+    case couch_util:get_value(doc, Row) of
```
davisp: Super minor nit, but `case ViewRow#view_row.doc of` would be slightly faster since you've already pulled it out of the Row value.

src/mango/src/mango_cursor_view.erl
```diff
@@ -332,17 +380,31 @@ apply_opts([{_, _} | Rest], Args) ->
...
             {ok, {DocProps}, {execution_stats, ExecutionStats1}};
         undefined ->
             ExecutionStats1 = mango_execution_stats:incr_quorum_docs_examined(ExecutionStats),
             Id = couch_util:get_value(id, RowProps),
             case mango_util:defer(fabric, open_doc, [Db, Id, Opts]) of
-                {ok, #doc{}=Doc} ->
-                    {ok, couch_doc:to_json_obj(Doc, []), {execution_stats, ExecutionStats1}};
+                {ok, #doc{}=DocProps} ->
+                    Doc = couch_doc:to_json_obj(DocProps, []),
+                    case mango_selector:match(Selector, Doc) of
```
davisp: You'll need to modify the ExecutionStats here to increment the examined doc count.

src/mango/src/mango_cursor_view.erl
```diff
@@ -209,22 +213,66 @@ choose_best_index(_DbName, IndexRanges) ->
...
+            Selector = couch_util:get_value(selector, Options),
+            case mango_selector:match(Selector, Doc) of
+                true ->
+                    ViewRow2 = ViewRow#view_row{
+                        value = get(mango_docs_examined) + 1
+                    },
+                    ok = rexi:stream2(ViewRow2),
+                    put(mango_docs_examined, 0);
+                false ->
+                    put(mango_docs_examined, get(mango_docs_examined) + 1)
```
davisp: Something to think about for the future will be how to tunnel a message through from here occasionally if we happen upon a highly restrictive filter. This doesn't change anything from the user POV currently as we could have the same effect from filtering at the collector level.

src/mango/src/mango_cursor_view.erl
```diff
@@ -28,6 +29,8 @@
 
 -include_lib("couch/include/couch_db.hrl").
 -include_lib("couch_mrview/include/couch_mrview.hrl").
+-include_lib("fabric/include/fabric.hrl").
```
davisp: I'm missing why we need to include fabric.hrl.

davisp:
Also, have you had a chance to do any sort of performance testing with this? It seems like it'd be helpful but I'm having a bit of premature-optimization-itis wondering if it makes any difference in practice. I can certainly dream up some scenarios where it'd help a cluster in a specific set of circumstances but I can't remember ever having seen such issues crop up in production.

src/mango/src/mango_cursor_view.erl
```diff
@@ -209,22 +213,66 @@ choose_best_index(_DbName, IndexRanges) ->
...
+    #view_row{
+        id = couch_util:get_value(id, Row),
+        key = couch_util:get_value(key, Row),
+        doc = couch_util:get_value(doc, Row)
+    }.
+
+
+view_cb({meta, Meta}, Acc) ->
+    % Map function starting
+    put(mango_docs_examined, 0),
```
tonysun83: having trouble figuring out why we need this

tonysun83: nevermind, it's for execution stats below

src/mango/src/mango_cursor_view.erl
garrensmith: We need it because it has the view_row record https://github.com/apache/couchdb/blob/master/src/fabric/include/fabric.hrl#L43

src/mango/src/mango_cursor_view.erl
garrensmith: I think we want to keep the quorum docs examined separate to the other docs examined.

garrensmith:
Thanks for reviewing @davisp and @tonysun83 I'm going to setup and run some perf tests on this, this week.

garrensmith:
I've run some perf tests on this. First a very basic performance test. Below are metrics for network traffic between nodes. The first image is for current mango. The second image is CouchDB with this PR.
old-mango2


new-mango2

The graphs are not at the same scale which is a bit frustrating. There is no regression and a slight improvement on network traffic. 
Then I ran a full regression test:
regression-test

There are no regressions and in some cases a nice performance improvement.

davisp:
From what I see there's basically no difference. For the network traffic graphs it looks maybe a bit less on the front end but then spikes badly on the second half. And if memory serves you were attempting to have a best case scenario for this test by restricting results to an empty set. Given there's not a dramatic difference I'm hesitant on whether this change is actually providing any benefit.

The regression tests I think are even more indicative of this not having much if any effect on the results. Your largest median change on a single test is only 23ms and it looks like nearly half of them are measuring nearly exactly the same suggesting this PR isn't providing a measurable difference.

Can you describe the tests you did for the network graphs? Perhaps we missed something there or we can figure out how to measure things better?

davisp:
Up to you whether you want to keep it open or closed. We can always re-open later if you get more time for perf testing as well. Also agree that I would have expected a much starker difference which suggests we're either missing something (like maybe that timeout you found) or there's some other bottleneck that prevents this from having a noticeable effect. Doesn't mean we don't want to pursue this, just that we need to investigate more to figure out what's up before we can merge it.

src/mango/test/20-no-timeout-test.py
```diff
@@ -0,0 +1,38 @@
...
+# License for the specific language governing permissions and limitations under
+# the License.
+
+import mango
+import copy
+import unittest
+
+class LongRunningMangoTest(mango.DbPerClass):
+
+  def setUp(self):
```
davisp: Python uses 4 space indents.

src/mango/src/mango_cursor_view.erl
```diff
@@ -240,9 +242,21 @@ view_cb({row, Row}, #mrargs{extra = Options} = Acc) ->
                         value = get(mango_docs_examined) + 1
                     },
                     ok = rexi:stream2(ViewRow2),
-                    put(mango_docs_examined, 0);
+                    put(mango_docs_examined, 0),
+                    set_mango_msg_timestamp();
                 false ->
-                    put(mango_docs_examined, get(mango_docs_examined) + 1)
+                    put(mango_docs_examined, get(mango_docs_examined) + 1),
```
davisp: Move this whole block to its own function.

tonysun83:
+1, great work @garrensmith

PR #1424, opened 2018-07-08
allow socket server configuration for TLS httpd

janl:
The https we optimally start doesn’t read any server options, that prevents users from applying things like `{max, 512}` sockets for mochiweb. This PR adds support for this:

```ini
[httpsd]
server_options = [{max, 512}]
```

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -104,10 +104,12 @@ start_link(https) ->
     end,
     SslOpts = ServerOpts ++ ClientOpts,
 
-    Options =
+    Options0 =
         [{port, Port},
          {ssl, true},
          {ssl_opts, SslOpts}],
+    CustomServerOpts = get_server_options("httpsd"),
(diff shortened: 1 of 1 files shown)


wohali:
minor quibble: I don't know how I feel about using `[chttpsd]` since elsewhere we refer to this daemon just as `[httpsd]`. Are you going to rename the option everywhere?

janl:
Good catch, will fix to the already existing section name.

janl:
Well, turns out, it is `[httpsd]` in the code. I just typo’d it in the description here: https://github.com/apache/couchdb/pull/1424/files#diff-b4ed04bdc6fda9ff5e838def1d1bf7a7R111

wohali:
+1 

Doesn't address #901, which would be nice to fix while you're in this part of the code, but not essential.

janl:
#901 is a lot more complicated to handle, as it is hooked into the architecture completely differently. I’d also argue that folks who need that should bind to 127.0.0.1 and maybe add a firewall setting to block access.

PR #1426, opened 2018-07-08
feat: avoid double builds on PRs

janl:
we don’t need to run the full test matrix twice for each PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,14 @@ script:
 after_failure:
   - build-aux/logfile-uploader.py
 
+# start a push build on master and release branches + PRs build on every branch
+# Avoid double build on PRs (See https://github.com/travis-ci/travis-ci/issues/1147)
+branches:
+  only:
+    - master
+    - /^\d+\.x\.x$/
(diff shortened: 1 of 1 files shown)


wohali:
-1, read on:

I think this means we won't run builds on PRs to release branches.

https://docs.travis-ci.com/user/customizing-the-build#Building-Specific-Branches 

If you want to stick with this approach, you'd explictly want to safelist using a regex per the docs.

However:

https://docs.travis-ci.com/user/pull-requests#double-builds-on-pull-requests
https://docs.travis-ci.com/user/web-ui/#build-pushed-branches

This suggests we just have Infra go into the web UI and disable "Build pushed branches", instead just leaving "Build pushed pull requests" turned on.

We might want to test the switches on a trivial repo of our own just to make sure

janl:
Ah yes, I'll add a release brachen reges, that might be easiest.

wohali:
OK, that'll do if we don't want to harass Infra. It also lets us turn it back on without bugging them anytime we want, if we want.

+1

PR #1432, opened 2018-07-11
Support callback module data provider

iilyak:
## Overview

This fixes the use case when data provider module for `couch_epi` is not a pure functional.
To illustrate the issue let's look at the following code:
```
-export([data/0]).

data() ->
   config:get("some_section").
```

We would want to regenerate dynamically compiled module when term returned from `data/0` is changed. However the previous implementation of `module` data provider was relying on `vsn` attribute of a module. Which is valid approach if the data term is hard coded in the module source.

This PR also renames `module` data provider into `static_module`.

## Testing recommendations

```
make eunit apps=couch_epi
```

## Checklist

### DIFF
--- a/src/couch_epi/README.md
+++ b/src/couch_epi/README.md
@@ -39,7 +39,8 @@ could add an entry in its implementation of couch_epi_plugin behaviour:
                 {priv_file, "stats_descriptions.cfg"}, [{interval, 5000}]}
             {{couch_stats, descriptions},
                 {file, "/tmp/extra_stats.cfg"}, [{interval, 5000}]},
-            {{couch_stats, descriptions}, {module, my_stats}}
+            {{couch_stats, descriptions}, {static_module, my_stats}},
+            {{couch_stats, descriptions}, {callback_module, my_stats}}
         ].
 
 When service provider wants to learn about all the installed config data for it to use
--- a/src/couch_epi/README.md
+++ b/src/couch_epi/README.md
@@ -66,6 +66,38 @@ There are also additional functions to get the same data in various formats
 - `couch_epi:keys(Handle)` - returns list of configured keys
 - `couch_epi:subscribers(Handle)` - return list of known subscribers
 
+The difference between `static_module` and `callback_module` providers is in how
+couch_epi detects the changes. `static_module` is designed for the cases when you
+have your data hardcoded in the module. For example you might have the following:
+
+```
+-export([data/0]).
+
+data() ->
+    [
+        {[complex, key, 2], [
+            {type, counter},
+            {desc, bar}
+        ]},
+        {[complex, key, 1], [
+            {type, counter},
+            {desc, updated_foo}
+        ]}
+    ].
+```
+
+The changes are detected by relying on `vsn` module attribute. Therefore we
+would notice the change only when data source module is recompiled.
+
+The `callback_module` provider uses the return value from `data/0` to detect
+changes and it is useful for cases when the data term is constructed dynamically.
+For example to cache values of CouchDB config one could use the following:
+
+```
+-export([data/0]).
+data() ->
+    config:get("dreyfus").
+```
 
 # Function dispatch example
 
--- a/src/couch_epi/src/couch_epi_data.erl
+++ b/src/couch_epi/src/couch_epi_data.erl
@@ -102,9 +102,9 @@ definitions({file, FilePath}) ->
         {error, Reason} ->
             {error, {FilePath, Reason}}
     end;
-definitions({module, Module}) when is_atom(Module) ->
-    definitions({module, [Module]});
-definitions({module, Modules}) ->
+definitions({static_module, Module}) when is_atom(Module) ->
+    definitions({static_module, [Module]});
+definitions({static_module, Modules}) ->
     Data = lists:append([M:data() || M <- Modules]),
     Hash = couch_epi_functions_gen:hash(Modules),
     {ok, Hash, Data}.
--- a/src/couch_epi/src/couch_epi_data.erl
+++ b/src/couch_epi/src/couch_epi_data.erl
@@ -107,7 +107,10 @@ definitions({static_module, Module}) when is_atom(Module) ->
 definitions({static_module, Modules}) ->
     Data = lists:append([M:data() || M <- Modules]),
     Hash = couch_epi_functions_gen:hash(Modules),
-    {ok, Hash, Data}.
+    {ok, Hash, Data};
+definitions({callback_module, Module}) ->
+    Data = Module:data(),
+    {ok, erlang:phash2(Data), Data}.
 
 hash_of_file(FilePath) ->
     {ok, Data} = file:read_file(FilePath),
--- a/src/couch_epi/src/couch_epi_sup.erl
+++ b/src/couch_epi/src/couch_epi_sup.erl
@@ -121,7 +121,7 @@ modules(#couch_epi_spec{kind = services, value = Module}) ->
     [Module];
 modules(#couch_epi_spec{kind = data_providers, value = Value}) ->
     case Value of
-        {module, Module} -> [Module];
+        {static_module, Module} -> [Module];
         _ -> []
     end;
 modules(#couch_epi_spec{kind = data_subscriptions, behaviour = Module}) ->
@@ -159,7 +159,7 @@ services() ->
 
 data_providers() ->
     [
-        {{test_app, descriptions}, {module, ?MODULE}, [{interval, 100}]}
+        {{test_app, descriptions}, {static_module, ?MODULE}, [{interval, 100}]}
     ].
 
 data_subscriptions() ->
--- a/src/couch_epi/src/couch_epi_sup.erl
+++ b/src/couch_epi/src/couch_epi_sup.erl
@@ -122,6 +122,7 @@ modules(#couch_epi_spec{kind = services, value = Module}) ->
 modules(#couch_epi_spec{kind = data_providers, value = Value}) ->
     case Value of
         {static_module, Module} -> [Module];
+        {callback_module, Module} -> [Module];
         _ -> []
     end;
 modules(#couch_epi_spec{kind = data_subscriptions, behaviour = Module}) ->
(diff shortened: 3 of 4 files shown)


src/couch_epi/src/couch_epi_data.erl
```diff
@@ -102,12 +102,15 @@ definitions({file, FilePath}) ->
...
-definitions({module, Modules}) ->
+definitions({static_module, Module}) when is_atom(Module) ->
+    definitions({static_module, [Module]});
+definitions({static_module, Modules}) ->
     Data = lists:append([M:data() || M <- Modules]),
     Hash = couch_epi_functions_gen:hash(Modules),
-    {ok, Hash, Data}.
+    {ok, Hash, Data};
+definitions({callback_module, Module}) ->
+    Data = Module:data(),
```
iilyak: I am not sure if there is a value in `static_module` provider. Because if your data term is hardcoded anyway why would you generate a wrapper module.
The only useful case is if we would decide to generate a versioned record access functions. Then the `data/0` would contain calls to pseudo function `record_info`. Even in this use case we can use `callback_module` logic. 

Let me know if you think we should just kill previous implementation of `module` provider.

tonysun83: It seems we are separating two different module types, but I'm not sure where we are actually regenerating the module. I still have to manually call  ```couch_epi_data_gen:generate/3``` to get values for new keys.

tonysun83: Upon reviewing the rest of the code base, I don't see ```couch_epi_data_gen:generate/3``` being called anywhere in codebase. Perhaps I'm misunderstanding how the process should work. My understanding is

1) I call ```couch_epi_data_gen:generate/3``` in my code to generate a new data module.
2) ```data/0``` is constantly being called every interval to check for new values. 
```
data() ->
    config:get("dreyfus").
```
3) If ```config:set("dreyfus", "foo", "bar").``` is called, then 
```
data() ->
   config:get("dreyfus", "foo").
```
will return a "bar" which is a new value. This should trigger some mechanism to call ```couch_epi_data_gen:generate/3``` to regenerate the module. This last piece
I am confused as to how it occurs. Perhaps something needs to be done in the ```notify/3```
function?

iilyak: You don't need to call `couch_epi_data_gen:generate/3`. Here is how it works. You define a data provider and optionally data subscriber (if you want notification on change):
```
dreyfus_epi.erl
 data_subscriptions() ->
-    [].
+    [{dreyfus, black_list}].

 data_providers() ->
-    [].
+    [
+        {{dreyfus, black_list}, {callback_module, dreyfus_config}, [{interval, 100}]}
+    ].
```

`rebar` reads [`coach_epi.config`](https://github.com/apache/couchdb/blob/master/rel/apps/couch_epi.config) [here](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi.app.src.script#L13). Vendors might have their own content for `coach_epi.config` in same location. That file should include `dreyfus_epi` plugin.

When `coach_epi` starts it reads list of plugins from `coach_epi.config` [in couch_epi_sup](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_sup.erl#L66).
It figures out [what kind of module](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_sup.erl#L66) we need to generate (we have two generators [`couch_epi_data_gen`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_data_gen.erl) and [`couch_epi_functions_gen`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_functions_gen.erl)). We start a module keeper as a supervisor child [`couch_epi_module_keeper:start_link(provider_kind(Kind), Key, CodeGenModule)`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_sup.erl#L101). The module keeper's job is to [generate and compile module](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_module_keeper.erl#L65) and [recompile it](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_module_keeper.erl#L108) when the definition has been changed. We detect the change by calling [CodeGenModule:data()](https://github.com/apache/couchdb/blob/master/src/couch_epi_module_keeper.erl#L116). It would be [couch_epi_data:data/1`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_data.erl#L30) for data provider case. Then following happens:
- we call [`collect_data/2`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_data.erl#L32) for every data provider definition (`dreyfus_epi:data_providers/1` returns a list if you recall). 
- [`collect_data/2`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_data.erl#L44) calls [`definitions/1`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_data.erl#L105)
- `definitions/1` [returns `{ok, Hash, Data}`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_data.erl#L110)

- We use returned hash to find out [if there are changes](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_module_keeper.erl#L117) in the definitions.
- If there is a change we generate a new version of module by calling [`CodeGenModule:generate/2`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_module_keeper.erl#L152) (in case of data provider `couch_epi_data_gen`)
- very first time the old Hash is undefined so we generate module. As you recall this is happening on couch_epi app start when we start application supervisor. That is why there is no need to manually generate module.
- if there was a change we call [`notify`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_module_keeper.erl#L121) which goes through [`couch_epi_plugin:notify/3`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_module_keeper.erl#L160) and [notify every subscriber](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_plugin.erl#L76) by calling [`Plugin:notify/3`](https://github.com/apache/couchdb/blob/master/src/couch_epi/src/couch_epi_plugin.erl#L85)

This is all quite complicated for the following reasons.
1. all plugins and their components need to start in order of their definition in `coach_epi.config`
2. every generated module need to be re-compiled if hot-code load is used (again in the right order)
3. modules need to be re-generated if definition changes

iilyak: In other words what you need is to add `dreyfus_epi` to `coach_epi.config` (if it is not there), `{callback_module, dreyfus_config}` into `dreyfus_epi:data_providers()`  and define `dreyfus_config:data/0` function. You don't need to worry about generating the module or re-generating it on change.

jaydoane: This is some great detail @iilyak, and I wonder if it wouldn't improve couch_epi if you copied some of it over to that repository's documentation?

src/couch_epi/README.md
```diff
@@ -65,6 +66,38 @@ There are also additional functions to get the same data in various formats
...
+would notice the change only when data source module is recompiled.
+
+The `callback_module` provider uses the return value from `data/0` to detect
+changes and it is useful for cases when the data term is constructed dynamically.
+For example to cache values of CouchDB config one could use the following:
+
+```
+-export([data/0]).
+data() ->
+    config:get("dreyfus").
```
tonysun83: change this to ```config:get("dreyfus", "foo")```?

src/couch_epi/src/couch_epi_data.erl
tonysun83: @iilyak :  I tend to agree with your assessment that we only need one type of "module":
**Because if your data term is hardcoded anyway why would you generate a wrapper module.**

I would just change callback_module to module and remove static_module.

The tests ran fine for me.

src/couch_epi/README.md
iilyak: It is up to the user. You can cache either values for a certain key or the whole section.

src/couch_epi/src/couch_epi_data.erl
iilyak: There is one use case when `static_module` is useful. There I think we should keep it. 
You can define same data provider in multiple applications. This is helpful for cases similar to use of `stats`. Currently stats definitions are loaded from plain configuration files in priv directory of every application. It can be rewritten using couch_epi as follows:
```
cat dreyfus_epi.erl
data_providers() ->
    [
        {{couch_stats, descriptions}, {static_module, dreyfus_stats}}
    ].
```
```
cat global_changes_epi.erl
data_providers() ->
    [
        {{couch_stats, descriptions}, {static_module, global_changes_stats}}
    ].
```

```
cat dreyfus_stats.erl
-module(dreyfus_stats).
-export([data/0]).
data() -> [
    {[dreyfus, httpd, search], [
        {type, histogram},
        {desc, <<"Distribution of overall search request latency as experienced by the end user">>}
    ]},
    {[dreyfus, rpc, search], [
        {type, histogram},
        {desc, <<"length of a search RPC worker">>}
    ]}
].
```

```
cat global_changes_stats.erl
-module(global_changes_stats).
-export([data/0]).

data() ->
   [
      {[global_changes, db_writes], [
          {type, counter},
          {desc, <<"number of db writes performed by global changes">>}
      ]},
      {[global_changes, event_doc_conflict], [
           {type, counter},
           {desc, <<"number of conflicted event docs encountered by global changes">>}
       ]}
   ].

```

In this case one single module would be generated which would include data from all applications. It would be more efficient to detect changes using `vsn` attribute of a module.

iilyak: @tonysun83: Are you OK if we keeping both providers?
 - `static_module`
- `callback_module`

iilyak:
@tonysun83: I am not sure if we can merge this PR before CouchDB release.

wohali:
@iilyak is there a strong motivation to merge it for 2.2.0?

iilyak:
> @iilyak is there a strong motivation to merge it for 2.2.0?

@wohali: We can wait. It is a bug fix for a feature which is not used yet.

tonysun83:
+1

wohali:
@iilyak you are unblocked, merge to master when ready

PR #1435, opened 2018-07-13
report git sha on welcome request

### LINKED ISSUE #1309: CouchDB should always report its git hash at runtime
## Current and expected behaviour
Currently, if you are running a CouchDB master build, you will see the matching git version when performing a `GET /`:
```
{"couchdb":"Welcome","version":"2.2.0-47a38d3d0","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```

And you see it in the log output at the `info` level:
```
[info] 2018-05-02T18:33:16.364837Z node1@127.0.0.1 <0.237.0> -------- Apache CouchDB 2.2.0-47a38d3d0 is starting.
```

When building a release, this information vanishes:
```
{"couchdb":"Welcome","version":"2.1.1","features":["scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```

It's not even in the log anymore:
```
[info] 2018-05-02T18:54:01.657777Z couchdb@127.0.0.1 <0.204.0> -------- Apache CouchDB 2.1.1 is starting.
```

This makes build audits painful, and makes release candidates impossible to distinguish from one actual releases. This, in turn, has lead to actual problems in the field, where people install an RC that has serious problems but it looks like it's an actual release.

## Proposal
1. Change the version number response as follows:
    1. When a build is built off of a release-like tag (regex: `^[0-9]+\.[0-9]\.[0-9]+(-RC[0-9]+)?$$`), give the build the version number `X.Y.Z`.
    1. When the regex doesn't match, label the build `X.Y.Z-dev`.
1. Always return the above version number, as well as the git hash (short or long, long might be better) info on `GET /` as a subkey. Examples:
```json
{"couchdb":"Welcome","version":"2.2.0-dev","git_hash":"47a38d3d0","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
{"couchdb":"Welcome","version":"2.1.1","git_hash":"f84faa532","features":["scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```
5. Always print the version number and the git hash at info level into the log on startup. Examples:
```
[info] 2018-05-02T18:33:16.364837Z node1@127.0.0.1 <0.237.0> -------- Apache CouchDB 2.2.0-dev (git hash 47a38d3d0) is starting.
[info] 2018-05-02T18:54:01.657777Z couchdb@127.0.0.1 <0.204.0> -------- Apache CouchDB 2.1.1 (git hash f84faa532) is starting.
```

## Context
As a release manager, I want to ensure that any and all builds of CouchDB are traceable to the exact source code used to build them. This extends to and includes all development builds, release candidate builds, and official release builds.

janl:
Closes #1309

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@
 include version.mk
 
 REBAR?=$(shell echo `pwd`/bin/rebar)
+COUCHDB_GIT_SHA=$(git_sha)
 IN_RELEASE = $(shell if [ ! -d .git ]; then echo true; fi)
 ifeq ($(IN_RELEASE), true)
 COUCHDB_VERSION = $(vsn_major).$(vsn_minor).$(vsn_patch)
@@ -22,8 +23,9 @@ IN_RC = $(shell git describe --tags --always --first-parent \
 							| grep -Eo -- '-RC[0-9]+' 2>/dev/null)
 RELTAG = $(shell git describe --dirty --abbrev=0 --tags --always --first-parent \
 				| grep -Eo '^[0-9]+\.[0-9]\.[0-9]+')
+COUCHDB_GIT_SHA=$(shell git rev-parse --short --verify HEAD)
 ifeq ($(RELTAG),)
-COUCHDB_VERSION_SUFFIX = $(shell git rev-parse --short --verify HEAD)
+COUCHDB_VERSION_SUFFIX = $(COUCHDB_GIT_SHA)
 COUCHDB_VERSION = $(vsn_major).$(vsn_minor).$(vsn_patch)-$(COUCHDB_VERSION_SUFFIX)
 else
 COUCHDB_VERSION = $(RELTAG)
@@ -82,7 +84,7 @@ help:
 .PHONY: couch
 # target: couch - Build CouchDB core, use ERL_OPTS to provide custom compiler's options
 couch: config.erl
-	@COUCHDB_VERSION=$(COUCHDB_VERSION) $(REBAR) compile $(COMPILE_OPTS)
+	@COUCHDB_VERSION=$(COUCHDB_VERSION) COUCHDB_GIT_SHA=$(COUCHDB_GIT_SHA) $(REBAR) compile $(COMPILE_OPTS)
 	@cp src/couch/priv/couchjs bin/
 
 
--- a/Makefile
+++ b/Makefile
@@ -13,22 +13,49 @@
 include version.mk
 
 REBAR?=$(shell echo `pwd`/bin/rebar)
+
+# Handle the following scenarios:
+#   1. When building from a tarball, use version.mk.
+#   2. When building from a clean release tag (#.#.#), use that tag.
+#   3. When building from a clean RC tag (#.#.#-RC#), use JUST the version
+#      number inside the tarball, but use the full name for the name of the
+#      tarball itself.
+#   4. When not on a clean tag, use version.mk + git sha + dirty status.
+
 COUCHDB_GIT_SHA=$(git_sha)
+
 IN_RELEASE = $(shell if [ ! -d .git ]; then echo true; fi)
 ifeq ($(IN_RELEASE), true)
+
+# 1. Building from tarball, use version.mk.
 COUCHDB_VERSION = $(vsn_major).$(vsn_minor).$(vsn_patch)
+
 else
-# IN_RC generates a tarball that has the -RCx suffix in the name if needed
+
+# Gather some additional information.
+# We do it this way so we don't bake shell-isms into Makefile
+# to make it easier to port to Windows. I know, I know. -jst
+# IN_RC contains the -RCx suffix in the name if present
 IN_RC = $(shell git describe --tags --always --first-parent \
-							| grep -Eo -- '-RC[0-9]+' 2>/dev/null)
-RELTAG = $(shell git describe --dirty --abbrev=0 --tags --always --first-parent \
-				| grep -Eo '^[0-9]+\.[0-9]\.[0-9]+')
+        | grep -Eo -- '-RC[0-9]+' 2>/dev/null)
+# ON_TAG matches *ONLY* if we are on a release or RC tag
+ON_TAG = $(shell git describe --tags --always --first-parent \
+        | grep -Eo -- '^[0-9]+\.[0-9]\.[0-9]+(-RC[0-9]+)?$$' 2>/dev/null)
+# RELTAG contains the #.#.# from git describe, which might be used
+RELTAG = $(shell git describe --tags --always --first-parent \
+        | grep -Eo -- '^[0-9]+\.[0-9]\.[0-9]+' 2>/dev/null)
+# DIRTY identifies if we're not on a commit
+DIRTY = $(shell git describe --dirty | grep -Eo -- '-dirty' 2>/dev/null)
+# COUCHDB_GIT_SHA is our current git hash.
 COUCHDB_GIT_SHA=$(shell git rev-parse --short --verify HEAD)
-ifeq ($(RELTAG),)
-COUCHDB_VERSION_SUFFIX = $(COUCHDB_GIT_SHA)
+
+ifeq ($(ON_TAG),)
+# 4. Not on a tag.
+COUCHDB_VERSION_SUFFIX = $(COUCHDB_GIT_SHA)$(DIRTY)
 COUCHDB_VERSION = $(vsn_major).$(vsn_minor).$(vsn_patch)-$(COUCHDB_VERSION_SUFFIX)
 else
-COUCHDB_VERSION = $(RELTAG)
+# 2 and 3. On a tag.
+COUCHDB_VERSION = $(RELTAG)$(DIRTY)
 endif
 endif
 
--- a/Makefile.win
+++ b/Makefile.win
@@ -14,15 +14,50 @@ include version.mk
 
 SHELL=cmd.exe
 REBAR?=$(shell where rebar.cmd)
+
+# Handle the following scenarios:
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -14,18 +14,19 @@ include version.mk
...
 RELTAG = $(shell git describe | grep -E '^[0-9]+\.[0-9]\.[0-9]+(-RC[0-9]+)?$$')
 ifeq ($(RELTAG),)
-COUCHDB_VERSION_SUFFIX = $(shell git rev-parse --short --verify HEAD)
 COUCHDB_VERSION = $(vsn_major).$(vsn_minor).$(vsn_patch)-$(COUCHDB_VERSION_SUFFIX)
 else
 COUCHDB_VERSION = $(RELTAG)
 endif
 endif
 
+
```
wohali: unnecessary whitespace

Makefile.win
```diff
@@ -15,11 +15,11 @@ include version.mk
 SHELL=cmd.exe
 REBAR?=$(shell where rebar.cmd)
 IN_RELEASE = $(shell if not exist .git echo true)
+COUCHDB_VERSION_SUFFIX = -$(shell git rev-parse --short --verify HEAD)
```
wohali: nitpick: why? the way it is now does less work

src/couch/rebar.config.script
```diff
@@ -39,6 +39,13 @@ Version = case os:getenv("COUCHDB_VERSION") of
         string:strip(Version0, right)
 end,
 
+VersionSuffix = case os:getenv("COUCHDB_VERSION_SUFFIX") of
```
wohali: I don't think this will t work when building from a release tarball, when we're not inside of git. Test case should be:

```
./configure -c
make dist
cd apache-couchdb-*
./configure -c
make couch
```

which I expect to fail.

janl: You’re right. lemme fix

wohali:
@janl No, this is still broken. `make dist` should produce a tarball at anytime that includes the git sha.

Running my test case above on your branch, with no tags, I get this:

```
$ curl localhost:15984
{"couchdb":"Welcome","version":"2.2.0","git_sha":"","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```

The git sha should be 'baked' into the result of `make dist` regardless of whether we're on a tag or not. This allows us to distinguish actual releases from RCs from merely snapshot builds.

janl:
@wohali can you show all steps preceding that curl?

The way it should™ work is that it gets baked into rebar at `make` time, but I might be misreading things.

wohali:
```
cd ~/couchdb
git reset --hard && git clean -ffdx
./configure -c
make dist
cd apache-couchdb-2.2.0-c8b0b0a2e
./configure -c
make couch
dev/run -n 1 >/dev/null 2>&1
sleep 10 && curl localhost:15984
```

janl:
I tried a good while trying to get this going, but it is not going to be a trivial patch.

The problem is as follows: the git sha is only known at `make dist` time. At that point, `make list` ruins `git archive` to build the tarball. That means that only files that have been committed to the source tree make it into the source archive. In order to get the git sha in there, is to make a commit. Or to ament the tarball with a new file that contains the git sha. The former doesn’t sound right to me as that would make the tarball dependent on a commit that will only exist on the release manager’s machine, plus those commits would never make sense to share. The second option might be worth investigating, but I ran out of time for this. Dropping from 2.2.0 unless someone picks this up.

janl:
turns out this was easier after all. After the `git archive`, we do a bunch of things to the tarball, like copying the docs and fauxton builds into it, so I’m now appending `git_sha` to `version.mk` and have `Makefile[.win]` -> `src/couch/rebar.config.script` -> `couch_server.erl` pick it up:

```
{"couchdb":"Welcome","version":"2.2.0","git_sha":"88e9b5e25","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```

janl:
@wohali no strong opinion either way. Feel free to amend my patch directly. 

One thing I could see people ask is to not make this info available on a non-admin endpoint (I know), what's your thinking on that?

wohali:
While reviewing this PR, I found that our fixes to improve the 2.1.2 RC process accidentally broke building "snapshot" dist tarballs from `master` (useful if you want to run bleeding edge releases, for instance). This, in turn, broke our Jenkins CI setup.

I've fixed the bug, and also added dirty detection (which marks a release as `-dirty` in the tag if it's been locally modified.)

Like the commits prior, this absolutely requires the repo to be inside of git. If we switch version control systems ever, we'll have to redo this part of the Makefile.

##  Examples

### Building this PR
```sh
# ./configure && make dist
Done: apache-couchdb-2.2.0-794ea816c.tar.gz
# dev/run -n 1 >/dev/null &
# curl localhost:15984
{"couchdb":"Welcome","version":"2.2.0-794ea816c","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
# When building and running from the generated tarball:
{"couchdb":"Welcome","version":"2.2.0","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```
I think that, perhaps, we want to keep the sha suffix on this one inside the tarball. It's not a tagged release of any sort. Then again, the Jenkins auto-built tarballs rarely make their way out of the developer community, and if they do, people use them at their own peril. The presence of the `git_sha` is all that's necessary to distinguish this from a real release.

This is a minor concern and I don't think blocks this PR.

### Tagging this branch (locally!!!) `2.2.0-RC1`:
```
# ./configure -c && make dist
Done: apache-couchdb-2.2.0-RC1.tar.gz
# dev/run -n 1 >/dev/null &
# curl localhost:15984
{"couchdb":"Welcome","version":"2.2.0","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
# When building and running from the generated tarball:
{"couchdb":"Welcome","version":"2.2.0","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```
Everything is as expected: the tarball clearly says 2.2.0-RC1, but the build acts like a real 2.2.0 release, from tarball or from the free.

### Tagging this branch (locally!!!) `2.2.0` *without* removing the `2.2.0-RC1` tag (mimicing a real release):
```
# ./configure -c && make dist
Done: apache-couchdb-2.2.0.tar.gz
# dev/run -n 1 >/dev/null &
# curl localhost:15984
{"couchdb":"Welcome","version":"2.2.0","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
# When building and running from the generated tarball:
{"couchdb":"Welcome","version":"2.2.0","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```
Again, everything is as expected. Tarball and both built binaries match.

### After making a trivial change to a file (dirty repo) and running the build again:
```
# ./configure -c && make dist
Done: apache-couchdb-2.2.0-dirty.tar.gz
# dev/run -n 1 >/dev/null &
# curl localhost:15984
{"couchdb":"Welcome","version":"2.2.0-dirty","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
# When building and running from the generated tarball:
{"couchdb":"Welcome","version":"2.2.0","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```
This is fine for a first pass. It'd be nice if the asset inside the tarball reflected in the sha that it's been messed with, but that's a new ticket in my mind.

### Removing all tags, then tagging this branch (locally!!!) `2.3.4-RC1`:
```
# ./configure -c && make dist
Done: apache-couchdb-2.3.4-RC1.tar.gz
# dev/run -n 1 >/dev/null &
# curl localhost:15984
{"couchdb":"Welcome","version":"2.3.4","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
# When building and running from the generated tarball:
{"couchdb":"Welcome","version":"2.2.0","git_sha":"794ea816c","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}
```
This was an invalid test I did to satisfy my curiosity. The result is unusual, but expected: we still rely on the `version.mk` checked into git to be accurate inside a tarball. (Removing the need to have `version.mk` checked into the tree is something we could conceivably do later.)

@janl I thought about it, and no one in our repo is likely to collide hashes, and basically ~everyone uses short git hashes, so let's stick with it as is.

As for the info not being visible on a non-admin endpoint...prior art shows Cloudant exposes their internal build numbers on `GET /`... Honestly I don't think it's worth worrying about. If people are that worried they can block the entire `GET /` endpoint because they won't want to expose they're running 2.1.0 still! :wink:

I still need to test this on Windows - which, incidentally, doesn't support `make dist` right now (bash dependency). That's up next.

wohali:
And the windows build is broken, because the new `hyper` app doesn't support it /cc @kocolosk 

I'll fix Windows issues in another ticket. For now this is +1

wohali:
And the windows build is broken, because the new `hyper` app doesn't support it /cc @kocolosk 

I'll fix Windows issues in another ticket. For now this is +1

PR #1440, opened 2018-07-14
Optionally prevent non-admins from accessing /_all_dbs

janl:
This is a leftover from the 1.x -> 2.x transition. Adds a config option (bike shed away pls) to block non-admins from accessing `/_all_dbs`, which we should set to `true` in 3.0.0

### DIFF
--- a/src/chttpd/src/chttpd_auth_request.erl
+++ b/src/chttpd/src/chttpd_auth_request.erl
@@ -34,7 +34,10 @@ authorize_request_int(#httpd{path_parts=[]}=Req) ->
 authorize_request_int(#httpd{path_parts=[<<"favicon.ico">>|_]}=Req) ->
     Req;
 authorize_request_int(#httpd{path_parts=[<<"_all_dbs">>|_]}=Req) ->
-    Req;
+   case config:get_boolean("chttpd", "admin_only_all_dbs", false) of
+       true -> require_admin(Req);
+       false -> Req
+   end;
 authorize_request_int(#httpd{path_parts=[<<"_dbs_info">>|_]}=Req) ->
     Req;
 authorize_request_int(#httpd{path_parts=[<<"_replicator">>], method='PUT'}=Req) ->
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -98,6 +98,9 @@ max_db_number_for_dbs_info_req = 100
 ; uncomment the next line to enable proxy authentication
 ; authentication_handlers = {chttpd_auth, proxy_authentication_handler}, {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, default_authentication_handler}
 
+; prevent non-admins from accessing /_all_dbs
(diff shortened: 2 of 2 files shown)


src/chttpd/src/chttpd_auth_request.erl
```diff
@@ -34,7 +34,10 @@ authorize_request_int(#httpd{path_parts=[]}=Req) ->
 authorize_request_int(#httpd{path_parts=[<<"favicon.ico">>|_]}=Req) ->
     Req;
 authorize_request_int(#httpd{path_parts=[<<"_all_dbs">>|_]}=Req) ->
-    Req;
+   case config:get("chttpd", "admin_only_all_dbs", "false") of
+       "false" -> Req;
+       _Else -> require_admin(Req)
```
rnewson: I'd rather that "true" was required for this explicitly.

rnewson:
+1 otherwise.

src/chttpd/src/chttpd_auth_request.erl
janl: Cool, done in latest push, added `Else -> couch_log:error` so folks know what’s up when this crashes, hope that was okay

src/chttpd/src/chttpd_auth_request.erl
```diff
@@ -34,7 +34,11 @@ authorize_request_int(#httpd{path_parts=[]}=Req) ->
 authorize_request_int(#httpd{path_parts=[<<"favicon.ico">>|_]}=Req) ->
     Req;
 authorize_request_int(#httpd{path_parts=[<<"_all_dbs">>|_]}=Req) ->
-    Req;
+   case config:get("chttpd", "admin_only_all_dbs", "false") of
+       "false" -> Req;
+       "true" -> require_admin(Req);
+       Else  -> couch_log:error("Invalid setting for admin_only_all_dbs: ~p. Must be true or false.", [Else])
```
rnewson: this will cause a badmatch or something in the response itself, no? In general I'm not in favour of error messages that only the couchdb administrator can see and not the person/agent making the request. 

Either don't have the clause, so we get a case_clause (and a 500 server error) or flip it around like;

"true" -> require_admin(Req);
_ -> Req

janl: @rnewson sorry for making my intentions not clear enough. I’d like to avoid people leaving `_all_dbs` open when they think they’ve closed it due to a config typo like `admin_only_all_dbs = True`.

So in my first attempt I’m allowing non-admin access is the config value is explicitly `"true"`, and require admin access at all other times.

In my second attempt, I matched `"true"` and `"false"`, but that would lead to a bad match, if the config value had a typo, so I wanted at least to leave a nice error message.

janl: @rnewson agreed, good point. I switched it back to true/false and added `:get_boolean` so at least we get a unified bad value thing from the config system, which then points to the right place.

rnewson:
+1, very tidy.

PR #1468, opened 2018-07-25
Port bulk_docs.js

jaydoane:
## Overview

Port bulk_docs.js

## Testing recommendations

```
$ mix test --trace test/bulk_docs_test.exs
Excluding tags: [pending: true]


BulkDocsTest
  * test bulk docs raises conflict error for combined update & delete (475.6ms)
  * test bulk docs can create, update, & delete many docs per request (858.6ms)
  * test bulk docs supplies `id` if not provided in doc (169.4ms)
  * test bulk docs can detect conflicts (469.8ms)
  * test bulk docs raises error for invlaid `docs` parameter (253.9ms)
  * test bulk docs raises error for missing `docs` parameter (157.7ms)
  * test bulk docs raises error for `all_or_nothing` option (237.6ms)
  * test bulk docs emits conflict error for duplicate doc `_id`s (202.7ms)
  * test bulk docs raises error for invlaid `new_edits` parameter (207.1ms)

CouchTestCase


Finished in 3.5 seconds
9 tests, 0 failures

Randomized with seed 858127
```

## Checklist


test/elixir/test/bulk_docs_test.exs
```diff
@@ -0,0 +1,147 @@
...
+
+  @doc_count 5
+
+  @tag :with_db
+  test "bulk docs can create, update, & delete many docs per request", ctx do
+    db = ctx[:db_name]
+    docs = make_docs(@doc_count)
+    resp = bulk_post(docs, db)
+    assert revs_start_with(resp.body, "1-")
+    docs = revise(docs, resp.body)
```
iilyak: Naming for `revise` is unfortunate. It is not clear what the function does without looking at its source. Should we name it `to_map`?

iilyak: `to_map` is bad as well since the function do something else.

test/elixir/test/bulk_docs_test.exs
```diff
@@ -0,0 +1,147 @@
...
+  @tag :with_db
+  test "bulk docs emits conflict error for duplicate doc `_id`s", ctx do
+    docs = [
+      %{_id: "0", a: 0},
+      %{_id: "1", a: 1},
+      %{_id: "1", a: 2},
+      %{_id: "3", a: 3}]
+    body = bulk_post(docs, ctx[:db_name]).body
+    assert length(body) == 4
+    assert Enum.count(body, &(Map.get(&1, "error") == "conflict")) == 1
```
iilyak: The original JS test do it differently. It checks that:
- second element has: `id = "1"` and `error = undefined`
- third element has: `error = "conflict"`

Your version of the test would pass if we change implementation of the endpoint to return (for example):
```
[
   %{_id: "0", error: "conflict"},
   %{_id: "1", a: 1},
   %{_id: "1", a: 2},
   %{_id: "3", a: 3}
]
```

I think the test need to be changed to make sure that:
- it is exactly on element with `error = "conflict"` and its `_id` is `1`

test/elixir/test/bulk_docs_test.exs
jaydoane: Hmm, `revise` seems like a decent name (to me), since it takes the value of the "rev" field returned from the server and inserts or updates it as `:_rev` into the doc (map).

test/elixir/test/test_helper.exs
```diff
@@ -180,6 +180,10 @@ defmodule CouchTestCase do
         }
       end
 
+      def make_docs(count) do
+        make_docs(1, count)
```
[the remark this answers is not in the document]
jaydoane: I was tempted to change `make_docs/2`, but that would also require checking all the other places it's used, and I was not convinced that indexing from `1` would really be an issue. In any case, that change should likely be handled in a separate PR.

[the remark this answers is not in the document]
jaydoane: @iilyak I pushed a commit that refactors the doc generation code, standardizing on the range-based version from replication_test.exs, and also including a new `create_docs` which generates docs having atomic keys for more idiomatic code.

test/elixir/test/bulk_docs_test.exs
jaydoane: Yes, you are right! I was trying to be clever, but the original assertions are much stronger. Fixed.

iilyak:
I am getting 
```
code:  assert resp.status_code() == 201
     left:  202
     right: 201
     stacktrace:
       test/bulk_docs_test.exs:2: BulkDocsTest.create_db/1
       test/bulk_docs_test.exs:2: BulkDocsTest.set_db_context/1
       (elixir) lib/enum.ex:1899: Enum."-reduce/3-lists^foldl/2-0-"/3
       test/bulk_docs_test.exs:2: BulkDocsTest.__ex_unit_setup_0/1
       test/bulk_docs_test.exs:1: BulkDocsTest.__ex_unit__/2
```

I think we need to update https://github.com/apache/couchdb/blob/elixir-suite/test/elixir/test/test_helper.exs#L157

jaydoane:
@iilyak those line numbers on your stack trace are confusing, but it sounds like you think we should allow 202 for successful db creation. The docs don't mention 202 as a valid response from PUT /{db} however: http://docs.couchdb.org/en/latest/api/database/common.html#put--db -- is that a documentation problem?

iilyak:
It was a problem with my setup. I had just one node running. In such case some operations return "no quorum".

jaydoane:
Hmm, I guess that begs the question of whether we should assume these tests will only pass on a fully connected 3-node cluster? That was my assumption in any case.

iilyak:
@jaydoane Agree it was a local problem so no need to change anything. We might add some safety checks at some point to detect possible problems with setup, but it out of the scope for this PR I think.

iilyak:
I think "test bulk docs can detect conflicts" is flaky. Out of 20 executions it failed 4 times with following exception:
```
  1) test bulk docs can detect conflicts (BulkDocsTest)
     test/bulk_docs_test.exs:45
     Expected truthy, got false
     code: assert resp.status_code() == 201 and length(resp.body()) == length(docs)
     stacktrace:
       test/bulk_docs_test.exs:135: BulkDocsTest.bulk_post/2
       test/bulk_docs_test.exs:57: (test)
```

I added a debug printout to print the response we get. It is the following:
```
%HTTPotion.Response{
   body: [
       %{"accepted" => true, "id" => "1", "ok" => true, "rev" => "2-59422af42db146a6a87b76e5edbeeb44"}, 
       %{"id" => "2", "ok" => true, "rev" => "2-0450643c5eb507fee080a4298e102c9a"}, 
       %{"id" => "3", "ok" => true, "rev" => "2-7c9bd31172bbb3d4097fa45e3a6e2c85"}, 
       %{"id" => "4", "ok" => true, "rev" => "2-2fd866cc031a23ac8ccccb49952a2666"}, 
       %{"id" => "5", "ok" => true, "rev" => "2-3af34b0749ed8322340556d497581836"}
   ], 
   headers: %HTTPotion.Headers{
       hdrs: %{
           "cache-control" => "must-revalidate", 
           "content-length" => "338", 
           "content-type" => "application/json", 
           "date" => "Thu, 02 Aug 2018 20:27:43 GMT", 
           "server" => "CouchDB/2.1.1 (Erlang OTP/19)", 
           "x-couch-request-id" => "5bd040d9ec", 
           "x-couchdb-body-time" => "0"
       }
    }, 
    status_code: 202
}
```

test/elixir/test/bulk_docs_test.exs
```diff
@@ -0,0 +1,148 @@
...
+    rows = bulk_post(docs, ctx[:db_name]).body
+    assert Enum.at(rows, 1)["id"] == "1"
+    assert Enum.at(rows, 1)["ok"]
+    assert Enum.at(rows, 2)["error"] == "conflict"
+  end
+
+
+  defp bulk_post(docs, db) do
+    resp = Couch.post("/#{db}/_bulk_docs", [body: %{docs: docs}])
+    assert resp.status_code == 201 and length(resp.body) == length(docs)
```
iilyak: To simplify debugging please explain `assert` failures. Using something like (I used 203 code on purpose to make it fail, so you can see results):
```
assert resp.status_code == 203 and length(resp.body) == length(docs),
    """
    Expected 201 and the same number of rows as in request, got
    #{inspect resp, pretty: true, width: 20, limit: :infinity, printable_limit: :infinity}
    """
```

You can extract this inspect into a function in `test_helper.ex` to simplify reuse.

jaydoane:
@iilyak interesting that you got a 202 from a POST to _bulk_docs, since the docs don't say that's an expected result: http://docs.couchdb.org/en/latest/api/database/bulk-api.html#db-bulk-docs. I take it that you double checked you had all nodes up and connected?

I ran that test over 100 times using `for i in {1..100}; do mix test --trace test/bulk_docs_test.exs:45 || break; done` and did not see any errors. However, if you think we should also expect 202 from _bulk_docs POST, I can relax the assertion constraints.

test/elixir/test/bulk_docs_test.exs
jaydoane: @iilyak I pushed a commit that implements `pretty_inspect` in `test_helper.ex`, and improves the assertion as requested.

jaydoane:
@iilyak if you are ok with the latest changes, please LMK, and I will squash and rebase as necessary.

jaydoane:
@iilyak I have squashed and rebased -- I think this is ready to be merged.

PR #1470, opened 2018-07-29
Fix calculation of external size for attachments

eiri:
## Overview

A database external size is growing on any operation with attachments.

The external size of a database defined as an uncompressed size of its content, so expectation would be that with updating or deleting a document's attachment the external size would change accordingly, i.e. increase or decrease to a size of the attachment.

The external size, however, _always_ growing and shrinks to the expected values only on a database compression.

Here is an example:
```bash
$ curl -X PUT http://127.0.0.1:15984/koi
{"ok":true}
$ curl -X PUT http://127.0.0.1:15984/koi/holder -d'{}'
{"ok":true,"id":"holder","rev":"1-967a00dff5e02add41819138abb3284d"}
$ curl http://127.0.0.1:15984/koi | jq .sizes.external
4
$ curl -X PUT http://127.0.0.1:15984/koi/holder/keeper.bin?rev=1-967a00dff5e02add41819138abb3284d -H'Content-Type:application/octet-stream' --data-binary @data2K.bin
{"ok":true,"id":"holder","rev":"2-9125078ed1a87fcfca3d32d0834e41c8"}
$ curl http://127.0.0.1:15984/koi | jq .sizes.external
2052
$ curl -X PUT http://127.0.0.1:15984/koi/holder/filler.bin?rev=2-9125078ed1a87fcfca3d32d0834e41c8 -H'Content-Type:application/octet-stream' --data-binary @data8K.bin
{"ok":true,"id":"holder","rev":"3-eb058b53377ede2720530b6ff073813a"}
$ curl http://127.0.0.1:15984/koi | jq .sizes.external
10244
```

So far everything goes as expected and external size is growing in sync with attachment sizes. Let's try to update big attachment with a smaller file and then just delete the document completely:
```bash
$ curl -X PUT http://127.0.0.1:15984/koi/holder/filler.bin?rev=3-eb058b53377ede2720530b6ff073813a -H'Content-Type:application/octet-stream' --data-binary @data4K.bin
{"ok":true,"id":"holder","rev":"4-e1ce5b2ae946a8a64f3503cb6d646686"}
$ curl http://127.0.0.1:15984/koi | jq .sizes.external
14340
$ curl -X POST http://127.0.0.1:15984/koi/_compact
{"ok":true}
$ curl http://127.0.0.1:15984/koi | jq .sizes.external
6148
$ curl -X DELETE http://127.0.0.1:15984/koi/holder?rev=4-e1ce5b2ae946a8a64f3503cb6d646686
{"ok":true,"id":"holder","rev":"5-548de2faf802ebcca8de0c3ca3c32c09"}
$ curl http://127.0.0.1:15984/koi | jq .sizes.external
6148
$ curl -X POST http://127.0.0.1:15984/koi/_compact
{"ok":true}
$ curl http://127.0.0.1:15984/koi | jq .sizes.external
4
```

## Testing recommendations

Surprisingly we don't have any tests for database info and I couldn't find an appropriate place to add new tests, so I just wrote a new test suite.
```bash
$ make eunit apps=couch suites=couch_db_info_tests
...
module 'couch_db_info_tests'
  couch_db_info_tests: db_info_sizes_test_...[0.014 s] ok
  couch_db_info_tests: db_info_sizes_test_...[0.023 s] ok
  couch_db_info_tests: db_info_sizes_test_...[0.015 s] ok
  couch_db_info_tests: db_info_sizes_test_...[0.010 s] ok
  couch_db_info_tests: db_info_sizes_test_...[0.010 s] ok
  couch_db_info_tests: db_info_sizes_test_...[0.010 s] ok
  couch_db_info_tests: db_info_sizes_test_...[0.101 s] ok
  [done in 0.204 s]
  All 7 tests passed.
```

Naturally Travis and Jenkins should pass too.

## Checklist

### DIFF
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -914,7 +914,18 @@ prep_and_validate_update(Db, #doc{id=Id,revs={RevStart, Revs}}=Doc,
         % new doc, and we have existing revs.
         % reuse existing deleted doc
         if OldFullDocInfo#full_doc_info.deleted orelse AllowConflict ->
-            {validate_doc_update(Db, Doc, fun() -> nil end), Doc};
+            case couch_doc:has_stubs(Doc) of
+                true ->
+                    #full_doc_info{rev_tree=RevTree} = OldFullDocInfo,
+                    Leafs = couch_key_tree:get_all_leafs(RevTree),
+                    [{Leaf, DiskRevs} | _] = lists:reverse(Leafs),
+                    #leaf{deleted=Deleted, ptr=DiskSp} = Leaf,
+                    DiskDoc = make_doc(Db, Id, Deleted, DiskSp, DiskRevs),
+                    Doc2 = couch_doc:merge_stubs(Doc, DiskDoc),
+                    {validate_doc_update(Db, Doc2, fun() -> DiskDoc end), Doc2};
+                false ->
+                    {validate_doc_update(Db, Doc, fun() -> nil end), Doc}
+            end;
         true ->
             {conflict, Doc}
         end
--- a/src/couch/src/couch_db_updater.erl
+++ b/src/couch/src/couch_db_updater.erl
@@ -387,7 +387,7 @@ flush_trees(#db{} = Db,
         rev_tree = Flushed,
         sizes = #size_info{
             active = FinalAS + TotalAttSize,
-            external = FinalES + TotalAttSize
+            external = FinalES
         }
     },
     flush_trees(Db, RestUnflushed, [NewInfo | AccFlushed]).
@@ -415,7 +415,7 @@ check_doc_atts(Db, Doc) ->
     end.
 
 
-add_sizes(Type, #leaf{sizes=Sizes, atts=AttSizes}, Acc) ->
+add_sizes(Type, #leaf{deleted=Deleted, sizes=Sizes, atts=AttSizes}, Acc) ->
     % Maybe upgrade from disk_size only
     #size_info{
         active = ActiveSize,
@@ -423,7 +423,15 @@ add_sizes(Type, #leaf{sizes=Sizes, atts=AttSizes}, Acc) ->
     } = upgrade_sizes(Sizes),
     {ASAcc, ESAcc, AttsAcc} = Acc,
     NewASAcc = ActiveSize + ASAcc,
-    NewESAcc = ESAcc + if Type == leaf -> ExternalSize; true -> 0 end,
+    NewESAcc = case {Type, Deleted} of
+        {leaf, false} ->
+            SumFun = fun({_, S}, A) -> S + A end,
+            TotalAttSizes = lists:foldl(SumFun, 0, AttSizes),
+            ESAcc + ExternalSize + TotalAttSizes;
+        {leaf, _} ->
+            ESAcc + ExternalSize;
+        _ -> 0
+    end,
     NewAttsAcc = lists:umerge(AttSizes, AttsAcc),
     {NewASAcc, NewESAcc, NewAttsAcc}.
 
--- a/src/couch/src/couch_db_updater.erl
+++ b/src/couch/src/couch_db_updater.erl
@@ -415,7 +415,7 @@ check_doc_atts(Db, Doc) ->
     end.
 
 
-add_sizes(Type, #leaf{deleted=Deleted, sizes=Sizes, atts=AttSizes}, Acc) ->
+add_sizes(Type, #leaf{sizes=Sizes, atts=AttSizes}, Acc) ->
     % Maybe upgrade from disk_size only
     #size_info{
         active = ActiveSize,
@@ -423,14 +423,12 @@ add_sizes(Type, #leaf{deleted=Deleted, sizes=Sizes, atts=AttSizes}, Acc) ->
     } = upgrade_sizes(Sizes),
     {ASAcc, ESAcc, AttsAcc} = Acc,
     NewASAcc = ActiveSize + ASAcc,
-    NewESAcc = case {Type, Deleted} of
-        {leaf, false} ->
+    NewESAcc = if
+        Type == leaf ->
             SumFun = fun({_, S}, A) -> S + A end,
             TotalAttSizes = lists:foldl(SumFun, 0, AttSizes),
             ESAcc + ExternalSize + TotalAttSizes;
-        {leaf, _} ->
-            ESAcc + ExternalSize;
-        _ -> 0
+        true -> 0
     end,
     NewAttsAcc = lists:umerge(AttSizes, AttsAcc),
     {NewASAcc, NewESAcc, NewAttsAcc}.
--- a/src/couch/test/couch_db_info_tests.erl
+++ b/src/couch/test/couch_db_info_tests.erl
@@ -53,91 +53,59 @@ db_info_sizes_test_() ->
 
 
 should_change_with_doc_insert(DbName, DocId) ->
-    {ok, PreSizes} = get_db_sizes(DbName),
-    PreActiveSize = couch_util:get_value(active, PreSizes),
-    PreExternalSize = couch_util:get_value(external, PreSizes),
+    {PreActiveSize, PreExternalSize, _} = get_db_sizes(DbName),
     {ok, _} = update_doc(DbName, DocId, ?l2b(lists:duplicate(60, $#))),
-    {ok, PostSizes} = get_db_sizes(DbName),
-    PostActiveSize = couch_util:get_value(active, PostSizes),
-    PostExternalSize = couch_util:get_value(external, PostSizes),
+    {PostActiveSize, PostExternalSize, _} = get_db_sizes(DbName),
     ?assert(PostActiveSize > PreActiveSize),
     ?assert(PostExternalSize > PreExternalSize).
 
 should_change_with_doc_update(DbName, DocId) ->
-    {ok, PreSizes} = get_db_sizes(DbName),
-    PreActiveSize = couch_util:get_value(active, PreSizes),
-    PreExternalSize = couch_util:get_value(external, PreSizes),
+    {PreActiveSize, PreExternalSize, _} = get_db_sizes(DbName),
     {ok, _} = update_doc(DbName, DocId, <<>>),
-    {ok, PostSizes} = get_db_sizes(DbName),
-    PostActiveSize = couch_util:get_value(active, PostSizes),
-    PostExternalSize = couch_util:get_value(external, PostSizes),
+    {PostActiveSize, PostExternalSize, _} = get_db_sizes(DbName),
     ?assert(PostActiveSize > PreActiveSize),
     ?assert(PostExternalSize < PreExternalSize).
 
 should_change_with_att_upload(DbName, DocId) ->
-    {ok, PreSizes} = get_db_sizes(DbName),
-    PreActiveSize = couch_util:get_value(active, PreSizes),
-    PreExternalSize = couch_util:get_value(external, PreSizes),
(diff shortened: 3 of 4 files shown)


eiri:
@davisp ping. Does this look like a valid fix or am I missing something?

eiri:
@davisp Hey Paul, sorry this took ages to follow up, I kept been distracted by shiny things in Kubernetes land.

It turned out that you right and it's quite possible to have a deleted (with `_deleted` field) doc with attachment, so I removed bits around deleted flag from external size calculation and changed the tests to be more explicit about been ordered.

I also fixed a crash on attempt to restore a doc with attachment and added even more confusing foreachx tests in an attachments tests suite.

Can you please take a look if this looks right to you?

eiri:
@davisp ping? do you want me to change anything else here?

davisp:
Sorry I missed this. I'll not be able to get to it until Monday.

src/couch/src/couch_db.erl
```diff
@@ -914,7 +914,18 @@ prep_and_validate_update(Db, #doc{id=Id,revs={RevStart, Revs}}=Doc,
         % new doc, and we have existing revs.
         % reuse existing deleted doc
         if OldFullDocInfo#full_doc_info.deleted orelse AllowConflict ->
-            {validate_doc_update(Db, Doc, fun() -> nil end), Doc};
+            case couch_doc:has_stubs(Doc) of
+                true ->
+                    #full_doc_info{rev_tree=RevTree} = OldFullDocInfo,
+                    Leafs = couch_key_tree:get_all_leafs(RevTree),
+                    [{Leaf, DiskRevs} | _] = lists:reverse(Leafs),
```
davisp: I can't convince myself that this is correct. Is there some sort of guarantee that we're getting on the order of leaves returned by `get_all_leafs/1`? If not it seems like we need be looking at adding a sort based on `{Depth, Rev}` for the returned leaves. Though that's also assuming that we're only allowing the un-deletion of the latest rev. 

The test here would be when the deleted doc has multiple deleted conflicts of which only some have the required missing attachments. Given that the new doc comes in without a revision it seems like we'd need to pick a deterministic one? Though that confuses me on when a different deleted leaf revision would satisfy the attachment conditions...

wohali: @eiri two questions:

1) Will this be ready in the next few days?

2) Are you subscribed to dev@couchdb.apache.org and seeing my regular emails about the 2.3.0 milestone?

eiri: 1. Maybe, but I can't guarantee that. I just closed another blocker yesterday and working on this now, but here is also inclusion of changes for `active_size` we've discussed on dev mail list which makes a scope of this PR bigger.
2. Yes I am. Sorry, I'm working on those issues as fast as I can.

wohali: @eiri thanks :) no problem. If this is going to be too big to get into 2.3.0, that's fine -- and I know people in the US are off until Monday -- so willing to be lenient here if this is still making progress.

Thank you again for all your hard work :)

eiri: @wohali I'm afraid I can't make it to 2.3.0 😞 I've spend half of the weekend on this, but `active` still not calculated as expected in my tests and I yet to find what I'm missing there. I'll continue to work on this, but I don't see myself wrapping it up today.

wohali: @eiri that's fine, i'd rather this be correct than rushed. Thanks for the heads up!

iilyak:
@eiri What is the status on this one?

AlexanderKaraberov:
@davisp @eiri  Hello guys. Sorry to disturb. I have a question regarding `sizes.active` vs `sizes.file`, because I'm confused about the algorithm of how they are calculated and CouchDB documentation is overly terse regarding this topic. 
I have a database with `revs_limit=1000`, it contains only one document with the size 113 bytes (two small text fields). This document has 1001 revisions and revisions 4 and 5 of this document had attachment which weighted 666624 bytes. In revision 6 attachment was removed and then I just generated 996 revisions. Compaction is disabled hence I still can get that attachment. Now when I query db size at this state I see:
```
sizes: {
file: 31364327,
external: 666701,
active: 799810
}
```
`active` size looks more or less correct: 113bytes*1000 + 666624  + (I guess) some db headers and meta info. But why `file` size is  31364327? This totally dumbfounded me. What are those 30Mb for? 

P.S. "Wrong"-looking `external` size I guess is the subject of this PR hence I'm not asking about it.

eiri:
@AlexanderKaraberov Hey, sorry I somehow completely missed the notification for your comment and only found in because I'm thinking of reviving this PR.

So, yes db `sizes` is a confusing topic, I'll try to guess here to best of my knowledge.

The issue itself is two folded. First, `file` represents the complete disk space a database's shards are taking on a single node, i.e. the value deduplicated by shard, but summed up for all the shards. So if say I have a single shard empty db my `file` is going to be ~8kB (2 headers with `0` padding to 4kB, as you already found out), but for the same empty db with 8 shards the value is going to be ~66kB. But this shouldn't matter much in your case, as you are updating a single shard.

Second, append-only approach is kind of wasteful by design. When a doc got added to the db we are writing attachment body (if presented), doc body, doc info, IDs b+tree node, update seqs b+tree node and the commit's header, some of those augmented with md5 hashes and zero padding. Updating the same doc effectively makes all of this "inactive" and the whole thing got appended again, there are no some smart deduplication or anything. A side note - adding a new doc is a bit different, as we don't need to update _whole_ b+trees, only the changed path, but that again is not your case.

Said all of that 30Mb does seems excessive indeed, I'm going to run some tests to see if I can reproduce and analyze this. Btw, do you happened to have a size of this db after compression?

AlexanderKaraberov:
Hi @eiri 

> I somehow completely missed the notification for your comment and only found in because I'm thinking of reviving this PR.

That's ok. I appreciate your detailed answer.

> Btw, do you happened to have a size of this db after compression?

I assume you meant an actual size of the shard file on the file system after `snappy` compression? In my test case it was almost matching the one reported by the `sizes.file` (~30KB less). 

I've just repeated exactly the same test case as I described above on a new database (with 537Kb attachment) and I see that `sizes.file` shows `file: 31266020` and the actual `db1.1551713236.couch` shard file on the FS is `31236291` bytes. Looks like this ~26 MB of unknown space occupied are pretty consistent.

eiri:
@AlexanderKaraberov Ah, no, slip of the tongue, sorry. I meant did you try to compare pre- and post-compaction sizes of that database?

For compression, iirc in production snappy shown itself to be noticeable worse that deflate_6 size-wise and didn't gave _that_ much advantage perf-wise to worth a switch over.

AlexanderKaraberov:
@eiri 

> I meant did you try to compare pre- and post-compaction sizes of that database?

So I've just run the compaction for the mentioned db. FS size became `49347` bytes and almost matches reported `active` (`39092`) size, even though `sizes.file` report is not as accurate:
```
sizes: {
file: 165113,
external: 60,
active: 39092
}
```
As expected compaction effectively "fixes" this file bloat. But what is concerning for me is a pre-compaction size. Even considering a wasteful append-only which you've described those 30MB of space for nothing seems excessive. Unfortunately I'm not aware of any approach to analyse the actual content of the `.couch` shard file to understand what is the main culprit of this bloat.

rnewson:
I'm struggling to reproduce that we should note that compaction is an essential maintenance task and it clearly solves the problem.

My attempt to isolate this was as follows;

1) I created a single shard database;
curl 'foo:bar@localhost:15984/db1?n=1&q=1' -XPUT

2) "sizes" is

curl 'foo:bar@localhost:15984/db1' -s | jq .sizes
{
  "file": 8348,
  "external": 0,
  "active": 0
}

which looks like two couch file headers and an empty b+tree.

2) I made a file exactly 10MB long
dd if=/dev/zero of=zeroes bs=1m count=10

3) sizes after inserting this with
curl 'foo:bar@localhost:15984/db1/doc1/att1' -XPUT -T zeroes

{
  "file": 10522824,
  "external": 10485764,
  "active": 10505529
}

4) I then updated the document, preserving the attachment

{
  "file": 10547400,
  "external": 10485791,
  "active": 10513692
}

This confirms we do _not_ write an additional copy of the attachment, we're simply pointing at the original chunks from rev 1.

5) I compact the db and sizes are;

{
  "file": 10526923,
  "external": 10485786,
  "active": 10505598
}

---

Conclusion: There's no unexpected size bloat at any stage. 

Over time, as databases open/close and when couchdb is restarted, couch will add a new 4k 'header' at the end of the file, but you'd need a lot of those to reach 30mb. Other changes to the database would lead to old versions of b+tree nodes being littered in the file, superseded by later writes.

Can you do this test and tell us if you get a different result? Or perhaps there are other things going on in your situation that explain it.

Again, compaction is a required maintenance step and it fixes the problem. We are concerned if there's excessive (and unexplained) write magnification so please let us know if you can reproduce this in isolation.

rnewson:
(and noting that this is not the right place for this as it's a pull request for something completely unrelated)

AlexanderKaraberov:
@rnewson Thank you for a detailed response and sorry for the noise I've created in this PR. I didn't want to create an issue for this as I'm still not sure whether this is a bug. With your test case I get the same result. But behaviour which I've described is steadily reproducible with a different test case (I mentioned it above) which has something to do with creating a lot of revisions (I can't say for sure that adding and deleting attachment is related to be honest). So here are exact steps which can be used to reproduce my scenario (perhaps some of them are redundant but I'll mention all of them):
1. Create a db (?n=1&q=8').
2. Create an empty document `doc1` (rev=1)
3.  Add attachment (in my tests it was a 500Kb png image or a random 600Kb pdf file) (rev=2)
4. Update document (rev=3)
5. Delete an attachment (rev=4)
6. Generate 996 revisions (`revs_limit`=1000 in my case).
7. Observe `files.size` and actual FS size is around 30 Mb.

Expected size (according to my estimates): 1000*doc_size(59 bytes)+500Kb+4096 block*1000

rnewson:
the 996 revisions will generate trash as the btree is updated plus a 4k header for each write (assuming you did them sequentially). You can confirm that by seeing the file size increasing by something in excess of 4k for each revision in step 6 of your routine.

AlexanderKaraberov:
@rnewson Looks like this "trash"/leftovers from the B+-tree updates are a culprit. But what's interesting this is not predictable. I created a doc and updated it 3 times and size growth was constant `previous_size` + 4096*3. But after I ran a script with 995 revisions the size became 51Mb. So at some point this size growth ceases to be constant. Attachments indeed have nothing to do with this as in this case I was not using them and size for a 999 revision is even 20Mb bigger.

rnewson:
think the mystery is solved then, will stop misusing this PR. If you want to go further, start a new issue or hit us up on the couchdb-user mailing list.

AlexanderKaraberov:
@rnewson @eiri  Thank you for pointing me into a right direction and I again apologise for creating an unrelated noise in this PR. When I have more free time I'll try to investigate myself first why B+-tree updates generate so much garbage and bloat a file size, and if in the dead end I'll get back in the couchdb-user mailing list or create an issue.

rnewson:
because they're appended to the file. we expect this behaviour and its why compaction exists (to create a new file without the old btree nodes).

wohali:
@eiri I would love to see this in CouchDB 3.1, before we completely retire the storage backend. Do you have time to finish up this PR?

eiri:
Hey @wohali, really sorry for the late reply. I'm busy for at least next two weeks, but I will try to get to this fix after that or at first opportunity.

I'm going to close this PR for now, it's really stale and not worth of rebase attempt, I'll just redo it anew.

PR #1482, opened 2018-08-01
add partition endpoint for views along with validation on which

garrensmith:
Creating a new PR because the previous PR #1480 was closed and then I force pushed which means I cannot open it again.

This is a PR for the partitioned-wip branch. It adds some validation around partition work and a new url. The final url is still TBD but and can be changed in this PR easily at a later date.

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -18,7 +18,8 @@
     db_req/2, couch_doc_open/4,handle_changes_req/2,
     update_doc_result_to_json/1, update_doc_result_to_json/2,
     handle_design_info_req/3, handle_view_cleanup_req/2,
-    update_doc/4, http_code_from_status/1]).
+    update_doc/4, http_code_from_status/1,
+    handle_partition_design_req/2]).
 
 -import(chttpd,
     [send_json/2,send_json/3,send_json/4,send_method_not_allowed/2,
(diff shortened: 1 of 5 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -250,20 +251,83 @@ handle_view_cleanup_req(Req, Db) ->
     ok = fabric:cleanup_index_files_all_nodes(Db),
     send_json(Req, 202, {[{ok, true}]}).
 
+
+handle_partition_design_req(#httpd{
+        path_parts=[DbName, <<"_partition">>, PartitionKey, _Design, Name, <<"_",_/binary>> = Action | _Rest]
```
rnewson: In my work I'm calling this simply "Partition". For example, you might put documents in a partition called foo. It seems redundant to call that a partition key.

src/chttpd/src/chttpd_db.erl
```diff
@@ -250,20 +251,83 @@ handle_view_cleanup_req(Req, Db) ->
...
+    }=Req, Db) -> 
+    validate_partition_key(PartitionKey),
+
+    case chttpd:qs_value(Req, "partition") of
+        undefined ->
+            ok;
+        _ ->
+            throw({
```
rnewson: right idea, cwazy indentation.

src/chttpd/src/chttpd_db.erl
```diff
@@ -250,20 +251,83 @@ handle_view_cleanup_req(Req, Db) ->
...
+                bad_request,
+                <<"Partition key is not allowed in the query string">>
+            })
+     end,
+
+    case mem3:is_partitioned(DbName) of
+        false -> throw(<<"Database is not partitioned">>);
```
rnewson: {bad_request, ..}

src/chttpd/src/chttpd_db.erl
```diff
@@ -250,20 +251,83 @@ handle_view_cleanup_req(Req, Db) ->
...
+    end;
+handle_partition_design_req(_Req, _Db) ->
+    throw({bad_request, <<"missing partition key">>}).
+
+
+validate_partition_key(<<"_", _/binary>>) ->
+    throw({bad_request, <<"partition keys must not begin with an underscore">>});
+validate_partition_key(<<" ">>) ->
+    throw({bad_request, <<"partition keys cannot begin with a space">>});
+validate_partition_key(_) ->
```
rnewson: this is ad-hoc. there is a function to validate document ids inside couch already, we should call that. As per the updated spec, we should call it for the partition half _and_ the docid half, each of which has to obey the docid rules.

src/chttpd/src/chttpd_db.erl
```diff
@@ -250,20 +251,83 @@ handle_view_cleanup_req(Req, Db) ->
...
+    ok.
+
+
+check_ddoc_can_use_partitionkey(#doc{body={DDocBody}}) ->
```
rnewson: this needs to be deeper down as we will have the same question in other places. As a placeholder, fine.

src/chttpd/src/chttpd_view.erl
```diff
@@ -101,6 +106,37 @@ handle_temp_view_req(Req, _Db) ->
...
+
+    Restrictions = [
+        {include_docs, Args#mrargs.include_docs, true},
+        {conflicts, Args#mrargs.conflicts, true},
+        {stable, Args#mrargs.stable, true}
+    ],
+    lists:foreach(fun ({Param, Field, Value}) ->
+        case Field =:= Value of
+            true ->
+                Msg = iolist_to_binary(io_lib:format("~s is not allowed for a partition query", [Param])),
```
rnewson: no need to flatten the iolist, I think? and possibly as simple as `Msg = [Param, <<" is not allowed for a partition query">>]` if Param is a list or binary.

src/chttpd/src/chttpd_view.erl
```diff
@@ -101,6 +106,37 @@ handle_temp_view_req(Req, _Db) ->
...
+                throw({
+                    bad_request,
```
rnewson: needless indentation for something that fits comfortably on a single line

garrensmith:
closing for https://github.com/apache/couchdb/pull/1577

PR #1568, opened 2018-08-20
Log error when changes forced to rewind to beginning

jaydoane:
## Overview

During a recent rolling upgrade, we observed the changes feed rewinding to a very old version, which caused performance issues for the client since it had to slog through millions of previously seen changes. Unfortunately, there is no logging when the changes endpoint is forced to rewind, so this adds such logging .

## Testing recommendations

Unfortunately, this is not trivial to test.

## Checklist


jaydoane:
I've squashed into a single commit. @rnewson do you mind merging this?

PR #1570, opened 2018-08-21
Simplify logic in mem3_rep

davisp:
Previously there were two separate database references and it was not
clear which was used where. This simplifies things by reducing it to a
single instance so that the logic is simpler.

Original PR accidentally merged to the wrong branch:

https://github.com/apache/couchdb/pull/1367


davisp:
+1

PR #1577, opened 2018-08-23
add /_partition/partition/designdoc/ endpoints

garrensmith:
Adding partition endpoints along with adding some validation

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -18,7 +18,8 @@
     db_req/2, couch_doc_open/4,handle_changes_req/2,
     update_doc_result_to_json/1, update_doc_result_to_json/2,
     handle_design_info_req/3, handle_view_cleanup_req/2,
-    update_doc/4, http_code_from_status/1]).
+    update_doc/4, http_code_from_status/1,
+    handle_partition_design_req/2]).
 
 -import(chttpd,
     [send_json/2,send_json/3,send_json/4,send_method_not_allowed/2,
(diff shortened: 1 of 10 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -250,24 +251,73 @@ handle_view_cleanup_req(Req, Db) ->
...
+            ok;
+        _ ->
+            throw({bad_request, <<"Partition is not allowed in the query string">>})
+    end.
+
+
 bad_action_req(#httpd{path_parts=[_, _, Name|FileNameParts]}=Req, Db, _DDoc) ->
     db_attachment_req(Req, Db, <<"_design/",Name/binary>>, FileNameParts).
 
+
```
rnewson: ws only change.

src/chttpd/src/chttpd_db.erl
```diff
@@ -250,24 +251,73 @@ handle_view_cleanup_req(Req, Db) ->
...
-    end;
+    DDoc = get_design_doc(DbName, Name),
+    Handler = chttpd_handlers:design_handler(Action, fun bad_action_req/3),
+    Handler(Req, Db, DDoc);
 
 handle_design_req(Req, Db) ->
     db_req(Req, Db).
 
+
+get_design_doc(DbName, Name) ->
```
rnewson: the extraction of this function could be in a separate commit to make intention clearer. (You extracted it in order to use it from an extra location, rather than duplicating the code).

src/chttpd/src/chttpd_view.erl
```diff
@@ -50,6 +54,7 @@ design_doc_view(Req, Db, DDoc, ViewName, Keys) ->
         fun couch_mrview_http:view_cb/2, VAcc, Args),
     {ok, Resp#vacc.resp}.
 
+
```
rnewson: ws only change.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -582,9 +582,6 @@ parse_param(Key, Val, Args, IsDecoded) ->
             Args#mrargs{callback=couch_util:to_binary(Val)};
         "sorted" ->
             Args#mrargs{sorted=parse_boolean(Val)};
-        "partition" ->
```
rnewson: are we firmly against ?partition= working against _view? I quite like it for its simplicity.

garrensmith: I think we against it otherwise its going to get really confusing having a ?partition and a new url

rnewson: I would rather keep it. It's a feature of the map/reduce code and should be exposed there, rather than only reachable obscurely through mango.

rnewson:
merged manually, closing.

Issue #1585, opened 2018-08-28
/_{db}/_ensure_full_commit returns success for non-existent databases


flimzy:
If I attempt to call the `_ensure_full_commit` endpoint for a database with a name beginning with the '_' character, I get (fake) success, rather than 404.

## Expected Behavior

I expect a 404, as I get for a db with a standard name. Example:

    curl -X POST http://admin:abc123@localhost:6004/foo/_ensure_full_commit -H 'Content-Type: application/json'
    {"error":"not_found","reason":"Database does not exist."}

## Current Behavior

Apparent success, even though the database `_foo` does not exist.

    curl -X POST http://admin:abc123@localhost:6004/_foo/_ensure_full_commit -H 'Content-Type: application/json'
    {"ok":true,"instance_start_time":"0"}

## Context

Observed on CouchDB 2.2.0, but not on 2.1.1 and older. Tested using Docker images.

wohali:
Any attempts to access `/_*` endpoints are not real databases, but system endpoints.

Port 6004 is unusual, can you confirm this is the standard clustered (port 5984) port and not the node-local (port 5986) port?

flimzy:
Yes, sorry, 6004 is mapped to 5984 on my docker container (the unusual port mapping is to accommodate multiple versions of CouchDB running).

eiri:
@flimzy thank you for reporting. This is a result of refactoring where previously the calls to a non-existent database were intercepted up the chain by auxiliary security check. I've opened PR apache/couchdb#1588 to address this issue.

By the way, I don't know if you aware, but in 2.x a call to `/db/_ensure_full_commit` is a noop operation that's always returning 201. I don't know for sure, but I suspect it was done just to keep API's back compatibility.

flimzy:
> By the way, I don't know if you aware, but in 2.x a call to /db/_ensure_full_commit is a noop operation 

I wasn't aware. It looks like this needs to be updated in the documentation. I'll try to find time to do that soon, if nobody beats me to it.

PR #1588, opened 2018-08-29
Check if db exists in /db/_ensure_full_commit call

eiri:
## Overview

We removed a security call in `do_db_req` in apache/couchdb#1246 to avoid duplicate authorization check and as a result there are now no db validation in noop call `/db/_ensure_full_commit`. This makes it always return a success code, even for the missing databases.

This fix places the security check directly in `_ensure_full_commit` call and adds eunit tests for a good measure.

## Testing recommendations

`make eunit apps=chttpd suites=chttpd_db_test` for isolated tests

Closes apache/couchdb#1585

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -375,8 +375,13 @@ db_req(#httpd{method='POST', path_parts=[DbName], user_ctx=Ctx}=Req, Db) ->
 db_req(#httpd{path_parts=[_DbName]}=Req, _Db) ->
     send_method_not_allowed(Req, "DELETE,GET,HEAD,POST");
 
-db_req(#httpd{method='POST',path_parts=[_,<<"_ensure_full_commit">>]}=Req, _Db) ->
+db_req(#httpd{method='POST', path_parts=[DbName, <<"_ensure_full_commit">>],
+        user_ctx=Ctx}=Req, _Db) ->
     chttpd:validate_ctype(Req, "application/json"),
+    %% use fabric call to trigger a database_does_not_exist exception
+    %% for missing databases that'd return error 404 from chttpd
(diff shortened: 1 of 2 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -375,8 +375,10 @@ db_req(#httpd{method='POST', path_parts=[DbName], user_ctx=Ctx}=Req, Db) ->
 db_req(#httpd{path_parts=[_DbName]}=Req, _Db) ->
     send_method_not_allowed(Req, "DELETE,GET,HEAD,POST");
 
-db_req(#httpd{method='POST',path_parts=[_,<<"_ensure_full_commit">>]}=Req, _Db) ->
+db_req(#httpd{method='POST', path_parts=[DbName, <<"_ensure_full_commit">>],
+        user_ctx=Ctx}=Req, _Db) ->
     chttpd:validate_ctype(Req, "application/json"),
+    fabric:get_security(DbName, [{user_ctx, Ctx}]),
```
iilyak: I think there is a need for a comment here to describe why the call is needed and the reason for `fabric:get_security/2` to be used for database existence check (get_security uses local shards while `fabric:get_db_info` for example uses non local shards as well)

eiri: I'm using `get_security` here because 1) it is what old code was using 2) get security calls `fabric_util:get_db` that actually does open remote shards, but gives priority to local ones, so it allows to avoid round-trip when possible 3) it _presumably_ also confirms user to be `reader`, that what old code was saying, but I don't see it in fabric, tbh.

I can drop a comment here or I can switch to `fabric:get_db_info` if you think that'd be less subtle.

eiri:
@iilyak ok, updated with the comment. does it look sufficient for you?

PR #1605, opened 2018-09-18
Feature/user partitioned databases

rnewson:
## Overview

This PR introduces a new feature, user-defined partitioned databases.

A new kind of database can be created with the `?partitioned=true` option. All documents within the database must have document ids of the following format;

  partition_name:doc_id

both partition_name and doc_id must follow the couchdb id format (can't begin with _, etc).

All documents with the same partition_name are guaranteed to be mapped to the same shard range. When querying an index, the new `/db/_partition/$partition/_view` endpoint can query the view more efficiently, by only consulting the single shard range holding the partition. This is much more efficient and scales the same way that primary key lookup (GET /dbname/docid) does (approximately linearly).

## Testing recommendations

The PR contains multiple tests for basic functionality and all existing tests  still pass. When testing the PR, it is important to try the feature yourself, interactively, with docs and views of your choosing, to give us confidence in this new feature.

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -285,12 +285,17 @@ create_db_req(#httpd{}=Req, DbName) ->
     Q = chttpd:qs_value(Req, "q", config:get("cluster", "q", "8")),
     P = chttpd:qs_value(Req, "placement", config:get("cluster", "placement")),
     EngineOpt = parse_engine_opt(Req),
+    Partitioned = parse_partitioned_opt(Req),
     Options = [
         {n, N},
         {q, Q},
-        {placement, P}
+        {placement, P},
+        {initial_props, [{partitioned, Partitioned}]}
     ] ++ EngineOpt,
     DocUrl = absolute_uri(Req, "/" ++ couch_util:url_encode(DbName)),
+
+    ok = validate_partition_database_create(DbName, Partitioned),
+
     case fabric:create_db(DbName, Options) of
     ok ->
         send_json(Req, 201, [{"Location", DocUrl}], {[{ok, true}]});
@@ -417,16 +422,19 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_bulk_docs">>], user_ctx=Ctx}=Req, 
     _ ->
         Options = [{user_ctx,Ctx}, {w,W}]
     end,
+    DbName = couch_db:name(Db),
+    VOpts = [{partitioned, mem3:is_partitioned(DbName)}],
     case couch_util:get_value(<<"new_edits">>, JsonProps, true) of
     true ->
         Docs = lists:map(
             fun(JsonObj) ->
-                Doc = couch_doc:from_json_obj_validate(JsonObj),
+                Doc = couch_doc:from_json_obj_validate(JsonObj, DbName),
                 validate_attachment_names(Doc),
                 Id = case Doc#doc.id of
                     <<>> -> couch_uuids:new();
                     Id0 -> Id0
                 end,
+                couch_doc:validate_docid(Id, DbName, VOpts),
                 Doc#doc{id=Id}
             end,
             DocsArray),
@@ -452,7 +460,11 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_bulk_docs">>], user_ctx=Ctx}=Req, 
             send_json(Req, 417, ErrorsJson)
         end;
     false ->
-        Docs = [couch_doc:from_json_obj_validate(JsonObj) || JsonObj <- DocsArray],
+        Docs = [begin
+            couch_doc:from_json_obj_validate(JsonObj)
+            end
+            || JsonObj <- DocsArray],
+        [couch_doc:validate_docid(D#doc.id, DbName, VOpts) || D <- Docs],
         [validate_attachment_names(D) || D <- Docs],
         case fabric:update_docs(Db, Docs, [replicated_changes|Options]) of
         {ok, Errors} ->
@@ -783,7 +795,7 @@ db_doc_req(#httpd{method='GET', mochi_req=MochiReq}=Req, Db, DocId) ->
 
 db_doc_req(#httpd{method='POST', user_ctx=Ctx}=Req, Db, DocId) ->
     couch_httpd:validate_referer(Req),
-    couch_doc:validate_docid(DocId, couch_db:name(Db)),
+    validate_docid(DocId, couch_db:name(Db)),
     chttpd:validate_ctype(Req, "multipart/form-data"),
 
     W = chttpd:qs_value(Req, "w", integer_to_list(mem3:quorum(Db))),
@@ -839,7 +851,7 @@ db_doc_req(#httpd{method='PUT', user_ctx=Ctx}=Req, Db, DocId) ->
         update_type = UpdateType
     } = parse_doc_query(Req),
     DbName = couch_db:name(Db),
-    couch_doc:validate_docid(DocId, DbName),
+    validate_docid(DocId, DbName),
 
     W = chttpd:qs_value(Req, "w", integer_to_list(mem3:quorum(Db))),
     Options = [{user_ctx,Ctx}, {w,W}],
@@ -1330,7 +1342,7 @@ db_attachment_req(#httpd{method=Method, user_ctx=Ctx}=Req, Db, DocId, FileNamePa
                 % check for the existence of the doc to handle the 404 case.
                 couch_doc_open(Db, DocId, nil, [])
             end,
-            couch_doc:validate_docid(DocId, couch_db:name(Db)),
+            validate_docid(DocId, couch_db:name(Db)),
             #doc{id=DocId};
         Rev ->
             case fabric:open_revs(Db, DocId, [Rev], [{user_ctx,Ctx}]) of
@@ -1452,6 +1464,20 @@ parse_engine_opt(Req) ->
                     throw({bad_request, invalid_engine_extension})
             end
     end.
+
+
+parse_partitioned_opt(Req) ->
+    case chttpd:qs_value(Req, "partitioned") of
+        undefined ->
+            false;
+        "true" ->
+            true;
+        "false" ->
+            false;
+        _ ->
+            throw({bad_request, <<"`partitioned` parameter can only be set to true or false.">>})
+    end.
+
 
 parse_doc_query({Key, Value}, Args) ->
     case {Key, Value} of
@@ -1592,6 +1618,18 @@ extract_header_rev(Req, ExplicitRev) ->
     end.
 
 
+% cannot partition a system database
+validate_partition_database_create(DbName, Partitioned) ->
+    SystemId = DbName =:= ?l2b(config:get("mem3", "shards_db", "_dbs")) orelse
+        lists:member(DbName, ?SYSTEM_DATABASES),
+    case {Partitioned, SystemId} of
+        {true, true} ->
+            throw({bad_request, <<"Cannot partition a system database">>});
+        {_, _} ->
+            ok
+    end.
+
+
 validate_attachment_names(Doc) ->
     lists:foreach(fun(Att) ->
         Name = couch_att:fetch(name, Att),
@@ -1672,7 +1710,7 @@ bulk_get_open_doc_revs(Db, {Props}, Options) ->
 
 
 bulk_get_open_doc_revs1(Db, Props, Options, {}) ->
-    case parse_field(<<"id">>, couch_util:get_value(<<"id">>, Props)) of
+    case parse_id_field(couch_util:get_value(<<"id">>, Props), Db) of
         {error, {DocId, Error, Reason}} ->
             {DocId, {error, {null, Error, Reason}}, Options};
 
@@ -1721,16 +1759,18 @@ bulk_get_open_doc_revs1(Db, Props, _, {DocId, Revs, Options}) ->
     end.
 
 
-parse_field(<<"id">>, undefined) ->
+parse_id_field(undefined, _Db) ->
     {ok, undefined};
-parse_field(<<"id">>, Value) ->
+parse_id_field(Value, Db) ->
     try
-        ok = couch_doc:validate_docid(Value),
+        ok = validate_docid(Value, couch_db:name(Db)),
         {ok, Value}
     catch
         throw:{Error, Reason} ->
             {error, {Value, Error, Reason}}
-    end;
+    end.
+
+
 parse_field(<<"rev">>, undefined) ->
     {ok, undefined};
 parse_field(<<"rev">>, Value) ->
@@ -1792,6 +1832,11 @@ bulk_get_json_error(DocId, Rev, Error, Reason) ->
                              {<<"rev">>, Rev},
                              {<<"error">>, Error},
                              {<<"reason">>, Reason}]}}]}).
+
+validate_docid(DocId, DbName) ->
+    Partitioned = mem3:is_partitioned(DbName),
+    Options = [{partitioned, Partitioned}],
+    couch_doc:validate_docid(DocId, DbName, Options).
 
 
 -ifdef(TEST).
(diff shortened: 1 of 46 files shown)


garrensmith:
The majority of tests for partitions are currently here https://github.com/cloudant/partition_tests
I wanted to write the tests in elixir so I've kept them in a separate repo until the elixir branch is merged in. Once that is merged in I can move the partition tests into the CouchDB repo

wohali:
@garrensmith would you consider merging the tests into our elixir branch?

garrensmith:
@wohali sure I can do that once this PR is a little further down the line

src/chttpd/src/chttpd_db.erl
```diff
@@ -1592,6 +1676,18 @@ extract_header_rev(Req, ExplicitRev) ->
     end.
 
 
+% cannot partition a system database
+validate_partition_database_create(DbName, Partitioned) ->
+    SystemId = DbName =:= ?l2b(config:get("mem3", "shards_db", "_dbs")) orelse
+        lists:member(DbName, ?SYSTEM_DATABASES),
```
iilyak: This will not work for cases when the database name contain `/`. For example the following would be broken (example is taken from https://github.com/apache/couchdb/pull/1647#issuecomment-427804738 where one of CouchDb users mention his use case).
```
account%2F85%2Fea%2F6075c6c1e266f8512e2233541bdb-201807
```

I think this should use `couch_db:is_systemdb(DbName)` which abstracts that logic (keep in mind that there is an open PR which fixes it a little https://github.com/apache/couchdb/pull/1647).

src/couch/src/couch_doc.erl
```diff
@@ -199,12 +199,30 @@ parse_revs(_) ->
 
 
 validate_docid(DocId, DbName) ->
-    case DbName =:= ?l2b(config:get("mem3", "shards_db", "_dbs")) andalso
-        lists:member(DocId, ?SYSTEM_DATABASES) of
+    validate_docid(DocId, DbName, []).
+
+validate_docid(DocId, DbName, Options) ->
+    SystemId = DbName =:= ?l2b(config:get("mem3", "shards_db", "_dbs")) andalso
+        lists:member(DocId, ?SYSTEM_DATABASES),
```
iilyak: Should we use `couch_db:is_system_db(DbName)` here?

src/chttpd/src/chttpd_db.erl
iilyak: We would need to make `is_systemdb` function public and possibly rename it since we have public `is_system_db` already.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -32,13 +32,20 @@
...
 -export([fold_changes/4]).
 -export([to_key_seq/1]).
 
 -define(MOD, couch_mrview_index).
 -define(GET_VIEW_RETRY_COUNT, 1).
 -define(GET_VIEW_RETRY_DELAY, 50).
+-define(LOWEST_KEY, null).
+-define(HIGHEST_KEY, {<<255, 255, 255, 255>>}).
+-define(PARTITION_START(P), <<P/binary, $:>>).
+-define(PARTITION_END(P), <<P/binary, $;>>).
```
iilyak: You use semicolon here is it intentional or a typo?

src/chttpd/src/chttpd_httpd_handlers.erl
```diff
@@ -43,3 +44,6 @@ design_handler(<<"_update">>)  -> fun chttpd_show:handle_doc_update_req/3;
 design_handler(<<"_info">>)    -> fun chttpd_db:handle_design_info_req/3;
 design_handler(<<"_rewrite">>) -> fun chttpd_rewrite:handle_rewrite_req/3;
 design_handler(_) -> no_match.
+
+partition_design_handler(<<"_view">>) -> fun chttpd_view:handle_partition_view_req/4;
+partition_design_handler(_) -> no_match.
```
iilyak: I didn't test it yet, but I am pretty sure we need `partition_handler(_) -> no_match` here (and export it as well).

src/chttpd/src/chttpd_view.erl
```diff
@@ -101,6 +112,56 @@ handle_temp_view_req(Req, _Db) ->
...
+        chttpd:send_method_not_allowed(Req, "GET").
+
+
+check_partition_restrictions(#mrargs{} = Args) ->
+    Restrictions = [
+        {<<"include_docs">>, Args#mrargs.include_docs, true},
+        {<<"stable">>, Args#mrargs.stable, true},
+        {<<"conflicts">>, Args#mrargs.conflicts, true}
+    ],
+    lists:foreach(fun ({Param, Field, Value}) ->
```
iilyak: The `Field` and `Value` naming is confusing in this context.

src/couch/src/couch_bt_engine.erl
```diff
@@ -323,6 +405,20 @@ set_security(#st{header = Header} = St, NewSecurity) ->
     {ok, increment_update_seq(NewSt)}.
 
 
+set_prop(#st{header = Header} = St, Key, Value) ->
+    OldProps = get_props(St),
+    NewProps = lists:ukeymerge(1, [{Key, Value}], OldProps),
+    Options = [{compression, St#st.compression}],
```
iilyak: We can call `copy_props/2` here. Since the lines are identical.

src/couch/src/couch_bt_engine.erl
```diff
@@ -752,6 +848,16 @@ copy_security(#st{header = Header} = St, SecProps) ->
         needs_commit = true
     }}.
 
+copy_props(#st{header = Header} = St, Props) ->
```
iilyak: We are not really copying props here. So the name of the function is not accurate. Should we call it `write_props` instead?

src/couch/src/couch_bt_engine.erl
```diff
@@ -273,6 +280,56 @@ get_size_info(#st{} = St) ->
...
+            {ok, {Partition, DCAcc + 1, DDCAcc,
+                reduce_sizes(FDI#full_doc_info.sizes, Acc)}};
+        {false, _} ->
+            {ok, {Partition, DCAcc, DDCAcc, Acc}}
+    end.
+
+
+get_partition_info(#st{} = St, Partition) ->
+    StartKey = <<Partition/binary, ":">>,
+    EndKey = <<Partition/binary, ";">>,
```
iilyak: I guess `;` is intentional since I see it second time.

iilyak:
How we can hide this feature behind a feature flag, so we can stabilize it?

src/couch/src/couch_bt_engine.erl
```diff
@@ -283,6 +340,31 @@ get_security(#st{header = Header} = St) ->
...
+    case lists:keyfind(Key, 1, Props) of
+        false -> {error, no_value};
+        {Key, Value} -> {ok, Value}
+    end.
+
+
+get_prop(St, Key, DefaultValue) ->
+    case get_prop(St, Key) of
+        {error, no_value} -> DefaultValue;
+        Value -> Value
```
iilyak: There is a discrepancy in return value here. `get_prop/2` returns `{ok, Value}`. So we either change return value for `get_prop/2` or change `{error, no_value} -> {ok, DefaultValue}`.

src/couch/test/couchdb_views_tests.erl
```diff
@@ -60,6 +60,7 @@ setup_legacy() ->
     lists:foreach(fun(File) -> file:delete(File) end, Files),
 
     % copy old db file into db dir
+    ?debugFmt("copy ~p ~p", [OldDbFilePath, NewDbFilePath]),
```
iilyak: might need to remove prior to merge

src/mango/src/mango_idx.erl
```diff
@@ -119,6 +147,16 @@ validate_new(Idx, Db) ->
     Mod:validate_new(Idx, Db).
 
 
+validate_design_opts(Props) ->
+    case lists:keyfind(<<"options">>, 1, Props) of
+        {<<"options">>, {[{<<"partitioned">>, P}]}}
```
iilyak: We are making a dangerous assumption here that options would contain a single element and it will be `{<<"partitioned">>, P}`.

src/mem3/src/mem3_shards.erl
```diff
@@ -689,6 +705,70 @@ t_cache_insert_ignores_stale_update_and_kills_worker() ->
...
+        Shards = [
+            #ordered_shard{
+                name = <<"shards/80000000-9fffffff/db1.1533630706">>,
+                node = 'node1@127.0.0.1',
+                dbname = <<"db1">>,
+                range = [2147483648,2684354559],
+                ref = undefined,
+                order = 1,
+                opts = [{partitioned,true}]
+            }
```
iilyak: should we add non matching shards into mock result as well?

src/mango/src/mango_cursor_view.erl
```diff
@@ -98,15 +100,31 @@ maybe_replace_max_json([H | T] = EndKey) when is_list(EndKey) ->
...
         include_docs = true,
         extra = [{callback, {?MODULE, view_cb}}, {selector, Selector}]
-    }.
+    },
+
+    DbPartitioned = mem3:is_partitioned(couch_db:name(Cursor#cursor.db)),
+    Partitioned = couch_util:get_value(partitioned, Idx#idx.design_opts, DbPartitioned),
+    Args2 = couch_mrview_util:set_extra(Args1, partitioned, Partitioned),
+    Args3 = case couch_util:get_value(partition, Opts) of
+        <<>> ->
```
iilyak: I think I am missing something. I cannot figure out how we can get `<<>>` here. I'll get back to it latter.

iilyak: How we handle the case when `partition` is not in `Opts`? Is it possible? If so we would call `couch_mrview_util:set_extra(Args2, partition, undefined)` which is not correct.

src/chttpd/src/chttpd_view.erl
garrensmith: I'm going to rename to `ArgValue` and `RestrictedValue`

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -136,6 +136,28 @@ go(DbName, _Options, Workers, QueryArgs, Callback, Acc0) ->
...
     end.
 
+shards(DbName, Args) ->
+    case couch_mrview_util:get_extra(Args, partitioned) of
+        true ->
+            StartKey = partition(Args#mrargs.start_key),
+            EndKey = partition(Args#mrargs.end_key),
+            case {StartKey, EndKey} of
+                {Same, Same} when Same =/= undefined ->
+                    mem3:shards(DbName, <<Same/binary, ":foo">>);
```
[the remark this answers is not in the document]
garrensmith: Yes. We just need a "fake" value to add to the key to make it a full id of the format `Partition:Id`

jaydoane:
@garrensmith I updated to the latest partition_tests commit, but am still getting the same 3 failures from `make couch` and the bevy of rexi timeouts which follow 5 minutes later. Next, I will try to repro on a different dev env.

src/mem3/src/mem3_util.erl
```diff
@@ -34,6 +35,27 @@ hash(Item) when is_binary(Item) ->
...
+docid_hash(DocId) when is_binary(DocId) ->
+    docid_hash(DocId, []).
+
+docid_hash(<<"_design/", _/binary>> = DocId, _Options) ->
+    erlang:crc32(DocId); % design docs are never placed by partition
+
+docid_hash(DocId, []) when is_binary(DocId) ->
+    docid_hash(DocId, [{partitioned, false}]);
+
+docid_hash(DocId, [{partitioned, false}]) when is_binary(DocId) ->
```
chewbranca: I know up the call stack this parameter is referred to as `HashOptions` but this current structure will prevent this from ever actually being an options list given that it's matching on a single element list. So adding another option to `HashOptions` will cause a function_clause error here, so why have it at all? This seems like unnecessary future proofing, especially given that to use this field as an actual options list will require restructuring the function definitions as you can't define a function head to match on _an_ element in a proplist.

I suggest either making this an actual proplist and using a variation of `couch_util:get_value(partioned, Options)`, or eliminating the list entirely and making this a boolean parameter. You could rewrite this series of functions as something like:

```erlang
docid_hash(DocId) ->
  docid_hash(DocId, false).

docid_hash(DocId, false) ->
  erlang:crc32(DocId);
docid_hash(<<"_design/", _/binary>>=DocId, _) ->
  erlang:crc32(DocId);
docid_hash(DocId, true) ->
    case binary:split(DocId, <<":">>) of
        [Partition, _Rest] ->
            erlang:crc32(Partition);
        _ ->
            throw({illegal_docid, <<"doc id must be of form partition:id">>})
    end.
```

src/mem3/src/mem3_util.erl
```diff
@@ -34,6 +35,27 @@ hash(Item) when is_binary(Item) ->
 hash(Item) ->
     erlang:crc32(term_to_binary(Item)).
 
+docid_hash(DocId) when is_binary(DocId) ->
+    docid_hash(DocId, []).
+
+docid_hash(<<"_design/", _/binary>> = DocId, _Options) ->
+    erlang:crc32(DocId); % design docs are never placed by partition
```
chewbranca: Assuming the motivation to not use the `hash/1` function is not wanting to call `term_to_binary` on the DocId, I think it would be wise to avoid redefining the use of `erlang:crc32/1` and we should instead call out to a default hash function that is used ubiquitously. Perhaps something like:

```erlang
hash_raw(Item) ->
  erlang:crc32(Item).

hash(Item) ->
  hash_raw(term_to_binary(Item)).

%% ...

%% example usage
docid_hash(DocId, ...) ->
  hash_raw(DocId).
```

src/couch_mrview/src/couch_mrview_util.erl
rnewson: intentional. it is the next character in order after `:` and therefore marks the end of the partition key range.

src/chttpd/src/chttpd_httpd_handlers.erl
rnewson: we've run our test suite and manual testing without needing that, what fails without it?

src/couch/src/couch_bt_engine.erl
rnewson: This was intentional but I see your point. get_prop/3 will always return a value whereas get_prop/2 might not, which is why the result shape is different. We could change get_prop/2 to return `Value` or `undefined` instead, which matches proplists:get_value behaviour.

src/couch/src/couch_btree.erl
```diff
@@ -133,7 +133,9 @@ make_group_fun(Bt, exact) ->
     end;
 make_group_fun(Bt, GroupLevel) when is_integer(GroupLevel), GroupLevel > 0 ->
     fun
-        ({[_|_] = Key1, _}, {[_|_] = Key2, _}) ->
+        GF({{p, _Partition, Key1}, Val1}, {{p, _Partition, Key2}, Val2}) ->
```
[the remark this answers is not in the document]
rnewson: If the partition for Key1 and Key2 were different, that's a bug and we should crash.

src/mango/src/mango_cursor_view.erl
rnewson: `<<>>` is the default_value in mango_opts:validate_idx_create

src/chttpd/src/chttpd_db.erl
```diff
@@ -1453,6 +1525,18 @@ parse_engine_opt(Req) ->
             end
     end.
 
+
+parse_partitioned_opt(Req) ->
+    case chttpd:qs_value(Req, "partitioned") of
```
smithsz: Can you add a `"false"` case here so we can support `PUT /db?partitioned=false` requests?
Also, it'd be nice to normalise the case using `string:to_lower` (or similar) like we do for other parameters.

wohali:
Please note, we are moving swiftly now into the 2.3.0 release cycle. Until 2.3.x has branched, _**do not merge this PR into `master`**_.

I will update this PR when that has happened. Thanks!

src/chttpd/src/chttpd_db.erl
rnewson: we now handle false but I don't think there's much precedence for case-insensitivity

wohali:
The `2.3.x` branch has now forked.

You are free to merge this to `master` at your convenience.

davisp:
Closed in favor of #1789

PR #1609, opened 2018-09-19
Mango match doc on co-ordinating node

garrensmith:
## Overview

This fixes an issue when doing a rolling upgrade of a CouchDB cluster
and adding commit a6bc72e the nodes that were not upgraded yet would
send through all the docs in the index and those would be passed through
to the user because the co-oridnator would assume it was matched at the
node level. This adds in a check to see if it has been matched at the
node level or not. And then performs a match if required.

## Testing recommendations

This is tricky to test. If you have nodes running an older version of CouchDB, this should stop unmatched docs from being returned to the user.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -405,30 +405,40 @@ doc_member(Cursor, RowProps) ->
     Opts = Cursor#cursor.opts,
     ExecutionStats = Cursor#cursor.execution_stats,
     Selector = Cursor#cursor.selector,
-    Incr = case couch_util:get_value(value, RowProps) of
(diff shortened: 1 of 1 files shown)


tonysun83:
Test scenarios verified:

1) Reproduced error (non-matching documents) on 6 node cluster that had 3 nodes with https://github.com/apache/couchdb/commit/a6bc72e76c56a1befa8675b06016ecda46ef3a2d and 3 nodes with an older version that did not have the commit.

2) Changed the environment 3 nodes with this PR's commit and and older version without this commit. Ran the same set of test suites several times without non-matching documents surfacing.

Eunit tests also pass.

tonysun83:
+1

PR #1616, opened 2018-09-20
Add downgrade function to downgrade database with clustered purge

jiangphcn:
## Overview

The PR is aimed to provide downgrade functionality in case there is need to downgrade database which was modified by builds with clustered purge feature. 

After database is modified by builds with clustered purge feature, the disk version is bumped to 7. Also, there are two newly introduced trees: `PurgeTreeState` and `PurgeSeqTreeState`.

Once downgrade function is called, the disk version is back to the latest version of current build. Also, `purge_seq` and `purged_docs` are re-introduced again. 

PS: current PR is based on commit `28ba48da5877a5d471253c1b587af7e3b3121fd9` which doesn't include clustered purge feature.
## Testing recommendations

```
make check skip_deps+=couch_epi apps=couch tests=downgrade_test_
Couch Bt Engine downgrade tests
  couch_bt_engine_downgrade_tests:64: t_no_downgrade_without_purge_req...[0.088 s] ok
  couch_bt_engine_downgrade_tests:86: t_downgrade_without_purge_req...[0.095 s] ok
  couch_bt_engine_downgrade_tests:108: t_downgrade_with_1_purge_req...[0.084 s] ok
  couch_bt_engine_downgrade_tests:130: t_downgrade_with_2_purge_req...[0.079 s] ok
  [done in 0.358 s]
  All 4 tests passed.

```

https://github.com/apache/couchdb/pull/1370
https://github.com/apache/couchdb/issues/1352

## Checklist


rnewson:
-1.

We need a couchdb release that can recognise a newer version of the database format and downgrade it to the current version. Then a subsequent release will introduce and use the newer version.

This ensures that all users can downgrade by going back to their previous release.

davisp:
@rnewson To be clear, you -1'ing that this isn't in a CouchDB release before clustered purge, not the actual logic as implemented right?

rnewson:
@davisp I would like to see the triggering of the downgrade logic happening in the same spot as all previous upgrade logic triggering (specifically, couch_bt_engine_header:upgrade/1). The actual downgrade logic appears right (and you would know better than I).

Since we have another change after this that also changes db_header and will require up- and downgrade logic, we need a mechanism that extends naturally.

davisp:
@rnewson Definitely agree its probably a good idea to consolidate where up/down-grade logic happens. This is a bit wonky because it requires file modifications so it was pulled out of couch_bt_engine_header where the traditional bits do that work since they were all just about changing a view values in a record.

wohali:
Will hold this until after 2.3.0 releases.

wohali:
Do we still need this @davisp @jiangphcn ?

jiangphcn:
No yet. All clusters in Cloudant are moved to build with clustered purge feature now. We can close this and can re-open if other customer encounters some issue and really need to downgrade.

PR #1620, opened 2018-09-21
1612 fix, js-rewritten PUTs now work

### LINKED ISSUE #1612: JS-rewritten post/put hangs or returns bad_request
## Expected Behavior
JS rewrite, returning `{path:'whatever', method:'PUT', body:'body'}` should send body to the rewritten endpoint.

## Current Behavior
The `body` field of rewrite fn returned obj is in some cases dropped. If source request was POST/PUT, the rewritten requests hangs, or, if source request was GET-like, `bad_request` error is sent as response.

## Steps to Reproduce
1. Create empty bucket named, say, `zz`
2. Put design doc with `rewrites` into the bucket. Below ddoc is intended to create new random doc on each request to `_design/n/_rewrite` endpoint:
````
{
  "_id": "_design/n",
  "rewrites": "function (r) {\n\tvar id = 'x'+(Date.now()+'').substr(-8);\n\treturn {\n\t\tpath:'../../'+id,\n\t\tmethod:'PUT',\n\t\theaders:{'content-type':'application/json',Accept:'application/json'},\n\t\tbody:JSON.stringify({_id:id,type:'test',val:Date.now().toString(36)})\n\t}\n}"
}
```` 
3. Call `http://couchdb_ip/zz/_design/n/_rewrite`.

## Solution
The issue is caused by cleaning-up rewritten request prematurely here:
https://github.com/apache/couchdb/blob/master/src/chttpd/src/chttpd_rewrite.erl#L67. 
It should only be cleaned if we have no body to send.

That line (L67) should be replaced with this code:
````
            Body = case couch_util:get_value(<<"body">>, Props) of
                undefined -> erlang:get(mochiweb_request_body);
                B -> B
            end,
            case Body of
                undefined -> NewMochiReq:cleanup();
                _ -> erlang:put(mochiweb_request_body, Body)
            end,
````
It seems to fix issue completely.

## Your Environment
* Version used: 2.1.2, 2.2.0

ermouth:
# Overview

Fixes #1612, sends body when `body` field is present in rewrite function result object.

## Testing recommendations

Example ddoc from original issue works as expected with this patch.

## Checklist


wohali:
Hi @ermouth , next release of CouchDB will be 2.3.0, can you re-do this PR against `master` please? Thanks.

wohali:
@ermouth  Perfect.

Can I get a +1 from someone familiar with this code, e.g. @janl ?

ermouth:
@wohali since no +1 so far, may be I could answer questions, if any? Since js rewrites was initially my patch refactored by @kxepal, I doubt there’s a lot of devs knowing internals.

kxepal:
Oh, btw, do we have any tests for js rewrites? Would be nice to cover that case for future changes.

ermouth:
@kxepal Thanks 😁

> Though I forgot reasons to do erlang:get(mochiweb_request_body)

Because rewrites should be fast, as fast as possible. So we better not stream body into newly born request (which is also hackish btw), but replace body using internal Mochiweb thread dict entry. 

Perf is also the only reason why `.info` and `.form` fields are removed from request object here.

wohali:
@ermouth if tests are possible, that'd be great, otherwise this might break again (esp. if others are not familiar with the code).

Otherwise I'll rebase this later tonight.

@kxepal nice to see you, and thanks!

ermouth:
@wohali it’s tested now, but not by CouchDB tests. 

There are some tests in CouchDB as well, https://github.com/apache/couchdb/blob/master/test/javascript/tests/rewrite_js.js, but they definitely not cover some cases. 

If above link is correct, I could expand coverage, but one question: how to run those tests locally?

Also, when do you plan 2.3?

wohali:
@ermouth you asked when 2.3.0 releases - when this list is done:

https://github.com/apache/couchdb/milestone/4

Feel free to fix one or more if you have time.

After that's fixed and the CI runs are green, I update the release notes and start the release process. :)

ermouth:
@wohali I added small test, but see failed CI. What should I do with this? Seems CI fails on smth completely unrelated to rewrites.

wohali:
@ermouth thanks, I'll just update the branch to master's latest and re-run them. Test suite can be a bit touchy sometimes, it's an endless battle to keep it green all the time.

wohali:
@ermouth Thanks for the patch!!

PR #1622, opened 2018-09-24
Explicit Python version in scripts

adrienverge:
Recent Linux distributions start defaulting to Python 3, and require
ambiguous scripts to be more explicit.
For example building for Fedora 30 (not released yet) fails with:

    ERROR: ambiguous python shebang in /opt/couchdb/bin/couchup:
    #!/usr/bin/env python. Change it to python3 (or python2) explicitly.

So this commit changes the four Python scripts to use `python2`.

Note: They seem to be Python-3-compatible, but I couldn't be sure. If
you know they are, please tell me, I'll change it to `python3`.

### DIFF
--- a/build-aux/logfile-uploader.py
+++ b/build-aux/logfile-uploader.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not
 # use this file except in compliance with the License. You may obtain a copy of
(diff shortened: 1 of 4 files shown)


wohali:
@adrienverge I'm having to revert the change to `dev/run` for now because it has broken our build. See #1673 for the details.

Once @nickva 's PR lands, `dev/run` will have an explicit callout for `python3` instead of `python`.

adrienverge:
@wohali thanks for letting me know!

I'm glad to ear that Python 3 support is on its way :+1:

PR #1629, opened 2018-09-27
handle db deletion in load validation funs

iilyak:
## Overview

Handle db deletion in couch_db:load_validation_funs

Previously there were quite a few problems with load_validation_funs
in the case when clustered database is deleted.

- the calls to load_validation_funs were failing with `internal_server` error [1]
- the deleted database stayed opened because:
  - the caller of the load_validation_funs (update_doc) stayed alive
  - the main_pid of the deleted database wasn't killed either
- there was an infinite loop in ddoc_cache_entry trying to recover ddoc from deleted database

The solution is:
- do not call `recover` for deleted database
- close `main_pid`
- use `erlang:error` to crash the caller

[1] - The stack trace was:
```
{database_does_not_exist,[
    {mem3_shards,load_shards_from_db,"bailey/meta",[
        {file,"src/mem3_shards.erl"},{line,394}]},
    {mem3_shards,load_shards_from_disk,1,[
        {file,"src/mem3_shards.erl"},{line,369}]},
    {mem3_shards,for_db,2,[
        {file,"src/mem3_shards.erl"},{line,54}]},
    {fabric_view_all_docs,go,5,[
        {file,"src/fabric_view_all_docs.erl"},{line,24}]},
    {ddoc_cache_entry_validation_funs,recover,1,[
        {file,"src/ddoc_cache_entry_validation_funs.erl"},{line,33}]},
    {ddoc_cache_entry,do_open,1,[
        {file,"src/ddoc_cache_entry.erl"},{line,294}]}]}
```

## Testing recommendations

```
make eunit apps=couch tests=clustered_db_test
make eunit apps=ddoc_cache
```

N/A

## Checklist

### DIFF
--- a/src/ddoc_cache/test/ddoc_cache_open_test.erl
+++ b/src/ddoc_cache/test/ddoc_cache_open_test.erl
@@ -1,0 +1,107 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(ddoc_cache_open_test).
+
+-export([
+    dbname/1,
+    ddocid/1,
+    recover/1,
+    insert/2
+]).
+
+-include_lib("couch/include/couch_db.hrl").
+-include_lib("eunit/include/eunit.hrl").
+-include("ddoc_cache_test.hrl").
+
+
+%% behaviour callbacks
+dbname(DbName) ->
+    DbName.
+
+
+ddocid(_) ->
+    no_ddocid.
+
+
+recover({deleted, _DbName}) ->
+    erlang:error(database_does_not_exist);
+recover(DbName) ->
+    ddoc_cache_entry_validation_funs:recover(DbName).
+
+
+insert(_, _) ->
+    ok.
+
+
+start_couch() ->
+    Ctx = ddoc_cache_tutil:start_couch(),
+    meck:new(ddoc_cache_entry_validation_funs, [passthrough]),
+    meck:expect(ddoc_cache_entry_validation_funs, recover,
+        ['_'], meck:passthrough()),
+    Ctx.
+
+
+stop_couch(Ctx) ->
+    meck:unload(),
+    ddoc_cache_tutil:stop_couch(Ctx).
+
+
+check_open_error_test_() ->
+   {
+       setup,
+       fun start_couch/0,
+       fun stop_couch/1,
+       ddoc_cache_tutil:with([
+           {"should_return_database_does_not_exist",
+               fun should_return_database_does_not_exist/1},
+           {"should_not_call_recover_when_database_does_not_exist",
+               fun should_not_call_recover_when_database_does_not_exist/1},
+           {"should_call_recover_when_needed",
+               fun should_call_recover_when_needed/1},
+           {"should_call_recover_when_needed",
+               fun should_not_crash_lru_process/1}
+       ])
+    }.
+
+
+should_return_database_does_not_exist({DbName, _}) ->
+    ?assertError(
+        database_does_not_exist,
+        ddoc_cache_lru:open({?MODULE, {deleted, DbName}})).
+
+
+should_not_call_recover_when_database_does_not_exist({DbName, _}) ->
+    meck:reset(ddoc_cache_entry_validation_funs),
+    ?assertError(
+       database_does_not_exist,
+       ddoc_cache_lru:open({?MODULE, {deleted, DbName}})),
+    ?assertError(
+        timeout,
+        meck:wait(1, ddoc_cache_entry_validation_funs, recover, '_', 100)).
+
+
+should_call_recover_when_needed({DbName, _}) ->
+    meck:reset(ddoc_cache_entry_validation_funs),
+    ddoc_cache_lru:open({?MODULE, DbName}),
+    ?assertEqual(
+        ok,
+        meck:wait(1, ddoc_cache_entry_validation_funs, recover, '_', 500)).
+
+
+should_not_crash_lru_process({DbName, _}) ->
+    LRUPid = whereis(ddoc_cache_lru),
+    ?assert(is_process_alive(LRUPid)),
+    ?assertError(
+        database_does_not_exist,
+        ddoc_cache_lru:open({?MODULE, {deleted, DbName}})),
+    ?assert(is_process_alive(LRUPid)).
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -872,6 +872,9 @@ load_validation_funs(#db{main_pid=Pid, name = <<"shards/", _/binary>>}=Db) ->
         {'DOWN', Ref, _, _, {ok, Funs}} ->
             gen_server:cast(Pid, {load_validation_funs, Funs}),
             Funs;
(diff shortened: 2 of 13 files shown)


jaydoane:
Comparing
```
$ make eunit skip_tests+=couch_epep apps=couch suites=couchdb_db_tests
module 'couchdb_db_tests'
  Checking clustered db API
    DB deletion
      couchdb_db_tests:53: should_close_deleted_db...[0.029 s] ok
      couchdb_db_tests:68: should_kill_caller_from_load_validation_funs_for_deleted_db...[0.033 s] ok
  2 tests passed.
```
with
```
$ make eunit skip_tests+=couch_epep apps=ddoc_cache suites=ddoc_cache_open_test
module 'ddoc_cache_open_test'
  ddoc_cache_open_test: check_open_error_test_...ok
  ddoc_cache_open_test: check_open_error_test_...[0.012 s] ok
  ddoc_cache_open_test: check_open_error_test_...[0.015 s] ok
  ddoc_cache_open_test: check_open_error_test_...ok
  All 4 tests passed.
```
I know it requires more boiler plate to accomplish, but I much prefer to see the actual names of the tests being run with the couchdb_db_tests module. Just seeing 4 identical lines of `ddoc_cache_open_test: check_open_error_test_` isn't as useful, IMO.

src/ddoc_cache/test/ddoc_cache_open_test.erl
```diff
@@ -0,0 +1,98 @@
...
+        ddoc_cache_lru:open({?MODULE, {deleted, DbName}})).
+
+should_not_call_recover_when_database_does_not_exist({DbName, _}) ->
+    meck:reset(ddoc_cache_entry_validation_funs),
+    ?assertError(
+       database_does_not_exist,
+       ddoc_cache_lru:open({?MODULE, {deleted, DbName}})),
+    ?assertError(
+        timeout,
+        meck:wait(ddoc_cache_entry_validation_funs, recover, '_', 500)).
```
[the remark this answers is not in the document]
iilyak: Good catch. Apparently the `meck:wait` function has two slightly different semantics depending on arity:
- `meck:wait(foo, '_', '_', 1000)` - wait any number of calls for 1 second
- `meck:wait(5, foo, '_', '_', 1000),` - wait for 5 calls and return (within 1 second)

I.e. to avoid waiting we would need to change it to wait for 1 call.

src/ddoc_cache/test/ddoc_cache_open_test.erl
```diff
@@ -0,0 +1,98 @@
...
+        ddoc_cache_lru:open({?MODULE, {deleted, DbName}})).
+
+should_not_call_recover_when_database_does_not_exist({DbName, _}) ->
+    meck:reset(ddoc_cache_entry_validation_funs),
+    ?assertError(
+       database_does_not_exist,
+       ddoc_cache_lru:open({?MODULE, {deleted, DbName}})),
+    ?assertError(
+        timeout,
+        meck:wait(1, ddoc_cache_entry_validation_funs, recover, '_', 500)).
```
jaydoane: @iilyak I'm still seeing this test always take > 500ms:
```
$ make eunit skip_tests+=couch_epep apps=ddoc_cache suites=ddoc_cache_open_test
module 'ddoc_cache_open_test'
  ddoc_cache_open_test: check_open_error_test_...[0.002 s] ok
  ddoc_cache_open_test: check_open_error_test_...[0.501 s] ok
  ddoc_cache_open_test: check_open_error_test_...[0.015 s] ok
  ddoc_cache_open_test: check_open_error_test_...ok
  [done in 0.530 s]
  All 4 tests passed.
```
Do we really need to wait this long for that timeout?

iilyak: I don't understand how it can be related to `meck`. Here is a result of my investigation:
- `meck:wait/5` is dispatched to `meck:wait/6` [see here](https://github.com/eproxus/meck/blob/master/src/meck.erl#L570)
- `meck:wait/6` call is dispatched to `meck_proc:wait/6` [see here](https://github.com/eproxus/meck/blob/master/src/meck.erl#L593)
- `meck_proc:wait/6` does a `gen_server` call [here](https://github.com/eproxus/meck/blob/master/src/meck_proc.erl#L161)
- the `handle_call` for wait message has following shape 
  ```
    case times_called(OptFunc, ArgsMatcher, OptCallerPid, History) of
        CalledSoFar when CalledSoFar >= Times ->
             {reply, ok, S};
  ```
- the `times_called/4` function is pretty simple and just [iterate through `History` term](https://github.com/eproxus/meck/blob/master/src/meck_proc.erl#L603:L612)  

The only explanation I have is it takes time to process `ddoc_cache_lru:open` and detect the call on `meck` side. 

I don't like reducing the timeout since there is a high chance of turning the test into a flaky one. Because shorter timeout could be a problem for slower CI systems.

I'll try to reduce the timeout locally to see if it has any effect on the run time of the test suite. If it does it would mean that I don't read `meck` code correctly. If it doesn't not it would mean that the test is slow.

iilyak:
> Comparing
> 
> ```
> $ make eunit skip_tests+=couch_epep apps=couch suites=couchdb_db_tests
> module 'couchdb_db_tests'
>   Checking clustered db API
>     DB deletion
>       couchdb_db_tests:53: should_close_deleted_db...[0.029 s] ok
>       couchdb_db_tests:68: should_kill_caller_from_load_validation_funs_for_deleted_db...[0.033 s] ok
> =======================================================
>   2 tests passed.
> ```
> 
> with
> 
> ```
> $ make eunit skip_tests+=couch_epep apps=ddoc_cache suites=ddoc_cache_open_test
> module 'ddoc_cache_open_test'
>   ddoc_cache_open_test: check_open_error_test_...ok
>   ddoc_cache_open_test: check_open_error_test_...[0.012 s] ok
>   ddoc_cache_open_test: check_open_error_test_...[0.015 s] ok
>   ddoc_cache_open_test: check_open_error_test_...ok
> =======================================================
>   All 4 tests passed.
> ```
> 
> I know it requires more boiler plate to accomplish, but I much prefer to see the actual names of the tests being run with the couchdb_db_tests module. Just seeing 4 identical lines of `ddoc_cache_open_test: check_open_error_test_` isn't as useful, IMO.

I don't know how to do it with `with`. All tests in ddoc_cache application rely on combination of `ddoc_cache_tutil:start_couch` and `{setup, _, _, {with, Tests}}` construct. The `{with, Tests}` construct is implemented via shortcut in https://github.com/erlang/otp/blob/master/lib/eunit/src/eunit_data.erl#L294:L295. The author didn't take into account that the `{with, X, Tests}` construction rely on the [`eunit_lib:fun_parent(A)`](https://github.com/erlang/otp/blob/master/lib/eunit/src/eunit_lib.erl#L379:L390). The reason the hack doesn't get correct name of the test is because they are wrapping the function one more time on dispatch. 
```
parse({setup, P, S, C, {with, As}}) when is_list(As) ->
     parse({setup, P, S, C, fun (X) -> {with, X, As} end});
```

I am trying few things to avoid rewriting of all tests for the whole application.

src/ddoc_cache/test/ddoc_cache_open_test.erl
iilyak: JFYI: It looks like there is a way.

iilyak: I ended up reducing the timeout value

jaydoane:
New changes look good:
```
$ make eunit apps=couch tests=clustered_db_test
  DB deletion
    couchdb_db_tests:53: should_close_deleted_db...[0.028 s] ok
    couchdb_db_tests:75: should_kill_caller_from_load_validation_funs_for_deleted_db...[0.032 s] ok
  2 tests passed.
```
```
$ make eunit apps=ddoc_cache
...
  All 51 tests passed.
```

Just a couple minor quibbles:
- The second line of the commit says "Previously there were few problems with load_validation_funs". I think you mean "*a* few problems", since omitting the article "a" makes it sound like there were not that many problems. In short, it inverts the meaning.
- Some functions in the new test modules don't have 2 lines between them in places.

PR #1642, opened 2018-10-04
Set io_priority for couch_index pids

iilyak:
## Overview

In production we noticed that `io_priority` is not set for view update and compaction. This PR adds appropriate `io_priority` to the processes involved.
 
## Testing recommendations

```
make eunit apps=couch tests=couch_index_ioq_priority_test_
```

N/A

## Checklist

### DIFF
--- a/src/couch/src/couch_debug.erl
+++ b/src/couch/src/couch_debug.erl
@@ -234,19 +234,17 @@ opened_files_contains(FileNameFragment) ->
         string:str(Path, FileNameFragment) > 0
     end, couch_debug:opened_files()).
 
+
 process_name(Pid) when is_pid(Pid) ->
-    case process_info(Pid, registered_name) of
-        {registered_name, Name} ->
+    Info = process_info(Pid, [registered_name, dictionary, initial_call]),
+    case Info of
+        undefined ->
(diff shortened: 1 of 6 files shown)


AlexanderKaraberov:
Hi @iilyak 

I wonder whether this change may result in better performance of index I/O operations (like view index updates)? 
If I understood correctly this `io_priority` flag is for [ioq supervisor](https://github.com/apache/couchdb-ioq) which prioritises I/O operations inside of CouchDB.
Or this change is pure informative/debugging one?

iilyak:
Hi @AlexanderKaraberov 

As you noticed already `io_priority` is used by `ioq` ([here](https://github.com/apache/couchdb-ioq/blob/master/src/ioq.erl#L110:L115)) to select appropriate requests queue. The bug fixed in this PR makes sure that `view_compact` requests go to the right queue. Database admin can influence which requests would have greater priority [see here](https://github.com/apache/couchdb-ioq/blob/master/src/ioq.erl#L139:L144).

wohali:
@iilyak would love to get this into 2.3.0 but it looks like your tests are failing...

iilyak:
Unfortunately due to changes in tracing mechanism in OTP this PR has to wait for deprecation of old erlang versions.

PR #1655, opened 2018-10-15
Fix ets_lru configuration in chttpd application

iilyak:
## Overview

The code was incorrect in a sense that it was using is_integer guard,
while `config:get` cannot return integer.

## Testing recommendations

1. Update default.ini to set max_lifetime to 10000
2. Restart chttpd application
3. Try manually insert keys into lru_cache and query them. They all should be removed after 10 seconds
```
lists:foreach(fun(Idx) -> ets_lru:insert(chttpd_auth_cache_lru, Idx, [1,2,3]) end, lists:seq(1, 1000)).
lists:filter(fun(Idx) -> ets_lru:lookup(chttpd_auth_cache_lru, Idx) /= not_found end, lists:seq(1, 1000)).
```
4. Try negative value for the setting check logs to make sure it complains
5. Try putting "some_string" as value check logs to make sure it complains

N/A

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_sup.erl
+++ b/src/chttpd/src/chttpd_sup.erl
@@ -80,21 +80,18 @@ maybe_replace(Key, Value, Settings) ->
     end.
 
 lru_opts() ->
-    case config:get("chttpd_auth_cache", "max_objects") of
-        MxObjs when is_integer(MxObjs), MxObjs > 0 ->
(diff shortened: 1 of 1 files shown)


rnewson:
the simpler version is to use config:get_integer

iilyak:
@rnewson Updated to use config:get_integer

PR #1656, opened 2018-10-16
Partition review fixes

garrensmith:
## Overview

Fixes to the partition work based off testing and reviews

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_view.erl
+++ b/src/chttpd/src/chttpd_view.erl
@@ -110,6 +110,25 @@ handle_temp_view_req(Req, _Db) ->
     chttpd:send_error(Req, 410, gone, Msg).
 
 
+handle_partition_view_req(#httpd{method='POST',
+        path_parts=[_, _, _, _, _, _, ViewName]} = Req, Db, DDoc, Partition) ->
+    chttpd:validate_ctype(Req, "application/json"),
+    Props = couch_httpd:json_body_obj(Req),
+    Keys = couch_mrview_util:get_view_keys(Props),
+    case Keys of
+        Keys when is_list(Keys) ->
+            couch_stats:increment_counter([couchdb, httpd, view_reads]),
+            Args0 = couch_mrview_http:parse_params(Req, Keys),
+            Args1 = couch_mrview_util:set_extra(Args0, partition, Partition),
+            Args2 = couch_mrview_util:set_extra(Args1, partitioned, true),
+            design_doc_view_int(Req, Db, DDoc, ViewName, Args2);
+        _ ->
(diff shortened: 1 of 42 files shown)


src/chttpd/src/chttpd_view.erl
```diff
@@ -151,8 +151,8 @@ check_partition_restrictions(#mrargs{} = Args) ->
         {<<"stable">>, Args#mrargs.stable, true},
         {<<"conflicts">>, Args#mrargs.conflicts, true}
     ],
-    lists:foreach(fun ({Param, Field, Value}) ->
-        case Field =:= Value of
+    lists:foreach(fun ({Param, ArgValue, RestrictedValue}) ->
+        case ArgValue =:= RestrictedValue of
             true ->
                 Msg = [<<"`">>, Param, <<"=true` is not supported in this view.">>],
```
rnewson: I'm confused here. The RestrictedValue would need to always be `true` for this error message to make sense?

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -148,16 +154,16 @@ shards(DbName, Args) ->
...
-partition(undefined) ->
-    undefined;
-partition(null) ->
-    null;
+
 partition(Key) when is_binary(Key) ->
-    hd(binary:split(Key, <<":">>)).
+    hd(binary:split(Key, <<":">>));
+partition(_) ->
+    undefined.
```
rnewson: let's be explicit here, I assume only `undefined` and `null` are likely to be passed, anything else is a bug we want to detect.

src/chttpd/src/chttpd_view.erl
```diff
@@ -147,12 +147,12 @@ handle_partition_view_req(Req, _Db, _DDoc, _Pk) ->
...
-        {<<"stable">>, Args#mrargs.stable, true},
-        {<<"conflicts">>, Args#mrargs.conflicts, true}
+        {<<"include_docs">>, Args#mrargs.include_docs},
+        {<<"stable">>, Args#mrargs.stable},
+        {<<"conflicts">>, Args#mrargs.conflicts}
     ],
-    lists:foreach(fun ({Param, Field, Value}) ->
-        case Field =:= Value of
+    lists:foreach(fun ({Param, ArgValue}) ->
+        case ArgValue =:= true of
```
rnewson: just say `case ArgValue of` since we match `true` below, and change the `false` below to `_`. Either that or use two clauses and match in the head.

garrensmith:
Closing this since we have another PR coming for this.

PR #1669, opened 2018-10-19
Enforce upper bound on partition size

rnewson:
Add a configurable upper bound to partition size (only applies to partitioned databases).


garrensmith:
Closing since this isn't needed anymore.

PR #1670, opened 2018-10-20
couchjs: show default runtime SIZE limit on help message

iblis17:


wohali:
Thank you for your contribution to CouchDB, @iblis17 !

PR #1676, opened 2018-10-23
Downgrade code for Database partitions

garrensmith:
## Overview

Code to run a downgrade on a non partitioned database after a cluster has been upgraded to use the new partition code. Currently we cannot downgrade a database that has been created with partitions as the documents will not be where CouchDB would expect. But with this code it would be possible to downgrade a non-partitioned database.

## Testing recommendations

* Run the unit tests. 
* Create databases with the partitioned feature #1605 
* Then switch to this branch. The databases should load.

#1605 

## Checklist


src/couch/src/couch_bt_engine.erl
```diff
@@ -857,6 +868,59 @@ init_state(FilePath, Fd, Header0, Options) ->
...
+        {id_tree_state, IdTreeState},
+        {seq_tree_state, SeqTreeState},
+        {local_tree_state, LocalTreeState},
+        {purge_tree_state, PurgeTreeState},
+        {purge_seq_tree_state, PurgeSeqTreeState},
+        {security_ptr, SecurityPtr},
+        {revs_limit, RevsLimit},
+        {uuid, Uuid},
+        {epochs, Epochs},
+        {compacted_seq, CompactedSeq}
```
jiangphcn: The value of `purge_infos_limit` can be changed and may not be default. So there is need to write current value to database header as well.

src/couch/src/couch_bt_engine_header.erl
```diff
@@ -210,6 +219,50 @@ indexes() ->
     lists:zip(Fields, Indexes).
 
 
+downgrade_partition_header(Header) ->
```
davisp: Why is there a duplicate of this in couch_bt_engine?

src/couch/src/couch_bt_engine.erl
```diff
@@ -112,6 +112,7 @@
 -include("couch_bt_engine.hrl").
 
 
+
```
davisp: Minor nit but remove this whitespace only hunk.

Issue #892, opened 2017-10-17
Building CouchDB on arm64-v8a


DQvsRA:
I'm trying to run CouchDB on my Odroid C2 (arm64)
After "./configure" I run "make release" but it gives me an error 

## Current Behavior
![image](https://user-images.githubusercontent.com/442111/31650424-8ea31262-b317-11e7-89ca-d1d8b916e04e.png)

## Your Environment
Linux odroid64 3.14.79-116 #1 SMP PREEMPT Tue Sep 26 01:19:06 BRT 2017 aarch64 aarch64 aarch64 GNU/Linux

kxepal:
Seems like the [`-m64` is not supported](https://bugzilla.redhat.com/show_bug.cgi?id=713685) by ARM build toolchain. Try to remove it or replace with analogue. As @wohali [saild](https://github.com/apache/couchdb/issues/885#issuecomment-335514010) we don't have ARM's in our infra to test and support builds for them. 

However, any contributions to fix that are welcome! Happy hacking!

koleto:
Just manage to comile on Linaro ARMv8.

1. Open the file ~/apache-couchdb-2.1.0$ vim src/rebar/src/rebar_port_compiler.erl
2. Remove the '-m64' flag from Line 608 and 609. The output should be:
`     %% Linux specific flags for multiarch
     {"linux.*-64$", "CFLAGS", "$CFLAGS"},
     {"linux.*-64$", "CXXFLAGS", "$CXXFLAGS"},
     {"linux.*-64$", "LDFLAGS", "$LDFLAGS"},` 

3. Recompile rebar:   cd src/rebar
`./bootstrap`

4. Rebuild couchdb: 
`make release`

janl:
That means this needs to be fixed in rebar, not CouchDB. cc @davisp any opinions?

wohali:
If it's a rebar problem, we should look at the rebar3 change using paul's [enc](https://github.com/davisp/erlang-native-compiler) and make this patch over there.

koleto:
But unfortunately after I was able to build couchdb, there are still some issues with the replication and views on ARMv8.

wohali:
Sounds like JavaScript didn't build correctly. You will need `libmozjs185` specifically, and it MUST be built 32-bit -- see https://bugzilla.mozilla.org/show_bug.cgi?id=589735 and https://bugzilla.mozilla.org/show_bug.cgi?id=577056 for the gory details.

wohali:
[...]

Issue #1428, opened 2018-07-09
Migrate to rebar3


wohali:
We're long past due upgrading to[rebar 3 (latest)](https://github.com/erlang/rebar3).

This requires using @davisp's [erlang-native-compiler](https://github.com/davisp/erlang-native-compiler) to continue to build our NIFs.

tsloughter:
You might find it easier to use https://github.com/blt/port_compiler/ instead of having to call out to the external `enc` escript just to build NIFs.

davisp:
The issues I had trying to use port_compiler is precisely what motivated me to write enc in the first place. I haven't paid any attention at all to changes to port_compiler in the meantime but that's precisely because I've not had any issues reported with enc in the nearly two years since I last touched it.

I'm not against changing back to port_compiler if someone wants to take a stab at it.

nickva:
I started on this a while back: 

https://github.com/cloudant/couchdb/commit/b0f3feefa8c1725b178641c024a06cf7b1defb2a

Besides the port compiler issue the other one was using raw dependencies (Fauxton and docs). `rebar3` doesn't do those by default (or didn't do when I checked): https://github.com/erlang/rebar3/issues/110 There might be a plugin to handle those. The other option could be to make a "fake" Erlang OTP directory structure in those dependencies.

tsloughter:
Yea, there is a raw resource plugin that can be used.

iilyak:
I think we should consider trying out `mix` (Elixir build tool) instead of `rebar3`. Although we would have same compatibility issues as with rebar3 (NIFs and raw deps). Nevertheless I find mix implementation faster and easier to hack on. I am  happy to go with any direction the community would decide to pursue.

garrensmith:
+1 to using mix


________________________________

eiri:
I'd +1 for trying `mix` too. I don't have an opinion on it itself, but I personally don't find `rebar3` a significant improvement over "classical" rebar.

tsloughter:
:( 

I don't work on couchdb so my opinion doesn't matter here, but I am always interested in ways to improve rebar3, so please pass along (probably not on this issue unless it somehow directly relates to the question of couchdb migration) any suggestions or complaints (particularly how it doesn't do better than rebar2 which was one of the main goals and why it was essentially a rewrite). 

rebar3 is fairly different from mix, so personal preference for one or the other over particular decisions we made differently is an understandable separation that can't necessarily be bridged, but always interested in hearing about them.

[...]

PR #1678, opened 2018-10-24
Use mainline Rebar (v2)

lag-linaro:
This PR uses the latest software provided by the upstream Rebar (v2) repo.

The most significant change allows this version of Rebar to work with compilers which do not support the -m64 flag, as reported in https://github.com/apache/couchdb/issues/892 amongst others.

The latest code base also contains @davisp's `Properly skip apps with a .app.src.script file` patch with CouchDB has been carrying for a while now.

Signed-off-by: Lee Jones <lee.jones@linaro.org>

### DIFF
--- a/configure
+++ b/configure
@@ -209,7 +209,7 @@ EOF
 install_local_rebar() {
     if [ ! -x "${rootdir}/bin/rebar" ]; then
         if [ ! -d "${rootdir}/src/rebar" ]; then
-            git clone --depth 1 --branch 2.6.0-couchdb https://github.com/apache/couchdb-rebar.git ${rootdir}/src/rebar
+            git clone --depth 1 --branch master https://github.com/rebar/rebar.git ${rootdir}/src/rebar
         fi
         make -C ${rootdir}/src/rebar
         mv ${rootdir}/src/rebar/rebar ${rootdir}/bin/rebar


wohali:
Unfortunately, we cannot accept this PR, as none of our dependencies can be on repositories outside of the Apache organization directly.

You can try a PR against the apache/couchdb-rebar repository if you want...

Sorry about this.

wohali:
Also, we have a strong preference to migrate the build process to rebar 3 - #1428. If you are able to help with this conversion process, we'll be especially grateful.

lag-linaro:
> Unfortunately, we cannot accept this PR, as none of our dependencies can be on repositories outside of the Apache organization directly.
> 
> You can try a PR against the apache/couchdb-rebar repository if you want...
> 
> Sorry about this.

That's okay.  I will update CouchDB's Rebar repo instead.

lag-linaro:
> Also, we have a strong preference to migrate the build process to rebar 3 - #1428. If you are able to help with this conversion process, we'll be especially grateful.

Yes, I am aware of that, and I'm happy to help.  I would still like to see Rebar v2 updated in the mean time please.

I did briefly investigate moving to Rebar3, but I didn't find the code to update.

How does one compile native code with @davisp's offering?  Where is the NIF code?

PR #1682, opened 2018-10-25
Feature flags

iilyak:
## Overview

Once in a while we merge a highly experimental features. It is beneficial to be able to have the codepath for these features effectivelly disabled by default until the implementation is stabilized. 
This PR introduces an ability to add feature flag checks for experimental features.   
This feature is for developers only and wouldn't need documentation.

### The feature configuration

```
[feature_flags]

my_feature||* = false
my_feature||mytestdatabaseprefix* = true
```

Notes:
- The `*` is only allowed at the end of the path
- In order to distinguish between shards and clustered db the following convention is used.
  - it is a shard if pattern starts with `/`
- It is also possible to enable multiple flags for the path using alternative syntax
  - `mytestdatabaseprefix* = [my_featureA, my_featureB]`

### Usage of the feature in the code

```
case couch_flags:is_enabled(my_feature, Subject) of
    true -> ...;
    false -> ...
end
```

Where Subject is one of `#db{}` | `#httpd{)` | `#shard{}` | `#ordered_shard{}` | `ShardName :: binary()` | `ShardName :: list()` | `DbName :: binary()` | `DbName :: list()`

### Performance

There is a naive performance test in the test suite it does 1 million of calls to `is_enabled/2`. One million of calls takes "1.884 s" on the laptop.

### Security

It is not safe to allow configuration of the feature flags over http since there might be a possibility of injecting arbitrary code via name of the configuration key within `feature_flags` section. Therefore the http access is disabled. 

## Testing recommendations

```
make eunit apps=couch suites=couch_flags_tests,couch_flags_config
```

N/A

## Checklist

### DIFF
--- a/src/couch/src/couch_db_epi.erl
+++ b/src/couch/src/couch_db_epi.erl
@@ -35,14 +35,15 @@ providers() ->
 
 services() ->
     [
-        {couch_db, couch_db_plugin}
+        {couch_db, couch_db_plugin},
+        {feature_flags, couch_flags}
     ].
 
 data_subscriptions() ->
     [].
 
 data_providers() ->
-    [].
+    [couch_flags_config:data_provider()].
 
 processes() ->
     [].
--- a/src/couch/src/couch_flags.erl
+++ b/src/couch/src/couch_flags.erl
@@ -1,0 +1,135 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+% http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+% This module serves two functions
+% - provides public API to use to get value for a given feature flag and subject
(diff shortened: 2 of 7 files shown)


nickva:
FYI: There is already a similar config-based "features" feature implementation: 

https://github.com/apache/couchdb-config/commit/f62d553b337ce975edb0fb68772d22bdd3bf6490

That's used by various components to indicated their presence and then those features are bubbled up to the top level API. 

https://github.com/apache/couchdb/blob/master/src/chttpd/src/chttpd_misc.erl#L53

iilyak:
> @nickva FYI: There is already a similar config-based "features" feature implementation:

There are few important differences:
1. config-based "features" are really supported capabilities. People use welcome message to check if new API is available on the cluster. While feature flags are totally internal things. The flags suppose to be deleted once feature is stable and officially released. 
2. config-based "features" are global. It is impossible to enable it for some databases and disable for others.

nickva:
Ah you're right. It was mostly a capabilities thing.  I just wanted to call it out in the PR so it's visible. As the old thing has the (confusing) "features" name.

Maybe it would be good to replace the old thing with this?  The "config" capabilities thing was a quick adhock thing, this seems better designed.

garrensmith:
@nickva just checking, are you happy with this PR? Are you ok with us merging it?

nickva:
@garrensmith I didn't review the PR contents in detail so as far as that I am +0 so if you reviewed and looks good, go ahead and merge.

garrensmith:
I think IIya wants to add one more thing before merging.

PR #1718, opened 2018-11-07
Perform explicit $exists:false on Index validation

garrensmith:
## Overview

When selecting an index for a mango query, the mango index selector can incorrectly choose an index when a selector has an exists: false for a field in the index along with another condition for that same field. E.g:

Index with fields = ["tags] and selector:

``` js
Selector: {
    $or: [
      {
        tags: {
          $elemMatch: {
            $eq: "tag1"
          }
        }
      },
      {
        tags: { $exists: False }
      }
    ]
  }
};
```

In the above case, mango will select the index instead of falling back to `_all_docs`.

## Testing recommendations

* Create similar queries like the above and make sure the index is not selected. 
* All Mango tests should pass.

## Checklist


davisp:
It seems like rather than adding a second pass we should just modify this clause:

https://github.com/apache/couchdb/blob/master/src/mango/src/mango_selector.erl#L636

There we can just do a lists:member tests and if the exists:false field is present we just return false from that test and reject the index.

davisp:
You could also start with a second empty list and when one is found we move it from one to the other.

garrensmith:
Yes. Nice idea. I’ll change it to that.

src/mango/src/mango_selector.erl
```diff
@@ -608,35 +607,42 @@ has_required_fields_int([{[{<<"$and">>, Args}]}], RequiredFields)
...
+    RequiredFields1 = case {Cond, lists:keymember(Field, 1, RequiredFields)} of
         % $exists:false is a special case - this is the only operator
         % that explicitly does not require a field to exist
-        {[{<<"$exists">>, false}]} ->
-            has_required_fields_int(Rest, RequiredFields);
-        _ ->
-            has_required_fields_int(Rest, lists:delete(Field, RequiredFields))
-    end.
+        {{[{<<"$exists">>, false}]}, true} ->
+            % case lists:keymember(Field, 1, RequiredFields) of
```
tonysun83: let's remove these extraneous comments

src/mango/src/mango_selector.erl
```diff
@@ -667,6 +673,53 @@ is_constant_field([{[{_UnMatched, _}]} | Rest], Field) ->
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 
+exists_false_basic_test() ->
+    RequiredFields = [{<<"A">>, false}],
+    Selector = normalize({[{<<"A">>,{[{<<"$exists">>,false}]}}]}),
```
tonysun83: just going to be a bit nitpicky here and say be consistent with space behind the comma. this goes for the rest of the testcases and the normalized selectors

tonysun83:
+1 after some of the formatting issues.

garrensmith:
Thanks @tonysun83 I've fixed the formatting issues

src/mango/src/mango_selector.erl
```diff
@@ -573,27 +573,26 @@ match({[_, _ | _] = _Props} = Sel, _Value, _Cmp) ->
...
+% If it is, set the field to true and continue
+% until run out of selector to
 % match against.
 
-has_required_fields(Selector, RequiredFields) ->
-    Remainder = has_required_fields_int(Selector, RequiredFields),
+has_required_fields(Selector, Fields) ->
+    RequiredFields = [{Field, false} || Field <- Fields],
+    RequiredFields1 = has_required_fields_int(Selector, RequiredFields),
+    Remainder = [Result || {_Field, Result} <- RequiredFields1, Result =:= false],
```
davisp: Style nit, but this is a bit subtle in that the intent isn't immediately obvious. `lists:all/2` seems like a better fit here.

```suggestion
    lists:all(fun({_, Status}) -> Status end, RequiredFields1).
```

src/mango/src/mango_selector.erl
```diff
@@ -608,35 +607,39 @@ has_required_fields_int([{[{<<"$and">>, Args}]}], RequiredFields)
...
-    case Cond of
+    RequiredFields1 = case {Cond, lists:keymember(Field, 1, RequiredFields)} of
         % $exists:false is a special case - this is the only operator
-        % that explicitly does not require a field to exist
-        {[{<<"$exists">>, false}]} ->
-            has_required_fields_int(Rest, RequiredFields);
-        _ ->
-            has_required_fields_int(Rest, lists:delete(Field, RequiredFields))
-    end.
+        % that explicitly does not require a field to exist and so if a field in
```
davisp: This comment is a bit awkward. How about something like:

```erlang
% $exists:false is a special case. During indexing, any document that
% does not have the field is not indexed, thus any index using this field
% is incompatible with $exists:false
```

src/mango/src/mango_selector.erl
```diff
@@ -608,35 +607,39 @@ has_required_fields_int([{[{<<"$and">>, Args}]}], RequiredFields)
...
+        % the index has a $exists: false in the selector for it. We cannot use this index
+        {{[{<<"$exists">>, false}]}, true} ->
+            [{exists, false}];
```
davisp: This seems like a super subtle hack that relies on exists being an atom so that it is never matched by the rest of the selector. And assuming I'm understanding that correctly I'm rather not fond of that approach. What I'd suggest instead would be to throw here instead which is then caught in `has_requred_fields/2` where it'd return false.

Something here along the lines of `throw(exists_false_mismatch)` and then catch that and just return false.

src/mango/src/mango_selector.erl
```diff
@@ -24,7 +24,6 @@
 -include_lib("couch/include/couch_db.hrl").
 -include("mango.hrl").
 
-
```
tonysun83: whitespace

tonysun83:
+1 again

wohali:
@garrensmith Any plans to get this in? I am preparing to mass-close old PRs that never got merged.

garrensmith:
Thanks, You can close it

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #1719, opened 2018-11-07
Don't use $not to validate use of index

garrensmith:
## Overview

When mango is selecting an index for a query, previously it would consider a field with a $not to be 
acceptable to validate a field and increase the chance of that index being selected. That isn't quite correct. It is better to ignore a field and condition that contains a $not. It isn't proof enough that the index is usable for the query.

## Testing recommendations

All tests should pass. Queries using $not should use _all_docs unless more selectors in the query help select an index.

## Checklist

### DIFF
--- a/src/mango/test/03-operator-test.py
+++ b/src/mango/test/03-operator-test.py
@@ -251,6 +251,16 @@ class OperatorTests:
             })
         user_ids = [12,13,14]
         self.assertUserIds(user_ids, docs)
+
+    def test_ne_returns_correct_docs(self):
+        selector = {
+            "twitter": {
+                "$ne": "@nonahorton"
+            }
+        }
+        docs = self.db.find(selector, fields=["user_id"])
+        ex = self.db.find(selector, fields=["user_id"], explain=True)
+        self.assertUserIds([0,1,9,13], docs)
         
 
 
--- a/src/mango/src/mango_selector.erl
+++ b/src/mango/src/mango_selector.erl
@@ -628,6 +628,9 @@ has_required_fields_int([{[{<<"$and">>, Args}]} | Rest], RequiredFields)
     Remainder = has_required_fields_int(Args, RequiredFields),
     has_required_fields_int(Rest, Remainder);
 
+has_required_fields_int([{[{_Field, {[{<<"$not">>, _Args}]}}]} | Rest], RequiredFields) ->
+    has_required_fields_int(Rest, RequiredFields);
+
 has_required_fields_int([{[{Field, Cond}]} | Rest], RequiredFields) ->
     case Cond of
         % $exists:false is a special case - this is the only operator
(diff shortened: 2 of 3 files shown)


willholley:
I'd like to see a test that compares the results of a $not selector with and without an index, since that was the issue that triggered investigation into this. Adding the index should not change the result set. Otherwise, the solution looks good to me.

tonysun83:
@willholley @garrensmith :  
I'm trying to wrap my head around NOT vs NE and if we're consistent with the results.

The fix for `$not` makes sense to me.

So `$ne` implies that the field exists. I think that's why @garrensmith's current test case with
`"somefield": { "$ne": "somevalue"}}` uses an index.

But what if we did:
`"somefield": { "$ne": { "$elemMatch": {"$regex": "expression"}}}` ?

This means the field should exist, and not equal to certain documents. It's just a bigger
set than `"somefield": { "$ne": "somevalue"}}`

 However, IIRC, $regex is a combination operator and we need to use _all_docs. So do
we choose an index or just _all_docs in this case? From the current code I think
we choose an index right?

willholley:
@tonysun83 I think the (admittedly subtle) difference is whether the field is specified before the operator or not. For instance:

```
{
    "$not": {
          "foo": "bar"
    }
}
```

does not require the `"foo"` field to exist in matches but

```
{
    "foo": {
          "$not": {
              ...
         }
    }
}
```

does. Since `"$ne"` can only be expressed in the latter form, I think the behaviour is correct with respect to index selection.

tonysun83:
If we can see a couple tests as Will suggested, then that would be great! Otherwise looks good.

garrensmith:
@willholley and @tonysun83 I've added a test to compare text, all_docs and json index.

src/mango/test/03-operator-test.py
```diff
@@ -251,6 +251,16 @@ def test_gte_respsects_unicode_collation(self):
...
+
+    def test_ne_returns_correct_docs(self):
+        selector = {
+            "twitter": {
+                "$ne": "@nonahorton"
+            }
+        }
+        docs = self.db.find(selector, fields=["user_id"])
+        ex = self.db.find(selector, fields=["user_id"], explain=True)
+        self.assertUserIds([0,1,9,13], docs)
```
willholley: We probably want to exercise `$not` here (even though it may be normalized to `$ne` under the hood). For example:
```
# gotcha here is that this implicitly adds { $exists: true } for the referenced field,
# so is not strictly the inverse of { "ordered":"A" }
def test_not_returns_expected_docs(self):
        selector = {
            "$not": {
                "ordered": "A"
            }
        }
        docs = self.db.find(selector, fields=["user_id"])
        ex = self.db.find(selector, fields=["user_id"], explain=True)

        # returns all docs that have an "ordered" field that is not equal to "A"
        self.assertUserIds([6, 7, 8, 9, 10, 11, 12, 14], docs)
```

as mentioned in the test, there is an underlying oddity/bug in Mango where we differ from MongoDB behaviour for $not and $ne operators. [MongoDB docs](https://docs.mongodb.com/manual/reference/operator/query/not/) state:
> $not performs a logical NOT operation on the specified <operator-expression> and selects the documents that do not match the <operator-expression>. This includes documents that do not contain the field.

Mango will currently only return documents that contain the field; to return documents that do not contain the field, the user needs to add an explicit `$exists: false` to the selector. I think this is incorrect but it's a tricky thing to change at this point; perhaps @davisp can remember why it is this way?

PR #1726, opened 2018-11-10
Support elixir 1.7

jaydoane:
## Overview

Reformat mix.lock, and suppress elixir 1.7 warnings.

## Testing recommendations

Using elixir 1.7, `mix test --trace` should not reformat the `mix.lock` file, and should not generate warnings like:
```
warning: found quoted keyword "_id" but the quotes are not required. Note that keywords are always atoms, even when quoted, and quotes should only be used to introduce keywords with foreign characters in them
  test/test_helper.exs:178
```

## Checklist


wohali:
Can you confirm this doesn't break support for 1.6.6? We currently have 1.6.6 in our regression testing environment and plan to start running `make elixir` as part of our test suite soon.

The regression environment is Erlang 19 so I don't even know if we can upgrade to 1.7 with that or not...

jaydoane:
@wohali, yes, I also ran the tests with elixir 1.6.6, and they all pass except for this one test (which also fails on 1.7.4):
```
  * test Standard config options are present (2.0ms)

  1) test Standard config options are present (ConfigTest)
     test/config_test.exs:70
     Expected truthy, got nil
     code: assert context[:config]["daemons"]["httpd"]
     stacktrace:
       test/config_test.exs:72: (test)
```
I hope to have a fix for that test shortly.

As for erlang/elixir version compatibility, there's this [nice table](https://hexdocs.pm/elixir/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp) which indicates elixir 1.7 should be compatible with OTP 19.

PR #1747, opened 2018-11-19
Set query server env variable in Makefile for isolated eunit tests

eiri:
## Overview

After #1602 we requite env variable `COUCHDB_QUERY_SERVER_JAVASCRIPT` to point to js query server. This patch adds the env var on Makefile's `eunit` target returning ability to run the isolated eunit tests without setting that var in each tests' setup.

## Testing recommendations

Isolated tests with query server, e.g. `make eunit apps=couch_mrview suites=couch_mrview_all_docs_tests` should run locally without throwing "unknown query server" exceptions. 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,7 @@ check: all
 # target: eunit - Run EUnit tests, use EUNIT_OPTS to provide custom options
 eunit: export BUILDDIR = $(shell pwd)
 eunit: export ERL_AFLAGS = -config $(shell pwd)/rel/files/eunit.config
+eunit: export COUCHDB_QUERY_SERVER_JAVASCRIPT = $(shell pwd)/bin/couchjs $(shell pwd)/share/server/main.js
 eunit: couch
 	@$(REBAR) setup_eunit 2> /dev/null
 	@$(REBAR) -r eunit $(EUNIT_OPTS)


Makefile
```diff
@@ -148,6 +148,7 @@ check: all
 # target: eunit - Run EUnit tests, use EUNIT_OPTS to provide custom options
 eunit: export BUILDDIR = $(shell pwd)
 eunit: export ERL_AFLAGS = -config $(shell pwd)/rel/files/eunit.config
+eunit: export COUCHDB_QUERY_SERVER_JAVASCRIPT = $(shell pwd)/bin/couchjs $(shell pwd)/share/server/main.js
```
iilyak: I think we can use `$BUILDDIR` instead of calling `$(shell pwd)` several times.

eiri: @iilyak I guess we can, but it's inconsistent with usage `pwd` everywhere else and feels kind of flimsy, setting env var in one step then depending on it in another.

Why do you want this change? It's not like here is a performance penalty we care about.

eiri: > It looked strange to see export BUILDIR and $(shell pwd) two lines bellow.

and one line above too 😄 

Well, I agree it's not the best way to write Makefiles and better way to go about it would be something like `ROOTDIR := $(shell pwd)` to expand current dir at top of `Makefile` and then just use make's var `${ROOTDIR}` everywhere else. But this is a bit bigger change for a quick fix I intended to do, so since you don't have strong feelings about it I'll go ahead with how it is.

PR #1754, opened 2018-11-22
Do not use [] in feature_flags configuration

iilyak:
## Overview

[] characters are fobidden in configuration keys for security reason.
Use list without [] to configure flags. Example configuration:

[feature_flags]
foo,bar,baz||test* = true

## Testing recommendations

```
%% run new eunit tests
make eunit apps=couch tests=parse_flags_term_test_
%% make sure the flag is not enabled 
false = couch_flags:is_enabled(foo, "anything").
%% make sure flags are not set
[] = couch_flags:enabled("anything").
%% set flag 
config:set("feature_flags", "foo,bar||*", "true", false).
%% wait 2 sec
timer:sleep(2000).
%% check that flag is enabled
true = couch_flags:is_enabled(foo, "anything").
%% check list of enabled flags
[bar,foo] = couch_flags:enabled("anything").
```

fixes  https://github.com/apache/couchdb/pull/1682

## Checklist

### DIFF
--- a/src/couch/src/couch_flags_config.erl
+++ b/src/couch/src/couch_flags_config.erl
@@ -21,6 +21,7 @@
 ]).
 
 -define(DATA_INTERVAL, 1000).
+-define(MAX_FLAG_NAME_LENGTH, 256).
 
 -type pattern()
     :: binary(). %% non empty binary which optionally can end with *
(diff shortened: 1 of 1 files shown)


src/couch/src/couch_flags_config.erl
```diff
@@ -120,16 +121,29 @@ parse_flags(_Tokens, _) ->
...
+split_by_comma(Binary) ->
+    case binary:split(Binary, <<",">>, [global]) of
+       [<<>>] -> [];
+       Tokens -> Tokens
+    end.
+
+parse_flag(FlagName) when size(FlagName) > ?MAX_FLAG_NAME_LENGTH ->
+	{too_long, FlagName};
+parse_flag(FlagName) ->
+	list_to_atom(string:strip(binary_to_list(FlagName))).
```
eiri: How about changing this into `list_to_existing_atom/1`, just to be super safe? If I understand correctly a flag should be defined by this time, so that should work. It can even act as a poor man's validator, if we wrap it in something like `try ... catch _:_ -> {invalid_flag, FlagName}`

eiri:
Looks good to me, just have a single suggestion. And thank you for listening to my concerns on using `parse_term` on an arbitrary config line!

PR #1755, opened 2018-11-22
feat: do not record server admin requests as cache misses

### LINKED ISSUE #1746: ini-file admin users always cause auth cache miss
When accessing CouchDB with an admin user defined ONLY in the ini file, the auth_cache_misses counter is incremented.

## Expected Behavior
The counter should not increment.

## Current Behavior
The counter does increment.

## Possible Solution
auth_cache should have a filter of ini admins before logging the stat.

A workaround is to create users in _users with identical credentials.

1. dev/run -n 1 -a user:pass
2. curl http://user:pass@localhost:15984/_node/_local/_stats | jq .couchdb.auth_cache_misses
3. Repeat and see the counter go up.

## Context
Dashboards tracking auth cache misses will see their own activity reflected in the graph. AUGH!

janl:
Closes #1746

### DIFF
--- a/src/chttpd/src/chttpd_auth_cache.erl
+++ b/src/chttpd/src/chttpd_auth_cache.erl
@@ -73,8 +73,7 @@ get_from_cache(UserName) ->
             couch_log:debug("cache hit for ~s", [UserName]),
             Props;
         _ ->
-            couch_stats:increment_counter([couchdb, auth_cache_misses]),
-            couch_log:debug("cache miss for ~s", [UserName]),
+            maybe_increment_auth_cache_miss(UserName),
             case load_user_from_db(UserName) of
                 nil ->
                     nil;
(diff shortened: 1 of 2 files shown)


src/chttpd/src/chttpd_auth_cache.erl
```diff
@@ -84,11 +83,20 @@ get_from_cache(UserName) ->
...
         error:badarg ->
-            couch_stats:increment_counter([couchdb, auth_cache_misses]),
-            couch_log:debug("cache miss for ~s", [UserName]),
+            maybe_increment_auth_cache_miss(UserName),
             load_user_from_db(UserName)
     end.
 
+maybe_increment_auth_cache_miss(UserName) ->
+    Admins = config:get("admins"),
+    case proplists:get_value(?b2l(UserName), Admins) of
```
eiri: If all we want is to check if user is in admin list `lists:member/2` works better here. It's faster too.

janl: @eiri good observation, however, `config:get(section)` returns a proplist like `[{Name, PassHash}],{Name2,Pass2Hash}]`, so it’d have to be at least `lists:keymember()`, but I’m not sure it is faster.

janl: Looks like it is: https://stackoverflow.com/a/30812279

eiri: It is, actually, `lists` operations are `nifs` and `proplists` are pure erlang module. Plus `keymember` still better semantically.

janl: fixed, h/t

PR #1760, opened 2018-11-23
test: port coffee.js to Elixir test suite

dottorblaster:
## Overview

I just ported `coffee.js` to the Elixir test suite.

## Testing recommendations

Issue `make elixir`

None 

## Checklist

### DIFF
--- a/test/elixir/test/coffee_test.exs
+++ b/test/elixir/test/coffee_test.exs
@@ -1,0 +1,68 @@
+defmodule CoffeeTest do
+  use CouchTestCase
+
+  @moduletag :coffee
+
+  @moduledoc """
+  Test basic coffeescript functionality.
+  This is a port of the coffee.js test suite.
+  """
+
+  @tag :with_db
+  test "CoffeeScript basic functionality", context do
+    db_name = context[:db_name]
+
+    docs = [
+      %{:_id => "a", :foo => 100},
+      %{:foo => 1},
+      %{:foo => 1},
+      %{:foo => 2},
+      %{:foo => 2},
+      %{:bar => 1},
+      %{:bar => 1},
+      %{:bar => 2},
+      %{:bar => 2}
+    ]
+
+    resp = Couch.post("/#{db_name}/_bulk_docs", body: %{docs: docs})
+
+    design_doc = %{
+      :_id => "_design/coffee",
+      :language => "coffeescript",
+      :views => %{
+        :myview => %{
+          :map => "(doc) -> if doc.foo\n  emit(doc.foo, 1)",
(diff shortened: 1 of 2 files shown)


test/elixir/test/coffee_test.exs
```diff
@@ -0,0 +1,82 @@
+defmodule CoffeeTest do
+  use CouchTestCase
+
+  @moduletag :coffee
+  @headers ["X-Couch-Full-Commit": "false"]
```
jaydoane: Not sure if we need this header any more. I ran this test with that header removed, and it also passed.

test/elixir/test/coffee_test.exs
```diff
@@ -0,0 +1,82 @@
...
+        "/#{db_name}/_bulk_docs",
+        body: %{docs: docs},
+        headers: @headers
+      )
+
+    assert resp.status_code === 201 and length(resp.body) === length(docs)
+
+    # Giving time to the database to complete sync, otherwise values
+    # would be messed up and the test would be so much flaky
+    :timer.sleep(100)
```
jaydoane: Using a fixed sleep time is still very fragile (since CI machines can be much slower than your dev env), and I would recommend using a more robust synchronization technique. You could do something like:
```elixir
  retry_until(fn ->
    %{"rows" => values} =
      Couch.get("/#{db_name}/_design/coffee/_view/myview",
        headers: @headers
      ).body
     assert 5 === hd(values)["value"]  
  end)
```

dottorblaster: I think we would have to discuss about an official way to check collection consistency right after an update 😬 this looks fine, I'll try. Thanks for pointing this out!

dottorblaster:
@jaydoane I think I came out with a cleverer approach in the end. Since we don't care about that specific execution order and we are testing the coffeescript query language here, now the test plays with documents first, then it receives the design document that builds the index. This way we work around the eventual consistency for the purpose of this test and we don't need any `retry_until` logic.

Thanks for pointing out that `retry_until` can be used anyway, I think it'll be useful for future ports! 👍 And thanks for the review man 😎

test/elixir/test/coffee_test.exs
dottorblaster: With the new approach, the need for this header became absolute zero. I removed it indeed. Thanks!

wohali:
@dottorblaster Keeping in mind #1767, please re-run `make elixir` on this PR after rebasing on `master` and ensure that your new test conforms to the coding standard.

Once I have a +1 from you I'll merge this PR. Thanks!

dottorblaster:
@wohali done. Thanks!

PR #1766, opened 2018-11-27
Add Global timeout for map/reduce queries

garrensmith:
## Overview

This makes the global timeout for a map/reduce and all_docs request
configurable via the config. It separates the config into global queries
and partition queries so that it is possible to make the global timeout
is for partitioned queries

## Testing recommendations

Add the `global_partition_view_timeout` to the `[fabric]` config. You can then make queries and see if they timeout based on the value set.

## Checklist

### DIFF
--- a/src/fabric/src/fabric_util.erl
+++ b/src/fabric/src/fabric_util.erl
@@ -15,7 +15,7 @@
 -export([submit_jobs/3, submit_jobs/4, cleanup/1, recv/4, get_db/1, get_db/2, error_info/1,
         update_counter/3, remove_ancestors/2, create_monitors/1, kv/2,
         remove_down_workers/2, doc_id_and_rev/1]).
--export([request_timeout/0, attachments_timeout/0, all_docs_timeout/0]).
+-export([request_timeout/0, attachments_timeout/0, all_docs_timeout/0, view_timeout/1]).
 -export([log_timeout/2, remove_done_workers/2]).
 -export([is_users_db/1, is_replicator_db/1]).
 -export([open_cluster_db/1, open_cluster_db/2]).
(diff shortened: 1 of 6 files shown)


davisp:
Looks good but obviously as currently written should wait till after we finish partition queries.

src/fabric/src/fabric_util.erl
```diff
@@ -152,6 +152,13 @@ all_docs_timeout() ->
 attachments_timeout() ->
     timeout("attachments", "600000").
 
+global_view_timeout(Args) ->
+    PartitionQuery = couch_mrview_util:get_extra(Args, partition, false),
+    case PartitionQuery of
+        false -> timeout("global_view_timeout", "infinity");
```
davisp: You're doubling up the _timeout suffix cause the timeout/2 call appends that as well. Also global_view_timeout should probably be shortened to just view_timeout because you're looking at global vs partitioned. Also global_partition is a bit of a misnomer and should just be partition_view_timeout.

Also I think the SOP is to add commented out config options with their default values in default.ini. If the other timeout config options aren't already present I'd add all five settings in a separate commit on this PR.

garrensmith: I've removed the `global_` part for the names. The reason I added that is that rexi has two timeouts. A timeout for how long to wait for each response from a node and then an overall global timeout for the node to send all responses. The view and partition_view are for the second situation where as the `all_docs` and `request` timeouts are fo the individual requests.

davisp:
Whoops. No idea how I managed to close this. Definitely wasn't on purpose.

src/fabric/src/fabric_util.erl
```diff
@@ -152,6 +152,13 @@ all_docs_timeout() ->
 attachments_timeout() ->
     timeout("attachments", "600000").
 
+global_view_timeout(Args) ->
```
davisp: Still seems odd to be calling this global_view_timeout when it applies to both global and partitioned.

garrensmith: I've changed the name

davisp:
+1

Issue #1756, opened 2018-11-22
Adopt Credo inside Elixir test suite


dottorblaster:
The Elixir test suite gets slowly broader and broader, and I think it could be a good idea to adopt Credo as the linter and static code analysis tool of choice. What do you think about it? 

Ref.: https://github.com/rrrene/credo

dottorblaster:
Wow, blazing fast @wohali! 🙂

jaydoane:
Informal +1 from me. I think it's a great idea!

It was trivial to add to mix.exs, to run via `mix credo -a`, and it found a bunch of stylistic problems:
```
$ mix credo -a
Checking 17 source files ...

  Software Design
┃
┃ [D] → Found a TODO tag in a comment: # TODO: port sever_port tests from
┃       config.js
┃       test/config_test.exs:66 #(ConfigTest.delete_config)
┃ [D] → Found a TODO tag in a comment: # TODO: In JS we re-fetch _changes with
┃       since_seq, is that
┃       test/replication_test.exs:573 #(ReplicationTest.run_since_seq_repl)
┃ [D] → Found a TODO tag in a comment: # TODO: Parameterize these
┃       test/replication_test.exs:9 #(ReplicationTest)
┃ [D] → Found a TODO tag in a comment: # TODO: switch this to _local when that's
┃       landed
┃       test/config_test.exs:12 #(ConfigTest)
┃ [D] → Found a TODO tag in a comment: # TODO: port remainder of
┃       security_validation.js suite
┃       test/security_validation_test.exs:185 #(SecurityValidationTest)
┃ [D] → Found a TODO tag in a comment: # TODO: port _list function tests and
┃       everything below in rewrite.js
┃       test/rewrite_test.exs:328 #(RewriteTest)
┃ [D] → Found a TODO tag in a comment: # TODO: port the two "bugged" tests from
┃       rewrite.js
┃       test/rewrite_test.exs:318 #(RewriteTest)
┃ [D] → Found a TODO tag in a comment: # TODO: make protocol check use defined
┃       protocol value
┃       test/basics_test.exs:204 #(BasicsTest)
┃ [D] → Found a TODO tag in a comment: # TODO: enable chunked encoding
┃       test/basics_test.exs:117 #(BasicsTest)
┃ [D] → Found a TODO tag in a comment: # TODO: do we need to bring this in?
┃       test/all_docs_test.exs:11 #(AllDocsTest)

  Code Readability
┃
[...]

PR #1767, opened 2018-11-27
Check code format before running elixir test suite

van-mronov:
## Overview

- Format elixir tests with `mix format`;
- Add make task checking code format before running elixir tests.

## Testing recommendations

`make elixir`

## Related Issues or Pull Requests

Issue #1756
PR #1745 
PR #1760 

## Checklist

- [ ] Code is written and works correctly;
- [ ] Changes are covered by tests;
- [ ] Documentation reflects the changes;

PR #1769, opened 2018-11-28
Add Credo to Elixir test suite

dottorblaster:
## Overview

This PR adds Credo as the static code analysis tool of choice for the Elixir test suite.

## Testing recommendations

Just issue `make elixir-credo` first, then `make elixir`

#1756 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,12 +194,18 @@ python-black-update: .venv/bin/black
 		. dev/run rel/overlay/bin/couchup test/javascript/run
 
 .PHONY: elixir
-elixir: elixir-check-formatted devclean
+elixir: elixir-check-formatted elixir-credo devclean
 	@dev/run -a adm:pass --no-eval test/elixir/run
 
 .PHONY: elixir-check-formatted
 elixir-check-formatted:
 	@cd test/elixir/ && mix format --check-formatted
+
+# Credo is a static code analysis tool for Elixir.
+# We use it in our tests
+.PHONY: elixir-credo
+elixir-credo:
+	@cd test/elixir/ && mix deps.get && mix credo
 
 .PHONY: javascript
 # target: javascript - Run JavaScript test suites or specific ones defined by suites option
--- a/test/elixir/.credo.exs
+++ b/test/elixir/.credo.exs
@@ -1,0 +1,153 @@
+# This file contains the configuration for Credo and you are probably reading
+# this after creating it with `mix credo.gen.config`.
+#
+# If you find anything wrong or unclear in this file, please report an
+# issue on GitHub: https://github.com/rrrene/credo/issues
+#
+%{
+  #
+  # You can have as many configs as you like in the `configs:` field.
+  configs: [
+    %{
+      #
+      # Run any exec using `mix credo -C <name>`. If no exec name is given
+      # "default" is used.
+      #
+      name: "default",
+      #
+      # These are the files included in the analysis:
+      files: %{
+        #
+        # You can give explicit globs or simply directories.
+        # In the latter case `**/*.{ex,exs}` will be used.
+        #
+        included: ["lib/", "src/", "test/", "web/", "apps/"],
+        excluded: [~r"/_build/", ~r"/deps/", ~r"/node_modules/"]
+      },
+      #
+      # If you create your own checks, you must specify the source files for
+      # them here, so they can be loaded by Credo before running the analysis.
+      #
+      requires: [],
+      #
+      # If you want to enforce a style guide and need a more traditional linting
+      # experience, you can change `strict` to `true` below:
+      #
+      strict: false,
+      #
+      # If you want to use uncolored output by default, you can change `color`
+      # to `false` below:
+      #
+      color: true,
+      #
+      # You can customize the parameters of any check by adding a second element
+      # to the tuple.
+      #
+      # To disable a check put `false` as second element:
+      #
+      #     {Credo.Check.Design.DuplicatedCode, false}
+      #
+      checks: [
+        #
+        ## Consistency Checks
+        #
+        {Credo.Check.Consistency.ExceptionNames, []},
+        {Credo.Check.Consistency.LineEndings, []},
+        {Credo.Check.Consistency.ParameterPatternMatching, false},
+        {Credo.Check.Consistency.SpaceAroundOperators, []},
+        {Credo.Check.Consistency.SpaceInParentheses, []},
+        {Credo.Check.Consistency.TabsOrSpaces, []},
+
+        #
+        ## Design Checks
+        #
+        # You can customize the priority of any check
+        # Priority values are: `low, normal, high, higher`
+        #
+        {Credo.Check.Design.AliasUsage,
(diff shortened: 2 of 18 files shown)


dottorblaster:
Here is a rationale of this PR.

I added Credo to the Mix project, then I disabled some "not so useful" checks. They are:

- `Credo.Check.Refactor.CyclomaticComplexity`
- `Credo.Check.Refactor.LongQuoteBlocks`
- `Credo.Check.Refactor.Nesting`

Then, I added Credo to the Makefile and Makefile.win with a proper goal, and I added that goal to `make elixir`. This way, `make elixir` formats the code according to #1767, then executes the `elixir-credo` goal erroring on basically everything.

The only check I maintained but I changed its exit code to `0` is `Credo.Check.Design.TagTODO`, because I want to be warned about TODOs and monitor them, but I still want to be able to use them as reminders.

@wohali I think your eyes could be useful here 🙂

Makefile
```diff
@@ -172,14 +172,18 @@ soak-eunit: couch
...
-elixir: elixir-check-formatted
+elixir: elixir-check-formatted elixir-credo
 	@rm -rf dev/lib
 	@dev/run -a adm:pass --no-eval test/elixir/run
 
 .PHONY: elixir-check-formatted
 elixir-check-formatted:
 	@cd test/elixir/ && mix format --check-formatted
 
+.PHONY: elixir-credo
```
wohali: Can you add a comment in the two Makefiles that explains what `credo` actually is? For people unfamiliar with Elixir it may not be clear that this is a linting/formatting tool.

test/elixir/test/test_helper.exs
```diff
@@ -1,3 +1,5 @@
+# credo:disable-for-this-file Credo.Check.Consistency.ParameterPatternMatching
```
wohali: Why wouldn't we disable this globally, or fix the problem in this one file? Keep in mind I am new to Elixir, but not to linting and formatting...exceptions in a single file are usually bad news.

jaydoane:
I have a general question about max line length. [This commit](https://github.com/apache/couchdb/commit/df37da6662be6cc62e9c866eee4544000f5329c4) explicitly sets the max line length to 120, up from its default of 96. For Erlang code, the current consensus (and also @davisp's [emelio](https://github.com/cloudant-labs/emilio/blob/master/priv/emilio.config#L23-L24)) appears to be 80, for which there is [ample precedent](https://softwareengineering.stackexchange.com/a/222998). Because it forces you to write more concise code, and it’s great being able to have two code windows open side by side (on a 15" laptop), I hope we also adopt 80 for Elixir code. But I'm curious to what others think it should be.

wohali:
@jaydoane also, as to your comment on "two code windows open side by side (on a 15" laptop)" is highly variable. I use a 15" laptop and often get 3 or 4 code windows open side by side, with >80 characters each. Resolution and font size (and your own visual acuity!) enter into it a lot :)

test/elixir/.credo.exs
```diff
@@ -0,0 +1,153 @@
...
+        {Credo.Check.Design.TagTODO, [exit_status: 0]},
+        {Credo.Check.Design.TagFIXME, []},
+
+        #
+        ## Readability Checks
+        #
+        {Credo.Check.Readability.AliasOrder, []},
+        {Credo.Check.Readability.FunctionNames, []},
+        {Credo.Check.Readability.LargeNumbers, []},
+        {Credo.Check.Readability.MaxLineLength, [priority: :low, max_length: 120]},
```
wohali: Please change this to 90 - see the discussion thread on the main PR.

wohali:
@jaydoane Hrm, I don't mind it...

Makefile
```diff
@@ -172,14 +172,18 @@ soak-eunit: couch
...
+elixir-credo:
+	@cd test/elixir/ && mix deps.get && mix credo
+
```
wohali: How about adding a

wohali: A concern I have is that credo is formatting the code. It shouldn't do a format - it should be doing a check, and failing if the code doesn't pass the check. This is in keeping with the philosophy that a `make check` run is a pass/fail thing, it shouldn't have side effects.

A separate target that actually does the change would be useful.

See apache/couchdb#1776 for what I did on a similar Python PR (inspired by you here!)

dottorblaster: @wohali Credo doesn't format code at all, and I know we are using `mix format`'s check for that 🙂

dottorblaster:
Some comments on the above discussion:

- Wow, you were great guys, that was such a beautiful debate 😬 
- I enforced max line length to `120` because our formatter enforces the coding style to 120-chars lines. I'm gonna take it to `90` in Credo, and of course I'm updating the `.formatter.exs` as well.
- Also, I love the 120 limit, and the community does. But if we want to enforce a 90 columns limit, that's the beautiful point of having a tool that you can configure 😹 
- About disabling `Credo.Check.Consistency.ParameterPatternMatching` and the long quotes one: `Consistency.ParameterPatternMatching` is ok, but I think: 1) that we need a long quote block for our test helper 2) that long quote blocks are smells and we should limit that smell to a known place.

Enforcing a 90 columns format will cause almost all of our tests to be reformatted. @wohali et al. do you want me to go further or do we want to do that in another subsequent PR?

davisp:
Reformatting all the lines that require it to pass a `make check` invocation is totally fine. Just keep it to an isolated commit and then make sure that `make check` fails after the fact so that we're not re-introducing breaking things after the fact.

dottorblaster:
@davisp thanks! I'm going further then

van-mronov:
Hi all!

Not long ago we also added checking elixir code format with `mix format` before running elixir test suite. Since line length is more code formatting issue rather than some logic error, like not used local var or something like that, I believe it's better to address this to `mix format`.

Here is formatter [options](https://hexdocs.pm/elixir/Code.html#format_string!/2), `:line_length` exactly what we need. We just need to add this in our `.formatter.exs` wuth reasonable value. In that case `make elixir` if new changes have incorrect line length. And developer just need to run `mix format` to reformat his changes.

In that case `mix credo` have to address only logic code errors.

van-mronov:
@dottorblaster sorry, looks like you are going to add `line_length` in formatter.

dottorblaster:
@van-mronov yeah `line_length` is already inside the `.formatter.exs` and I'm gonna enforce that to `90` as I said before, no problem

dottorblaster:
@wohali @davisp @jaydoane @van-mronov I added two more commits for that, gonna edit the Makefile soon, can you provide me feedbacks?

@van-mronov I still prefer to maintain max length checks in Credo too because it's more developer friendly. I enforced the value formatter-side as well.

test/elixir/test/test_helper.exs
wohali: @dottorblaster This still needs addressing before I can approve this PR.

dottorblaster: Done! Sorry, I missed it 🙇

wohali:
+1 LGTM

Issue #1737, opened 2018-11-13
Performance regression querying changes using _doc_ids filter


garethbowen:
When upgrading from v1.7.1 to v2.2.0 I noticed our replication was taking longer. I investigated further and found the problem was specifically in relation to the initial request for changes. We use the _doc_ids filter so we can replicate only certain documents and this has been stable and performant on the 1.x versions.

## Steps to Reproduce

Use [this node script](https://github.com/apache/couchdb/files/2578449/index.js.txt) to create a database and fill it with 1 million docs and then query it for specific IDs.

## Context

In my testing using the above script I got responses in 1 to 2ms on v1.7.1 and 2500 to 2600ms on 2.2.0.
In our production database with real world data and about 8 million docs it takes less than a second on v1.7.1 and around 40 seconds on v2.2.0.

This has affected real world performance for users trying to replicate their data.

## Your Environment

* Version used: 2.2.0 (from 1.7.1)
* Operating System: Ubuntu 1804
* Link to your project: https://github.com/medic/medic-webapp/

## What I've tried

- changing clustering from n=3 q=8 to n=1 q=1 - no improvement
- changing the `changes_doc_ids_optimization_threshold` at 1, 100, and very large - no improvement
- using a very large `seq_interval` parameter - no improvement
- using a mango selector - worse performance

rnewson:
ok, I think I get this finally. Before 2.0, we had an optimization for _doc_ids and _design filter;

```
commit bfa0a8900163edd4f85c7bbf5b595de9885cfbf9
Author: Filipe David Borba Manana <fdmanana@apache.org>
Date:   Tue Sep 20 22:55:29 2011 +0000

    Efficient implementation of builtin filters

    Currently, the builtin changes filters "_doc_ids" and "_design"
    are not very efficient because they fold the entire seq btree
    and then filter the values by document ID.
    This implementation avoids that by doing direct lookups against
    the id btree, and then, for continuous changes requests, it
    just listens for database update events and does partial seq
    btree folds.

    COUCHDB-1288
```

The clustered code for _changes does not use it.

rnewson:
we pass `{default,main_only}` down as the filter, so the optimization does not fire.

It's not yet clear if this can be restored given how the clustered version of changes feed has to work, but at least this is progress.

rnewson:
so fabric_rpc:changes always calls `couch_db:fold_changes` .

wohali:
[...]

PR #1771, opened 2018-11-28
fix: re-enable _doc_id filter optimisation path

janl:
wip fix for #1737 just to see how CI likes this, initial comments welcome

### DIFF
--- a/src/couch/include/couch_db.hrl
+++ b/src/couch/include/couch_db.hrl
@@ -203,6 +203,14 @@
     atts = []
 }).
 
+-record (fabric_changes_acc, {
+    db,
+    seq,
+    args,
+    options,
+    pending,
+    epochs
+}).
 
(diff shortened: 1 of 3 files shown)


wohali:
Could use a test case so we don't regress on this again.

janl:
@wohali you mean when we add another clustering later? ;D

This needs conceptual approval before moving on.

Should this approach be acceptable, I'd also re-enable the design doc filter optimisation  the same way.

wohali:
@janl Elixir now tests against a cluster, so, yes, we have the infrastructure in place :)

janl:
@wohali tried to make a joke, got foiled by a typo, nvm :/

janl:
cleaned up and ready for review

wohali:
Holding for 2.3.0 per @rnewson. Thanks Bob!

PR #1776, opened 2018-11-29
Format and check all Python code using python black

wohali:
Inspired by our formatting work on [Elixir](https://github.com/apache/couchdb/pull/1769), this PR proposes doing the same thing for our Python code.

[Python Black](https://github.com/ambv/black) is an opinionated code formatter now adopted by prominent Python projects such as `pip`. It is easy to use and provides both formatting and check modes.

The `python-black` target builds a python3 venv at .venv and installs black if possible. Since black is Python 3.6 and up only, we skip the check on systems with an older Python 3.x and print a warning. Our CI systems will still catch problems from people who submit PRs who don't have 3.6+ since CI is  configured to run this check automatically on at least one platform (e.g., Jenkins Ubuntu Bionic, or Python 3.6 explicitly requested under Travis CI.)

Assuming the check runs, it fails if any file would be changed by the formatter.

A Makefile target is provided that actually _does_ the updating (`make python-black-update`, it simply omits the `--check` flag) for happy dev fingers to fix up python files.

The change to the `docs` dependency in `rebar.config.script` is necessary because of embedded Python files in that repo that have only just been re-formatted. Once docs is tagged with 2.3.0, this dep will be reverted to a tag-type reference, not a commit.

@nickva I had to change the location of the Mango venv to `src/mango/.venv` so it would be automatically ignored by the formatter, otherwise it tries to format the entire venv. Oops :)

## Testing recommendations

If you have Python 3.6+, try `make python-black` and see if it works.

if you don't, try the same command and make sure it is a noop (with an exit code of 0).

## Checklist

### DIFF
--- a/build-aux/logfile-uploader.py
+++ b/build-aux/logfile-uploader.py
@@ -13,7 +13,6 @@
 # the License.
 
 
-
 import datetime
 import glob
 import json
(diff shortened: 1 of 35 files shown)


build-aux/logfile-uploader.py
```diff
@@ -26,96 +25,111 @@
 COUCH_URL = "https://couchdb-vm2.apache.org/ci_errorlogs"
 TARFILE = "couchlog.tar.gz"
 
+
 def _tojson(req):
     """Support requests v0.x as well as 1.x+"""
-    if requests.__version__[0] == '0':
```
jaydoane: I wonder if it would be better to `--skip-string-normalization` and avoid all the conversions from single to double quotes? I don't personally find their [opinions on preferring double quotes](https://github.com/ambv/black#strings) entirely compelling.

wohali:
Any comments from @janl @davisp @iilyak @eiri or anyone else who does Python-y things?

PR #1778, opened 2018-11-29
Cache query servers in ets table

### LINKED ISSUE #1772: Slow operation in couch_proc_manager
We've noticed a couple couch_proc_manager backups when deploying the latest version. Inspecting code I ended up stumbling over the new config code. Its theoretically quite a bit slower than the old code since its doing OS env checking but its also doing things quite inefficiently beyond that.

https://github.com/apache/couchdb/blob/master/src/couch/src/couch_proc_manager.erl#L375-L383

The three big things to notice here are that we're joining the variable names inside the loop which is a bit wasteful. Given the size of the strings I'm not sure how bad that is though. I'd wager it likely depends on how many environment variables there are to know whether its making a difference. The tokens call with a pattern match also seems like an expensive thing to be doing. However the killer bit is probably that we're looping over `os:getenv()` every time we create a new process which is likely what's caused the backups.

I'd propose two changes. The first simpler of the two is to just create our expected environment variable and use `os:getenv/1`.

Something like:

```erlang
get_env_for_spec(Spec, Target) ->
    SpecStr = Spec ++ Target,
    os:getenv(SpecStr, undefined).
```

We're still going to be hitting `os:getenv/1` every time we create a new process though so I'd also recommend caching succcessful finds in an ets table or similar. That would allow us to have similar speed to before the patch while still maintaining the OS env config approach.

iilyak:
## Overview

The configuration of query servers is done via environment variables.
Calling os:getenv every time we need a new query process is expensive.
Instead we extact all configured query servers from the environemt on
`couch_proc_manager` startup and cache them in ets table.

fixes #1772

## Testing recommendations

1. make eunit
2. remsh into running dev cluster and check the following
 - `ets:tab2list(couch_proc_manager_servers).` - returns 
   - COFFEESCRIPT
   - QUERY
   - JAVASCRIPT
 - update config to enable native query server `config:set("native_query_servers", "enable_erlang_query_server", "true", false).`
  - make sure `ets:tab2list(couch_proc_manager_servers).` includes ERLANG

- issue https://github.com/apache/couchdb/issues/1772
- problem introduced in https://github.com/apache/couchdb/pull/1602

## Checklist

### DIFF
--- a/src/couch/src/couch_proc_manager.erl
+++ b/src/couch/src/couch_proc_manager.erl
@@ -43,6 +43,7 @@
 -define(PROCS, couch_proc_manager_procs).
 -define(WAITERS, couch_proc_manager_waiters).
 -define(OPENING, couch_proc_manager_opening).
+-define(SERVERS, couch_proc_manager_servers).
 -define(RELISTEN_DELAY, 5000).
 
 -record(state, {
@@ -104,6 +105,11 @@ init([]) ->
     ets:new(?PROCS, TableOpts ++ [{keypos, #proc_int.pid}]),
     ets:new(?WAITERS, TableOpts ++ [{keypos, #client.timestamp}]),
     ets:new(?OPENING, [public, named_table, set]),
+    ets:new(?SERVERS, [public, named_table, set]),
+    ets:insert(?SERVERS, get_servers_from_env("COUCHDB_QUERY_SERVER_")),
+    ets:insert(?SERVERS, get_servers_from_env("COUCHDB_NATIVE_QUERY_SERVER_")),
+    ets:insert(?SERVERS, [{"QUERY", {mango_native_proc, start_link, []}}]),
+    maybe_configure_erlang_native_servers(),
 
     {ok, #state{
         config = get_proc_config(),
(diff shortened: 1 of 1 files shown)


src/couch/src/couch_proc_manager.erl
```diff
@@ -372,23 +382,36 @@ new_proc(Client) ->
...
     end, os:getenv()).
 
 get_query_server(LangStr) ->
-    % look for COUCH_QUERY_SERVER_LANGSTR in env
-    % if exists, return value, else undefined
-    UpperLangString = string:to_upper(LangStr),
-    case get_env_for_spec("COUCHDB_QUERY_SERVER_", UpperLangString) of
-        [] -> undefined;
-        [Command] -> Command
+    case ets:lookup(?SERVERS, string:to_upper(LangStr)) of
```
janl: doubtful it’s a performance issue, but we could do `to_lower()` when populating the ets table so we can avoid `to_upper()` here on each call. not blocking a merge, just leaving this here.

iilyak: in this case languages with mixed case will not work.

PR #1782, opened 2018-11-30
Only validate partitioned db name if db is partitioned

jaydoane:
## Overview

Maybe there's already code to check whether the {partitioned, true} prop is present, but I didn't see
it. In any case, this seems like one way to do it.

## Testing recommendations

```
$ mix test --trace test/partition_crud_test.exs:299
Including tags: [line: "299"]
Excluding tags: [:test]


PartitionCrudTest
  * test can create unpartitioned system db (218.2ms)
  * test _bulk_docs saves docs with partition key (skipped)
  * test Sets partition in db info (skipped)
  * test create database with bad `partitioned` value (skipped)
  * test _bulk_docs errors with bad partition key (skipped)
  * test _bulk_get bad partitioned document (skipped)
  * test POST and GET document (skipped)
  * test _bulk_docs errors with bad doc key (skipped)
  * test saves attachment with partitioned doc (skipped)
  * test POST and _bulk_get document (skipped)
  * test cannot create partitioned system db (skipped)
  * test PUT fails if a partition key is not supplied (skipped)
  * test POST fails if a partition key is not supplied (skipped)
  * test PUT fails for partitions with _ (skipped)
  * test PUT fails for bad partitions (skipped)
  * test _bulk_docs errors with missing partition key (skipped)
  * test PUT and GET document (skipped)

CouchTestCase

PartitionHelpers


Finished in 0.3 seconds
17 tests, 0 failures, 16 skipped
```

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -905,6 +905,8 @@ error_info({error, {illegal_database_name, Name}}) ->
     {400, <<"illegal_database_name">>, Message};
 error_info({illegal_docid, Reason}) ->
     {400, <<"illegal_docid">>, Reason};
+error_info({illegal_partition, Reason}) ->
+    {400, <<"illegal_partition">>, Reason};
 error_info({_DocID,{illegal_docid,DocID}}) ->
     {400, <<"illegal_docid">>,DocID};
 error_info({error, {database_name_too_long, DbName}}) ->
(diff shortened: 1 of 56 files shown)


eiri:
In my (bias) opinion it'd be cleaner to contain this logic within partition module, i.e. change  `couch_partition:validate_dbname/1` to `/2`, pass Opts and do this check there.

Also, `proplists:get_value/2` is as long as it is slow, let's use `couch_util:get_value/3` as we do everywhere else :)

jaydoane:
@eiri thanks for the suggestions! I've incorporated them into a fixup commit. Keep in mind this is Paul's development branch I'm basing this on, so he'll probably just refactor the whole think anyway 😀

jaydoane:
This has been incorporated into https://github.com/apache/couchdb/pull/1789

PR #1783, opened 2018-11-30
fix empty queries

tonysun83:
## Overview

Mango text indexes currently throw function clauses when queries in two
situations:

1) Empty Selector
2) Operators with empty arrays.

We fix this issue and change the behavior as follows:

1) Any empty selector will fall back on _all_docs (like json indexes).
2) $or, $and, $in, $all that specify an empty arrays
will be treated as no-op when combined with other operators.
3) A single no-nop query such as {"$or": []} will not be executed and
return nothing just like json indexes.
4) $nin with an empty array will return all docs that match the field,
just like json indexes.

## Testing recommendations

New tests were added to the python suite and can be run with
MANGO_TEXT_INDEXES=1 nosetests test/21-empty-selector-tests.py

This requires a clouseau setup. 

The issue is this PR.

## Checklist

### DIFF
--- a/src/mango/src/mango_selector.erl
+++ b/src/mango/src/mango_selector.erl
@@ -399,10 +399,16 @@ negate({[{Field, Cond}]}) ->
     {[{Field, negate(Cond)}]}.
 
 
+% We need to treat an empty array as always true. This will be applied
+% for $or, $in, $all, $nin as well.
+match({[{<<"$and">>, []}]}, _, _) ->
+    true;
 match({[{<<"$and">>, Args}]}, Value, Cmp) ->
     Pred = fun(SubSel) -> match(SubSel, Value, Cmp) end,
     lists:all(Pred, Args);
 
+match({[{<<"$or">>, []}]}, _, _) ->
+    true;
 match({[{<<"$or">>, Args}]}, Value, Cmp) ->
     Pred = fun(SubSel) -> match(SubSel, Value, Cmp) end,
     lists:any(Pred, Args);
@@ -410,6 +416,8 @@ match({[{<<"$or">>, Args}]}, Value, Cmp) ->
 match({[{<<"$not">>, Arg}]}, Value, Cmp) ->
     not match(Arg, Value, Cmp);
 
+match({[{<<"$all">>, []}]}, _, _) ->
+    true;
 % All of the values in Args must exist in Values or
 % Values == hd(Args) if Args is a single element list
 % that contains a list.
@@ -493,6 +501,8 @@ match({[{<<"$gte">>, Arg}]}, Value, Cmp) ->
 match({[{<<"$gt">>, Arg}]}, Value, Cmp) ->
     Cmp(Value, Arg) > 0;
 
+match({[{<<"$in">>, []}]}, _, _) ->
+    true;
 match({[{<<"$in">>, Args}]}, Values, Cmp) when is_list(Values)->
     Pred = fun(Arg) ->
         lists:foldl(fun(Value,Match) ->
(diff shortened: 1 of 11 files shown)


src/mango/src/mango_selector.erl
```diff
@@ -399,17 +399,25 @@ negate({[{Field, Cond}]}) ->
...
+    true;
 match({[{<<"$or">>, Args}]}, Value, Cmp) ->
     Pred = fun(SubSel) -> match(SubSel, Value, Cmp) end,
     lists:any(Pred, Args);
 
 match({[{<<"$not">>, Arg}]}, Value, Cmp) ->
     not match(Arg, Value, Cmp);
 
+match({[{<<"$all">>, []}]}, _, _) ->
+    true;
```
ksnavely: Is this a duplicate clause from L404?

src/mango/test/06-basic-text-test.py
```diff
@@ -438,6 +438,41 @@ def test_regex(self):
...
+
+    def test_empty_array_and_with_age(self):
+        resp = self.db.find({"age": 22, "$and": [{"b": {"$all":[]}}]},
+            explain=True)
+        self.assertEqual(resp["index"]["type"], "text")
+        docs = self.db.find({"age": 22, "$and": []})
+        assert len(docs) == 1
+
+    def test_empty_arrays_complex(self):
+        resp = self.db.find({"$or": [], "a": {"$in" : []}}, explain=True)
```
ksnavely: "simple" combinations of empty queries will also short-circuit to no execution right?

michaelbreslin:
Is there a fix for this going in anytime soon?

wohali:
Hi @michaelbreslin , probably not before the holiday break at this point - but I expect it'll land shortly thereafter. Looks like we may need some more tests before this can be merged.

src/mango/src/mango_selector.erl
tonysun83: 404 is $and, this is $all

src/mango/test/06-basic-text-test.py
tonysun83: correct

tonysun83:
@willholley : Your additional testing suggestion revealed a conflicting issue with special/view indexes vs text indexes.

```
{"$or": []} 
{"$all": []} 
{"$in": []} 
{"$and": []} 
```

Special/View - We do a full range scan, then filter.
Text Index - The current implementation doesn't execute the query. 

I think the solution is to make special/view indexes ignore the query(not execute the fabric call)  as well. But it get's a little ugly as the _explain endpoint needs to reflect this as well. 

cc @michaelbreslin

tonysun83:
@willholley : It doesn't generate an error and will be an improvement. My only concern is that it will break existing functionality. I have a test suite for that so that shouldn't be a problem. The issue I'm currently running into is getting all 3 to produce the same result.

Ex: 

```
1) {"age": 22, "$or": []}

2) {"$or": []}
```
produce different results in json, _all_docs, and text depending on where part of the code is implemented. The correct solution should be IMO should be

1) We get all documents back where the age is 22
2) We get 0 documents back.

for all 3 index types. I'm debugging why it's currently not the case

src/mango/src/mango_selector.erl
davisp: Also of note, a duplicate clause would have triggered a compilation error.

janl:
After fixing the linter errors and allowing for CI flakes in view_compaction.js, this now fails with an error in the new tests:

```
8269  File "/home/travis/build/apache/couchdb/src/mango/test/21-empty-selector-tests.py", line 42, in test_empty_array_and_with_age
8270    self.assertEqual(resp["index"]["type"], klass.INDEX_TYPE)
8271AssertionError: 'special' != 'text'
8272- special
8273+ text
```

@tonysun83 do you have time to look into this?

jiangphcn:
I would like to continue, and found that `export MANGO_TEXT_INDEXES=1` is helpful after some investigation. Will add new commit.

PR #1790, opened 2018-12-03
Move tests to suite

iilyak:
## Overview

Implementation of eunit tests in the same module with the code we are trying to test is not a good idea. The test suites in the module are conditionally compiled using `-ifdef(TEST).`. This means that depending on phase of the moon the tests can be skipped. They are certainly skipped when `eunit` is used in combination with `meck` library. This problem manifested itself in https://github.com/apache/couchdb/pull/1754 where the tests were originally introduced. The PR was merged because Travis was green and it had a `+1`. However close look at test logs (https://travis-ci.org/apache/couchdb/builds/458820868?utm_source=github_status&utm_medium=notification) on Travis reveal that the tests didn't run (search for couch_flags_config):
```
module 'couch_bt_engine_stream'
module 'couch_flags_config'
module 'couch_changes'
  module 'couch_changes_tests'
    Changes feed
      Filter _selector
        couch_changes_tests:368: should_select_basic...[0.055 s] ok
        couch_changes_tests:382: should_select_with_since...[0.001 s] ok
``` 

The only correct solution is to move tests into a suite in `test` directory. In this case compilation will be unconditional so the tests wouldn't be skipped.

This PR moves the tests and updates the expected return value so the test pass.

## Testing recommendations

1. Run `make eunit > eunit.log` without reducing the scope of testing (using apps= or tests=).
2. Consult `eunit.log` to see if tests for couch_flags_config has been run.

- https://github.com/apache/couchdb/pull/1754

## Checklist

### DIFF
--- a/src/couch/src/couch_flags_config.erl
+++ b/src/couch/src/couch_flags_config.erl
@@ -20,6 +20,11 @@
     data_provider/0
 ]).
 
+%% for test suite only
+-export([
+    parse_flags_term/1
+]).
+
 -define(DATA_INTERVAL, 1000).
 -define(MAX_FLAG_NAME_LENGTH, 256).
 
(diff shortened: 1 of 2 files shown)


kxepal:
I was think that tests within a module are supposed to test some internal things, which cannot be exported to be tested from the suite. There are no such cases?

iilyak:
@kxepal: Unfortunately tests within a module don't work in some cases. There are at least two problems with such tests:
- when `meck` mocks the module it doesn't define `TEST` macro. When eunit stops the test where meck was used it is not smart enough to unload the module. BEAM uses that version of the module which doesn't have exported test functions. When eunit framework scans the loaded modules to find what tests to run it doesn't see the tests (because they are not define in the loaded module). 
- when module is used from a dependency. Let's assume the eunit decided to run tests for applications in following order `chttpd`, `couch`. `chttpd` uses some functions from `couch` application. Therefore `couch` application is compiled. The problem is `TEST` is not defined. Which means that modules are compiled without test functions. This set of modules loaded in the VM. Again eunit cannot unload the module and load correct version when it moves on to `couch` app testing.

Therefore tests in the same module is anti-pattern, unfortunately.

kxepal:
@iilyak TIL something useful. Thank you!

iilyak:
Here is a list of issues which are solved by moving tests into separate module
- https://github.com/eproxus/meck/issues/133#issuecomment-262028993
- https://github.com/eproxus/meck/issues/123#issuecomment-47282605
- https://github.com/eproxus/meck/issues/158#issuecomment-160589223

I think this could be the  problem https://github.com/eproxus/meck/blob/38b25cc60e0a6d869c6fad0446517ac2f6e3a88a/src/meck_proc.erl#L593 (but I am not sure)

purge can return false and in this case it need to be called second time see http://erlang.org/doc/man/code.html#purge-1

Issue #1704, opened 2018-10-29
Test suite `delayed_commits.js` unreliable.


eiri:
It became quite often that JS tests fail on Travis with the following

```bash 
test/javascript/tests/delayed_commits.js                       
    Error: Failed to execute HTTP request: Failed to connect to 127.0.0.1 port 15984: Connection refused
Trace back (most recent call first):
    
  37: test/javascript/couch_http.js
      ("\"false\"\n")
 468: 127.0.0.1/_config/couchdb/delayed_commits",[object Object])@test/javascript/couch.js
      ("PUT","/_node/node1
 408: test/javascript/couch_test_runner.js
      run_on_modified_server([object Array],(function () {sleep(15000);T(db.
  29: test/javascript/tests/delayed_commits.js
      ()
  37: test/javascript/cli_runner.js
      runTest()
  48: test/javascript/cli_runner.js
      
fail
```

## Expected Behavior
JS tests should pass.

## Current Behavior
JS tests fail often locally in `delayed_commits` suite.

## Possible Solution
It's either some changes introduced with new purge functionality or some tweaking happened around JS suite itself. A bit of digging what exactly `delayed_commits` is testing and how that part of functionality was affected in the latest changes is required.

## Steps to Reproduce (for bugs)
Run `make javascript` until the failure. Note that isolated run `make javascript suites=delayed_commits` most of the time finishes fine.

wohali:
@eiri note that Jenkins is passing recently, all the time. If you're trying to reproduce this, maybe try on a heavily CPU-loaded or RAM-constrained system?

eiri:
@wohali Annoyingly this particular test fails for me locally 4 times out of 5, even on the current master, and my MBP not very starved on the resources. Initially I thought this is my local issue, something in my env, but now when I've seen the same failure on Travis, my guess is that this test just became finicky with some of the recent changes.

eiri:
So, a minimal set of the js suites I can reproduce this with is `make javascript suites=coffee,compact,delayed_commits` which leads me to believe it triggered by some interaction between a compaction ran on a freshly built view (from `coffee` suite) and `delayed_commits`.

My guess is that there are a race somewhere that's _less_ visible on a setup with slower HD, that's why Jenkins not affected that much.

wohali:
@eiri now that Elixir tests have landed, maybe focus on porting this to Elixir so we can ditch the JS test suite ASAP? :) Unless you think there's an actual race condition here that needs addressing.

[...]

PR #1794, opened 2018-12-06
Fix get minimum purge seq

davisp:
## Overview

This fixes a silly mistake of using a clustered _all_docs lookup to get a single design document.

## Testing recommendations

Code is covered by `make check`

Clustered purge added the bad code.

## Checklist

### DIFF
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -77,6 +77,7 @@
     get_full_doc_info/2,
     get_full_doc_infos/2,
     get_missing_revs/2,
+    get_design_doc/2,
     get_design_docs/1,
     get_design_doc_count/1,
     get_purge_infos/2,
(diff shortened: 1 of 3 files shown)


tonysun83:
+1

src/couch/src/couch_db.erl
```diff
@@ -608,6 +609,25 @@ get_db_info(Db) ->
     ],
     {ok, InfoList}.
 
+get_design_doc(#db{name = <<"shards/", _/binary>> = ShardDbName}, DDocId0) ->
+    DDocId = case DDocId0 of
```
eiri: Maybe abstract this cut'n'paste bit in a helper, something like `normalize_ddoc_id/1`?
```erlang
normalize_ddoc_id(<<"_design/", _/binary>> = DDocId) ->
    DDocId;
normalize_ddoc_id(DDocId) ->
   <<"_design/", DDocId>>.
```

davisp: Yeah, I wasn't super happy with that bit. I guess adding a couch_util function might be better.

davisp: Fixed.

src/couch_mrview/src/couch_mrview_index.erl
```diff
@@ -226,16 +226,15 @@ verify_index_exists(DbName, Props) ->
         if Type =/= <<"mrview">> -> false; true ->
             DDocId = couch_util:get_value(<<"ddoc_id">>, Props),
             couch_util:with_db(DbName, fun(Db) ->
-                {ok, DesignDocs} = couch_db:get_design_docs(Db),
-                case get_ddoc(DbName, DesignDocs, DDocId) of
-                    #doc{} = DDoc ->
+                case couch_db:get_design_doc(Db, DDocId) of
```
eiri: This bit here is kind of inconsistent (to be honest it was inconsistent before). We are matching on `not_found`, but clause crashing on `{error, _...`, yet catching and discarding the exception below.

We can either remove this deep `case` all together and always crash on not matching to `{ok, Doc}`, since we are treating errors as `not_found` or go other way around and add `Error -> throw(Error)` clause to make error trace a bit more readable.

jaydoane:
All tests passing for me with `make eunit`. Getting this failure from `make check`
```
    Error: Failed to execute HTTP request: Failed to connect to 127.0.0.1 port 15984: Connection refused
```
from presumably flaky JS test.

eiri:
@jaydoane yep, `delayed_commits` not awesome, see #1704

src/couch_mrview/src/couch_mrview_index.erl
davisp: Hrm. I see your point but I'd think that'd be a second PR if you want to change things. I'm hesitant to change orthogonal logic as part of this work.

PR #1796, opened 2018-12-06
Port delayed_commits test to Elixir

### LINKED ISSUE #1704: Test suite `delayed_commits.js` unreliable.

eiri:
## Overview

This PR adds a new helper function to `test_helper` allowing to restart a node or the whole cluster and ports `delayed_commits` test.

## Testing recommendations

`make elixir` should pass, output log should include:
```
...
DelayedCommitsTest
  * test delayed commit (9208.3ms)
...
```

Closes #1704 

## Checklist

### DIFF
--- a/test/elixir/test/delayed_commits_test.exs
+++ b/test/elixir/test/delayed_commits_test.exs
@@ -1,0 +1,31 @@
+defmodule DelayedCommitsTest do
+  use CouchTestCase
+
+  @moduledoc """
+  Test CouchDB delayed commits
+  This is a port of the delayed_commits.js suite
+
+  Note that delayed_commits is deprecated in 2.0, so this is a minimal
+  test to show it still works. delayed_commits will be removed in 3.0.
+  """
+
+  @tag config: [
+         {"couchdb", "delayed_commits", "true"}
+       ]
+  @tag :with_db
+  test "delayed commit", context do
+    db_name = context[:db_name]
+    doc_id = "doc-1"
+    resp = Couch.put("/#{db_name}/#{doc_id}", body: %{a: 2, b: 4})
+    assert resp.status_code in 201..204
+    assert resp.body["ok"]
+
+    resp = Couch.get("/#{db_name}/#{doc_id}")
+    assert resp.status_code == 200, "The new doc should be in the database"
+
+    restart_cluster()
+
+    resp = Couch.get("/#{db_name}/#{doc_id}")
+    assert resp.status_code == 404, "The new doc should be missing"
+  end
+end
--- a/test/elixir/README.md
+++ b/test/elixir/README.md
@@ -33,7 +33,7 @@ X means done, - means partially
   - [ ] Port conflicts.js
   - [ ] Port cookie_auth.js
   - [ ] Port copy_doc.js
-  - [ ] Port delayed_commits.js
+  - [X] Port delayed_commits.js
   - [ ] Port design_docs.js
   - [ ] Port design_options.js
   - [ ] Port design_paths.js
--- a/test/elixir/lib/couch.ex
+++ b/test/elixir/lib/couch.ex
@@ -50,6 +50,10 @@ defmodule Couch do
   CouchDB library to power test suite.
   """
 
+  def process_url("http://" <> _ = url) do
+    url
+  end
+
   def process_url(url) do
     "http://127.0.0.1:15984" <> url
   end
--- a/test/elixir/lib/couch/db_test.ex
+++ b/test/elixir/lib/couch/db_test.ex
@@ -239,4 +239,57 @@ defmodule Couch.DBTest do
     opts = [pretty: true, width: 20, limit: :infinity, printable_limit: :infinity]
     inspect(resp, opts)
   end
+
+  def restart_cluster() do
+    resp = Couch.get("/_membership")
+    assert resp.status_code == 200
+    nodes = resp.body["all_nodes"]
+
+    nodes_ports =
+      Enum.reduce(nodes, [], fn node, acc ->
+        port = node_to_port(node)
+        [{node, port} | acc]
+      end)
+
+    tasks =
+      Enum.map(nodes_ports, fn {node, port} ->
+        Task.async(fn -> restart_node(node, port) end)
+      end)
+
+    Task.yield_many(tasks, length(nodes) * 5000)
+  end
+
+  def restart_node(node \\ "node1@127.0.0.1") do
+    port = node_to_port(node)
+    restart_node(node, port)
+  end
+
+  defp restart_node(node, port) do
+    url = "http://127.0.0.1:#{port}/_node/#{node}/_restart"
+    resp = Couch.post(url)
+    assert HTTPotion.Response.success?(resp)
+    assert resp.body["ok"]
+    # make sure node went down. we assuming the node can't bounce quick
+    # enough to inroduce a race here
+    retry_until(fn -> !node_is_running(port) end)
+    # wait utill node is back
+    retry_until(fn -> node_is_running(port) end, 500, 10_000)
+  end
+
+  defp node_is_running(port) do
+    url = "http://127.0.0.1:#{port}/_up"
+    resp = Couch.get(url)
+
+    case HTTPotion.Response.success?(resp) do
+      true -> resp.status_code in 200..399
+      false -> false
+    end
+  end
+
+  defp node_to_port(node) do
+    url = "/_node/#{node}/_config/chttpd/port"
+    resp = Couch.get(url)
+    assert HTTPotion.Response.success?(resp)
+    resp.body
+  end
 end


test/elixir/test/delayed_commits_test.exs
```diff
@@ -0,0 +1,31 @@
...
+
+  Note that delayed_commits is deprecated in 2.0, so this is a minimal
+  test to show it still works. delayed_commits will be removed in 3.0.
+  """
+
+  @tag config: [
+         {"couchdb", "delayed_commits", "true"}
+       ]
+  @tag :with_db
+  test "delayed commit", context do
```
iilyak: The implementation of delayed commits is prune to race conditions since it is implemented using send_after see https://github.com/apache/couchdb/blob/master/src/couch/src/couch_db_updater.erl#L742. The timeout is only one second. Which makes the test a flaky one. I would recommend wrapping all steps with `retry_until(fn -> ... true end)`

test/elixir/test/delayed_commits_test.exs
```diff
@@ -0,0 +1,31 @@
...
+    db_name = context[:db_name]
+    doc_id = "doc-1"
+    resp = Couch.put("/#{db_name}/#{doc_id}", body: %{a: 2, b: 4})
+    assert resp.status_code >= 201
```
iilyak: This would also return true for error codes. You can use `resp.status_code in (201..204)` or something like that.

test/elixir/test/delayed_commits_test.exs
```diff
@@ -0,0 +1,31 @@
...
+    assert resp.body["ok"]
+
+    resp = Couch.get("/#{db_name}/#{doc_id}")
+    assert resp.status_code == 200, "The new doc should be in the database"
```
iilyak: thank you for using explanation argument for assertion.

test/elixir/test/delayed_commits_test.exs
iilyak: There is a slight problem with using `retry_until`. Most likely it would swallow the exception when timeout is reached so we wouldn't see nice error message. Maybe we need to talk to @jaydoane to extend it to re-raise original error.

test/elixir/test/test_helper.exs
```diff
@@ -259,6 +259,59 @@ defmodule CouchTestCase do
...
+            port = node_to_port(node)
+            [{node, port} | acc]
+          end)
+
+        tasks =
+          Enum.map(nodes_ports, fn {node, port} ->
+            Task.async(fn -> restart_node(node, port) end)
+          end)
+
+        Task.yield_many(tasks, length(nodes) * 5000)
```
iilyak: Good idea to make timeout depend on number of nodes.

test/elixir/test/delayed_commits_test.exs
eiri: This test is racy pretty much by definition, though most of the problems with its javascript version coming not from the fact that it takes more than one second to _send_ a node to a restart, but that the cluster doesn't come back in time.

I'm honestly sceptical about usage of `retry_until` here. It has default timeout at 5 secs and this test takes whooping 9 sec on my MBP with the whole cluster bounce. I can wrap it with something like 10 sec retry interval and total timeout 30 secs, but this is pretty big band aid, so I'd prefer to actually see that race where doc saved before cluster bounce before extending the suites' runtime that much.

test/elixir/test/delayed_commits_test.exs
eiri: Good idea. Fixed.

test/elixir/test/test_helper.exs
```diff
@@ -259,6 +259,59 @@ defmodule CouchTestCase do
...
+        port = node_to_port(node)
+        restart_node(node, port)
+      end
+
+      defp restart_node(node, port) do
+        url = "http://127.0.0.1:#{port}/_node/#{node}/_restart"
+        resp = Couch.post(url)
+        assert HTTPotion.Response.success?(resp)
+        assert resp.body["ok"]
+        # make sure node went down. we assuming the node can't bounce quick
```
iilyak: If we want to be extra cautious we could use `uptime` to detect node restart.
```
$ http --auth=adm:pass :15984/_node/node2@127.0.0.1/_system | jq .uptime
218
$ http --auth=adm:pass POST :15984/_node/node2@127.0.0.1/_restart
HTTP/1.1 200 OK
Content-Length: 12
Date: Fri, 07 Dec 2018 14:59:18 GMT
Server: CouchDB/2.3.0-fcd62ed-dirty (Erlang OTP/19)
X-Couch-Request-ID: b5b6d5b302
X-CouchDB-Body-Time: 0

{
    "ok": true
}

$ http --auth=adm:pass :15984/_node/node2@127.0.0.1/_system | jq .uptime
null
$ http --auth=adm:pass :15984/_node/node2@127.0.0.1/_system | jq .uptime
null
$ http --auth=adm:pass :15984/_node/node2@127.0.0.1/_system | jq .uptime
null
$ http --auth=adm:pass :15984/_node/node2@127.0.0.1/_system | jq .uptime
null
$ http --auth=adm:pass :15984/_node/node2@127.0.0.1/_system | jq .uptime
0
$ http --auth=adm:pass :15984/_node/node2@127.0.0.1/_system | jq .uptime
1
$ http --auth=adm:pass :15984/_node/node2@127.0.0.1/_system | jq .uptime
2
```

test/elixir/test/test_helper.exs
```diff
@@ -259,6 +259,59 @@ defmodule CouchTestCase do
...
+        # wait utill node is back
+        retry_until(fn -> node_is_running(port) end, 500, 10000)
+      end
+
+      defp node_is_running(port) do
+        url = "http://127.0.0.1:#{port}/_up"
+        resp = Couch.get(url)
+
+        case HTTPotion.Response.success?(resp) do
+          true -> resp.status_code >= 200 && resp.status_code < 400
```
iilyak: `resp.status_code in (200 .. 399)`

test/elixir/test/test_helper.exs
eiri: I think this is an overkill. Checking that node went down, then got back should be enough to make sure we have our restart.

test/elixir/test/test_helper.exs
eiri: Hm.. This does the same thing, you know. I don't see ranges used for status code assertion anywhere else in existing elixir tests, so are you just expressing your personal preferences?

eiri: Well, whatever, I think my variant is more readable, but I don't care. Changed

eiri:
@iilyak sorry, it's not clear to me - are you blocking this PR over `retry_until` and `uptime` or just got distracted by something else?

test/elixir/test/delayed_commits_test.exs
van-mronov: Could it be useful to define a bunch of helper functons(`json_response/2`, `text_response/2`) like [Phoenix](https://github.com/phoenixframework/phoenix/blob/4da40a29f968b6e834ab55849af4a7337815ea50/lib/phoenix/test/conn_test.ex#L417) did?
But pass `%HTTPotion.Response{}` instead of `Plug.Conn`.
So we can do something like:

```elixir
resp = Couch.get("/#{db_name}/#{doc_id}")
assert %{"ok" => true} = json_response(resp, 200)
```
or
```elixir
resp = Couch.get("/#{db_name}/#{doc_id}")
assert %{"error" => _error} = json_response(resp, 404)
```

eiri: In my opinion that would do more harm than good. I believe that writing tests is inherently different from writing "normal" code and in former the simplicity and readability should be always prioritized over tersity.

When I write
```
assert resp.body == ${"ok" => true}
```
I can scan these lines and instantly understand what's going on here.

If i hide it into wrapper
```
assert %{"ok" => true} = json_response(resp, 200)
```
sure I saved one line, but now I have to stop and parse what exactly happening here. Unless I remember exactly what `json_response/2` does I have to guess: is that `200` a status code? Is the function going to raise an exception if status code not `200` or return `false`? Is any body transformation happening "under the hood"?

In general I'd prefer us to introduce code obfuscating changes to tests only when it's absolutely unavoidable.

eiri: 1) I don't find the second example more readable. 2) A thing that'd require a new or occasional test writer to stop and look into documentation or helper's code is an unnecessary hurdle. 3) I don't see a point to discuss here our personal opinion on the test writing. I suggest you to open a separate GH issue on this topic for a broader audience.

PR #1798, opened 2018-12-07
Suppress compiler warnings

jaydoane:
## Overview

Suppress warnings when compiled with OTP 20.3

## Testing recommendations

After applying these changes,
`make clean && make`
using OTP 20 should have many fewer compiler warnings than before. Of course, numerous dependencies still emit compiler warnings.

## Persistent behaviour warnings

I'm still seeing the following behaviour warnings
```
couchdb/src/couch/src/couch_bt_engine.erl:14: Warning: behaviour couch_db_engine undefined
couchdb/src/couch_replicator/src/couch_replicator_clustering.erl:31: Warning: behaviour mem3_cluster undefined
couchdb/src/couch_peruser/src/couch_peruser.erl:15: Warning: behaviour mem3_cluster undefined
```
but it's not obvious why. The dependent applications are included in .app.src files, and it appears that the callbacks are implemented correctly. Any help figuring these out would be appreciated.

https://github.com/apache/couchdb-ets-lru/pull/7

## Checklist


src/mem3/src/mem3.erl
```diff
@@ -196,7 +196,7 @@ choose_shards(DbName, Nodes, Options) ->
     Q = mem3_util:q_val(couch_util:get_value(q, Options,
         config:get("cluster", "q", "8"))),
     %% rotate to a random entry in the nodelist for even distribution
-    {A, B} = lists:split(crypto:rand_uniform(1,length(Nodes)+1), Nodes),
+    {A, B} = lists:split(couch_rand:uniform(length(Nodes)+1), Nodes),
```
nickva: Note that `couch_rand:uniform(N)` has a `[1, N]` range, while `crypto:rand_uniform(1, N)` has `[1, N)` so we'd want to subtract 1 as in `...(couch_rand:uniform(length(Nodes))...`

nickva:
For behavior undefined issues try checking application dependencies in the .app.src file and also remove explicit `{modules, [...]}` lists as those should be automatically created.

Now in the case of couch and mem3 it's a tricky one as they would essentially need to depend on each other. Not sure how and if circular dependencies are handled.

src/couch_pse_tests/src/cpse_test_fold_changes.erl
```diff
@@ -170,25 +171,11 @@ do_mutation_ordering(Db, Seq, [{DocId, _OldSeq} | Rest], DocSeqAcc) ->
 
 
 shuffle(List) ->
-    random:seed(os:timestamp()),
-    Paired = [{random:uniform(), I} || I <- List],
+    Paired = [{couch_random:uniform(), I} || I <- List],
```
jiangphcn: `couch_rand:uniform/0` instead of `couch_random:uniform/0`?

src/couch_pse_tests/src/cpse_util.erl
```diff
@@ -28,6 +29,7 @@
     cpse_test_fold_purge_infos,
     cpse_test_purge_docs,
     cpse_test_purge_replication,
+    cpse_test_purge_seqs,
```
jiangphcn: Good catch

src/mem3/src/mem3.erl
jaydoane:
```
(node1@127.0.0.1)2> couch_rand:uniform(2).
2
(node1@127.0.0.1)3> couch_rand:uniform(2).
2
(node1@127.0.0.1)4> couch_rand:uniform(2).
1
(node1@127.0.0.1)5> couch_rand:uniform(2).
1
(node1@127.0.0.1)6> couch_rand:uniform(2).
1

(node1@127.0.0.1)7> crypto:rand_uniform(1,2).
1
(node1@127.0.0.1)8> crypto:rand_uniform(1,2).
1
(node1@127.0.0.1)9> crypto:rand_uniform(1,2).
1
(node1@127.0.0.1)10> crypto:rand_uniform(1,2).
1
(node1@127.0.0.1)11> crypto:rand_uniform(1,2).
1
(node1@127.0.0.1)12> crypto:rand_uniform(1,2).
1
```
whoa that is sneaky as heck, and will hopefully teach me to not make assumptions about the uniformity of word meaning. Thank you for pointing out what would have been a horrific bug.

src/couch_pse_tests/src/cpse_test_fold_changes.erl
jaydoane: That is indeed a typo. Good catch. It appears that those tests don't normally run using `make eunit` -- is that correct? How do you normally run them?

jaydoane: Yep, I see them now, thanks. I was trying to invoke them with the `make eunit apps=` syntax and having no luck.

src/couch/src/couch.erl
```diff
@@ -13,6 +13,7 @@
 -module(couch).
 
 -compile(export_all).
+-compile(nowarn_export_all).
```
iilyak: I would explicitly export only things we really need:
- start/0
- stop/0
- restart/0
- start_apps/1 (hmm I didn't know about it. We have something similar in test_util.)

jaydoane: Agreed, although it seems like `start_apps/1` should maybe not be exported, since I'm not finding it in use anywhere else.

jaydoane:
@nickva wrote:
> For behavior undefined issues try checking application dependencies in the .app.src file and also remove explicit {modules, [...]} lists as those should be automatically created.

I found an explicit `{modules, [...]}` term in numerous couchdb applications' .app.src files, but removing them did not suppress the warnings. Perhaps deleting those would be best left to a separate PR.

nickva:
@jaydoane 

Thanks for checking. We'd basically want mem3 compiled before say couch_peruser. I tried removing module from mem3 and moving it before couch and couch_peruser app in rebar config file. That cleared those 3 behavior warning. However because of the circular dependency between mem3 and couch another one popup up as expected:

```
src/mem3/src/mem3_sync_event_listener.erl:14: Warning: behaviour couch_event_listener undefined
```

```
diff --git a/rebar.config.script b/rebar.config.script
index 5f17c29e0..598097459 100644
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -26,6 +26,7 @@ SubDirs = [
     "src/couch_epi",
     "src/couch_log",
     "src/chttpd",
+    "src/mem3",
     "src/couch",
     "src/couch_index",
     "src/couch_mrview",
@@ -40,7 +41,6 @@ SubDirs = [
     "src/fabric",
     "src/global_changes",
     "src/mango",
-    "src/mem3",
     "src/rexi",
     "src/setup",
     "rel"
diff --git a/src/mem3/src/mem3.app.src b/src/mem3/src/mem3.app.src
index 99a9eed88..15efc64ca 100644
--- a/src/mem3/src/mem3.app.src
+++ b/src/mem3/src/mem3.app.src
@@ -13,20 +13,6 @@
 {application, mem3, [
     {description, "CouchDB Cluster Membership"},
     {vsn, git},
-    {modules, [
-        mem3,
-        mem3_app,
-        mem3_httpd,
-        mem3_nodes,
-        mem3_rep,
-        mem3_shards,
-        mem3_sup,
-        mem3_sync,
-        mem3_sync_event,
-        mem3_sync_nodes,
-        mem3_sync_security,
-        mem3_util
-    ]},
     {mod, {mem3_app, []}},
     {registered, [
         mem3_events,
```

so we trade 3 warning for 1. Still a bit better!

jaydoane:
During this process, I discovered two test suites that were not being run, but that would require additional work before they can be added to the suite.

1. [This test name is missing an underscore suffix](https://github.com/apache/couchdb/blob/master/src/couch_mrview/test/couch_mrview_index_changes_tests.erl#L28), which prevents it from being run correctly. However, modifying the name to `changes_index_test_` results in all tests failing, and it seems clear it never worked. Given the reliance on explicit `sleep` calls and the use of undefined functions, it seems like these tests will require some effort to get working robustly.

2. [This test module](https://github.com/apache/couchdb/blob/master/src/couch_pse_tests/src/cpse_test_purge_seqs.erl) is missing from [this list](https://github.com/apache/couchdb/blob/master/src/couch_pse_tests/src/cpse_util.erl#L22-L33), and thus does not get run. When added, there is a failure:
```
    cpse_gather: make_test_fun (cpse_increment_purge_seq_on_partial_purge)...*failed*
in function cpse_util:save_doc/3
  called as save_doc(<<"db-18b60b9572e53026d7326183fdc57786">>,{[{'_id',foo},{vsn,2}]},[replicated_changes])
in call from cpse_test_purge_seqs:cpse_increment_purge_seq_on_partial_purge/1 (src/cpse_test_purge_seqs.erl, line 104)
in call from eunit_proc:run_test/1 (eunit_proc.erl, line 510)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 335)
in call from eunit_proc:handle_test/2 (eunit_proc.erl, line 493)
in call from eunit_proc:tests_inorder/3 (eunit_proc.erl, line 435)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 325)
**error:undef
  output:<<"">>
```
which will need further investigation to fix.

jaydoane:
@nickva would you mind looking at this again, particularly the crypto & random stuff?

src/couch/src/couch_rand.erl
```diff
@@ -14,11 +14,34 @@
...
+%% where LoInclusive =< N < HiExclusive [1]
+%%
+%% rand:uniform(HiInclusive) and random:uniform(HiInclusive) both
+%% generate a random number N where 1 =< X =< HiInclusive [2][3].
+%%
+%% [1] http://erlang.org/doc/man/crypto.html#rand_uniform-2
+%% [2] http://erlang.org/doc/man/rand.html#uniform-1
+%% [3] http://erlang.org/doc/man/random.html#uniform-1
+
+crypto_to_rand(LoInclusive, HiExclusive)
```
nickva: Do we use this function anywhere? I did a page search in the browser and couldn't find any instances.

nickva:
@jaydoane 

Great cleanups! Rand stuff looks good. Made a minor comment about a function I couldn't find, otherwise +1.

It's unfortunate all those tests don't always run.

Also normally, I don't like commented blocks of code, but I see reason, perhaps someone would want to fix them in the future. If not, they can be removed later.

I'd wait for @iilyak or @jiangphcn to have another look at PSE tests and exports and then

src/couch/src/couch_rand.erl
jaydoane: No, but I used it for the crypto:rand_uniform/2 to rand:uniform/1 conversions, and thought I should "show my work". Happy to remove it, though, since it's really only related to this PR.

nickva: Yeah let's remove if it is not used

jiangphcn:
+1

src/couch_log/src/couch_log_writer_file.erl
```diff
@@ -38,6 +38,7 @@
 
 -ifdef(TEST).
 -compile(export_all).
+-compile(nowarn_export_all).
```
iilyak: Most of the functions from this module are private. I think it would be better to convert all `compile(export_all).` to explicit `-import([]).` of the functions which really need to be exported. I'll mark such places as well just so I wouldn't forget.

src/couch_log/src/couch_log_writer_syslog.erl
```diff
@@ -40,6 +40,7 @@
 
 -ifdef(TEST).
 -compile(export_all).
+-compile(nowarn_export_all).
```
iilyak: We don't need `export_all` here. Following functions have explicit export above:
- `init/0`
- `terminate/2`
- `write/2`

Following functions are used only internally (there are no external calls to these functions in the codebase):
- `send/2` used from `write/2`

Functions used from the test suite (see: https://github.com/apache/couchdb/blob/master/src/couch_log/test/couch_log_writer_syslog_test.erl):
- `get_facility/1` 
- `get_level/1` 

Please replace `-compile(export_all)` with:
```
-export([
    get_facility/1, 
    get_level/1
]).

File: src/couch_epi/src/couch_epi_sup.erl [hunk 8]
```diff
@@ -145,6 +145,7 @@ merge([{Id, _, _, _, _, _} = Spec | Rest], Children) ->
 %% BEGIN couch_epi_plugin behaviour callbacks
 
 -compile([export_all]).
+-compile(nowarn_export_all).
```
iilyak: Please explicitly export following functions:
```
-export([
   app/0,
   providers/0,
   services/0,
   data_providers/0,
   data_subscriptions/0,
   processes/0,
   notify/3
]).
```

src/couch_log/test/couch_log_test_util.erl
```diff
@@ -12,6 +12,7 @@
 
 -module(couch_log_test_util).
 -compile(export_all).
+-compile(nowarn_export_all).
```
iilyak: Please replace `-compile(export_all)` with:
```
-export([
   start/0,
   stop/1,
   with_level/2,
   with_config_listener/1,
   wait_for_config/0,
   last_log_key/0
]).
```

src/couch_mrview/src/couch_mrview_test_util.erl
```diff
@@ -13,6 +13,7 @@
 -module(couch_mrview_test_util).
 
 -compile(export_all).
+-compile(nowarn_export_all).
```
iilyak: In this case `-compile(export_all).` and `-compile(nowarn_export_all).` is appropriate. Since all implemented functions are used externally in the test suite.

src/couch_pse_tests/src/cpse_test_attachments.erl
```diff
@@ -12,6 +12,7 @@
 
 -module(cpse_test_attachments).
 -compile(export_all).
+-compile(nowarn_export_all).
```
iilyak: All functions in this module need to be exported. Since we have a test discovery based on `cpse_` prefix (see https://github.com/apache/couchdb/blob/master/src/couch_pse_tests/src/cpse_gather.erl#L31). Which means that it is ok to use `export_all` here.

src/mem3/src/mem3_rep.erl
```diff
@@ -137,7 +137,6 @@ verify_purge_checkpoint(DbName, Props) ->
             Target = binary_to_existing_atom(TargetBin, latin1),
 
             try
-                Shards = mem3:shards(DbName),
```
iilyak: Good catch.

src/couch_log/src/couch_log_writer_file.erl
jaydoane: I agree, and thanks for pointing out where it's appropriate to do so. In this case, it looks like we can simply replace the `export_all` with just
```
-export([
    maybe_reopen/1
]).
```

src/couch_log/src/couch_log_writer_syslog.erl
jaydoane: Done

src/couch_epi/src/couch_epi_sup.erl
jaydoane: After consulting with @iilyak, I moved the test in couch_epi_sup to a new file couch_epi_basic_test , changing it as necessary to pass, and exporting `plugin_childspecs/3` from couch_epi_sup for testing.

src/couch_log/test/couch_log_test_util.erl
jaydoane: Actually, both `last_log/0` and `with_meck/0` also need to be exported, so the full export attribute becomes:
```
-export([
    start/0,
    stop/1,
    last_log/0,
    last_log_key/0,
    wait_for_config/0,
    with_config_listener/1,
    with_level/2,
    with_meck/2
]).
```

PR #1804, opened 2018-12-10
Make view query limits configurable

garrensmith:
## Overview

This allows us to set a maximun allowed number of documents
to be returned for a global or a partitioned query.

## Testing recommendations

Set a view limit in the default.ini. 

* A query with a limit larger than the max will error
* A query with the limit value below the max will run the query. A
* A query with no limit will use the max limit set.

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -293,6 +293,8 @@ os_process_limit = 100
 ; Timeout for how long a response from a busy view group server can take.
 ; "infinity" is also a valid configuration value.
 ;group_info_timeout = 5000
+;query_limit = 268435456
+;partition_query_limit = 268435456
 
 [mango]
 ; Set to true to disable the "index all fields" text index, which can lead
--- a/src/couch_mrview/src/couch_mrview_util.erl
+++ b/src/couch_mrview/src/couch_mrview_util.erl
@@ -496,9 +496,10 @@ fold_reduce({NthRed, Lang, View}, Fun,  Acc, Options) ->
     couch_btree:fold_reduce(Bt, WrapperFun, Acc, Options).
 
 
-validate_args(Db, DDoc, Args) ->
+validate_args(Db, DDoc, Args0) ->
     {ok, State} = couch_mrview_index:init(Db, DDoc),
-    validate_args(State, Args).
+    Args1 = apply_limit(State#mrst.partitioned, Args0),
+    validate_args(State, Args1).
 
 
 validate_args(#mrst{} = State, Args0) ->
@@ -523,6 +524,28 @@ validate_args(#mrst{} = State, Args0) ->
     end.
 
 
+apply_limit(ViewPartitioned, Args) ->
+    LimitType = case ViewPartitioned of
+        true -> "partition_query_limit";
+        false -> "query_limit"
+    end,
+
+    MaxLimit = config:get_integer("query_server_config",
+        LimitType, ?MAX_VIEW_LIMIT),
+
+    % Set the highest limit possible if a user has not
+    % specified a limit
+    Args1 = case Args#mrargs.limit == ?MAX_VIEW_LIMIT of
+        true -> Args#mrargs{limit = MaxLimit};
+        false -> Args
+    end,
+
+    if Args1#mrargs.limit =< MaxLimit -> Args1; true ->
+        Fmt = "Limit is too large, must not exceed ~p",
+        mrverror(io_lib:format(Fmt, [MaxLimit]))
+    end.
+
+
 validate_all_docs_args(Db, Args0) ->
     Args = validate_args(Args0),
 
@@ -533,7 +556,8 @@ validate_all_docs_args(Db, Args0) ->
         {false, <<_/binary>>} ->
             mrverror(<<"`partition` parameter is not supported on this db">>);
         {_, <<_/binary>>} ->
-            apply_all_docs_partition(Args, Partition);
+            Args1 = apply_limit(true, Args),
+            apply_all_docs_partition(Args1, Partition);
         _ ->
             Args
     end.
--- a/test/elixir/test/partition_all_docs_test.exs
+++ b/test/elixir/test/partition_all_docs_test.exs
@@ -116,36 +116,82 @@ defmodule PartitionAllDocsTest do
     assert ids == ["foo:22"]
   end
 
+  test "partition all docs can set query limits", context do
+    set_config({"query_server_config", "partition_query_limit", "2000"})
+
+    db_name = context[:db_name]
+    create_partition_docs(db_name)
+    create_partition_ddoc(db_name)
+
+    url = "/#{db_name}/_partition/foo/_all_docs"
+
+    resp =
+      Couch.get(url,
+        query: %{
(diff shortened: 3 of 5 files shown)


src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -525,6 +527,39 @@ validate_args(#mrst{} = State, Args0) ->
     end.
 
 
+apply_limit(ViewPartitioned, Args) ->
+    MaxLimitType = case ViewPartitioned of
+        true -> "partition_";
```
davisp: I think it'd be easier to understand this bit if we just type out the config names rather than doing the concatenation down below.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -525,6 +527,39 @@ validate_args(#mrst{} = State, Args0) ->
...
+        false -> ""
+    end,
+    
+    MaxLimit = case config:get("query_server_config",
```
davisp: Using `config:get_integer("query_server_config", MaxLimit, ?MAX_VIEW_LIMIT)` would clean up the logic and avoid having to do the whole infinity dance.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -525,6 +527,39 @@ validate_args(#mrst{} = State, Args0) ->
...
+        andalso MaxLimit =/= infinity of
+        false ->
+            Args;
+        true ->
+            Args#mrargs{
+                limit = MaxLimit
+            }
+    end,
+
+    case Args1#mrargs.limit > MaxLimit of
```
davisp: The if statement would make this a bit cleaner I think:

```erlang
if Args1#mrargs.limit < MaxLimit -> ok; true ->
    Fmt = "...".
    mrverror(...)
end,

Args1.
```

rel/overlay/etc/default.ini
```diff
@@ -283,6 +283,8 @@ os_process_limit = 100
 ; Timeout for how long a response from a busy view group server can take.
 ; "infinity" is also a valid configuration value.
 ;group_info_timeout = 5000
+;query_limit = 2000
```
davisp: The value that's commented out should match the default of `?MAX_VIEW_LIMIT`.

src/couch_mrview/src/couch_mrview_util.erl
garrensmith: From what I can see `config:get_integer` cannot handle `infinity` as an input.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -525,6 +527,33 @@ validate_args(#mrst{} = State, Args0) ->
...
+        true -> "partition_query_limit";
+        false -> "query_limit"
+    end,
+    
+    MaxLimit = config:get_integer("query_server_config",
+        LimitType, ?MAX_VIEW_LIMIT),
+
+    % set the highest limit possible if a user has not 
+    % specified a limit
+    Args1 = case Args#mrargs.limit =:= ?MAX_VIEW_LIMIT
```
davisp: Super minor nit but I'd probably use `==` instead of `=:=` here. Only because I had to stop and think on whether we needed the stronger `=:=` for any reason or not.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -525,6 +527,33 @@ validate_args(#mrst{} = State, Args0) ->
...
+        andalso MaxLimit =/= infinity of
```
davisp: No need for this clause checking `infinity` anymore.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -523,6 +525,33 @@ validate_args(#mrst{} = State, Args0) ->
...
+        false ->
+            Args;
+        true ->
+            Args#mrargs{
+                limit = MaxLimit
+            }
+    end,
+
+    if Args1#mrargs.limit =< MaxLimit -> Args1; true ->
+        Fmt = "Limit is too large, must not exceed ~p",
```
davisp: Something is wrong here but I'm not immediately seeing why. The view_multi_key_design.js test has failed with this error message on each worker. Perhaps we left a bad config from an unrelated test?

garrensmith: I found the issue. It was because of the `fix_skip_and_limit` function https://github.com/apache/couchdb/blob/master/src/fabric/src/fabric_view.erl#L380

We made the max limit greater than the defined max limit.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -116,6 +116,7 @@ get_view_index_state(Db, DDoc, ViewName, Args0, RetryCount) ->
         UpdateSeq = couch_util:with_db(Db, fun(WDb) ->
             couch_db:get_update_seq(WDb)
         end),
+        io:format("GET VIEW ~p ~n", [Args]),
```
davisp: Debug logging needs to be removed.

src/couch_mrview/src/couch_mrview_util.erl
davisp: I see the issue there but I'm not seeing what changed to account for it.

src/couch_mrview/src/couch_mrview_util.erl
garrensmith: I changed this line from `if Args1#mrargs.limit =:= MaxLimit -> Args1; true ->` to
`if Args1#mrargs.limit =< MaxLimit -> Args1; true ->`

test/elixir/test/partition_all_docs_test.exs
```diff
@@ -116,6 +116,43 @@ defmodule PartitionAllDocsTest do
     assert ids == ["foo:22"]
   end
 
+  test "partition all docs can set query limits", context do
+    resp = Couch.get("/_membership")
+    %{:body => body} = resp
+
+    Enum.each(body["all_nodes"], fn node ->
+      resp = Couch.put("/_node/#{node}/_config/query_server_config/partition_query_limit", body: "\"2000\"")
```
davisp: Re-reading again, and just noticed this. You should use set_config which takes care of removing the config change after the test finishes. There's an example here:

https://github.com/apache/couchdb/blob/8d2af71f2896d5f75ceb0f780993493fcecdf49d/test/elixir/test/partition_size_limit_test.exs#L15

davisp: And samesies for partition_view_test

davisp:
Also, to force the bug, the skip would have to be set to `Q * $max_rows` because the current apply_limit throws away the skip bump to the limit we'd end up returning 0 rows when there was actual data to return.

That is of course an unlikely scenario and goes against the general recommendation to not use a large skip value but it'd still be an incorrect response if someone did get into that situation where maybe `$max_rows` was set low enough to be not super crazy.

davisp:
Turns out fabric_util was too far up. However we can just move it up one call so the RPC workers don't call it again and that fixes the issue.

PR #1808, opened 2018-12-13
Update before_doc_update/2 to before_doc_update/3

jiangphcn:
## Overview

Pass update type to before_doc_update/3 to allow additional check. The possible update type includes  
 - interactive_edit
 - replicated_changes

## Testing recommendations

```
make check skip_deps+=couch_epi apps=couch tests=callback_test_

```
## Checklist

### DIFF
--- a/src/couch/src/couch_db_plugin.erl
+++ b/src/couch/src/couch_db_plugin.erl
@@ -14,7 +14,7 @@
 
 -export([
     validate_dbname/3,
-    before_doc_update/2,
+    before_doc_update/3,
     after_doc_read/2,
     validate_docid/1,
     check_is_admin/1,
@@ -34,11 +34,13 @@
 validate_dbname(DbName, Normalized, Default) ->
     maybe_handle(validate_dbname, [DbName, Normalized], Default).
 
-before_doc_update(Db, Doc0) ->
+before_doc_update(Db, Doc0, UpdateType) ->
     Fun = couch_db:get_before_doc_update_fun(Db),
-    case with_pipe(before_doc_update, [Doc0, Db]) of
-        [Doc1, _Db] when is_function(Fun) -> Fun(Doc1, Db);
-        [Doc1, _Db] -> Doc1
+    case with_pipe(before_doc_update, [Doc0, Db, UpdateType]) of
+        [Doc1, _Db, UpdateType1] when is_function(Fun) ->
+            Fun(Doc1, Db, UpdateType1);
+        [Doc1, _Db, _UpdateType] ->
+            Doc1
     end.
 
 after_doc_read(Db, Doc0) ->
--- a/src/couch_replicator/src/couch_replicator_docs.erl
+++ b/src/couch_replicator/src/couch_replicator_docs.erl
@@ -18,7 +18,7 @@
     parse_rep_db/3,
     parse_rep_doc_without_id/1,
     parse_rep_doc_without_id/2,
-    before_doc_update/2,
+    before_doc_update/3,
     after_doc_read/2,
     ensure_rep_db_exists/0,
     ensure_rep_ddoc_exists/1,
@@ -615,10 +615,10 @@ ssl_verify_options(false) ->
     [{verify, verify_none}].
 
 
--spec before_doc_update(#doc{}, Db::any()) -> #doc{}.
-before_doc_update(#doc{id = <<?DESIGN_DOC_PREFIX, _/binary>>} = Doc, _Db) ->
+-spec before_doc_update(#doc{}, Db::any(), couch_db:update_type()) -> #doc{}.
+before_doc_update(#doc{id = <<?DESIGN_DOC_PREFIX, _/binary>>} = Doc, _Db, _UpdateType) ->
     Doc;
-before_doc_update(#doc{body = {Body}} = Doc, Db) ->
+before_doc_update(#doc{body = {Body}} = Doc, Db, _UpdateType) ->
     #user_ctx{
        roles = Roles,
        name = Name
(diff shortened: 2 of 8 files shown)


src/fabric/src/fabric_doc_update.erl
```diff
@@ -106,11 +106,11 @@ before_doc_update(DbName, Docs, Opts) ->
         {true, _} ->
             %% fake db is expensive to create so we only do it if we have to
             Db = fabric_util:fake_db(DbName, Opts),
-            [couch_replicator_manager:before_doc_update(Doc, Db) || Doc <- Docs];
+            [couch_replicator_manager:before_doc_update(Doc, Db, []) || Doc <- Docs];
```
iilyak: According to update description of the PR the update type can be either:
- replicated_changes
- interactive_updates

src/couch/src/couch_db_plugin.erl
```diff
@@ -34,11 +34,13 @@
...
     maybe_handle(validate_dbname, [DbName, Normalized], Default).
 
-before_doc_update(Db, Doc0) ->
+before_doc_update(Db, Doc0, UpdateType) ->
     Fun = couch_db:get_before_doc_update_fun(Db),
-    case with_pipe(before_doc_update, [Doc0, Db]) of
-        [Doc1, _Db] when is_function(Fun) -> Fun(Doc1, Db);
-        [Doc1, _Db] -> Doc1
+    case with_pipe(before_doc_update, [Doc0, Db, UpdateType]) of
+        [Doc1, _Db, UpdateType] when is_function(Fun) ->
```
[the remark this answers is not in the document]
iilyak: this is still an issue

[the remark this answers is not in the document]
jiangphcn: No problem, Ilya. I think that I got your point now. Totally make sense and I should use different name after calling `with_pipe/2`. The change was reflected in recent commit.

src/couch/src/couch_db_plugin.erl
```diff
@@ -34,11 +34,13 @@
...
 
-before_doc_update(Db, Doc0) ->
+before_doc_update(Db, Doc0, UpdateType) ->
     Fun = couch_db:get_before_doc_update_fun(Db),
-    case with_pipe(before_doc_update, [Doc0, Db]) of
-        [Doc1, _Db] when is_function(Fun) -> Fun(Doc1, Db);
-        [Doc1, _Db] -> Doc1
+    case with_pipe(before_doc_update, [Doc0, Db, UpdateType]) of
+        [Doc1, _Db, _UpdateType] when is_function(Fun) ->
+            Fun(Doc1, Db, _UpdateType);
```
iilyak: Ahh. I didn't notice earlier that you are passing `_UpdateType` to Fun. In this case it is better to rename it into `UpdateType1` or something. According to erlang's conventions (roughly).

- Use new name for variable when the value need to be changed (usually the integer is added as suffix)
- Use `_Name` style for variables you don't care about.

In this case we do care about the variable since we are passing it to another function. Also we cannot use `UpdateType` since the value returned from `with_pipe` can differ from the passed value. In this case we would have a badmatch.

src/couch_replicator/src/couch_replicator_docs.erl
```diff
@@ -615,10 +615,10 @@ ssl_verify_options(false) ->
     [{verify, verify_none}].
 
 
--spec before_doc_update(#doc{}, Db::any()) -> #doc{}.
-before_doc_update(#doc{id = <<?DESIGN_DOC_PREFIX, _/binary>>} = Doc, _Db) ->
+-spec before_doc_update(#doc{}, Db::any(), any()) -> #doc{}.
```
iilyak: we shouldn't use `any()` here. By using `any` we disable dializer. Try defining `update_type()` in `couch_db` and refer to it using `couch_db:update_type()`.

```
diff --git a/src/couch/src/couch_db.erl b/src/couch/src/couch_db.erl
index 1f8bc42..43ed562 100644
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -135,6 +135,10 @@
 -include("couch_db_int.hrl").

+-type update_type()
+    :: replicated_changes
+        | interactive_edit.
+
```

src/fabric/src/fabric_doc_update.erl
```diff
@@ -106,11 +106,13 @@ before_doc_update(DbName, Docs, Opts) ->
         {true, _} ->
             %% fake db is expensive to create so we only do it if we have to
             Db = fabric_util:fake_db(DbName, Opts),
-            [couch_replicator_docs:before_doc_update(Doc, Db) || Doc <- Docs];
+            [couch_replicator_docs:before_doc_update(Doc, Db, interactive_updates)
```
iilyak: Hmm. I couldn't find where `interactive_updates` is handled. I think it is typo and should be `interactive_edit`.

src/fabric/src/fabric_doc_update.erl
```diff
@@ -106,11 +106,13 @@ before_doc_update(DbName, Docs, Opts) ->
...
+                || Doc <- Docs];
         {_, true} ->
             %% fake db is expensive to create so we only do it if we have to
             Db = fabric_util:fake_db(DbName, Opts),
-            [couch_users_db:before_doc_update(Doc, Db) || Doc <- Docs];
+            [couch_users_db:before_doc_update(Doc, Db, interactive_updates)
```
iilyak: I think it should be `interactive_edit` unless `interactive_updates` is a new thing defined in some other dependent PR I am not aware of.

src/fabric/src/fabric_doc_update.erl
jiangphcn: Good catch. Because this is for placeholder and I directly use https://github.com/apache/couchdb/pull/1808#discussion_r242498749. I already change it to replicated_changes.

src/fabric/src/fabric_doc_update.erl
jiangphcn: yes, change it to interactive_edit

jiangphcn:
Thanks @iilyak for careful review. I addressed your comments. I will rebase and merge them if getting approval later.

PR #1811, opened 2018-12-15
Add new /{db}/_sync_shards endpoint (admin-only)

### LINKED ISSUE #1807: Add /{db}/_shard_sync endpoint to force internal sync of all shards
## Current and expected behavior
Right now, when replacing or introducing new nodes into a cluster, we have a [defined process in our documentation](http://docs.couchdb.org/en/stable/cluster/sharding.html?#monitor-internal-replication-to-ensure-up-to-date-shard-s) that states:

> After you complete the previous step, as soon as CouchDB receives a write request for a shard on the target node, CouchDB will check if the target node’s shard(s) are up to date. If it finds they are not up to date, it will trigger an internal replication job to complete this task. You can observe this happening by triggering a write to the database (update a document, or create a new one), while monitoring the /_node/<nodename>/_system endpoint, which includes the internal_replication_jobs metric.
> 
> Once this metric has returned to the baseline from before you wrote the document, or is 0, the shard replica is ready to serve data and we can bring the node out of maintenance mode.

This is all well and good, but if you have a system with any number of idle databases (especially true for db-per-user setups or similar) or a number of databases that are mostly read-only, the sync will never be triggered. Should another node failure occur (assuming `n=3`) you would start failing quorum checks; lose a third node and you'd definitely experience data loss.

This isn't good.

By introducing a new `/{db}/_shard_sync` server-admin-only endpoint, a server admin should be able to force these internal replication jobs to trigger. One approach to making this work was provided by @chewbranca in [this gist](https://gist.github.com/chewbranca/055d50be1173cb016982b21aa2c3674e) (untested!) which should immediately trigger internal shard replications for a given shard between all nodes. This code still needs to be passed a list of shards to replicate, so to work at the db level, you'd want to wrap this in another comprehension that walks across all shards of a given DB.

Writing a test case to validate this given our current eunit framework will be a bit painful, but I bet it could be done using the without-quorum type tests we currently have in JS (and @jjrodrig is working to convert to elixir).

We could then update our documentation to include the following text:

> After you complete the previous step, *as server admin, `POST /{db}/_compact` for each database moved, one database at a time. This will tell CouchDB to trigger internal replication jobs between all copies of all shards in that database. You can then monitor the /_node/<nodename>/_system endpoint, which includes the internal_replication_jobs metric, indicating how many of these jobs are currently running. 
>
> Once this metric has returned to the baseline from before you wrote the document, or is 0, the new node is ready to serve data for that database. We can then repeat for other databases, or, if done, we can bring the node out of maintenance mode and return the cluster to normal operation.

We may also want to include advice about allowing more sync concurrency temporarily while these operations occur:

> During the sync operation, one internal sync job is started for each permutation of 2 shard copies. For a default cluster (`n=3, q=8`) this means that syncing all shards for a given database will start 48 sync jobs ( 8 shards × P(3,2) = 48). However, by default, CouchDB will perform up to 10 simultaneous shard sync operations. This can create a bottleneck. You can temporarily increase the `[mem3] sync_concurrency` value to a larger figure to allow the cluster to consume more resources for the sync operation. Be sure to lower this value back to default after synchronization is complete.

## Workaround
Today's (CouchDB 2.0.0 thru 2.3.0) workaround would be to follow [this section in our documentation](http://docs.couchdb.org/en/stable/cluster/sharding.html?#copying-shard-files) which relies on copying shard files over from known-good nodes to the newly added nodes.

As this isn't automated, this process is error-prone, and requires additional tooling on the part of the administrator we don't currently provide. It also doesn't help if there are writes during the copy, but the DB is once again quiescent after the new node is brought out of maintenance mode. We can do better.

## Possible Solution
Start from [this gist](https://gist.github.com/chewbranca/055d50be1173cb016982b21aa2c3674e), and wrap another list comprehension around it to get all of the shard names for a given db. Expose at the new `/{db}/_sync` endpoint, and ensure that endpoint's access is server admin-level access only.

## Context
Users of CouchDB with many databases who lose a node and don't want to scp shards around have a hard time ensuring the newly added node to a database has been fully "caught up" with all of the data necessary.

wohali:
## Overview

This server admin-only endpoint forces an n-way sync of all shards
across all nodes on which they are hosted.

This can be useful for an administrator adding a new node to the
cluster, after updating _dbs so that the new node hosts an existing db
with content, to force the new node to sync all of that db's shards.

Users may want to bump their `[mem3] sync_concurrency` value to a
larger figure for the duration of the shards sync.

Closes #1807

## Testing recommendations

Testing this requires a non-standard setup, so I have not written an
automated test case for this. (We should think about how this could
possibly be done in the Elixir test suite in the future.)

Manual test script:

```console
# set up 2 nodes not joined together and put data in node1
dev/run -n 2 --no-join &
export AUTH="root:PASSWORD_GOES_HERE"
curl -X PUT http://$AUTH@localhost:15984/foo
curl -X PUT http://$AUTH@localhost:15984/foo/abc -d '{"a":"b"}'
curl -X PUT http://$AUTH@localhost:15984/foo/def -d '{"g":"h"}'

# see the shards with data in node1, here it's shards 00 and 20
ls -lR dev/lib/node1/data/shards/*/foo.*.couch

# join the nodes together
curl -X PUT http://$AUTH@localhost:15986/_nodes/node2@127.0.0.1 -d '{}'
curl http://$AUTH@localhost:15984/_membership
curl http://$AUTH@localhost:25984/_membership

# have node2 participate in serving foo for all shards
curl -X PUT http://$AUTH@localhost:15986/_dbs/foo -d '{"_id":"foo","_rev":"1-d0f95fcbe46dd04f66bc61906e202d56","shard_suffix":[46,49,53,52,52,56,51,52,53,49,51],"changelog":[["add","00000000-1fffffff","node1@127.0.0.1"],["add","20000000-3fffffff","node1@127.0.0.1"],["add","40000000-5fffffff","node1@127.0.0.1"],["add","60000000-7fffffff","node1@127.0.0.1"],["add","80000000-9fffffff","node1@127.0.0.1"],["add","a0000000-bfffffff","node1@127.0.0.1"],["add","c0000000-dfffffff","node1@127.0.0.1"],["add","e0000000-ffffffff","node1@127.0.0.1"]],"by_node":{"node1@127.0.0.1":["00000000-1fffffff","20000000-3fffffff","40000000-5fffffff","60000000-7fffffff","80000000-9fffffff","a0000000-bfffffff","c0000000-dfffffff","e0000000-ffffffff"],"node2@127.0.0.1":["00000000-1fffffff","20000000-3fffffff","40000000-5fffffff","60000000-7fffffff","80000000-9fffffff","a0000000-bfffffff","c0000000-dfffffff","e0000000-ffffffff"]},"by_range":{"00000000-1fffffff":["node1@127.0.0.1","node2@127.0.0.1"],"20000000-3fffffff":["node1@127.0.0.1","node2@127.0.0.1"],"60000000-7fffffff":["node1@127.0.0.1","node2@127.0.0.1"],"80000000-9fffffff":["node1@127.0.0.1","node2@127.0.0.1"],"a0000000-bfffffff":["node1@127.0.0.1","node2@127.0.0.1"],"c0000000-dfffffff":["node1@127.0.0.1","node2@127.0.0.1"],"e0000000-ffffffff":["node1@127.0.0.1","node2@127.0.0.1"]}}'

# double check shards in node2 are still empty (all should be same file size)
ls -lR dev/lib/node2/data/shards/*/foo.*.couch

# do a shard sync
curl -X POST http://$AUTH@localhost:15984/foo/_sync_shards

# check that shards 00 and 22 on node2 now have data from node1
ls -lR dev/lib/node2/data/shards/*/foo.*.couch

# optionally manually check docs in those shards
# substitute the correct timestamp in the shard name here
curl http://$AUTH@localhost:25986/shards%2F00000000-1fffffff%2Ffoo.1544834513/_all_docs
curl http://$AUTH@localhost:25986/shards%2F20000000-3fffffff%2Ffoo.1544834513/_all_docs

```

## Checklist

Documentation PR will be issued once this PR (and the bikeshedding over the name of the new endpoint) is approved.


wohali:
While I was in `mem3_httpd` I noticed the (currently undocumented) `/{db}/_shards` and `/{db}/_shards/{docid}` endpoints.

We should document these.

I dithered briefly over putting this new endpoint at `/{db}/_shards/_sync`. Happy to oblige if others feel strongly.

davisp:
+1

I don't think I'd nest it since its not quite the same.

src/mem3/src/mem3_httpd.erl
```diff
@@ -52,6 +53,16 @@ handle_shards_req(#httpd{path_parts=[_DbName, <<"_shards">>]}=Req, _Db) ->
 handle_shards_req(#httpd{path_parts=[_DbName, <<"_shards">>, _DocId]}=Req, _Db) ->
     chttpd:send_method_not_allowed(Req, "GET").
 
+handle_sync_req(#httpd{method='POST',
+        path_parts=[_DbName, <<"_sync_shards">>]} = Req, Db) ->
+    DbName = mem3:dbname(couch_db:name(Db)),
+    ShardList = [S#shard.name || S <- mem3:ushards(DbName)],
+    [ sync_shard(S) || S <- ShardList ],
+    chttpd:send_json(Req, 202, {[{ok, true}]});
+handle_sync_req(#httpd{method='GET'}=Req, _Db) ->
```
rnewson: best to catch all non-POST methods here so we send an error response for every method.

wohali:
TY @davisp - will leave it as is.

Documentation PR to land after this one.

rnewson:
rebase before merging though

PR #1820, opened 2018-12-19
Increase timeout on restart in JS/elixir tests to 30s

wohali:
## Overview

Currently, when a JS or Elixir test restarts a node, it waits 15 or 10 seconds respectively for it to come back.

CI systems have shown this is not enough. Most recent failure is here:
    https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/448/pipeline
but it's been happening semi-regularly since we reworked the restart code.

This PR bumps the timeout to 30s in both places.


wohali:
thanks @rnewson

Issue #1821, opened 2018-12-19
Cryptic error for chttpd_auth_cache


akircher:
I apologize if this is documented elsewhere, but I am seeing a lot of the error below on our production server.  Is this okay?  If not, is this an error in our code or in couchdb?  To my knowledge none of our views/validation use a function called "username". Thank you in advance for your time.

```[notice] 2018-12-19T23:46:43.281783Z couchdb@127.0.0.1 <0.332.0> -------- chttpd_auth_cache changes listener died no function clause matching chttpd_auth_cache:username(<<"e43caea614fb6e2250df47e4da043d64">>)(line:18$
[error] 2018-12-19T23:46:43.281838Z couchdb@127.0.0.1 emulator -------- Error in process <0.8490.6970> on node 'couchdb@127.0.0.1' with exit value:
{function_clause,[{chttpd_auth_cache,username,[<<"e43caea614fb6e2250df47e4da043d64">>],[{file,"src/chttpd_auth_cache.erl"},{line,183}]},{chttpd_auth_cache,changes_callback,2,[{file,"src/chttpd_auth_cache.erl"},{line,$```


## Expected Behavior
<!--- If you're describing a bug, tell us what should happen -->
No notice or error is logged


## Your Environment
This is running on Couch DB 2.2.0, Node 10.9.0, Ubuntu 16.04.5

wohali:
That's a definite bug - are you able to try Couch 2.3.0 to see if this is fixed?

eiri:
This is not from a db's VDU, but from [auth_cache](https://github.com/apache/couchdb/blob/master/src/chttpd/src/chttpd_auth_cache.erl#L191). The field `username` of user's doc [has to have `org.couchdb.user:` prefix](https://docs.couchdb.org/en/latest/intro/security.html?highlight=username#why-the-org-couchdb-user-prefix).

This is not exactly a bug, more of exception not been handled gracefully, because assumption here is that validation on `_users` database should prevent creation of users without that prefix, so they can't get into the cache.

@akircher Check out for a doc `e43caea614fb6e2250df47e4da043d64` in `_users` database or whatever database set in config's `[chttpd_auth]` as `authentication_db`.

wohali:
@eiri agreed there should only be docs in `_users` with the `org.couchdb.user:` prefix, but we probably should catch the situation where the wrongly named doc is in that DB and not toss a `function_clause` error.

eiri:
[...]

PR #1827, opened 2018-12-22
Handle invalid user docid in changes_callback

popojargo:
## Overview

Handle user document id without correct id. I simply catch any errors and log a debug message (perhaps we want a info or warn level message?)

## Testing recommendations

1. Disable the VDU in _users database (I removed the validate function)
2. Set the log level to Debug
3. Add an valid user without an id (let Couch generate one)
4. Replicate the user database to another database to trigger the changes_callback function.
5. You should see the debug level saying that the id is invalid and the cache reader should not throw a function clause. 

#1821 

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_auth_cache.erl
+++ b/src/chttpd/src/chttpd_auth_cache.erl
@@ -160,9 +160,12 @@ changes_callback({change, {Change}}, _) ->
         <<"_design/", _/binary>> ->
             ok;
         DocId ->
-            UserName = username(DocId),
-            couch_log:debug("Invalidating cached credentials for ~s", [UserName]),
-            ets_lru:remove(?CACHE, UserName)
+            try
+                UserName = username(DocId),
+                couch_log:debug("Invalidating cached credentials for ~s", [UserName]),
+                ets_lru:remove(?CACHE, UserName)
+            catch _:_ -> couch_log:info("Invalid user document id for ~s", [DocId])
+            end
     end,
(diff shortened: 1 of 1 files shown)


src/chttpd/src/chttpd_auth_cache.erl
```diff
@@ -160,9 +160,12 @@ changes_callback({change, {Change}}, _) ->
         <<"_design/", _/binary>> ->
             ok;
         DocId ->
-            UserName = username(DocId),
-            couch_log:debug("Invalidating cached credentials for ~s", [UserName]),
-            ets_lru:remove(?CACHE, UserName)
+            try
+                UserName = username(DocId),
+                couch_log:debug("Invalidating cached credentials for ~s", [UserName]),
+                ets_lru:remove(?CACHE, UserName)
```
jaydoane: Isn't it a problem that we don't remove the doc from the cache if it has an invalid name? It seems like it should be removed if it's stale, despite it having an unexpected prefix.

Is it possible (or desirable) to prevent such a doc from even entering the cache at all?

popojargo: I'll look into this this weekend. I would be indeed better to simply prevent the doc to enter the cache. Altought, if the user document is invalid, can we still use the user credentials to login? That's something I will check today.

wohali:
@popojargo any plans to get this in? I am preparing to mass-close old PRs that never got merged.

Issue #1558, opened 2018-08-14
Prevent modification of /_users/_security object


wohali:
## Expected Behavior
Modifying the `_security` object inside of `_users` is unsupported, and can lead to some unusual behaviour - see #1556.

## Current Behavior
We allow people to shoot themselves in the foot by modifying `_users/_security`.

## Possible Solution
Always return a `403` on write attempts for `_users/_security`. @rnewson do you have any comment on this?

## Steps to Reproduce (for bugs)
1. `dev/run -n 1 --with-admin-party-please`
2. `curl -X PUT http://localhost:15984/_users/_security -d '{"foo": "bar"}'

## Context
People are trying to change the rules for who can read/write documents in `_users` and it goes very badly.

## Your Environment
* Version used: 2.2.0

rnewson:
It's a difficult thing to remove at this late stage, though I agree with the proposal.

Should we fold this into the broader push to hide these databases behind an API that enforces only the semantics we wish?

wohali:
@rnewson If a naive block is a problem for 2.x, we can just say this is backwards-incompatible and fold it into #1504 if you want, sure. I just thought maybe a 2.3 that blocked `_users/_security` writes wouldn't be too hard to implement, and shouldn't affect TOO many people...

janl:
Alternatively, like we've done with default db security and making _all_dbs an opt-in config value could help folks until we have a proper fix.

rnewson:
yes, I can see that. [couchdb] users_db_security_editable=false|true or something, defaulting to false in the new default.ini

wohali:
If we're OK with that alternative for 2.x, I'll re-open this.

PR #1828, opened 2018-12-22
Allow to set the user db security object to readonly

popojargo:
## Overview

- Add the default config 
- Deny update on _security if the database is the user db and if the config is to false
- Add unit test

#1558 

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -658,6 +658,8 @@ db_req(#httpd{path_parts=[_,<<"_revs_diff">>]}=Req, _Db) ->
 
 db_req(#httpd{method='PUT',path_parts=[_,<<"_security">>],user_ctx=Ctx}=Req,
         Db) ->
+    DbName = ?b2l(couch_db:name(Db)),
+    validate_security_can_be_edited(DbName),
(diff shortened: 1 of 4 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -1696,6 +1698,15 @@ extract_header_rev(Req, ExplicitRev) ->
         throw({bad_request, "Document rev and etag have different values"})
     end.
 
+validate_security_can_be_edited(DbName) ->
+    UserDbName = config:get("chttpd_auth", "authentication_db", "_users"),
+    CanEditUserSecurityObject = config:get("couchdb","users_db_security_editable","false"),
+    case {DbName,CanEditUserSecurityObject} of
+        {UserDbName,"false"} ->
+            Msg = "You can't edit the security object of the user database.",
```
popojargo: Normally the error message is something like this: not_found
Should I replace with a similar message? Do you have any suggestions?

janl: this is fine, `not_found` here is the `forbidden` in the next line. The message is fine.

janl:
Merged in https://github.com/apache/couchdb/pull/2395

PR #1842, opened 2019-01-03
Implement partition size limits

davisp:
## Overview

This implements configurable size limits for partitions (which are not yet merged). This is a separate PR due to the complexity involved.

The point of enforcing limits is so that users don't accidentally footgun themselves by trying to store an entire db using a single partition key. It can be disabled by setting the limit to 0.

## Testing recommendations

`make check`
`make elixir`

#1789

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -873,6 +873,11 @@ error_info(conflict) ->
     {409, <<"conflict">>, <<"Document update conflict.">>};
 error_info({conflict, _}) ->
     {409, <<"conflict">>, <<"Document update conflict.">>};
+error_info({partition_overflow, DocId}) ->
+    Descr = <<
+            "Partition limit exceeded due to update on '", DocId/binary, "'"
+        >>,
+    {403, <<"partition_overflow">>, Descr};
 error_info({{not_found, missing}, {_, _}}) ->
     {409, <<"not_found">>, <<"missing_rev">>};
 error_info({forbidden, Error, Msg}) ->
--- a/src/couch/src/couch_db_updater.erl
+++ b/src/couch/src/couch_db_updater.erl
@@ -23,6 +23,15 @@
 -define(IDLE_LIMIT_DEFAULT, 61000).
 
 
+-record(merge_acc, {
+    revs_limit,
+    merge_conflicts,
+    add_infos = [],
+    rem_seqs = [],
+    cur_seq
+}).
+
+
 init({Engine, DbName, FilePath, Options0}) ->
     erlang:put(io_priority, {db_update, DbName}),
     update_idle_limit_from_config(),
@@ -450,11 +459,18 @@ doc_tag(#doc{meta=Meta}) ->
         Else -> throw({invalid_doc_tag, Else})
     end.
 
-merge_rev_trees(_Limit, _Merge, [], [], AccNewInfos, AccRemoveSeqs, AccSeq) ->
-    {ok, lists:reverse(AccNewInfos), AccRemoveSeqs, AccSeq};
-merge_rev_trees(Limit, MergeConflicts, [NewDocs|RestDocsList],
-        [OldDocInfo|RestOldInfo], AccNewInfos, AccRemoveSeqs, AccSeq) ->
-    erlang:put(last_id_merged, OldDocInfo#full_doc_info.id), % for debugging
+merge_rev_trees([], [], Acc) ->
+    {ok, Acc#merge_acc{
+        add_infos = lists:reverse(Acc#merge_acc.add_infos)
+    }};
+merge_rev_trees([NewDocs | RestDocsList], [OldDocInfo | RestOldInfo], Acc) ->
+    #merge_acc{
+        revs_limit = Limit,
+        merge_conflicts = MergeConflicts
+    } = Acc,
+
+    % Track doc ids so we can debug large revision trees
+    erlang:put(last_id_merged, OldDocInfo#full_doc_info.id),
     NewDocInfo0 = lists:foldl(fun({Client, NewDoc}, OldInfoAcc) ->
         merge_rev_tree(OldInfoAcc, NewDoc, Client, MergeConflicts)
     end, OldDocInfo, NewDocs),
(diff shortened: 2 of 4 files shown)


src/couch/src/couch_db_updater.erl
```diff
@@ -466,13 +468,22 @@ merge_rev_trees([], [], Acc) ->
...
     } = Acc,
 
     % Track doc ids so we can debug large revision trees
     erlang:put(last_id_merged, OldDocInfo#full_doc_info.id),
     NewDocInfo0 = lists:foldl(fun({Client, NewDoc}, OldInfoAcc) ->
-        merge_rev_tree(OldInfoAcc, NewDoc, Client, MergeConflicts)
+        NewInfo = merge_rev_tree(OldInfoAcc, NewDoc, Client, MergeConflicts),
+        case is_overflowed(NewInfo, OldInfoAcc, FullPartitions) of
+            true when not MergeConflicts ->
+                DocId = NewInfo#doc.id,
```
eiri: NewInfo is not a `#doc` record, it's `#full_doc_info`

src/chttpd/src/chttpd.erl
```diff
@@ -873,6 +873,9 @@ error_info(conflict) ->
     {409, <<"conflict">>, <<"Document update conflict.">>};
 error_info({conflict, _}) ->
     {409, <<"conflict">>, <<"Document update conflict.">>};
+error_info({partition_overflow, DocId}) ->
+    Descr = <<"'", DocId/binary, "' exceeds partition limit">>,
```
eiri: To be honest an error message that looks like `"'0000:0043' exceeds partition limit"` looks confusing to me, I think generic "Document update exceeds partition limit" looks more natural and similar to the rest of the error messages we are returning.

src/couch/src/couch_db_updater.erl
```diff
@@ -617,13 +638,31 @@ update_docs_int(Db, DocsList, LocalDocs, MergeConflicts, FullCommit) ->
...
+
+    %% Get the list of full partitions
+    FullPartitions = case couch_db:is_partitioned(Db) of
+        true ->
+            case max_partition_size() of
+                N when N =< 0 ->
+                    [];
+                Max ->
+                    Partitions = lists:usort(lists:map(fun(Id) ->
+                        couch_partition:from_docid(Id)
```
eiri: Same as above, `from_docid/1` throws an exception on design docs id.

src/couch/src/couch_db_updater.erl
```diff
@@ -656,6 +724,40 @@ increment_local_doc_revs(#doc{revs = {0, [RevStr | _]}} = Doc) ->
...
+                        atts = AttSizeInfo
+                    },
+                    add_sizes(leaf, Leaf, SizesAcc);
+                #leaf{} ->
+                    add_sizes(leaf, Value, SizesAcc)
+            end;
+        (_Rev, _Value, branch, SizesAcc) ->
+            SizesAcc
+    end,
+    {_, FinalES, FinalAtts} = couch_key_tree:fold(Fun, {0, 0, []}, RevTree),
```
tonysun83: Could this be a costly call since we're doing this for every doc update in that partition and if the rev trees get large?

src/couch/src/couch_db_updater.erl
tonysun83: Oh eric beat me to the punch, it should be `NewDoc#doc.id`?

davisp: Switched this to NewInfo#full_doc_info.id

src/couch/src/couch_db_updater.erl
```diff
@@ -466,13 +468,22 @@ merge_rev_trees([], [], Acc) ->
...
+        full_partitions = FullPartitions
     } = Acc,
 
     % Track doc ids so we can debug large revision trees
     erlang:put(last_id_merged, OldDocInfo#full_doc_info.id),
     NewDocInfo0 = lists:foldl(fun({Client, NewDoc}, OldInfoAcc) ->
-        merge_rev_tree(OldInfoAcc, NewDoc, Client, MergeConflicts)
+        NewInfo = merge_rev_tree(OldInfoAcc, NewDoc, Client, MergeConflicts),
+        case is_overflowed(NewInfo, OldInfoAcc, FullPartitions) of
+            true when not MergeConflicts ->
```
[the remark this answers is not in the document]
davisp: Good catch. Fixed.

src/couch/src/couch_db_updater.erl
```diff
@@ -595,6 +606,16 @@ merge_rev_tree(OldInfo, NewDoc, _Client, true) ->
     {NewTree, _} = couch_key_tree:merge(OldTree, NewTree0),
     OldInfo#full_doc_info{rev_tree = NewTree}.
 
+is_overflowed(_New, _Old, []) ->
+    false;
+is_overflowed(Old, Old, _FullPartitions) ->
+    false;
+is_overflowed(New, Old, FullPartitions) ->
+    Partition = couch_partition:from_docid(New#full_doc_info.id),
```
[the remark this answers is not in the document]
davisp: Good catch. Fixed.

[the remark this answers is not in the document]
davisp: Actually, take a look at how I fixed this one @eiri. I'm not 100% happy but it was the least worst I could think of to filter our _design docs.

src/chttpd/src/chttpd.erl
```diff
@@ -873,6 +873,9 @@ error_info(conflict) ->
...
+    {403, <<"partition_overflow">>, Descr};
```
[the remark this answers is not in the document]
davisp: I don't think so. 413 is when the request body itself is too big.

src/couch/src/couch_db_updater.erl
davisp: Fixed.

src/couch/src/couch_db_updater.erl
davisp: Well it'll be `O($num_revs_in_tree)` so its linear in that respect. However I added the obvious optimization to only call estimate_size when we find a document in a partition that is overflowed so we'll only be calling it when absolutely necessary.

src/couch/src/couch_db_updater.erl
davisp: Fixed.

src/couch/src/couch_db_updater.erl
```diff
@@ -576,6 +606,16 @@ merge_rev_tree(OldInfo, NewDoc, _Client, true) ->
...
+    NewSize = estimate_size(New),
+    OldSize = estimate_size(Old),
+    lists:member(Partition, FullPartitions) andalso NewSize > OldSize.
```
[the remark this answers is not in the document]
davisp: The key bit is that we're just looking at external sizes for the threshold which only includes leaves the revision tree which means that deleting docs and fields on docs does decrease the used space. An example of a test showing this behavior is here:

https://github.com/apache/couchdb/blob/feature/database-partitions/test/elixir/test/partition_size_test.exs#L102

tonysun83:
+1

src/chttpd/src/chttpd.erl
eiri: What I was thinking is that for a single insert/update it would be obvious what doc id caused the overflow and on a  bulk update with a hundred of docs we'll get a hundred of the error messages iterating over all the keys like:
```
...
 {
    "id": "0000:0095",
    "error": "partition_overflow",
    "reason": "Partition limit exceeded due to update on '0000:0095'"
  },
  {
    "id": "0000:0096",
    "error": "partition_overflow",
    "reason": "Partition limit exceeded due to update on '0000:0096'"
  },
...
```
which seems repetitive.

But I agree that knowing the partition is useful and extracting the partition out of id just for the error message is a bit excessive, so let's go with your version.

eiri:
@davisp Here is a thing - I see three corner cases where this approach is... not perfect.

1) A bulk update that tips a partition over will insert all its docs anyway, because we gather `FullPartitions` before `merge_rev_trees` and the partition that was not already overflown is not going to be considered for a doc's size check.

2) The ddocs can be inserted into overflown partition freely and there are some big ddocs out there. I believe this is what you've meant about not 100% happy, but to be honest I can't think of a clever way to do this better.

3) The attachments written before the overflow check, so a rouge (or a malicious) client with an aggressive attachment's upload retry can really bloat a database.

I don't see a good way to address any of these without really overcomplicated things, so this is just an awareness note to come along with my +1

src/chttpd/src/chttpd.erl
davisp: Fixed.

davisp:
@eiri Yeah, though this isn't intended to be a security feature to have hard constraints. This is just a foot-gun preventative so that a user doesn't get too far out into production before realizing that storing all docs under a single partition doesn't scale. With that in mind it seems forgivable to be able to go over some percentage of a bulk_docs call.

PR #1844, opened 2019-01-08
Add chttpd stats

davisp:
## Overview

This adds an advanced API for users that want to write some Erlang to be able to collect request level metrics for reporting.

## Testing recommendations

`make check`

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -265,6 +265,7 @@ handle_request_int(MochiReq) ->
 
 before_request(HttpReq) ->
     try
+        chttpd_stats:init(),
         chttpd_plugin:before_request(HttpReq)
     catch Tag:Error ->
         {error, catch_error(HttpReq, Tag, Error)}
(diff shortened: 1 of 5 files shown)


wohali:
This is a huge diff and includes a bunch of partition stuff that might be unrelated?

Can you check that your PR'ed the right thing, and if so, possibly add even developer-level documentation on how this works? I couldn't find the needle in the haystack of 69 changed files.

davisp:
@wohali Sorry, on a meeting and did this half brained and forgot to set the correct base branch. I should have also mentioned I have yet to write an example module for use with this but wanted to have a place for people to start responding to the general idea.

wohali:
Still looks broken @davisp, maybe start again?

davisp:
Sorry. I've been updating the other PRs and it gets all weird. Soon as the partition limit PR is merged here in a bit I'll update this to be off master so that its not constantly getting all weird.

davisp:
Fixed

PR #1868, opened 2019-01-21
Update couchup

pealthoff:
## Overview

The couchup script was not running if using python2, missing the quote from urllib

According to the python3 docuymentation:

>  The urllib module has been split into parts and renamed in Python 3 to urllib.request, urllib.parse, and urllib.error. 

At the couchup script, the 2 alternatives at lines 22 and 24 are both importing the python3 alternative.

## Testing recommendations

To test is just run the script on python2 before and after the change, because the error its just the import statement.

### DIFF
--- a/rel/overlay/bin/couchup
+++ b/rel/overlay/bin/couchup
@@ -22,7 +22,7 @@ import sys
 try:
     from urllib.parse import quote
 except ImportError:
-    from urllib.parse import quote
+    from urllib import quote
 import requests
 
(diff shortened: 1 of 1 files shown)


wohali:
@pealthoff Thanks, we'll take this, but keep in mind we don't support Python 2 for the script anymore, see line 1:

https://github.com/apache/couchdb/blob/master/rel/overlay/bin/couchup#L1

PR #1883, opened 2019-01-29
Elixir travis

garrensmith:
## Overview

configure Travis to run the Elixir tests. Since all partition tests are in Elixir it's important we start getting them to run with each build. I've kept the elixir command separate from `make check` until we are able to run the Elixir tests on Jenkins.

## Testing recommendations

Travis should pass

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-language: erlang
+language: elixir
+elixir: 1.6.6
 os: linux
 dist: trusty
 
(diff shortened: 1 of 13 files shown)


wohali:
We have, apparently, a lot of failing elixir tests...they are more unstable than the old JavaScript ones because we are now trying to do race-y things on a 3-node cluster instead of a single node.

iilyak:
> We have, apparently, a lot of failing elixir tests...they are more unstable than the old JavaScript ones because we are now trying to do race-y things on a 3-node cluster instead of a single node.

By looking at Jenkins log I see that there are some problems with elixir install.

wohali:
@iilyak Jenkins isn't running Elixir tests yet - it's not in the default targets. Do you mean Travis?

.travis.yml
```diff
@@ -2,10 +2,26 @@ language: erlang
...
-   - 21.2.3
-   - 20.3.8.5
-   - 19.3
+matrix:
+  include:
+  - language: elixir
+    elixir: 1.6
+    env: COMMAND=elixir
+    otp_release: 21.2.3
+  - language: erlang
```
wohali: Eventually, we're going to want elixir tests running against all OTP releases. I see this as a temporary workaround, but it's ugly. Using the `before_script` to install elixir, or just installing elixir in all versions, might be a better approach. See other comments as well.

test/elixir/mix.exs
```diff
@@ -5,7 +5,7 @@ defmodule Foo.Mixfile do
     [
       app: :foo,
       version: "0.1.0",
-      elixir: "~> 1.5",
+      elixir: "~> 1.6",
```
wohali: FYI Jenkins images right now have Elixir 1.6.6, so this should be fine.

PR #1887, opened 2019-01-30
Make it possible to override elixir couchdb settings via environmental variables

garrensmith:
## Overview

Make it possible to override elixir couchdb settings via environmental variables

## Testing recommendations

Spin up a local CouchDB cluster with different settings to what Elixir requires by default.
Run `USER=myusername PASS=password COUCH_URL=http://my-couchdb.com mix test `
and make sure the tests will run.

## Checklist

### DIFF
--- a/test/elixir/README.md
+++ b/test/elixir/README.md
@@ -11,7 +11,16 @@ mix deps.get
 mix test --trace
 ```
 
-# Tests to port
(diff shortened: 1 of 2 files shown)


jiangphcn:
`USER=foo PASS=bar mix test` works in my environment

PR #1897, opened 2019-02-01
Fix _bulk_docs crash when provided revision can't be parsed

### LINKED ISSUE #1895: _bulk_docs/put - damaged _rev causes uncaught exception

Putting document with damaged _rev can cause exception returning 500 Internal Error instead of relevant error message

## Description

It is possible to craft _revs by imitating replication and putting documents along with new_edits=false. However, if the _revs looks like revision generated by CouchDB, but it is somehow damaged, the couchdb crashes with exception. This also happens when _bulk_docs is used - rest of the documents are not stored and result of the execution is unknown

## Steps to Reproduce

1. create DB "test"
2.
```
$ curl -iH "Content-Type: application/json" -d '{"docs":[{"_id":"test","aaa":"bbb","_revisions":{"start":1,"ids":["7ea7f65fbd9ddcd547969ca540fed66w"]}},{"_id":"other","_revisions":{"start":1,"ids":["aaa"]}}],"new_edits":false}' "http://localhost:5984/test/_bulk_docs"
```
```
{"error":"unknown_error","reason":"badarg","ref":1621251896}
```

Notice 'w' in revision of the first document. The second document is not stored.

There is log record:

```
 req_err(1621251896) unknown_error : badarg
    [<<"erlang:list_to_integer/2">>,<<"couch_doc:parse_revid/1 L165">>,<<"couch_doc:-transfer_fields/3-lc$^2/1-1-/1 L280">>,<<"couch_doc:transfer_fields/3 L280">>,<<"couch_doc:from_json_obj_validate/2 L135">>,<<"chttpd_db:-db_req/2-fun-1-/1 L424">>,<<"lists:map/2 L1239">>,<<"chttpd_db:db_req/2 L423">>]
```

When a crafted _rev is used, the request is executed correctly

```
$ curl -iH "Content-Type: application/json" -d '{"docs":[{"_id":"test","aaa":"bbb","_revisions":{"start":1,"ids":["crafted"]}},{"_id":"other2","_revisions":{"start":1,"ids":["aaa"]}}],"new_edits":false}' "http://localhost:5984/test/_bulk_docs"
```
```
HTTP/1.1 201 Created
[]
```

This bug also appears, when new_edits=false is not used and the revision is damaged.

```
$ curl -X PUT -iH  "Content-Type: application/json" -d '{"_id":"test2"}' "http://localhost:5984/test/test2"
```
```
{"ok":true,"id":"test2","rev":"1-967a00dff5e02add41819138abb3284d"}
```
```
$ curl -X PUT -iH  "Content-Type: application/json" -d '{"_id":"test2","_rev":"1-967a00dff5e02add41819138abb3284w"}' "http://localhost:5984/test/test2"
```
```
{"error":"unknown_error","reason":"badarg","ref":531824669}
```

However, if the revision is not damaged, just different, correct error message is generated
```
$ curl -X PUT -iH  "Content-Type: application/json" -d '{"_id":"test2","_rev":"1-967a00dff5e02add41819138abb32841"}' "http://localhost:5984/test/test2"
```
```
{"error":"conflict","reason":"Document update conflict."}
```

It seems that CouchDB is trying to parse the hash part of the revision when it looks like a valid hash and it fails without handling the exception. The uncaught exception causes error 500

## Expected Behaviour

Uncaught exception should be handled

## Your Environment

* CouchDB Version used:

{"couchdb":"Welcome","version":"2.3.0","git_sha":"07ea0c7","uuid":"3246e174be789358f6e0855023ea8731","features":["pluggable-storage-engines","scheduler"],"vendor":{"name":"The Apache Software Foundation"}}

* Operating System and version: Ubuntu 18.04

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

Not blocking issue, I just found this by mistake during performing tests on my project.

eiri:
## Overview

If a doc's revision provided with `_bulk_docs` matches the expected length it assumed to be a parsable hexadecimal. When this is not the case `couch_doc:couch_doc:from_json_obj_validate/2` function crashes with `badarg`.

The fix catches the exception and re-throws it as a `doc_validation` exception to raise "400 Bad Request" error.

## Testing recommendations

`make eunit apps=couch suites=couch_doc_json_tests` or the manual stps from case #1895 

Closes #1895 

## Checklist

### DIFF
--- a/src/couch/src/couch_doc.erl
+++ b/src/couch/src/couch_doc.erl
@@ -275,9 +275,16 @@ transfer_fields([{<<"_revisions">>, {Props}} | Rest], Doc, DbName) ->
     true ->
         ok
     end,
-    [throw({doc_validation, "RevId isn't a string"}) ||
-            RevId <- RevIds, not is_binary(RevId)],
-    RevIds2 = [parse_revid(RevId) || RevId <- RevIds],
+    RevIds2 = lists:map(fun(RevId) ->
+        try
+            parse_revid(RevId)
+        catch
+            error:function_clause ->
+                throw({doc_validation, "RevId isn't a string"});
+            error:badarg ->
+                throw({doc_validation, "RevId isn't a valid hexadecimal"})
+        end
+    end, RevIds),
     transfer_fields(Rest, Doc#doc{revs={Start, RevIds2}}, DbName);
 
 transfer_fields([{<<"_deleted">>, B} | Rest], Doc, DbName) when is_boolean(B) ->
--- a/src/couch/test/couch_doc_json_tests.erl
+++ b/src/couch/test/couch_doc_json_tests.erl
@@ -270,6 +270,12 @@ from_json_error_cases() ->
             "Revision ids must be strings."
         },
         {
+            {[{<<"_revisions">>, {[{<<"start">>, 0},
+                {<<"ids">>, [<<"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx">>]}]}}]},
+            {doc_validation, "RevId isn't a valid hexadecimal"},
+            "Revision ids must be a valid hex."
+        },
+        {
             {[{<<"_something">>, 5}]},
             {doc_validation, <<"Bad special document member: _something">>},
             "Underscore prefix fields are reserved."
--- a/src/couch/test/couch_doc_json_tests.erl
+++ b/src/couch/test/couch_doc_json_tests.erl
@@ -288,18 +288,14 @@ from_json_error_cases() ->
 
     lists:map(fun
         ({Fun, Expect, Msg}) when is_function(Fun, 0) ->
-            Error = (catch couch_doc:from_json_obj_validate(Fun())),
-            {Msg, ?_assertMatch(Expect, Error)};
+            {Msg,
+            ?_assertThrow(Expect, couch_doc:from_json_obj_validate(Fun()))};
         ({EJson, Expect, Msg}) ->
-            Error = (catch couch_doc:from_json_obj_validate(EJson)),
-            {Msg, ?_assertMatch(Expect, Error)};
+            {Msg,
+            ?_assertThrow(Expect, couch_doc:from_json_obj_validate(EJson))};
         ({EJson, Msg}) ->
-            try
-                couch_doc:from_json_obj_validate(EJson),
-                {"Conversion failed to raise an exception", ?_assert(false)}
-            catch
-                _:_ -> {Msg, ?_assert(true)}
-            end
+            {Msg,
+            ?_assertThrow(_, couch_doc:from_json_obj_validate(EJson))}
     end, Cases).
 
 from_json_with_dbname_error_cases() ->


src/couch/src/couch_doc.erl
```diff
@@ -275,9 +275,16 @@ transfer_fields([{<<"_revisions">>, {Props}} | Rest], Doc, DbName) ->
     true ->
         ok
     end,
-    [throw({doc_validation, "RevId isn't a string"}) ||
-            RevId <- RevIds, not is_binary(RevId)],
-    RevIds2 = [parse_revid(RevId) || RevId <- RevIds],
+    RevIds2 = lists:map(fun(RevId) ->
+        try
```
tonysun83: any reason the check isn't  in `parse_revid` given the other `parse_revs` also throw errors?

eiri: > given the other parse_revs also throw errors?
what do you mean? as far as I see `parse_revid/1` doesn't have `throw` in any clauses.

tonysun83: @eiri : I should be a little specific. I just meant other `parse_rev` functions such as 

https://github.com/apache/couchdb/blob/57c0f04c681cd69806a525801a9431a4f9161cde/src/couch/src/couch_doc.erl#L185

and 

https://github.com/apache/couchdb/blob/57c0f04c681cd69806a525801a9431a4f9161cde/src/couch/src/couch_doc.erl#L198

were throwing errors.

eiri: @tonysun83 hm, ok. Well it is technically possible to convert `parse_revid/1` to do the same, but that'd require to add some kind of validation step there or introduce re-throw block to raise proper error type and will spread `doc_validation` exception cases across the module instead of keeping them within `transfer_fields/3` as it is now. In my opinion these two are pretty much equal approaches to fix this, so why do you think your idea is better than what I have here?

tonysun83: @eiri: My nitpick is only for readability purposes. Like you said, if we move it to `parse_revid/1`, you'd have to add a validation step and re-throw. The advantage of that is that you'd know we're throwing it because of `erlang:list_to_integer/2` and can logically conclude that it's not a valid hex. Whereas inside `transfer_fields/3`, and first glance, you can't assume there's only one possibility for `bad_args` and have to read `parse_revid`. Having said that, you do make a good point of keeping it inside `transfer_fields/3` so that it doesn't permeate other parts of the codebase, especially since `parse_rev/1` calls `parse_revid`. We don't want a change in behavior if 400 errors thrown suddenly in unexpected places (even if it is correct). I'll leave it up to you if you want to add a comment to make it a little clearer or if you think it should be self-evident as it is.

eiri: there are two possibilities for `badarg`, actually, an inconvertible list to `?l2b` is the second, but honestly, I think we (server side) don't care _why_ exactly parse failed as long as we can pass back a message that the rev format is invalid and then let the user to figure it out. this is a simple fix for a simple bug and I don't feel that overcomplicating things with validators and whatnot will add any value here.

src/couch/test/couch_doc_json_tests.erl
```diff
@@ -269,6 +269,12 @@ from_json_error_cases() ->
             {doc_validation, "RevId isn't a string"},
             "Revision ids must be strings."
         },
+        {
+            {[{<<"_revisions">>, {[{<<"start">>, 0},
+                {<<"ids">>, [<<"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx">>]}]}}]},
+            {doc_validation, "RevId isn't a valid hexadecimal"},
+            "Revision ids must be a valid hex."
```
ronag: Since when is this an invariant?

ronag: And it's not about being valid hex, but being a valid MD5 hex hash

eiri: Sorry, I'm not sure what are you trying to ask. A doc's revision been [md5 hash prefixed with doc's update count](https://docs.couchdb.org/en/latest/intro/api.html#revisions) for a pretty long time and as you can see this is documented.

This test _is_ kind of about validity of hex. Technically when `new_edits` set to `false` revision can be anything, but if it happens to be 16 bytes long rev's parser trying to treat it as a hex number, hence [the original issue](https://github.com/apache/couchdb/issues/1895).

Again, sorry, it might be just my perception, but your comments seems to be somehow aggressive. Is there a particular reason you are leaving them on this half-year old PR? If you have a problem with it, can you, please, open a new issue and describe it?

ronag: Um, I'm not sure how that's aggressive. Anyway it's not meant to be in anyway.

I was going through the changelog for 2.3 and ended up on this PR.

>  but if it happens to be 16 bytes long rev's parser trying to treat it as a hex number

Yes, this has been a huge problem for us (https://github.com/apache/couchdb/issues/2015) and I thought this PR made it worse. This strange behaviour still only applies if it's 16 bytes long? 

Anyway I still think this quirk should be documented since even though custom rev ids are discouraged they are allowed.

eiri: Look at this from my perspective: a stranger came out of nowhere and with zero context complaint about a message on a unit test of my ancient PR. This is a bit unusual for me and didn't seem to be well-intended. Glad to hear this is not the case.

Anyway, this PR does what it says in description, it trades a crash with `badarg` message for an error that, I hoped, would hint on a reason for such crash. It doesn't change rev's parsing logic in any way.

> This strange behaviour still only applies if it's 16 bytes long? i.e. not all revids must be a hex.

In short - yes. It distills from the fact that `new_edits=false` is a technical feature (in sense akin Apple's private APIs) that turns off system's assumption about MVCC, but if sent revision _looks_ like a "real" one it tricks system into applying some of these assumptions back. I see it was discussed pretty extensively on the issue you linked, not sure if I can add anything new here.

> Anyway I still think this quirk should be documented since even though custom rev ids are discouraged they are allowed.

Maybe? Sorry, it feels strange to discuss it in a middle of a merged PR. I suggest you to nudge on the original issue or even just change the documentation as you find desired and open a PR for it, it's an open source project after all.

Issue #1902, opened 2019-02-05
Fix flaky tests on Jenkins


eiri:
## Summary

Now that elixir test suite was added to Travis it proved to be rather unreliable (for me it's failing 4 times out of 5 on the random tests). This is a problem as it notably delays merging _any_ pull request and in general not causing a great deal of confidence in elixir tests to indicate anything useful.

## Desired Behaviour

Ideally we'd want to elixir tests to fail only for a legit reasons, but rare and insignificant amount of random failures is permissible.

## Possible Solution

We can tag all flaky tests as `flaky` as an initial stage of #1885 and make `make elixir` to skip them until Travis will start reliably (or at least more reliably) pass. This will give us a complete list of the tests to improve and as we go through them one-by-one we'll remove the `flaky` tag enabling them back.

garrensmith:
+1 to this.

jaydoane:
I have also found that in many cases it's possible to de-flake a test simply by wrapping the flaky parts in a retry like so:
```elixir
retry_until(fn ->
  flaky_code
end)
```
Of course, you have to include the request as well as the failing assertion in the fn, but as long the condition is met within the timeout (default 5 seconds), it should help.

janl:
how about we run the tests, but don't fail the build if they fail until we worked out the kinks?

eiri:
@janl That'd work too, the only thing is that with the tagging we'll have a list of the things to fix and ability to turn them back one-by-one, where with your approach it's all-or-nothing. But it's much faster and simpler to do, that's important :)

@jaydoane I guess that'd work, but this is a bit different from the topic, it's not about how to fix the tests, but how to progress with other job while they not fixed.

eiri:
An obvious concern with both tagging and turning off or ignoring all the failures approaches is that it's way too easy to do that and then just forget and never come back to the fixing of the failing tests 🙈 

I don't have a solution for this apart from somebody taking and driving this work from start till end.

janl:
@eiri fully agree that this needs to be a process, the not-fail-if-test-fail is a stop-gap to allow a process to gradually get there

jaydoane:
@eiri I mentioned the `retry_until` hack since trying that is about as time consuming as tagging, and if it fixes the issue, the tag and subsequent fixing are no longer necessary.

garrensmith:
[...]

PR #1904, opened 2019-02-05
Import smoosh from Cloudant

rnewson:
Remove couch_compaction_daemon and related tests too.


Makefile
```diff
@@ -143,11 +143,10 @@ fauxton: share/www
 check: all
 	@$(MAKE) test-cluster-with-quorum
 	@$(MAKE) test-cluster-without-quorum
+	@$(MAKE) python-black
 	@$(MAKE) eunit
 	@$(MAKE) javascript
-	@$(MAKE) python-black
 	@$(MAKE) mango-test
-	@$(MAKE) elixir
```
iilyak: I think changes to Makefile are accidental. 
- Why are we moving `python-black`? 
- Why are we removing `elixir`?

rnewson: python-black is moved to the front because it's a quick test and we get early feedback and short build times if it fails. 

elixir is removed because the test suite fails spuriously.

janl: there are separate commits for both including reasoning.

the black tests should fail as soon as possible, so we don’t have to wait 25 minutes for them to fail.

elixir tests are too flakey at the moment, so this is a tmp disable to get this PR green. larger discussion is on https://github.com/apache/couchdb/issues/1902

PR #1907, opened 2019-02-06
test: run elixir tests, but down fail entire build if they fail

janl:
My compromise suggestion for #1902

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,7 @@ check: all
 	@$(MAKE) eunit
 	@$(MAKE) javascript
 	@$(MAKE) mango-test
+	@$(MAKE) elixir
 #	@$(MAKE) build-test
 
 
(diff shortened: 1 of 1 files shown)


eiri:
Failure in tests:
```
* test GET /dbname/_design_docs (565.1ms)

  1) test GET /dbname/_design_docs (PartitionDDocTest)
     test/partition_ddoc_test.exs:159
     Assertion with == failed
     code:  assert length(body["rows"]) == 1
     left:  0
     right: 1
     stacktrace:
       test/partition_ddoc_test.exs:168: (test)
```
and green Travis, so works as expected.

That'd be +1 from me to immediately unblock release pipe, but I think we need @wohali opinion on this as well.

garrensmith:
We can close this since the elixir tests are merged in

PR #1914, opened 2019-02-12
Add security item to the RFC template

wohali:
per @rnewson 's request


wohali:
@rnewson Done - please approve?

PR #1927, opened 2019-02-20
build: stop propagating RC tags into erlang apps

dch:
This ensures that the only place an RC tag is visible is in the final
tarball. To ensure we can identify CouchDB versions in the wild, each
OTP app has the original git sha placed in the `{id, :hash}` field.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,13 +45,13 @@ IN_RC = $(shell git describe --tags --always --first-parent \
 # ON_TAG matches *ONLY* if we are on a release or RC tag
 ON_TAG = $(shell git describe --tags --always --first-parent \
         | grep -Eo -- '^[0-9]+\.[0-9]\.[0-9]+(-RC[0-9]+)?$$' 2>/dev/null)
-# RELTAG contains the #.#.# from git describe, which might be used
-RELTAG = $(shell git describe --tags --always --first-parent \
+# REL_TAG contains the #.#.# from git describe, which might be used
+REL_TAG = $(shell git describe --tags --always --first-parent \
         | grep -Eo -- '^[0-9]+\.[0-9]\.[0-9]+' 2>/dev/null)
 # DIRTY identifies if we're not on a commit
 DIRTY = $(shell git describe --dirty | grep -Eo -- '-dirty' 2>/dev/null)
 # COUCHDB_GIT_SHA is our current git hash.
-COUCHDB_GIT_SHA=$(shell git rev-parse --short --verify HEAD)
+COUCHDB_GIT_SHA=$(shell git rev-parse --short=7 --verify HEAD)
 
 ifeq ($(ON_TAG),)
 # 4. Not on a tag.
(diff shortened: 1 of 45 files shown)


src/chttpd/src/chttpd.app.src
```diff
@@ -12,7 +12,7 @@
 
  {application, chttpd, [
     {description, "HTTP interface for CouchDB cluster"},
-    {vsn, git},
+    {vsn, "2.3.0-f18d602"},{id, "f18d602"},
```
wohali: Isn't committing these to the repo bad practice, since any commit to that tree will require mucking with it and PRs will have commits to this file every time that app is touched?

wohali:
Really not comfortable with having to bump each .app.src file with every change to that revision tree. This looks like a recipe for unnecessary churn in the repo.

iilyak:
I agree with @wohali this is going to be a pain to update. @dch are you planing to implement some tooling which would rely on `id`?

iilyak:
We could embed git sha into vsn. But we would need to write a simple rebar plugin and place it in `rel/plugins`. Since we don't want to manually edit app files. The plugin could hook into `post_compile`. Then it would call `git rev-parse --verify HEAD` or `git rev-list -1 HEAD`. There is similar plugin here https://github.com/erlware/rebar_vsn_plugin/blob/master/src/rebar_vsn_plugin.erl. However ours could be way simpler.

src/chttpd/src/chttpd.app.src
dch: yep this isn't intended, but I'm not sure what has pulled it in.

dch:
hey folks I thought I'd done this as a fancy invisible drafts PR so I didn't lose it in my browser tabs.... didn't realise it would be publically visible already otherwise I'd have added a "wip" tag or something. sorry!

The diff itself is from last year so I'm a bit fuzzy on the details. I am `93.7%` certain that the `.app.src` diffs are only supposed to be in the release tarball and not in the repo/diff etc.

wohali:
@dch I'll take a look at this before the first 3.0.0 RC1. If you have time to help, though...would love to work together on it.

Makefile
```diff
@@ -121,10 +134,13 @@ help:
...
 .PHONY: couch
 # target: couch - Build CouchDB core, use ERL_OPTS to provide custom compiler's options
-couch: config.erl
+couch: config.erl derived set_otp_vsn
 	@COUCHDB_VERSION=$(COUCHDB_VERSION) COUCHDB_GIT_SHA=$(COUCHDB_GIT_SHA) $(REBAR) compile $(COMPILE_OPTS)
 	@cp src/couch/priv/couchjs bin/
 
+# target: set_otp_vsn - ensure that vsn field in OTP app is set appropriately
+set_otp_vsn:
+	@sed -i$(sed v < /dev/null 2> /dev/null || echo -n " ''")  -e "s/{vsn, git}/${SUB_VSN}/" $(shell find ./src -type f -name \*.app.src)
```
wohali: The crazy subshell here is to detect if we are using GNU sed (which does not require a file extension to be specified for `-i`, or BSD sed, which _does_ require a file extension (which we then set to null).

wohali:
This has broken building from a `make release` build. Shouldn't be hard to fix.

wohali:
@dch Hey, I'm sorry, I've run out of steam on this - the code seems broken in a fundamental way that makes the results of `make dist` uncompilable, even after my latest pushed change.

`rebar` keeps complaining from the result that it thinks it's in git, but isn't.

I think I'm going to punt on this for the 3.0.0 RCs, unless you have time to look. Let me know in the next 24 hours or I'll be removing this from the 3.0.0 worklist.

wohali:
Replaced by #2506.

PR #1936, opened 2019-02-25
Send correct 400 for missing partition with _find

garrensmith:
## Overview

When sending an incorrect partition query e.g:
/parition/_find, send a 400 with the message:

Partition must not start with an underscore

This makes it consistent with all the other partition requests

## Testing recommendations

Tests should pass and correct message should be returned for `_explain` and `_find`

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -274,6 +274,11 @@ handle_partition_req(#httpd{method='GET', path_parts=[_,_,PartId]}=Req, Db) ->
             throw({bad_request, <<"database is not partitioned">>})
     end;
 
+handle_partition_req(#httpd{method='POST',
+    path_parts=[_, <<"_partition">>, <<"_", _/binary>>]}, _Db) ->
+    Msg = <<"Partition must not start with an underscore">>,
+    throw({illegal_partition, Msg});
+
 handle_partition_req(#httpd{path_parts = [_, _, _]}=Req, _Db) ->
     send_method_not_allowed(Req, "GET");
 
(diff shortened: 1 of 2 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -274,6 +274,11 @@ handle_partition_req(#httpd{method='GET', path_parts=[_,_,PartId]}=Req, Db) ->
             throw({bad_request, <<"database is not partitioned">>})
     end;
 
+handle_partition_req(#httpd{method='POST',
```
iilyak: Why do we allow POST if we are not going to handle it? If I read the code correctly. Without this change any attempt to POST would fail with `method_not_allowed`. Nice and clean. With this change we allow POST only for :
- `_index`
- `_find`
- `_explain`

and fail it anyway. I believe that this would confuse the user. Since when the user would get this error he would think that the POST is allowed and he just having a problem with underscore in the name.

garrensmith: @iilyak that isn't quite right. What happens is a user can accidentally do a `POST /db/_partition/_find` which is missing a partition field. That post should be

`POST /db/_partition/:partition_value/_find` so we need to send a correct error message which this PR does.

iilyak: I still think that `method_not_allowed` is better response to post into `/db/_partition/_find`.

iilyak: For some reason I thought that `handle_partition_req` doesn't have a clause for POST. However [it does](https://github.com/apache/couchdb/pull/1936/files#diff-f984ed397efedb5250e67c88ede72837R285). 
With this new information I think you are right and it is the right change in general. Should we make it more generic though: `path_parts=[_, <<"_partition">>, <<"_", _/binary>>]`?

garrensmith: Good idea. I've made the change

PR #1952, opened 2019-02-27
Warn people to edit both Makefiles.

wohali:
Also fix dumb S&R \ vs / bug in URL.


iilyak:
Should we try adding an automated check into either of the files?:
- .travis.yml
- Jenkinsfile

Something which checks the output of the following command to make sure it is either 0 or 2:
```
git diff --name-only `git merge-base master HEAD` | grep Makefile | wc -l
```

wohali:
@iilyak hard to do since we don't know the target branch will always be `master`. but, a decent idea.

iilyak:
> @iilyak hard to do since we don't know the target branch will always be master. but, a decent idea.

I think it could be solved by using `TRAVIS_BRANCH` env which would be set to target branch.

PR #1963, opened 2019-03-04
Add stats in fabric for partition and normal views

garrensmith:
## Overview

Record the number of partition and global view queries along
with the number of timeouts that occur.

With partition queries, it is possible to set a much stricter global timeout value. We want to be able to record the number of timeouts that would occur as the number of total requests.

## Testing recommendations

Run the partition tests and then check stats

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -811,8 +811,7 @@ send_delayed_error(#delayed_resp{req=Req,resp=nil}=DelayedResp, Reason) ->
     {ok, Resp} = send_error(Req, Code, ErrorStr, ReasonStr),
     {ok, DelayedResp#delayed_resp{resp=Resp}};
 send_delayed_error(#delayed_resp{resp=Resp, req=Req}, Reason) ->
-    PathParts = req#httpd.requested_path_parts,
-    update_timeout_stats(Reason, PathParts),
+    update_timeout_stats(Reason, Req),
     log_error_with_stack_trace(Reason),
     throw({http_abort, Resp, Reason}).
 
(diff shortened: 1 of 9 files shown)


pszczypka:
Looks ok for metrics.

src/chttpd/src/chttpd.erl
```diff
@@ -1045,6 +1051,39 @@ send_error(Req, Code, Headers, ErrorStr, ReasonStr, Stack) ->
         case Stack of [] -> []; _ -> [{<<"ref">>, stack_hash(Stack)}] end
     ]}).
 
+update_timeout_stats(Error, PathParts)
```
iilyak: General rambling...

I don't like that we are exposing endpoint paths in `chttpd.erl`. I know we already do request specific handling in `chttpd:error_info`. This makes implementation of an endpoint coupled with correct update in `chttpd`. I would rather prefer not to leak this detail. However I don't see an easy way to achieve that without changing the way how http handlers are plugged. Changing that aspect is clearly out of scope of the PR. 

Back to this PR...
I think we should convert this function to use function clauses. It might look something like:
```
update_timeout_stats(<<"timeout">>, #httpd{requested_path_parts = RequestedPathParts} = Req) ->
   update_timeout_stats(RequestedPathParts);
update_timeout_stats(timeout, #httpd{requested_path_parts = RequestedPathParts} = Req) ->
   update_timeout_stats(RequestedPathParts).

update_timeout_stats([_, <<"_partition">>, _, <<"_design">>, _, <<"_view">> | _]) ->
     couch_stats:increment_counter([couchdb, httpd, partition_view_timeouts]);
...
```

PR #1970, opened 2019-03-06
Added more info for a mango sort error

garrensmith:
## Overview

The mango sort error can be a bit confusing with using a partitioned
database this gives a little more detail on why mango could not find
a specific index when a sort is used.

## Testing recommendations

Elixir and python mango tests should pass. Create a partitioned database, have a selector that has a sort in it:

```
selector = %{
      selector: %{
        some: "field"
      },
      sort: ["some"],
      limit: 50
    }
```

You should get back different errors depending on a normal database, partitioned database with a partitioned query or a global query on a partitioned database.

## Checklist

### DIFF
--- a/src/mango/src/mango_error.erl
+++ b/src/mango/src/mango_error.erl
@@ -25,13 +25,15 @@ info(mango_idx, {no_usable_index, missing_sort_index}) ->
     {
         400,
         <<"no_usable_index">>,
-        <<"No index exists for this sort, try indexing by the sort fields.">>
+        <<"No index exists for this sort, "
+            "try indexing by the sort fields.">>
     };
 info(mango_idx, {no_usable_index, missing_sort_index_partitioned}) ->
     {
         400,
         <<"no_usable_index">>,
-        <<"No partitioned index exists for this sort, try indexing by the sort fields.">>
+        <<"No partitioned index exists for this sort, "
+            "try indexing by the sort fields.">>
     };
 info(mango_idx, {no_usable_index, missing_sort_index_global}) ->
     {
--- a/src/mango/src/mango_error.erl
+++ b/src/mango/src/mango_error.erl
@@ -27,6 +27,18 @@ info(mango_idx, {no_usable_index, missing_sort_index}) ->
         <<"no_usable_index">>,
         <<"No index exists for this sort, try indexing by the sort fields.">>
     };
+info(mango_idx, {no_usable_index, missing_sort_index_partitioned}) ->
+    {
+        400,
+        <<"no_usable_index">>,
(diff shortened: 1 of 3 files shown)


src/mango/src/mango_idx.erl
```diff
@@ -410,12 +421,20 @@ get_idx_partitioned(Db, DDocProps) ->
             Default
     end.
 
+is_opts_partitioned(Opts) ->
+    case couch_util:get_value(partition, Opts) of
```
eiri: This call here makes a brave assumption that `partition` attribute is always presented in `Opts` (which might be true, I don't know), but I think `couch_util:get_value(partition, Opts, <<>>)` might be a bit safer here, just to avoid possibility of a crash on no case clause matching.

src/mango/src/mango_error.erl
```diff
@@ -27,6 +27,18 @@ info(mango_idx, {no_usable_index, missing_sort_index}) ->
         <<"no_usable_index">>,
         <<"No index exists for this sort, try indexing by the sort fields.">>
     };
+info(mango_idx, {no_usable_index, missing_sort_index_partitioned}) ->
+    {
+        400,
+        <<"no_usable_index">>,
+        <<"No partitioned index exists for this sort, try indexing by the sort fields.">>
```
jiangphcn: More than 80 characters

src/mango/src/mango_error.erl
```diff
@@ -27,6 +27,18 @@ info(mango_idx, {no_usable_index, missing_sort_index}) ->
...
+    };
+info(mango_idx, {no_usable_index, missing_sort_index_global}) ->
+    {
+        400,
+        <<"no_usable_index">>,
+        <<"No global index exists for this sort, try indexing by the sort fields.">>
```
jiangphcn: More than 80 characters

test/elixir/test/partition_mango_test.exs
```diff
@@ -633,4 +633,31 @@ defmodule PartitionMangoTest do
     %{:body => %{"reason" => reason}} = resp
     assert Regex.match?(~r/Partition must not start/, reason)
   end
+
+  @tag :with_partitioned_db
+  test "partitioned query sends correct errors for sort errors", context do
+    db_name = context[:db_name]
```
jiangphcn: Current test covers most of change on Partitioned database/index. However, it would be nicer to have one test for `{no_usable_index, missing_sort_index}`

garrensmith: There is a mango test in the python tests that exercises that error message.

PR #1983, opened 2019-03-18
More precise calculation of external docs' size

eiri:
## Overview

We define a database's external size as:
> Number of bytes that would be required to represent the contents outside of the database (for capacity and backup planning)

Given that we are using JSON as a data serialization format on REST API I assume that `external` should represent a size of uncompressed JSON serialized data.

However we are using `?term_size` for calculating doc's sizes and that gives this definition of `external` a noticeable overhead.

Here is a demo:
```bash
# movies.json contents a simplified imdb's movies info, i.e. reasonably complex json structure
$ cat movies.json | jq .docs[2000]
{
  "title": "The Love Light",
  "year": 1921,
  "cast": [
    "Mary Pickford",
    "Raymond Bloomer"
  ],
  "genres": [
    "Drama"
  ]
}

$ stat -f%z movies.json 
3386172 

$ curl -X PUT http://127.0.0.1:15984/test -G -d n=1 q=1
{"ok":true}

$ curl -X POST http://127.0.0.1:15984/test/_bulk_docs --data @movies.json > /dev/null

$ curl http://127.0.0.1:15984/test | jq .sizes.external
5856324
```

So `external` here gives ~58% overhead over "pure" JSON data.

I've switched from `?term_size` to `couch_ejson_size:encoded_size/1` for calculating external size and this returns much closer values. For the same test:
```bash
$ curl http://127.0.0.1:15984/test | jq .sizes.external
3357139
```

Another change is a switch from recalculation of external size during compaction to carrying it from leaf's info, so `external` size doesn't change after compaction:
```bash
$ curl -X POST http://127.0.0.1:15984/test/_compact
{"ok":true}
$ curl http://127.0.0.1:15984/test | jq .sizes.external
3357139
```

## Testing recommendations

All the existing `couch` app's test should pass. The manual test could be done following the steps in the example above.

## Checklist

### DIFF
--- a/src/couch/src/couch_bt_engine_compactor.erl
+++ b/src/couch/src/couch_bt_engine_compactor.erl
@@ -297,13 +297,13 @@ copy_docs(St, #st{} = NewSt, MixedInfos, Retry) ->
         {NewRevTree, FinalAcc} = couch_key_tree:mapfold(fun
             ({RevPos, RevId}, #leaf{ptr=Sp}=Leaf, leaf, SizesAcc) ->
                 {Body, AttInfos} = copy_doc_attachments(St, Sp, NewSt),
-                % In the future, we should figure out how to do this for
-                % upgrade purposes.
-                ExternalSize = case is_binary(Body) of
-                    true ->
+                #size_info{external = OldExternalSize} = Leaf#leaf.sizes,
+                ExternalSize = case OldExternalSize of
+                    0 when is_binary(Body) ->
                         couch_compress:uncompressed_size(Body);
-                    false ->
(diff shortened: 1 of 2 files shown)


eiri:
@davisp Can you take a look please?

I don't understand why it is necessary to recalculate doc's external size during compaction, given we are interested in uncompressed value, not changing doc's body and already have this info on `#leaf`, so I'm afraid I might be missing something subtle here.

eiri:
@davisp I've made the requested change, sorry it took so long, travis hung on the update for some reason.

PR #2003, opened 2019-04-11
Don't reset_index if read_header fails

rnewson:
This decision is an old one, I think, and took the view that an error
from read_header meant something fatally wrong with this index
file. These days, it's far more likely to be something else, a process
crash at the wrong moment, and therefore wiping the index out is an
overreaction.

### DIFF
--- a/src/couch_mrview/src/couch_mrview_index.erl
+++ b/src/couch_mrview/src/couch_mrview_index.erl
@@ -122,7 +122,7 @@ open(Db, State0) ->
 
     case couch_mrview_util:open_file(IndexFName) of
         {ok, Fd} ->
-            case (catch couch_file:read_header(Fd)) of
+            case couch_file:read_header(Fd) of
                 % upgrade code for <= 1.2.x
                 {ok, {OldSig, Header}} ->
                     % Matching view signatures.
@@ -135,7 +135,7 @@ open(Db, State0) ->
                     NewSt = couch_mrview_util:init_state(Db, Fd, State, Header),
                     ensure_local_purge_doc(Db, NewSt),
                     {ok, NewSt};
-                _ ->
+                no_valid_header ->
                     NewSt = couch_mrview_util:reset_index(Db, Fd, State),
                     ensure_local_purge_doc(Db, NewSt),
                     {ok, NewSt}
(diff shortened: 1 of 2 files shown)


janl:
@rnewson which process would crash and why that would cause read_header() to throw?

rnewson:
It's a good question. We have logs where this reset_indexes function has been called but it failed because the couch_file pid had terminated (and therefore read_header returned a noproc error). We've looked at this piece of code and don't think it's appropriate to reset the index quite so easily.

janl:
@rnewson would be worth adding a dedicated log line there so we find out what’s wrong? an/or catch a `noproc`, but reset when a parsing error occurs?

chewbranca:
> @rnewson which process would crash and why that would cause read_header() to throw?

I suspect there are multiple ways in which this could happen through the code path, but one highly plausible scenario is when the `couch_file:read_header` call is made, that request goes through an IOQ pid [1], and if that IOQ pid exits, we'll get a `gen_server:call` error that falls into the `_` codepath here and triggers an index reset. When that happens, the `couch_file:truncate` call (along with sync) go directly to the Fd pid and skip going through IOQ [2]. So a read_header requests sitting in IOQ that then exits could trigger a faulty index reset.

One approach is we could look for `gen_server:call` errors and skip resetting in those scenarios, but I think we would be better off explicitly resetting _only_ when we get the `no_valid_header` error. Also, the `couch_mrview_util:reset_index` call both truncates the index and writes a new base header [3], so it's quite possible that this code path is used to initialize empty indexes, and as such, I think we should keep the reset_index logic in here unless we can verify that indexes are bootstrapped always elsewhere and that this logic isn't needed.


[1] https://github.com/apache/couchdb/blob/17aec1f3339f864eb08869d6235753f0fdd28259/src/couch/src/couch_file.erl#L326-L332
[2] https://github.com/apache/couchdb/blob/17aec1f3339f864eb08869d6235753f0fdd28259/src/couch/src/couch_file.erl#L206-L223
[3] https://github.com/apache/couchdb/blob/17aec1f3339f864eb08869d6235753f0fdd28259/src/couch_mrview/src/couch_mrview_util.erl#L679

janl:
This looks good.

src/couch_mrview/src/couch_mrview_index.erl
```diff
@@ -122,7 +122,7 @@ open(Db, State0) ->
 
     case couch_mrview_util:open_file(IndexFName) of
         {ok, Fd} ->
-            case (catch couch_file:read_header(Fd)) of
+            case couch_file:read_header(Fd) of
```
chewbranca: So what happens when there's an error? I definitely think we should only trigger the index reset upon `no_valid_header`, but dropping the `catch` changes behavior to where errors can now bubble up out of this function, which may or may not be a good idea.

PR #2007, opened 2019-04-17
Port javascript tests auth cache, cookie auth and users db to elixir

jjrodrig:
## Overview

This PR ports the following javascript test suites into elixir
- auth_cache
- cookie_cache
- users_db

## Testing recommendations

```
 make elixir tests=test/auth_cache_test.exs
 make elixir tests=test/cookie_cache_test.exs
 make elixir tests=test/users_db_test.exs
```
## Checklist

- [] Documentation reflects the changes;


garrensmith:
This one test keeps failing https://travis-ci.org/apache/couchdb/jobs/521187100

jjrodrig:
@garrensmith I've been working on this trying to identify the source of test inconsistency.

Most of the problems appears after a request timeout, I've checked the JS test suite and it seems to no define any request timeout but the default request timeout in the Elixir test suite was set to 5 secs. I 've set the default timeout to 20 secs. This increases the success chances but it is still failing in some executions. 

I've also noticed that the most unstable build is the one that uses the 21.2.3 OTP Release. This build requires 40%-50% more time to complete compared to the other two builds. Seems to have some performance degradation that increases the chance of a request timeout. 

I've also modified some test checks to make it more robust, retry, check 202 responses... 

Thanks.

garrensmith:
Thanks for trying @jjrodrig I'll try and take a look later this week.

test/elixir/test/auth_cache_test.exs
```diff
@@ -0,0 +1,215 @@
+defmodule AuthCacheTest do
+  use CouchTestCase
+
+  @moduletag :authentication
+  @tag :with_db
+  test "test auth cache management", context do
```
iilyak: Nitpick: the output looks like (notice `test` mentioned twice):
```
AuthCacheTest
  * test test auth cache management (3692.6ms)
```

test/elixir/test/auth_cache_test.exs
```diff
@@ -0,0 +1,215 @@
...
+        :section => "chttpd_auth",
+        :key => "secret",
+        :value => generate_secret(64)
+      }
+    ]
+
+    run_on_modified_server(server_config, fn -> test_fun(db_name) end)
+  end
+
+  defp generate_secret(length) do
```
iilyak: Nitpick: avoid using names which have similarly named functions in Kernel.

test/elixir/test/auth_cache_test.exs
```diff
@@ -0,0 +1,215 @@
...
+    tab = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+
+    for _i <- 1..length, into: "" do
```
iilyak: This function can be written differently:
```
defp generate_secret(len) do
  tab = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
  Enum.take_random(tab, len)
end
```

test/elixir/test/auth_cache_test.exs
```diff
@@ -0,0 +1,215 @@
...
+    assert resp.error, "Login error is expected."
+  end
+
+  defp login(user, password) do
+    sess = Couch.login(user, password)
+    assert sess.cookie, "Login correct is expected"
+    sess
+  end
+
+  defp wait_until_compact_complete(db_name) do
```
iilyak: This is problematic for two reasons:
- there is no :timer.sleep(delay) in the loop
- there is no limit on either amount of time or number of attempts

I would recommend to implement it using `retry_until` similarly to https://github.com/apache/couchdb/blob/master/test/elixir/test/reshard_helpers.exs#L102:L104

test/elixir/test/auth_cache_test.exs
```diff
@@ -0,0 +1,215 @@
...
+    resp = Couch.post("/#{db_name}/_compact")
+    assert resp.status_code == 202
+    resp.body
+  end
+
+  def save_doc(db_name, body) do
+    resp = Couch.put("/#{db_name}/#{body["_id"]}", body: body)
+    assert resp.status_code in [201, 202]
+    assert resp.body["ok"]
+    {:ok, resp}
```
iilyak: Did you consider returning the document itself instead of `{:ok, resp}`? It might simplify things a bit in the `test_fun`.
```
def save_doc(db_name, body) do
    resp = Couch.put("/#{db_name}/#{body["_id"]}", body: body)
    assert resp.status_code in [201, 202]
    assert resp.body["ok"]
    Map.put(body, "_rev", resp.body["rev"])
end
```

test/elixir/test/auth_cache_test.exs
```diff
@@ -0,0 +1,215 @@
...
+
+  defp wait_until_compact_complete(db_name) do
+    info = info(db_name)
+
+    if info["compact_running"] do
+      wait_until_compact_complete(db_name)
+    end
+  end
+
+  defp assert_cache(event, user, password, expect \\ :expect_login_success) do
```
iilyak: nice!

test/elixir/test/auth_cache_test.exs
```diff
@@ -0,0 +1,215 @@
...
+    # assert_cache(:expect_hit, "fdmanana", "foobar", :expect_login_fail)
+    # assert_cache(:expect_hit, "fdmanana", "javascript")
+
+    delete_doc(db_name, fdmanana)
+
+    assert_cache(:expect_hit, "fdmanana", "javascript", :expect_login_fail)
+
+    # login, compact authentication DB, login again and verify that
+    # there was a cache hit
+    assert_cache(:expect_hit, "johndoe", "123456")
```
iilyak: **comment for other reviewers**: The correspondent lines in the JavaScript test [are](https://github.com/apache/couchdb/blob/master/test/javascript/tests/auth_cache.js#L248:L249):
```
// XXX BUGGED T(misses_after === (misses_before + 1));
// XXX BUGGED T(hits_after === hits_before);
```
- the check is commented out
- the commented out lines expect a cache **miss**

test/elixir/test/auth_cache_test.exs
```diff
@@ -0,0 +1,215 @@
...
+    compact(db_name)
+    wait_until_compact_complete(db_name)
+    assert_cache(:expect_hit, "johndoe", "123456")
```
iilyak: **comment for other reviewers**: The correspondent lines in the JavaScript test [are](https://github.com/apache/couchdb/blob/master/test/javascript/tests/auth_cache.js#L264:L265):
```
// XXX BUGGED T(misses_after === misses_before);
// XXX BUGGED T(hits_after === (hits_before + 1));
```
- the check is commented out

test/elixir/test/cookie_auth_test.exs
```diff
@@ -0,0 +1,415 @@
...
+    sess = Couch.login(user, password)
+    assert sess.cookie, "Login correct is expected"
+    sess
+  end
+
+  defp logout(session) do
+    assert Couch.Session.logout(session).body["ok"]
+  end
+
+  defp loginAsUser(user) do
```
iilyak: I think this function should be named  `login_as_user` or `login_as` (to match with `open_as`).

test/elixir/test/cookie_auth_test.exs
```diff
@@ -0,0 +1,415 @@
...
+    resp
+  end
+
+  defp open_as(db_name, doc_id, options) do
+    use_session = Keyword.get(options, :use_session)
+    user = Keyword.get(options, :user)
+    expect_response = Keyword.get(options, :expect_response, 200)
+    expect_message = Keyword.get(options, :error_message)
+
+    session =
```
iilyak: There is a shortcut to express it:
```
session = Keyword.get(options, :use_session)
  || loginAsUser(user)
```

iilyak: or since you need value of `use_session` below. We can say: `session = use_session || loginAsUser(user)`

test/elixir/test/cookie_auth_test.exs
```diff
@@ -0,0 +1,415 @@
...
+    resp.body
+  end
+
+  defp save_as(db_name, doc, options) do
+    use_session = Keyword.get(options, :use_session)
+    user = Keyword.get(options, :user)
+    expect_response = Keyword.get(options, :expect_response, [201, 202])
+    expect_message = Keyword.get(options, :error_message)
+
+    session =
```
iilyak: Same as above `session = use_session || loginAsUser(user)`.

test/elixir/test/cookie_auth_test.exs
```diff
@@ -0,0 +1,415 @@
...
+    resp
+  end
+
+  defp delete_as(db_name, doc, options) do
+    use_session = Keyword.get(options, :use_session)
+    user = Keyword.get(options, :user)
+    expect_response = Keyword.get(options, :expect_response, [200, 202])
+    expect_message = Keyword.get(options, :error_message)
+
+    session =
```
iilyak: Same as above `session = use_session || loginAsUser(user)`

test/elixir/test/cookie_auth_test.exs
```diff
@@ -0,0 +1,415 @@
...
+
+    create_doc_expect_error(@users_db, bad_id_user_doc, 403, "forbidden")
+
+    # login works
+    session = loginAsUser("Jason Davies")
+    info = Couch.Session.info(session)
+    assert info["userCtx"]["name"] == "Jason Davies"
+    assert not Enum.member?(info["userCtx"]["roles"], "_admin")
+
+    jason_user_doc =
```
iilyak: JavaScript test has a comment [here](https://github.com/apache/couchdb/blob/master/test/javascript/tests/cookie_auth.js#L153)  `// update one's own credentials document`

iilyak:
This PR is almost there. You did an enormous amount of work. It just needs little refactoring. My worry is that in the current form it is too close to JavaScript's big monolithic test. It would be nice to split big test into multiple smaller ones. The elixir framework in its current shape has few features which can be used here. For example instead of relying on `run_on_modified_server` it might be possible to use `@moduletag config`

```
defmodule CookieAuthTest do
  ...
  @moduletag config: [
     {"chttpd_auth", "authentication_db", db_name},
     {"chttpd_auth", "auth_cache_size", "3"},
     {"httpd", "authentication_handlers", "{couch_httpd_auth, default_authentication_handler}"},
     {"chttpd_auth", "secret", generate_secret(64)}  
  ]
  test "that the users db is born with the auth ddoc" do
     ...
  end

  test "we can't create _names" do
    ...
  end

  test "we can't create malformed ids" do
    ...
  end

  test "login works" do
    ...
  end

  test "can't delete another users doc unless you are admin" do
    ...
  end

  test "redirect on success" do
    ...
  end

  test "redirect on failure" do
    ...
  end

  ...and so on...
end
```

jjrodrig:
Thanks @iilyak  for your review. I've updated the PR with the changes you proposed.

I agree with you that the current form of the test is not ideal but I tried to do just a transcription of the original test keeping the previous test structure. 

I'll check if we can use new elixir test suite features for test splitting

jjrodrig:
Hi @iilyak,

 I've been working around your proposal of split the tests. I found it quite difficult for auth_cache
and cookie_cache but for users_db seems to be affordable. 

 It requires major rework of the test. Do you think that we can follow with this PR as it is now and split the tests in different PRs? 

 I would like to continue contributing to the rewrite of the test suite into elixir. 

Thanks, 
Juanjo

test/elixir/test/cookie_auth_test.exs
```diff
@@ -0,0 +1,401 @@
...
+
+    resp =
+      save_as(
+        @users_db,
+        jchris_user_doc,
+        user: "jan"
+      )
+
+    # wait for auth cache invalidation
+    :timer.sleep(500)
```
iilyak: this would be a source of flakiness

test/elixir/test/cookie_auth_test.exs
```diff
@@ -0,0 +1,401 @@
...
+      |> Map.put("foo", 3)
+
+    save_as(
+      @users_db,
+      jason_user_doc,
+      use_session: session,
+      expect_response: 404,
+      error_message: "not_found"
+    )
+
```
iilyak: JavaScript test has "// make sure we cant create duplicate users" comment.

test/elixir/test/cookie_auth_test.exs
```diff
@@ -0,0 +1,401 @@
...
+          :key => "jchris",
+          :value => "funnybone"
+        }
+      ],
+      &test_change_admin_fun/0
+    )
+
+    # log in one last time so run_on_modified_server can clean up the admin account
+    login("jan", "apple")
+  after
```
iilyak: This is usually a job for test teardown. In ExUnit it is accomplished via [`on_exit`](https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#on_exit/2)

iilyak:
Hi @jjrodrig, 

> I've been working around your proposal of split the tests. I found it quite difficult for auth_cache
and cookie_cache but for users_db seems to be affordable.

I agree with you observation that `auth_cache` seem to be harder to split. The `cookie_cache` on the other hand is not that hard (unless I am missing some subtle detail).

>Do you think that we can follow with this PR as it is now and split the tests in different PRs?

Could you convert from  `run_on_modified_server` to use of @moduletag config?
In my opinion the current shape of the `test_fun` is written in not conventional Elixir. Which makes it harder to understand and hard to maintain. I also understand the reason why it is like this. There is a positive side of it as well. Due to similarities with JavaScript test it makes the PR easier to review. Let's try to convert test setup/teardown logic into idiomatic ExUnit:
- setup uses moduletag
- teardown is done using `on_exit` 

Then in separate commit try to split some tests when possible. Keep the rest as is and ask @garrensmith to review as well. To get his opinion on whether it is acceptable or not.

jjrodrig:
Hi @iilyak, I've modified users_db and cookie_auth test and now are using moduletag and setup/teardown mechanisms from ExUnit.

@garrensmith, please could you look at this PR to now whether the tests are acceptable in this form?

Thanks,
Juanjo

janl:
Merged in https://github.com/apache/couchdb/pull/2394

PR #2023, opened 2019-05-03
Jenkins: Add ARM64, drop trusty, improve Jenkinsfile

wohali:
Because JENKINS-46277 and JENKINS-47106 got fixed, the Jenkinsfile can
be made much simpler and more readable since we are able to use docker
agents directly now.

Other changes:

* We now have an ARM64 build ready to go.
  * It's currently failing some couch_btree tests with timeouts.
    I've seen these failures on x86_64, though.

* If any one of the builds fails, Jenkins now kills all the rest.
  This is useful because, in Jenkins, you can currently only
  restart an entire pipeline stage, not just a single entry in the
  pipeline. So there's no point in not killing the build fast, for
  now. You can see which build actually failed by clicking through each
  parallel stage and looking for a failed build step.

* We're trying to save space on the drives:
  * We now only keep the last 10 builds and their artifacts around.
  * We also forcibly clean up our temp directories after any build.

* Build's a bit faster since we're no longer multiply checking out
  the source code for each parallel build.

* Improved stage naming.

* Formatting is all nice nice.

https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/jenkins-arm64-trusty/10/pipeline is the latest build attempt, which (just like master) is failing and there's nothing we can do about it for now.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,21 +1,56 @@
 #!groovy
-/*
-Licensed under the Apache License, Version 2.0 (the "License"); you may not
-use this file except in compliance with the License. You may obtain a copy of
-the License at
+//
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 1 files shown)


wohali:
Going to merge since no one ever seems to review my Jenkins changes anyway.

PR #2029, opened 2019-05-07
Fix epoch mismatch errors (#2027)

wohali:
Backport of fix to master for 2.3.x. Original commit follows:

Originally we posited that duplicate UUIDs could never be created.
However due to operations interventions its possible to unintentionally
copy UUIDs in ways that violate this assumption. Instead of crashing the
process we just reset the seq to zero and log a warning as was done in
the other instances when we have mismatches in the epoch history.


wohali:
Thanks!

PR #2037, opened 2019-05-25
Add "Dreyfus" (Erlang side of Search) to CouchDB

kocolosk:
## Overview

This contribution pulls in the "Dreyfus" OTP application, which teaches CouchDB how to communicate with the "Clouseau" Java application for Lucene-powered Search and Mango text indexes. Once this PR is merged it will no longer be necessary to use a custom fork of CouchDB in order to enable Search.

## Testing recommendations

In order to fully test this PR you'll need a build of Clouseau. Those can be found at

* https://github.com/cloudant-labs/clouseau/releases

There's also a public Maven repo:

* https://maven.cloudant.com/repo/com/cloudant/clouseau/

I've gone ahead and built a Docker image which can be used with Kubernetes:

* https://cloud.docker.com/repository/docker/kocolosk/couchdb-search

and updated the upstream Helm chart to use it. If you want a one-liner to install CouchDB+Search on a Kube cluster of your choice you can try:

```
helm install incubator/couchdb --set image.repository=kocolosk/couchdb,image.tag=dreyfus-by-default,clusterSize=1,enableSearch=true --name test-search
```

apache/couchdb-documentation#418
helm/charts#14822

## Checklist

Posting a draft PR for discussion. Will lift the draft flag once documentation and a Clouseau build are available.


wohali:
Will this cause all CI runs or `make check` to fail unless Clouseau is present? If so, please don't merge until this is addressed.

One possibility is to comment out all tests until a solution can be determined.

As a reminder, it is very challenging to get Java <=7 into [Debian](https://packages.debian.org/search?keywords=openjdk-7&searchon=names&suite=all&section=all) and [Ubuntu](https://packages.ubuntu.com/search?keywords=openjdk-7&searchon=names&suite=all&section=all) now technically, and there are issues with Java <=7 [potentially legally](https://www.infoworld.com/article/3269332/oracle-sets-date-for-end-of-java-8-updates.html) for our users.

Do you intend to bring Clouseau into the ASF as well, or is the intent to stay with independent binary releases of Clouseau (i.e., direct from IBM)?

kocolosk:
`make check` succeeds without Clouseau.

I’m happy to solicit input from the community, but my initial thought is to provide Clouseau builds directly from IBM. It’s a low-traffic project and not one we’re actively trying to build a developer community around, in part because the FoundationDB work will require a ~complete rewrite of it.

My understanding Clouseau is running fine and passing all tests on Java 8 in our dev environments. It just can’t be built using that JDK.

wohali:
@kocolosk Two more things occurred to me in the shower:

1. This will let users create `index` secondary indexes even if Clouseau isn't running. What sort of feedback could  Couch-With-Dreyfus-Without-Clouseau give when it's never able to build that index - either at create or query time? Or, is this feature hide-able behind a feature flag that is disabled by default?

2. We need docs :smile:

kocolosk:
Agree on both of those. The Cloudant Search docs are a good start; I’m going to see if we have any tooling that helps translate from the IBM Cloud format to the Apache CouchDB one.

I do think some “Clouseau isn’t running” feedback would be very useful; currently I think Dreyfus just times out with a generic error message if it fails to connect in 30 seconds. Do you think that improvement needs to happen before the merge?

wohali:
I'd prefer a feature flag that defaults to disabled. If it could auto-enable if it sees Clouseau in the cluster, that'd be amazing, but over and above the call of duty.

Envisioned user install experience would be something like:
1. Install CouchDB
1. Install Clouseau and daemonize it
1. (if not automatic) Turn on feature flag

or
1. Install CouchDB - nothing different from today if you don't have Clouseau installed, no "surprises"

kocolosk:
@nowisesys pointed out an interesting issue with this Pull Request. We have an ancient pattern in Mango which does a `module_loaded(dreyfus_index)` to check for the presence of Dreyfus before advertising text indexes. In release mode that module is not loaded on startup, so even if Dreyfus is included in a release the check can fail. Oops!

A trivial switch here would be to use `dreyfus_sup` instead of `dreyfus_index` as the former will be loaded as part of the supervision tree, but that doesn't actually confirm a working Search system since this PR proposes to always start the Dreyfus app. I think we should come up with a smarter approach -- perhaps testing for a connected Clouseau node instead?

/cc @rnewson @tonysun83 for comment

kocolosk:
Also, lifting the draft PR flag since we now have an associated documentation PR in apache/couchdb-documentation#418, and we also have a public Clouseau release:

* https://github.com/cloudant-labs/clouseau/releases
* https://maven.cloudant.com/repo/com/cloudant/clouseau/2.17.0/
* https://cloud.docker.com/repository/docker/kocolosk/couchdb-search

wohali:
-1 on landing until this gets cleared up and my previous comment (re: 
principle of least astonishment and Search not working if Clouseau isn't 
connected) is addressed.

+1 once it is. :)

kocolosk:
Yep understood and agreed.

rnewson:
Testing for clouseau is a better approach but be aware that dreyfus intentionally tolerates clouseau not being up, any search request is retried a few times (with a sleep inbetween) if it’s down, in order to mask a clouseau reboot (typically for upgrades, but JVM’s do sometimes OOM).

kocolosk:
I made a small change to replace the check for the `dreyfus_index` module with a test message to the Clouseau node. Here's a summary how CouchDB behaves when Clouseau is not connected:

### Search

* Updates to a Search design document (i.e., one with a top-level `"indexes"` field) **succeed**. I doubt we want to change this as it would have significant implications for replication.
* Search requests fail after ~30 seconds with a 500 status code and `{"error":"timeout","reason":"The request could not be processed in a reasonable amount of time."}`. This could be improved.

### Mango

* Attempts to create a Mango text index via the `_index` endpoint fail ~immediately with a 503 status and the following response: `{"error":"required index service unavailable","reason":"text"}`. I think this is the right behavior.
* The Mango query optimizer will skip any existing "text" indexes if the Clouseau subsystem is not available at query time. This could result in a previously fast query falling back to a full database scan. I'm OK with this.

I think there are two followups:

1. Reject the Search request immediately instead of timing out with a generic error
1. Consider some retry logic to mask Clouseau reboots during upgrades as suggested by @rnewson. I could see doing a short bounded exponential backoff for a couple of seconds.

rnewson:
for ref this is what dreyfus does for retries: https://github.com/cloudant-labs/dreyfus/blob/master/src/dreyfus_httpd.erl#L585 but obviously that is only for direct _search http requests, not via mango.

chintan-mishra:
As a user, I have a question.

Why not include Clouseau as well?

kocolosk:
@chintan-mishra good question.

The Closeau codebase is a very different thing. It's all Scala/Java, so the release packaging machinery that we use for the main Erlang codebase wouldn't be applicable. It runs as a completely separate process, so the scripts that we use to start/stop/manage CouchDB wouldn't apply. It's also a relatively mature project and not one where the current IBM maintainers foresee a lot of additional contributions.

I'd like to hear your opinion as a user. I can imagine you want the Search add-on to be enabled as simply as possible. I'd like to see that too, but personally I think the right approach is to ensure that when the Clouseau service is installed using the convenience binaries CouchDB recognizes it and uses it automatically, rather than pulling the whole codebase "in-tree".

AlexanderKaraberov:
Hello @kocolosk 

> It's also a relatively mature project and not one where the current IBM maintainers foresee a lot of additional contributions.

The most fantastic thing would be to update Clouseau to Java 8+ and Lucene 8.  I've already opened an issue for this: https://github.com/cloudant-labs/clouseau/issues/22 
but still didn't get an answer on what is the approximate timeframe for this to land. Are there any plans in the Cloudant roadmap to update Clouseau. JDK8 compatibility would be a major boost for this project as well as Lucene upgrade.

rnewson:
Updating to Java 8 is non-trivial as several dependencies need updating and there is no upstream support for them anymore. a Pull Request for all that work would move things forward, of course, but an Issue reporting what we already know isn't going to.

As the author of dreyfus/clouseau, I should point out that those projects apply to CouchDB 2.0 specifically, they are tightly coupled/aligned with that architecture.

Future versions of CouchDB that use FoundationDB as the base layer will need an alternative solution, which I, and my team, are working on right now (http://github.com/cloudant-labs/fdblucene is the fundamental piece). This work is pure Java and targets JDK 8, though I personally develop against JDK 11.

kocolosk:
The latest commit preserves the exponential backoff for Search requests, but if the final error message indicates an inability to connect to Clouseau the error message will be

```
{"error":"ou_est_clouseau","reason":"Could not connect to the Clouseau Java service at clouseau@127.0.0.1"}
```

That seems like a reasonable improvement to me.

I also did an ad hoc measurement of the time required to do the `clouseau_rpc:connected()` check. Unfortunately, on a Kubernetes deployment with CouchDB and Clouseau in the same Pod this check cost more than **3 milliseconds**, which seems an unacceptably long amount of time to waste on the happy path. I will think a bit more about how to guard those Mango requests in a way that doesn't introduce too much latency to a well-configured system.

kocolosk:
OK, so now the `connected()` check first looks at the hidden nodes to see if we already have a connection to Clouseau. This check is much cheaper, around 0.1 ms. If no connection exists already, it tries the `version()` RPC to see if a connection can be made.

Personally I'm satisfied with the PR at this point.

chintan-mishra:
@kocolosk yes I would prefer Full-Text Search to be one of the config options in the CouchDB config.

janl:
very happy to see this, thanks everyone!

kocolosk:
Doing a traditional merge commit here to be consistent with our past process on `git-subtree` merges. Thanks for the review everyone!

PR #2056, opened 2019-06-19
Blacklist all 21.x releases older than 21.2.3

davisp:
This basically just extends the black list to cover the `21.{0,1}`
release range. This is due to a compiler bug [1] which is a duplicate of
[2].

[1] https://bugs.erlang.org/browse/ERL-981
[2] https://bugs.erlang.org/browse/ERL-807

---

I'm not 100% on this. Technically this range contains the bug as mentioned in the tickets, however I've got no idea if anywhere in our code base actually triggers the bug. I can't think of anyone that's reported mysterious errors on these versions but we're not running things against it in Travis either. So basically a big ol' shrug either way from me at this point. Though my risk adversity leans toward blacklisting.

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -86,7 +86,7 @@ case VerList of
     [19 | _] -> NotSupported(VerString);
 
     [20 | _] = V20 when V20 < [20, 3, 8, 11] -> BadErlang(VerString);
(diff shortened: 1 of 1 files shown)


davisp:
Related PR suggested by @wohali: https://github.com/apache/couchdb-documentation/pull/422

PR #2077, opened 2019-07-21
Make sure that fsync errors are raised

davisp:
This changes `couch_file` to ensure that errors are raised when a call
to `fsync` fails. It will also stop the couch_file process to ensure
that anything handling a failed `fsync` won't attempt to retry the
operation and experience issues discovered by Postgres [1].

[1] http://danluu.com/fsyncgate/

## Overview

See http://danluu.com/fsyncgate/

## Testing recommendations

`make eunit`

However, I don't have a good test to assert that the `couch_file` pid exits since the `file` module is quite difficult to get mocked. I'm all ears on how to make this better. My only thought was that I'd have to export a function that can be mocked but it seems weird to have `handle_call` calling a fully qualified exported function just to test that.

## Checklist

- [x] N/A - A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation

### DIFF
--- a/src/couch/src/couch_file.erl
+++ b/src/couch/src/couch_file.erl
@@ -215,12 +215,26 @@ truncate(Fd, Pos) ->
 sync(Filepath) when is_list(Filepath) ->
     case file:open(Filepath, [append, raw]) of
         {ok, Fd} ->
-            try ok = file:sync(Fd) after ok = file:close(Fd) end;
+            try
+                case file:sync(Fd) of
+                    ok ->
+                        ok;
+                    {error, Reason} ->
+                        erlang:error({fsync_error, Reason})
+                end
+            after
+                ok = file:close(Fd)
+            end;
         {error, Error} ->
             erlang:error(Error)
     end;
 sync(Fd) ->
-    gen_server:call(Fd, sync, infinity).
+    case gen_server:call(Fd, sync, infinity) of
+        ok ->
+            ok;
+        {error, Reason} ->
+            erlang:error({fsync_error, Reason})
+    end.
 
 %%----------------------------------------------------------------------
 %% Purpose: Close the file.
@@ -462,7 +476,16 @@ handle_call({set_db_pid, Pid}, _From, #file{db_monitor=OldRef}=File) ->
     {reply, ok, File#file{db_monitor=Ref}};
 
 handle_call(sync, _From, #file{fd=Fd}=File) ->
-    {reply, file:sync(Fd), File};
+    case file:sync(Fd) of
+        ok ->
+            {reply, ok, File};
+        {error, _} = Error ->
+            % We're intentionally dropping all knowledge
+            % of this Fd so that we don't accidentally
+            % recover in some whacky edge case that I
+            % can't fathom.
+            {stop, Error, Error, #file{fd = nil}}
+    end;
 
 handle_call({truncate, Pos}, _From, #file{fd=Fd}=File) ->
     {ok, Pos} = file:position(Fd, Pos),
--- a/src/couch/test/couch_file_tests.erl
+++ b/src/couch/test/couch_file_tests.erl
@@ -498,3 +498,36 @@ make_delete_dir_test_case({RootDir, ViewDir}, DeleteAfterRename) ->
 remove_dir(Dir) ->
     [file:delete(File) || File <- filelib:wildcard(filename:join([Dir, "*"]))],
     file:del_dir(Dir).
+
+
+fsync_error_test_() ->
+    {
+        "Test fsync raises errors",
+        {
+            setup,
+            fun() ->
+                test_util:start(?MODULE, [ioq])
+            end,
+            fun(Ctx) ->
+                test_util:stop(Ctx)
+            end,
+            [
+                fun fsync_raises_errors/0
+            ]
+        }
+    }.
+
+
+fsync_raises_errors() ->
+    Fd = spawn(fun() -> fake_fsync_fd() end),
+    ?assertError({fsync_error, eio}, couch_file:sync(Fd)).
+
+
+fake_fsync_fd() ->
+    % Mocking gen_server did not go very
+    % well so faking the couch_file pid
+    % will have to do.
+    receive
+        {'$gen_call', From, sync} ->
+            gen:reply(From, {error, eio})
+    end.


src/couch/src/couch_file.erl
```diff
@@ -215,12 +215,26 @@ truncate(Fd, Pos) ->
 sync(Filepath) when is_list(Filepath) ->
     case file:open(Filepath, [append, raw]) of
         {ok, Fd} ->
-            try ok = file:sync(Fd) after ok = file:close(Fd) end;
+            try
+                case file:sync(Fd) of
+                    ok ->
+                        ok;
+                    Else ->
+                        erlang:error({fsync_error, Else})
```
wohali: Was going to complain about the "surprise" of `erlang:error` here, but I see we use it in a couple of other places, and this is serious enough that it's the only thing I can think of that makes sense.

`erlang:throw` might get caught higher up (I didn't look carefully) and it's good enough for our `couch_file` to die and restart.

If the disk subsystem is really broken, we'll get throttled on trying to restart `couch_file`.

At what point will we fully attempt to restart?

davisp: What do you mean by "fully restart"? In terms of the database and outstanding writes and re-applying writes that'd be at the first error since couch_file exiting will take down couch_db_updater and friends.

In terms of restarting the entire `beam.smp` process I'd think it'd depend on the error itself. Normally when EIO happens and the filesystem panics into read-only we tend to spew a lot of errors from various places which ends up causing supervision to restart the process eventually. In terms of ENOSPC or similar I think we'd never completely restart unless we happen to bubble up through the auth cache trying to write users or being unable to write docs into the dbs database or some other system related write activity.

src/couch/src/couch_file.erl
```diff
@@ -462,7 +476,12 @@ handle_call({set_db_pid, Pid}, _From, #file{db_monitor=OldRef}=File) ->
     {reply, ok, File#file{db_monitor=Ref}};
 
 handle_call(sync, _From, #file{fd=Fd}=File) ->
-    {reply, file:sync(Fd), File};
+    case file:sync(Fd) of
+        ok ->
+            {reply, ok, File};
+        Error ->
+            {stop, Error, Error, File}
```
rnewson: any reason to keep `File` in state? save another accident by setting it to `nil` here?

src/couch/src/couch_file.erl
davisp: Also I should say that unlike Postgres we don't have to do a complete restart since our append-only design basically means any successful couch_file open on a shard is basically equivalent to their "restart from valid WAL segment" or whatever they call it.

Or rather, the recovery mode thing they discussed at length we basically perform every time we open a shard since we're seeking back to the last valid header and skipping over any partially written data.

src/couch/src/couch_file.erl
```diff
@@ -215,12 +215,26 @@ truncate(Fd, Pos) ->
 sync(Filepath) when is_list(Filepath) ->
     case file:open(Filepath, [append, raw]) of
         {ok, Fd} ->
-            try ok = file:sync(Fd) after ok = file:close(Fd) end;
+            try
+                case file:sync(Fd) of
+                    ok ->
+                        ok;
+                    Else ->
```
rnewson: I'd prefer to match`{error, Reason}` and then `erlang:error({fsync_error, Reason}).` Should file:sync return something else, the crash from the function_clause is still an error.

src/couch/src/couch_file.erl
wohali: Right, my concern is if a disk/volume entirely goes away, and _every single `couch_file`_ starts EIOing out, the entire VM will probably tailspin spewing lots of 500s. I guess this is closer to the read-only panic mode. (I don't want to get into what happens if an NFS-mounted volume goes away, that usually just results in the entire process/kernel hanging on file I/O.)

I think what you're saying is that the supervision tree might try and completely restart in this case, yes? Or does it actually exit, leaving it to `runit`/`systemd`/whatever?

davisp: Right, if enough of the supervision tree dies often enough eventually `beam.smp` will exit and then rely on a service supervisor to restart it. But that requires taking an entire application down so with IO errors that basically means some sort of CouchDB "system" level activity that ends up taking down an supervised process fast enough.

Normal databases hitting errors will just add churn through couch_server and spew 500s for IO activity hitting IO issues.

src/couch/src/couch_file.erl
davisp: Done.

janl:
make test complains:

```
Test fsync raises errors
      couch_file_tests: fsync_error_test_...*failed*
in function couch_file_tests:'-fsync_raises_errors/0-fun-1-'/1 (test/couch_file_tests.erl, line 523)
in call from eunit_proc:run_test/1 (eunit_proc.erl, line 510)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 335)
in call from eunit_proc:handle_test/2 (eunit_proc.erl, line 493)
in call from eunit_proc:tests_inorder/3 (eunit_proc.erl, line 435)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 325)
in call from eunit_proc:run_group/2 (eunit_proc.erl, line 549)
**error:{assertException,
    [{module,couch_file_tests},
     {line,523},
     {expression,"couch_file : sync ( Fd )"},
     {pattern,"{ error , { fsync_error , eio } , [...] }"},
     {unexpected_exception,
         {error,
             {case_clause,eio},
             [{couch_file,sync,1,[{file,...},{...}]},
              {couch_file_tests,'-fsync_raises_errors/0-fun-1-',1,[{...}|...]},
              {eunit_test,run_testfun,1,[...]},
              {eunit_proc,run_test,1,...},
              {eunit_proc,with_timeout,...},
              {eunit_proc,...},
              {...}|...]}}]}
  output:<<"">>
```

src/couch/src/couch_file.erl
```diff
@@ -215,12 +215,26 @@ truncate(Fd, Pos) ->
 sync(Filepath) when is_list(Filepath) ->
     case file:open(Filepath, [append, raw]) of
```
janl: > 21:50:18]  <+davisp>	The first version of couch_file:sync that takes a file path isn't called anymore. That was a remnant from when couch_db_updater had two couch_file processes

Why not remove this then?

davisp:
I’m not against it but given the already sneaky bits of fsync I hesitated
to do too much in a single PR.

davisp:
For make check I probably forgot to run it after responding to Bob’s comments. I’ll fix it this afternoon.

src/couch/src/couch_file.erl
janl: > I’m not against it but given the already sneaky bits of fsync I hesitated
to do too much in a single PR.

Fair, carry on.

janl:
> Our delayed commits option, is hard-coded to 1 second "only". And later fixes in the kernel do ensure that if we keep the FD open the error will be reported eventually.

Good extra info! I'm just glad, this is on the deprecation track already

davisp:
For `close` I knew that was true but our algorithm is gated around fsync so I mostly ignored it. If we did get into the situation where we had written data but then closed before an fsync it'd be because we were shutting down. The lack of header write there would save us from any issues.

As for the "fsync will forget about errors" issue... well ain't that a kick in the pants? I agree there's not a whole lot we can do. Although I will have to say that given the length of time multiple popular databases have run on Linux without even a whisper of "fsync is totally broken" makes me think that actually encountering that particular issue in the wild is likely fairly miniscule.

davisp:
@janl My phone replied weird earlier so not sure if you saw that. Let me know if you have a preference on including the removal of the filename based fsync approach.

As for the test that was me forgetting I'd mocked something. Just needed to update the test.

wohali:
@davisp that "forget about errors" issue is precisely why I reported this to the project. Terrifying.

janl:
@davisp happy to leave the remove task for later.

PR #2083, opened 2019-07-28
Increase timeouts on two slow btree tests

kocolosk:
## Overview

These two tests are reliably timing out on ARM hardware in Jenkins. They do a lot of individual btree operations so this is not entirely surprising. Appropriate course of action here is to raise the timeout.

## Testing recommendations

I don't know how to make Jenkins build a PR branch. If someone can do that, great. Otherwise we should probably just merge this and see how it does. I've confirmed that the timeout setting works correctly locally (e.g. by lowering it to 0.1 and watching those tests time out).

## Checklist

### DIFF
--- a/src/couch/test/couch_btree_tests.erl
+++ b/src/couch/test/couch_btree_tests.erl
@@ -16,6 +16,7 @@
 -include_lib("couch/include/couch_db.hrl").
 
 -define(ROWS, 1000).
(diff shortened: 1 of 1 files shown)


wohali:
@kocolosk if you rename this branch anything starting with `jenkins-` then it'll get autobuilt.

kocolosk:
Cool. Closing in favor of #2084 , then.

PR #2084, opened 2019-07-28
Increase timeouts on two slow btree tests

kocolosk:
## Overview

These two tests are reliably timing out on ARM hardware in Jenkins. They do a lot of individual btree operations so this is not entirely surprising. Appropriate course of action here is to raise the timeout.

## Testing recommendations

This branch should get built automatically by Jenkins so we can confirm that it fixes the build there.

I've confirmed that the timeout setting works correctly locally (e.g. by lowering it to 0.1 and watching those tests time out).

## Checklist

### DIFF
--- a/src/couch/test/couch_btree_tests.erl
+++ b/src/couch/test/couch_btree_tests.erl
@@ -16,6 +16,7 @@
 -include_lib("couch/include/couch_db.hrl").
 
 -define(ROWS, 1000).
+-define(TIMEOUT, 60). % seconds
 
 
 setup() ->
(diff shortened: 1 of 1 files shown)


wohali:
I had to poke the Jenkins config...recent ASF Infra change.

https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/jenkins-bump-btree-test-timeouts/1/pipeline

is being built.

wohali:
@kocolosk looks like Jenkins is still failing on a lot of platforms, thanks to other timeouts, but this is probably a necessary (if not sufficient) change.

+1

kocolosk:
Yep, got it -- looks like it's on to `mem3` now.

PR #2088, opened 2019-07-30
Refactor Jenkins setup to enable test result capture

kocolosk:
## Overview

This is a non-trivial refactor of the Jenkins setup:

* Builds are moved back into the workspace. Parallel builds run in unique sub-folders which seems to avoid the clashes we had observed in the past.
* The large build_script is split out into smaller parts as sequential stages. Cleanup operations are deferred to `post` actions to enable retrieval of the test results.
* Test results are captured and saved in the Jenkins UI.
* Various fixes to platform labels and configuration that should get the nightly builds running again.

## Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -144,6 +144,9 @@ pipeline {
             } // withEnv
           } // steps
           post {
+            always {
+              junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+            }
             cleanup {
               sh 'rm -rf $COUCHDB_IO_LOG_DIR'
             }
@@ -172,6 +175,11 @@ pipeline {
               steps {
                 unstash 'tarball'
                 sh( script: build_and_test )
+              }
+              post {
+                always {
+                  junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+                }
               }
             }
             stage('Build CouchDB packages') {
@@ -214,6 +222,11 @@ pipeline {
               steps {
                 unstash 'tarball'
                 sh( script: build_and_test )
+              }
+              post {
+                always {
+                  junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+                }
               }
             }
             stage('Build CouchDB packages') {
@@ -258,6 +271,11 @@ pipeline {
                 unstash 'tarball'
                 sh( script: build_and_test )
               }
+              post {
+                always {
+                  junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+                }
+              }
             }
             stage('Build CouchDB packages') {
               steps {
@@ -299,6 +317,11 @@ pipeline {
               steps {
                 unstash 'tarball'
                 sh( script: build_and_test )
+              }
+              post {
+                always {
+                  junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+                }
               }
             }
             stage('Build CouchDB packages') {
@@ -342,6 +365,11 @@ pipeline {
                 unstash 'tarball'
                 sh( script: build_and_test )
               }
+              post {
+                always {
+                  junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+                }
+              }
             }
             stage('Build CouchDB packages') {
               steps {
@@ -384,6 +412,11 @@ pipeline {
                 unstash 'tarball'
                 sh( script: build_and_test )
               }
+              post {
+                always {
+                  junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+                }
+              }
             }
             stage('Build CouchDB packages') {
               steps {
@@ -425,6 +458,11 @@ pipeline {
               steps {
                 unstash 'tarball'
                 sh( script: build_and_test )
+              }
+              post {
+                always {
+                  junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+                }
               }
             }
             stage('Build CouchDB packages') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -246,7 +246,7 @@ pipeline {
             timeout(time: 90, unit: "MINUTES")
           }
           environment {
-            platform = 'xenial'
+            platform = 'bionic'
           }
           steps {
             sh 'rm -f apache-couchdb-*.tar.gz'
@@ -300,7 +300,7 @@ pipeline {
             timeout(time: 90, unit: "MINUTES")
           }
           environment {
-            platform = 'jessie'
+            platform = 'stretch'
           }
           steps {
             sh 'rm -f apache-couchdb-*.tar.gz'
@@ -327,7 +327,7 @@ pipeline {
             timeout(time: 90, unit: "MINUTES")
           }
           environment {
-            platform = 'jessie'
+            platform = 'aarch64-debian-stretch'
           }
           steps {
             sh 'rm -f apache-couchdb-*.tar.gz'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -457,7 +457,9 @@ pipeline {
             stage('Build from tarball & test') {
               steps {
                 unstash 'tarball'
-                sh( script: build_and_test )
+                withEnv(['MIX_HOME='+pwd(), 'HEX_HOME='+pwd()]) {
+                  sh( script: build_and_test )
+                }
               }
               post {
                 always {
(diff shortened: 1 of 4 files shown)


kocolosk:
Confirmed that the new `Jenkinsfile` passes the validator:

```
✗ curl -X POST -F "jenkinsfile=<Jenkinsfile" $JENKINS_URL/pipeline-model-converter/validate
Jenkinsfile successfully validated.
```

wohali:
@kocolosk can you rename this so it starts with `jenkins-` so we can test it? thx.

kocolosk:
Absolutely -- I assumed Jenkins would not be able to do branch tests of the `Jenkinsfile` itself.

kocolosk:
I just pushed a copy to `jenkins-junit-reports-and-jenkins-env`

wohali:
@kocolosk on the contrary, it can and does! Changes to `Jenkinsfile` are best run thru a `jenkins-*` branch for testing.

https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/jenkins-junit-reports-and-jenkins-env/1/pipeline

picked up and just started running.

kocolosk:
I believe I'm done making changes here, would love a review when feasible.

kocolosk:
OK I lied, I popped one more timeout fix on top :)

src/couch/test/eunit/couch_db_split_tests.erl
```diff
@@ -16,6 +16,7 @@
 -include_lib("couch/include/couch_db.hrl").
 
 -define(RINGTOP, 2 bsl 31).
+-define(TIMEOUT, 60). % seconds
```
wohali: Can you please leave the timeout bumps in a separate commit to the Jenkins setup / junit formatting stuff?

src/ddoc_cache/test/eunit/ddoc_cache_no_cache_test.erl
```diff
@@ -35,45 +35,43 @@ return_error(_DDocId) ->
     {error, timeout}.
 
 
-start(Resp) ->
-    Ctx = ddoc_cache_tutil:start_couch(),
+no_cache_test_() ->
```
wohali: Same comment as before, please leave these test changes in a separate commit from the others for posterity (don't squash all together)

wohali: Good change tho

test/elixir/lib/couch.ex
```diff
@@ -179,13 +179,13 @@ defmodule Couch do
       Keyword.get(
         options,
         :timeout,
-        Application.get_env(:httpotion, :default_timeout, 5000)
+        Application.get_env(:httpotion, :default_timeout, 60_000)
```
wohali: Would prefer to see these parameterized higher up in the file somewhere, magic numbers hidden in code always give me the twitches :)

Jenkinsfile
```diff
@@ -14,35 +14,33 @@
...
 ./configure --with-curl
 make check || (build-aux/logfile-uploader.py && false)
+'''
 
-echo
-echo "Build CouchDB packages"
-cd ${builddir}
+make_packages = '''
 git clone https://github.com/apache/couchdb-pkg
+rm -rf couchdb
```
wohali: Are you confident that this will not persist workspaces between runs? Leftover files from previous runs were why I had this in there previously; it caused problems with glob matching since multiple `apache-couchdb-*.tar.gz` files would be present down on L34-35.

Jenkinsfile
```diff
@@ -14,35 +14,33 @@
 // the License.
 
 // DRYing out the Jenkinsfile...
-build_script = '''
-mkdir -p ${COUCHDB_IO_LOG_DIR}
 
-echo
-echo "Build CouchDB from tarball & test"
```
wohali: I found these echos easier to grep for in the logfile than the commands themselves. Is there a similar keyword I can search on now?

Jenkinsfile
```diff
@@ -138,18 +132,25 @@ pipeline {
...
-                cd $builddir
+                mkdir build
+                cd build
                 tar -xf $WORKSPACE/apache-couchdb-*.tar.gz
                 cd apache-couchdb-*
                 ./configure --with-curl
                 gmake check || (build-aux/logfile-uploader.py && false)
 
                 # No package build for FreeBSD at this time
-                rm -rf $builddir $COUCHDB_IO_LOG_DIR
```
wohali: FreeBSD isn't running inside a container, and you didn't add a separate workspace here. Without using a `/tmp` directory, and without cleaning up the builddir, the next build will have files remaining in the folder. Are you 100% sure this isn't an issue on multiple successive builds? We can always restore it if necessary, I guess.

Jenkinsfile
```diff
@@ -167,14 +170,33 @@ pipeline {
           environment {
             platform = 'centos6'
           }
-          steps {
-            sh 'rm -f apache-couchdb-*.tar.gz'
-            unstash 'tarball'
-            sh( script: build_script )
-          } // steps
+          stages {
```
wohali: stages within a stage! :)

Jenkinsfile
```diff
@@ -167,14 +170,33 @@ pipeline {
...
+              post {
+                always {
+                  junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml'
+                }
+              }
+            }
+            stage('Build CouchDB packages') {
+              steps {
+                sh( script: make_packages )
+                sh( script: cleanup_and_save )
```
wohali: Much cleaner, thanks.

Jenkinsfile
```diff
@@ -167,14 +170,33 @@ pipeline {
...
+                  archiveArtifacts artifacts: 'pkgs/**', fingerprint: true
+                }
+              }
+            }
+          } // stages
           post {
-            success {
-              archiveArtifacts artifacts: 'pkgs/**', fingerprint: true
+            cleanup {
+              sh 'rm -rf ${WORKSPACE}/*'
```
wohali: This leaves the actual customWorkspace directory. Are those going to pile up? We can ask Infra to look at our workspaces if necessary. Have you tried an `ls -la ${WORKSPACE}/..` just to see?

src/couch/test/eunit/couch_db_split_tests.erl
kocolosk: Can do.

Jenkinsfile
kocolosk: These are separate build stages now, so the log files are entirely separate (at least in the Jenkins UI, I haven't checked the log file uploader).

Jenkinsfile
kocolosk: Yeah that's a new feature :) I only did it to provide a little extra visibility into the current progress of each parallel build.

Jenkinsfile
kocolosk: My thought was the `deleteDir()` at the beginning should take care of things for us, but I'll admit I didn't specifically check that it was doing what it said.

Jenkinsfile
kocolosk: Ah, you mean the original workspace before the `customWorkspace()` invocation changes it?

The way I observed this working was that the invocation of `customWorkspace()` would change the `${WORKSPACE}` environment variable as well, so this `cleanup` action will remove the contents of e.g. the `centos6` sub-folder of the original workspace. You're right, the original workspace directory will still exist. It won't take up any space. I don't know if that's a problem. I have to believe Jenkins manages workspace reclamations in general, but that's an assumption on my part.

Jenkinsfile
kocolosk: The contents of the workspace do get cleaned up, and I definitely ran into the globbing problem when I wasn't doing that correctly.

Jenkinsfile
kocolosk: 👍

test/elixir/lib/couch.ex
kocolosk: So, there are already multiple layers of configurable settings on top of this. Maybe it'd be better to figure out how to set the `:httpotion` `:default_timeout` in the app environment, or pass the `:timeout` in the `options`? I find that a little confusing for a simple setup like this because if you see the 5000 down here you'll have to go hunting for the config that overrides it.

Or are you saying it'd be better to have one more layer of indirection that leads back up to a constant at the top of the module?

I'll try a couple of things and see If something makes sense (in the next PR).

kocolosk:
Sweet, thanks for the review Joan. I've moved the timeout stuff out of this PR and will plan to merge once Travis gives the green light. LMK if any of my explanations don't pass muster.

wohali:
All good, +1 on this and the future PR.

PR #2092, opened 2019-08-02
[WIP] Deprecate port 5986

### LINKED ISSUE #1523: Retire the node-local interface (port 5986)
This can be done in two phases:

1. Move relevant 5986 port functions under `/_node/<nodename>/...` such as:
    * _dbs
    * _nodes
    * per-shard DB operations (all)
1. Actually eliminate `couch_httpd`.

@davisp thought this might go under the new HTTP/2 API/internals but I think they're distinct efforts.

wohali:
## Overview

This introduces a new `chttpd_node` module, under which all unclustered
functions now live. System DB and shard-level requests bypass mem3
and fabric, using `rpc:call` to call into `couch_db`, `couch_server`,
etc.

The idea is to transplant the `couch_httpd` functionality into
`chttpd` since those functions have `send_json` calls baked into them
and can't directly be `rpc:call`ed.

Endpoints appear under `/_node/$node/` or the special token `/_node/_local/`.

## Endpoint moving status

| *Status & Endpoint* | *Comment* |
|---| ---
 :ballot_box_with_check: `/_node/$node/_all_dbs` | All shards + system DBs on the node
 :ballot_box_with_check: `/_active_tasks`| top-level endpoint already provides all per-node info
 :ballot_box_with_check: `/_node/$node/_config` | node-level config settings
 :ballot_box_with_check: `/_node/$node/_stats` | node-level statistics
 :ballot_box_with_check: `/_node/$node/_system ` | node-level system
--- | --- | ---
:ballot_box_with_check: `GET /_node/$node/{db}` | info blob about system DB or shard
:black_square_button: `GET/POST /_node/$node/{db}/_all_docs` | for system DBs only
:black_square_button: `POST /_node/$node/{db}/_compact` | for system DBs and all shards
:black_square_button: `/_node/$node/{db}/{doc}` | CRUD for all system DB docs, **forbidden** for shards | 
 :black_square_button: `GET /_node/$node/{db}/{ddoc}/_info` | info stats for a view shard

## Testing recommendations

By-hand test results shown here:

  https://gist.github.com/wohali/c2b290845ca45fb3302925c7fbb8c7b9

Closes #1523 

## TODO
- [ ] Finish unimplemented endpoints
- [ ] Write test cases
- [ ] Consider more abstraction of `chttpd_node:handle_node_db_req/2`,
  but probably not to the point of externalising handlers to the app
  env.
- [ ] Nuke `couch_httpd` from orbit. It's the only way to be sure.
- [ ] Document new endpoints properly, scrubbing all 5986 refs from the docs too


wohali:
Status update:

I have some local work on the next part that I've not pushed yet. The idea for the `_all_docs` query is to grab the shard's `Db` record in the same way as we do for `GET /_node/$node/{db}` via the new `couch_node:do_db_req/4`, then pass it to https://github.com/apache/couchdb/blob/master/src/fabric/src/fabric_view_all_docs.erl#L23 which as far as I can tell isn't actually reachable today. (Thanks, dev who wrote this!)

If this makes sense, then my intent is to do the same thing for the other endpoints, adding methods in fabric as necessary that take actual `Db` records rather than `DbName` as a convenient way to re-use fabric to access individual shards rather than relying on couch_* functions. This will make it easier to remove couch_httpd in the last part of the PR.

rnewson:
Hi all, just a note that I'm working on this and hope to push an update in the next day or two.

Where I've gravitated to for 3.0 is the following;

1) All the _node enhancements that @wohali mentions above
2) leave couch_http modules in the codebase
3) Disable the 5986 port by default, allow it to be enabled in config

in 4.0 both the _node endpoint and the 5986 port are completely gone, so I don't see a lot of value in deleting couch_http modules today, and plenty of downside.

davisp:
@rnewson You merged a thing I thought? Does this need to be closed?

janl:
This happened in https://github.com/apache/couchdb/pull/2336

PR #2104, opened 2019-08-08
More timeouts fixes for Jenkins

kocolosk:
## Overview

Extends timeouts in places where we have seen test failures on Jenkins.

## Testing recommendations

This branch will be tested by Jenkins automatically.

## Checklist


kocolosk:
Got a pre-emptive 👍 from @wohali over in #2088. Commit c98f068 has been updated to use module constants instead of magic values per the review over there.

I expect we may still see some timeouts in Jenkins because I don't think #2098 quite covered all the issues there yet.

PR #2114, opened 2019-08-14
Ensure EUnit inherits appropriate environment variables

kocolosk:
## Overview

Omitting COUCHDB_VERSION caused the EUnit build of the replicator to have a corrupted User-Agent header. It tried to construct a version using git, but when building from a release tarball there is no git repo so the UA had a git error message in it. This error message contained a newline, which plausibly confused some part of the HTTP stack and caused replicator HTTP requests to hang.

The second commit removes the attempt to publish packages for Ubuntu 14.04, as we're no longer building them for that (EOL) platform.

## Testing recommendations

See https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/jenkins-cpse-debugging/18/pipeline/, which is building this branch before my cleanup rebase (the final diff from master is identical).

The specific test that was quite regularly failing without this patch is `cpse_purge_http_replication`.

## Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -529,8 +529,6 @@ pipeline {
             reprepro -b couchdb-pkg/repo includedeb jessie pkgs/jessie/*.deb
             cp js/debian-stretch/*.deb pkgs/stretch
             reprepro -b couchdb-pkg/repo includedeb stretch pkgs/stretch/*.deb
-            cp js/ubuntu-trusty/*.deb pkgs/trusty
-            reprepro -b couchdb-pkg/repo includedeb trusty pkgs/trusty/*.deb
             cp js/ubuntu-xenial/*.deb pkgs/xenial
             reprepro -b couchdb-pkg/repo includedeb xenial pkgs/xenial/*.deb
             cp js/ubuntu-bionic/*.deb pkgs/bionic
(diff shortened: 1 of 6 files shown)


kocolosk:
I'm being sloppy again and using this branch for whack-a-mole timeout extensions where needed. I can move those elsewhere if folks want to keep the PR clean.

Jenkinsfile
```diff
@@ -529,8 +529,6 @@ pipeline {
             reprepro -b couchdb-pkg/repo includedeb jessie pkgs/jessie/*.deb
             cp js/debian-stretch/*.deb pkgs/stretch
             reprepro -b couchdb-pkg/repo includedeb stretch pkgs/stretch/*.deb
-            cp js/ubuntu-trusty/*.deb pkgs/trusty
```
wohali: Yeah, we dropped the trusty builds, so this is fine.

Makefile
```diff
@@ -168,11 +168,11 @@ eunit: export BUILDDIR = $(shell pwd)
 eunit: export ERL_AFLAGS = -config $(shell pwd)/rel/files/eunit.config
 eunit: export COUCHDB_QUERY_SERVER_JAVASCRIPT = $(shell pwd)/bin/couchjs $(shell pwd)/share/server/main.js
 eunit: couch
-	@$(REBAR) setup_eunit 2> /dev/null
+	@COUCHDB_VERSION=$(COUCHDB_VERSION) COUCHDB_GIT_SHA=$(COUCHDB_GIT_SHA) $(REBAR) setup_eunit 2> /dev/null
```
wohali: Please make the same edit to `Makefile.win`.

kocolosk: Done. Rebased into a fixup of the original commit.

PR #2131, opened 2019-08-22
add trace test for couch_views

garrensmith:
Add a trace test for couch_views


src/couch_views/test/couch_views_trace_index_test.erl
```diff
@@ -0,0 +1,143 @@
...
+                ?EUNIT_FTW([
+                    fun trace_single_doc/1
+                ])
+            }
+        }
+    }.
+
+
+setup() ->
+    Ctx = test_util:start_couch([
```
davisp: This whole thing can be a one liner.

PR #2173, opened 2019-09-05
Remove old multi-query path

### LINKED ISSUE #2168: Remove support for ?queries parameter on views
From our release notes:

* #820, #1032: Multiple queries can now be made at the
  ``POST /{db}/_all_docs/queries``, ``POST /{db}/_design_docs/queries`` and
  ``POST /{db}/_local_docs/queries`` endpoints. Also, a new endpoint
  ``POST /{db}/_design/{ddoc}/_view/{view}/queries`` has been introduced to replace
  the ``?queries`` parameter formerly provided for making multiple queries to a view.
  The old ``?queries`` parameter *is now deprecated and will be removed in a future
  release of CouchDB.*

kocolosk:
## Overview

Users should send requests with multiple queries to the new endpoint:

/db/_design/{ddoc}/_view/{view}/queries

Closes #2168

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_view.erl
+++ b/src/chttpd/src/chttpd_view.erl
@@ -91,19 +91,19 @@ handle_view_req(#httpd{method='POST',
     Keys = couch_mrview_util:get_view_keys(Props),
     Queries = couch_mrview_util:get_view_queries(Props),
     case {Queries, Keys} of
-        {Queries, undefined} when is_list(Queries) ->
-            [couch_stats:increment_counter([couchdb, httpd, view_reads]) || _I <- Queries],
-            multi_query_view(Req, Db, DDoc, ViewName, Queries);
         {undefined, Keys} when is_list(Keys) ->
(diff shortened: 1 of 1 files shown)


kocolosk:
I left the code in `couch_mrview_http` that does this for 5986 in the expectation that it will be removed separately as part of #1523

src/chttpd/src/chttpd_view.erl
```diff
@@ -91,19 +91,19 @@ handle_view_req(#httpd{method='POST',
     Keys = couch_mrview_util:get_view_keys(Props),
     Queries = couch_mrview_util:get_view_queries(Props),
     case {Queries, Keys} of
```
rnewson: rather than keep the 2-tuple and have it always been an error to specify the first one, I think it would be cleaner to have if clauses for the two errors (Queries not being undefined, Keys not being a list) and then the two lines of actual implementation?

PR #2101, opened 2019-08-07
Refactor fabric:cleanup_index_files

iilyak:
#  Overview

Previous implementation of `fabric:cleanup_index_files` assembled a regexp by concatenating active signatures.
The approach caused regexp to exceed system limit in the case of huge number of them. 
The stacktrace looked like follows: 
```
2> fabric:cleanup_index_files().
** exception error: no match of right hand side value {error,{"regular expression is too large",37555}}
     in function  fabric:cleanup_index_files/1 (src/fabric.erl, line 519)
     in call from fabric:'-cleanup_index_files/0-lc$^0/1-0-'/1 (src/fabric.erl, line 501)
     in call from fabric:'-cleanup_index_files/0-lc$^0/1-0-'/1 (src/fabric.erl, line 501)
```

This implementation uses different approach. It relies on the fact that view files are named as `${signature}.view`. This means that the signature of the file can be easily extracted using `filename:basename(FilePath, ".view")`. 

## Testing recommendations

```
make exunit tests=src/couch/test/exunit/fabric_test.exs
```

## Related Issues or Pull Requests

N/A

## Checklist

- [x] Code is written and works correctly
- [x] Changes are covered by tests
- [ ] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation

PR #2178, opened 2019-09-09
do not cleanup ongoing compact files using fabric:cleanup_index_files/1

jiangphcn:
## Overview

This PR is not to cleanup ongoing compact files using fabric:cleanup_index_files/1

## Testing recommendations

make exunit tests=src/couch/test/exunit/fabric_test.exs

```
17:55:56.045 [debug] Calling 'teardown/3' for 'Elixir.Couch.Test.Setup.Step.Config'
  * test Fabric miscellaneous API Get inactive_index_files (1146.4ms)


Finished in 1.2 seconds
1 test, 0 failures
```

https://github.com/apache/couchdb/pull/2101
## Checklist

### DIFF
--- a/src/fabric/src/fabric.erl
+++ b/src/fabric/src/fabric.erl
@@ -524,8 +524,11 @@ inactive_index_files(DbName) ->
     end, mem3:local_shards(dbname(DbName))),
 
     if ActiveSigs =:= [] -> FileList; true ->
+        %% <sig>.view and <sig>.compact.view where <sig> is in ActiveSigs
+        %% will be excluded from FileList because they are active view
+        %% files and should not be deleted.
         lists:filter(fun(FilePath) ->
-            not maps:is_key(filename:basename(FilePath, ".view"), ActiveSigs)
+            not maps:is_key(get_view_sig_from_filename(FilePath), ActiveSigs)
         end, FileList)
     end.
 
@@ -662,6 +665,8 @@ kl_to_record(KeyList,RecName) ->
 set_namespace(NS, #mrargs{extra = Extra} = Args) ->
     Args#mrargs{extra = [{namespace, NS} | Extra]}.
 
+get_view_sig_from_filename(FilePath) ->
+    filename:basename(filename:basename(FilePath, ".view"), ".compact").
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").


src/fabric/src/fabric.erl
```diff
@@ -525,7 +525,8 @@ inactive_index_files(DbName) ->
 
     if ActiveSigs =:= [] -> FileList; true ->
         lists:filter(fun(FilePath) ->
-            not maps:is_key(filename:basename(FilePath, ".view"), ActiveSigs)
+            not maps:is_key(filename:basename(
+                filename:basename(FilePath, ".view"), ".compact"), ActiveSigs)
```
iilyak: this doesn't seem right. You check for `<sig>.view.compact`. However view compaction files are named `<sig>.compact.view` see [here](https://github.com/apache/couchdb/blob/master/src/couch_mrview/src/couch_mrview_util.erl#L834).

jiangphcn: No. What I check is `<sig>. compact.view`.

For `sig1.view`, the result of `filename:basename(filename:basename(FilePath, ".view"), ".compact")` is sig1.

For `sig2.compact.view`, the result of `filename:basename(filename:basename(FilePath, ".view"), ".compact")` is sig2.

So all view files and ongoing compact files are considered.

iilyak: Indeed:
```
Eshell V9.3.3.6  (abort with ^G)
1> FilePath = "sig2.compact.view".
"sig2.compact.view"
2> filename:basename(filename:basename(FilePath, ".view"), ".compact").
"sig2"
3> f(FilePath), FilePath = "sig2.view".
"sig2.view"
4> filename:basename(filename:basename(FilePath, ".view"), ".compact").
"sig2"
```

iilyak:
The description of the PR is misleading. It should be updated to "remove view compaction files on view cleanup".

iilyak:
Is there a reason you think `couch_mrview_cleanup:run(DbName).` shouldn't be used here?

iilyak:
> The description of the PR is misleading. It should be updated to "remove view compaction files on view cleanup".

I was wrong you description is good.

jiangphcn:
> Is there a reason you think couch_mrview_cleanup:run(DbName). shouldn't be used here?
It is `couch_mrview_cleanup:run(db)` instead of ` couch_mrview_cleanup:run(DbName)`. So it is shard db level instead of db level.

src/fabric/src/fabric.erl
iilyak: I think it would be easier to understand if we would abstract the nested filebase operations into a separate function and add comment there which mention both possible filename formats. The function could be called `get_view_signature_from_filename` or something shorter.

jiangphcn:
Thank you. I also take time to get better understanding on logic :-)

PR #2195, opened 2019-09-19
feat: do not run stats aggregations on an interval

janl:
## Overview

Similar to 448be7996999a706464d8f7429a56dc9e9c87c3a (hello 0.10.1),
`timer:{send,apply}_interval()` will apply functions / send messages
for all intervals that match the time that a machine was in sleep /
hibernation mode that is common on desktop systems.

In a typical office scneario, a laptop system that sleeps over a
weekend , when woken up on a monday, issue thousands of function
calls, that together with other, unrelated wake-up activity, make
a machine top out its CPU for no good reason.

The change addresses this by instead of relying on an interval to
start a given task, on startup, start the task once after a timeout,
and then start a fresh timer after the task is done.

Other than the 0.10-era patch, this one does not account for a system
waking up before the timeout. I’m happy to add that behaviour, if a
reviewer insists on it.

As a result, no matter how long the sleep period is, we only run
the desired function _once_ after we wake up again. In the never-
sleep scenario, the existing behaviour is retained.

This might impact metrics that have a time component, but I think
that’s a fair compromise, so I didn’t investigate that further.

## Testing recommendations

Start CouchDB on a system that supports sleep or hibernation, have CPU instrumentation set up for the CouchDB process, and observe CPU load after prolonged sleep times.

Alternatively, add a couch_log:info() line to the interval’d functions and see log entries for each Interval during the sleep period, but with a wake-up-time era timestamp.

https://issues.apache.org/jira/browse/COUCHDB-204
https://issues.apache.org/jira/browse/COUCHDB-539

## Checklist


janl:
Got +1s on IRC (@kocolosk @chewbranca)

kocolosk:
@janl heya sorry I didn't get a chance to review this yesterday. The idea is a good one, and the PR is an improvement over the current implementation for sure. One small enhancement would be to use `erlang:send_after/3` instead of `timer:send_after/3` per the Erlang Efficiency Guide:

http://erlang.org/doc/efficiency_guide/commoncaveats.html#timer-module

janl:
@kocolosk good find! Since this the reason:

> The timer module uses a separate process to manage the timers. That process can easily become overloaded if many processes create and cancel timers frequently (especially when using the SMP emulator).

And at most this creates ONE timer once every 1000ms, it doesn’t sound like a big deal.

I’m happy to do the switch, though :) https://github.com/apache/couchdb/pull/2207/

PR #2206, opened 2019-09-24
Include search in the list of advertised features

kocolosk:
## Overview

The reason we don't do this in `config:features()` directly is because this one is a dynamic check for the presence of a connected clouseau node. Calling `enable_feature` every time we conduct that check seemed too heavyweight, but I didn't see a good opportunity to just call it once and be confident that it would reliably advertise the feature.

The downside here is that CouchDB will not advertise the "search" feature if Clouseau is disconnected for maintenance or whatever, although technically it's accurate since search requests submitted during that interval would fail.

## Testing recommendations

Wire up CouchDB with Clouseau and see if "search" shows up in the list of features.

#2205 

## Checklist


willholley:
verified with and without search running. +1.

Issue #2115, opened 2019-08-14
Update default config settings (Q, max_document_size, etc.)


kocolosk:
I would like to take the opportunity to revisit a few default configuration settings for 3.0 deployments:

`cluster.q` (currently 8) - we have had mailing list threads about this is too large for many deployments. The introduction of shard splitting gives us more flexibility to add shards to a database after the fact. I'd suggest setting this to 1 or 2 as a default.

`couchdb.max_document_size` - the current default of 4GB is quite insane. The view engine already cannot handle anything much larger than ~10MB without a lot of tweaking. Moreover, 4.0 is going to force our hand on making documents smaller. I'd suggest we lower this default to 1MB.

rnewson:
+1 on lowering max_document_size to 1MB.

default q is more controversial. How well is the shard splitting working? If it's seamless (no interruption to reads or writes) and efficient, then lowering to 2 is ok by me (I'd not be keen on 1). That said, q vanishes entirely in 4.0 and we've lived with this default for a long time. Is it worth changing at this late stage?

wohali:
There was significant discussion on both of these changes in other tickets and on the mailing list.

I'm -1 until we at least review that feedback and consider the implications.

I believe we got closer to q=2 or 4 on the lists than 1 or 2. Or we could be more intelligent and try and autopick on first run or something, based on # of CPU cores and nodes in the cluster, and th absence of a q= value in the local ini files...spitballing here.

As for the max document size, I think this is actually more controversial, and will break anyone who has larger attachments in CouchDB today who upgrades! That seems to violate the principle of least surprise. See #1253 and #1200 for the latest on this.

#attach

/cc @janl

kocolosk:
I erred on the side of fewer shards because we have splitting but not merging, and because the initial setting of 8 was derived from some of our deployments which I'd wager are larger than average. But I'm happy to go along with whatever consensus (including retaining the current default if that's what folks want).

Attachments aren't included in the computation for `couchdb.max_document_size`, only the actual JSON fields and values. That said, there are absolutely users in the wild with JSON docs larger than 1 MB, and this change in defaults could certainly surprise them. In 3.x they'll be able to work around it by changing the default on their server, but this is meant as a signal to re-think the data model. My experience has been that those large docs are problematic for a number of reasons; the somewhat fuzzy limit at which the view engine breaks, and the intense write amplification when updating any field of the document are two examples that come to mind.

wohali:
[...]

PR #2228, opened 2019-10-03
Update default config settings

rnewson:
## Overview

Update default config settings

## Testing recommendations

In a fresh install, create a database, confirm that it is q=2. Try to create a >8mb document and confirm that it fails.

https://github.com/apache/couchdb/issues/2115


wohali:
Still not happy with 2MB as the default document size but I can see that I'm going to be outvoted by the Cloudant contingent on this.

rnewson:
discussed in couchdb-dev on IRC, Wohali is not comfortable dropping all the way to 1MB. Given the upcoming FDB restriction on a txn to 10MB, we've compromised on a new default of 8MB (8 million bytes). This is still a vast improvement over the unattainable 4 GiB default of old.

PR #2286, opened 2019-10-28
Opentracing

iilyak:
## Overview

This PR implements opentracing support. It mostly implements 

## Testing recommendations

0. run test suite 
  ```
  make setup-eunit
  make && ERL_LIBS=`pwd`/src BUILDDIR=`pwd` mix test --trace src/chttpd/test/exunit/ src/ctrace/test/exunit/
  ```
1. start FDB server
2. write connection string into fdb.cluster file
3. update rel/overlay/etc/default.ini as follows
  ```
  index 79555f3c0..50d060c6f 100644
  @@ -564,3 +564,25 @@ min_priority = 2.0
   ;
   ; Jitter applied when checking for new job types.
   ;type_check_max_jitter_msec = 5000
  +
  +[erlfdb]
  +cluster_file = fdb.cluster
  +
  +[tracing]
  +enabled = true
  +thrift_format = compact ; compact | binary
  +agent_host = 127.0.0.1
  +agent_port = 6831
  +; app_name is the value which would be used for
  +; `location.application` tag
  +app_name = couchdb
  +
  +[tracing.samplers]
  +
  +database-info.read = all
  +
  +[tracing.database-info.read]
  +
  +select = (#{'http.method' := Method}) when Method == 'GET' -> [report]
  +; fabric = (#{component := Component}) when Component == <<"couchdb.fabric">> -> [report]
  +details = (#{parent := Parent}) when Parent == <<"database-info.read">> -> [report]
  ```
4. start jaeger docker run -d -p 6831:6831/udp   -p 16686:16686   jaegertracing/all-in-one:1.14
5. start couchdb `dev/run --admin=adm:pass`
6. create database `curl -X PUT -u adm:pass http://127.0.0.1:15984/test`
7. query database `curl -X GET -u adm:pass http://127.0.0.1:15984/test`
8. open http://localhost:16686/ in the browser and search for spans

This PR would be rebased after https://github.com/apache/couchdb/pull/2279

## Checklist


iilyak:
The jaeger produces something like the following

Screen Shot 2019-10-28 at 5 41 13 AM

Screen Shot 2019-10-28 at 4 51 01 AM

Screen Shot 2019-10-28 at 4 51 23 AM

davisp:
Awesome work on this @iilyak! I think this will be a super useful tool to have for a whole bunch of different scenarios.

I do have a few questions on the filtering aspect though. I've been reviewing a bit and also trying to keep in mind how I might expect to use a tool like this. So far my main use cases look something like such:

1. Profiling all requests of a certain type (i.e., show me profiles for all _changes requests)
2. I have a specific request (dbname, docid/view/etc, query string) that is slow, please profile this one specific HTTP request.
3. Show me all HTTP requests that take longer than N seconds (that aren't changes).

For 1, I think that's easy enough. We'd just add a rule for `tracing.database.changes` (or whatever name we give it) that is a `match all` for something like `(#{}) -> [report]`. The only thing there is it seems somewhat complicated to explain why not just having a sampler of `all` isn't enough for that use case.

For 2, I think we could add a specific header as a tag to each HTTP span, except this get a bit tricky when selecting spans which I'll expand on below.

For 3, I'm pretty sure this is impossible with the current library. Given that child spans are decided/filtered before the parent span we'd have to build and store all spans until the very end when we decide whether or not to report them which would get into us having to invent and maintain some tracing system outside the library. However, the alternative would be to just trace all requests with a restrictive sampler and hope we eventually capture a long one. Its a bit wasteful in that we're reporting normal/acceptable traces but given I'd expect this to be in the "lets record traces for 30 seconds" it's not the end of the world.

I also think that the span selection/filtering for reporting isn't quite the right shape.

```
[tracing.view.build]
a_select = (#{'view.name' := Name}) when Name == "blablabla" -> [report]
details = (#{parent := Parent}) when Parent == <<"view.build">> -> [report]
```

Take this example from the docs. Unless I'm mistaken in my reading, won't this end up reporting all child spans of any view build regardless of whether `Name == "blablablah`? As each child span finishes the `details` rule would match and report, even for different values of `Name`. In Jaeger we'd end up just seeing a bunch of disconnected spans in addition to spans related the the `"blablabla"` view build.

Also from my reading it suggests that we'd not see child-of-child spans? I'm pretty sure `parent` isn't transitive, so if we had spans for something like `view.build` -> `doc.read` -> `fdb.get_range`, we'd never see the `fdb.get_range` spans since their parent is `doc.read`. Granted we could also add more rules to match those, but then we have to know every single span name possible.

So the two main issues are that it seems easy to trace unrelated spans while also not tracing all the spans of interest.

When trying to trace I'd roughly expect two main options: trace the entire tree or trace nothing at all. Beyond that I could see adding extensions that could add more granularity about skipping certain spans due to performance or noise requirements but that seems awfully advanced beyond the basic use cases.

The last thing that occurs to me about the rules and filtering is that it seems a bit wasteful that we delay filtering until the end of a span. It doesn't appear that the underlying passage library has this concept, but for instance, in an HTTP response we can generally make a pretty good guess on whether we want to trace or not all the way up in `chttpd:handle_request_int/1`. It seems like having a pre-emptive trace/don't trace decision would prevent us from generating and discarding a whole bunch of spans for unrelated requests.

Hopefully that makes sense. And its more than possible that I've missed a subtlety in there as this is almost exclusively just from reading the code without setting up jaeger and starting to work with a real setup yet.

iilyak:
Thank you for a detailed write up of the issues.

> 3. Show me all HTTP requests that take longer than N seconds (that aren't changes).

We can implement it easily if we pass request duration and operation_id to `match` function of generated module.

```
diff --git a/src/ctrace/src/ctrace_filter.erl b/src/ctrace/src/ctrace_filter.erl
index dc2d1854f..59a155075 100644
--- a/src/ctrace/src/ctrace_filter.erl
+++ b/src/ctrace/src/ctrace_filter.erl
@@ -46,7 +46,12 @@ new(Id, Module) ->
     ) -> ok.

 report(#{'__struct__' := ?MODULE, module := Module} = _State, PSpan) ->
-    Tags = passage_span:get_tags(PSpan),
+    Duration = passage_span:get_finish_time(PSpan) - passage_span:get_start_time(PSpan),
+    Operation = passage_span:get_operation_name(PSpan),
+    Tags = maps:merge(passage_span:get_tags(PSpan), #{
+        duration = Duration,
+        operation_id = Operation
+    }),
     case Module:match(Tags) of
         false ->
```

In this case we can write following rule:
```
select = (#{operation_id := Operation, duration = Duration}) when Operation != '_changes' andalso Duration > 15000 -> [report]
```

However you are right because this would only be possible if operation names are matching the http endpoint. Which could be not the case. For example we might assign `document.read` action for `_changes` endpoint depending on how granular we want it to be.

iilyak:
> Also from my reading it suggests that we'd not see child-of-child spans? I'm pretty sure parent isn't transitive, so if we had spans for something like view.build -> doc.read -> fdb.get_range, we'd never see the fdb.get_range spans since their parent is doc.read. Granted we could also add more rules to match those, but then we have to know every single span name possible.

Parent is probably not the right name here. I couldn't come up with better name though. It is an operation_name of a root span. I've considered:
- ancestor
- parent

Maybe I should name it `origin`. 

Currently we propagate `parent` to all descendants. We [take the current #span{} record]( https://github.com/apache/couchdb/pull/2286/files#diff-89ac3e12a9462a9e6c44129ae34d3a0eR135). We only replace a passage span object.

iilyak:
> I also think that the span selection/filtering for reporting isn't quite the right shape.
> ```
> [tracing.view.build]
> a_select = (#{'view.name' := Name}) when Name == "blablabla" -> [report]
> details = (#{parent := Parent}) when Parent == <<"view.build">> -> [report]
> ```
> Take this example from the docs. Unless I'm mistaken in my reading, won't this end up reporting all child spans of any view build regardless of whether Name == "blablablah? As each child span finishes the details rule would match and report, even for different values of Name. In Jaeger we'd end up just seeing a bunch of disconnected spans in addition to spans related the the "blablabla" view build.

True. I think we can fix it by adding `rule` tag. In this case we would write the following rules:
```
[tracing.view.build]
a_select = (#{'view.name' := Name}) when Name == "blablabla" -> [report]
details = (#{rule := Rule}) when Rule == <<"a_select">> -> [report]
```

iilyak:
> The last thing that occurs to me about the rules and filtering is that it seems a bit wasteful that we delay filtering until the end of a span. It doesn't appear that the underlying passage library has this concept, but for instance, in an HTTP response we can generally make a pretty good guess on whether we want to trace or not all the way up in chttpd:handle_request_int/1. It seems like having a pre-emptive trace/don't trace decision would prevent us from generating and discarding a whole bunch of spans for unrelated requests.

There is a basic support for samplers. However we only can use:
- all
- probalistic(Rate :: float())
- null

I thought about adding one more `by_operation(operation)`. However there is a problem. We don't know operation name when we start span. We start span in the beginning of `handle_request_int` and set operation name in `before_request` (when we have `#httpd{}` record). The sampler logic is called on start_span.

iilyak:
> I thought about adding one more by_operation(operation).

Another idea I had is to use generated matching module instead of  chttpd_handlers:handler_info(Req). This would allow as to have action names configurable. It might look as follows:
```
[httpd.actions]

all_dbs_read = (#{path_parts := [<<"_all_dbs">>], method := 'GET'}) -> {'all-dbs.read', [tag1, tag2]}
bulk = (#{path_parts := [<<"_bulk">>], method := 'GET'}) -> {chttpd_operations, operation, [extra_args]}
```


However it makes things more complex and doesn't solve our problem of knowing operation id when we call start_span. Also we would have to make DSL parser less restrictive to allow direct matching of values in the parameters (since we need to support complex types (in this case list of binaries)).

iilyak:
I wish passage library had a `passage:set_start_time` function. This would allow us to postpone starting span till we have `#httpd{}` record.

Should we fork it? We also could break an abstraction and manipulate the passage_span record. However in this case we might have problems if passage would change internal representation.

iilyak:
Actually I found that there is support to pass start_time into start_span function [here](https://github.com/sile/passage/blob/master/src/passage.erl#L127)

davisp:
> > Also from my reading it suggests that we'd not see child-of-child spans? I'm pretty sure parent isn't transitive, so if we had spans for something like view.build -> doc.read -> fdb.get_range, we'd never see the fdb.get_range spans since their parent is doc.read. Granted we could also add more rules to match those, but then we have to know every single span name possible.
> 
> Parent is probably not the right name here. I couldn't come up with better name though. It is an operation_name of a root span. I've considered:
> 
> * ancestor
> * parent
> 
> Maybe I should name it `origin`.
> 
> Currently we propagate `parent` to all descendants. We [take the current #span{} record](https://github.com/apache/couchdb/pull/2286/files#diff-89ac3e12a9462a9e6c44129ae34d3a0eR135). We only replace a passage span object.

Unless I'm reading this wrong, the `parent` tag isn't transitive:

https://github.com/cloudant/couchdb/blob/0b41779f029e474d344b9e0444048fa464dac7f0/src/ctrace/src/ctrace.erl#L131

In the `view.build` -> `doc.read` -> `fdb.get_range` example, the `fdb.get_range` would have a parent of `doc.read` which would then not be reported.

We could obviously make that transitive though I'd agree that it'd probably need a new name such as `origin` or `root`. Though I'm still thinking that's a bit more complex than if we had a "log all spans" default (once we decided to log a request/view build what have you).

davisp:
> > I thought about adding one more by_operation(operation).
> 
> Another idea I had is to use generated matching module instead of `chttpd_handlers:handler_info(Req)`. This would allow as to have action names configurable. It might look as follows:
> 
> ```
> [httpd.actions]
> 
> all_dbs_read = (#{path_parts := [<<"_all_dbs">>], method := 'GET'}) -> {'all-dbs.read', [tag1, tag2]}
> bulk = (#{path_parts := [<<"_bulk">>], method := 'GET'}) -> {chttpd_operations, operation, [extra_args]}
> ```
> 
> However it makes things more complex and doesn't solve our problem of knowing operation id when we call start_span. Also we would have to make DSL parser less restrictive to allow direct matching of values in the parameters (since we need to support complex types (in this case list of binaries)).

I'd contemplated firing the matching rules when we set tags but I was worried that it'd get super confusing if we ended up with spans where we set tags multiple times.

davisp:
I played around with this a bit on Friday and today to try and get a better feel for how this would work in practice. A couple of take aways so far:

1. Switching to the `passage_pd` (i.e., process dict) model cleans up the PR significantly.
2. Using this as an investigation tool is super duper awesome.
3. With the passage_pd approach I can drop in spans anywhere in the request process without having to worry about the extra plumbing making sure I carry the passage span around which (while I haven't tried directly) makes me think it'd be not so awesome as a debug tool.
4. Locally being able to add spans is a massive boon. If we were doing any of this dynamically in production we'll want tooling that will allow us to insert spans around various bits of code. I'm going to try and see if there's a way to do that by (ab)using Erlang's native tracing perhaps.
5. On macOS Docker isn't always the easiest. However, there's a binary package that has a `jaeger-all-in-one` binary that can be run locally trivially. That might make startup for others a bit easier in the instructions.
6. The filtering/selection/whatever is definitely not quite right. I ended up basically disabling it and just logging everything to be able to finally see some usable data. This obviously won't work in a production setting. It certainly feels like having a binary toggle at the start of an HTTP request for "yes, trace this entire request" would be the logical first step. And beyond that, just a "trace all the things!" setting because looking at all of the requests and then filtering for long requests in the jaeger UI was super duper useful for me. Also after using it for a bit I'm having a hard time where filtering individual spans would be useful. If anything I was basically spending most of my time adding a whole bunch of spans when investigating long requests. But in a local dev sort of way and not something I'd expect merged to master.

All in all, I got a bit carried away looking at some of the long requests during a run of `make elixir`. Here's a fun one I ended up with after investigating why the update docs transaction was taking so long. Patching erlfdb to show the commit time was super neat.

Screen Shot 2019-11-04 at 4 16 38 PM

iilyak:
The reason why passage_pd was not used is to be able to pass root span information and operation_name across process boundaries and potentially across nodes if we decide to port this PR into master. 

Also being able to tell which URL was used for given function call in lover layers of the stack is extremely valuable. Having request context would allow us to extend our logging statements to add additional information about the initial request.

iilyak:
> With the passage_pd approach I can drop in spans anywhere in the request process without having to worry about the extra plumbing making sure I carry the passage span around which (while I haven't tried directly) makes me think it'd be not so awesome as a debug tool.

There is a wrapper interface for `passage_pd` backend. 
```
ctrace:start_span(OperationName),
ctrace:trace(fun(S) -> 
    ctrace:tag(S, #{
        tag1: value1
    })
end),
ctrace:finish_span(),
```
There is one inconvenience, though the root span is started using `passage` backend. 
We need to find a way to pass `child_of` option to `ctrace:start_span/1`.

iilyak:
> The reason why passage_pd was not used is to be able to pass root span information and operation_name across process boundaries and potentially across nodes if we decide to port this PR into master.

Another words I wanted to introduce request_ctx for a long time regardless of tracing. It was a showstopper in my earlier work on distributed tracing tool for Cloudant (which was never released).

davisp:
Yeah, I got confused on how the backend stuff was supposed to work so I just cut out the variability and did only the passage_pd approach.

As to connecting spans across process and node boundaries I'm pretty sure that should be trivial? It should just be a matter of grabbing the trace_id and setting that when we start spans. Grabbing enough context to do child_of or follows_from seems doable. Though looking through passage_pd we might need to add an API since `current_span` appears to drop what gets set in start_span.

As to the request_ctx for tracking the request URL we can always introspect the trace state and/or just rely on open tracing to do its thing, no? If we're connecting everything through the trace_id I don't see a compelling reason to have to do all of that plumbing. Especially if that plumbing cripples our ability to dynamically add spans for debugging.

davisp:
Also another note on something I tried that failed. I tried to be clever and use `erlang:fun_info/2` to set the operation_name for HTTP requests like you had done for fabric2_fdb transactions. It turns out that the top level handlers are sufficiently opaque that you end up with a bunch of `chttpd_db:handle_request/N` operations which isn't super useful. So we'll definitely want to go through and fill out the `chttpd_handlers:handler_info/1` approach like you have.

iilyak:
The latest series of commits introduces following changes in the logic:
- `parent` is renamed into `origin` and made transitive
- the span is checked against matching rules on span_start
- the chained reporters are eliminated (we used to have report -> reporter:operation_id -> reporter:main)

The duration tag idea was replaced with `report_longer_than` action (It needs more work).
I decided against using passage_pd in chttpd (for now). Since it would make impossible to start multiple spans in a single process.

davisp:
Oooh! Good call on reusing begin_ts to delay that span start. That's a lot cleaner than anything i had contemplated doing so that we can have all of the tags and operation name ready to go when we initialize that first span.

I don't think that report_longer_than action is going to be much use. Say we want to report all spans over 5s. But all child spans take less than 5s. When we report that long span all of the child spans that collectively added up to more than 5s will have already been discarded as not meeting the threshold. So we're basically getting a splunk query for HTTP requests that exceeded a threshold. Which seems not useful for investigating *why* the request took more than 5s. Unfortunately given that spans are reported on individually makes me think that there's not a good approach here other than getting our sampling and limiting down so we can record a trace without flooding a loaded production system.

I like the change to matching on span start as that will clean up a lot of the logic for how those rules affect trace collection. However, I think that means the sampler configurations are no longer being used. That should be trivial to extend in the match function though. And one thing to double check on that `function_exported` call in match is whether that triggers a file system scan. Years ago we had a similar check and it crippled nodes on startup because until the module was dynamically compiled Erlang would scoure all of its code paths looking for it.

I don't agree that passage_pd makes anything impossible. Its a helper for the "main" set of related spans in a given process. If there did happen to be a case where you needed to track an unrelated set of spans or some sort of non-nested tree of related spans you can always use the regular passage functions. But not being able to slap in a `ctrace:with_span(foo, fun() -> ... end)` would severely hamper the utility of tracing. I've already managed to find a number of places to optimize fdb transactions by wrapping a few lines here and there with `ctrace:with_span/2`.

iilyak:
There is a simplified version if this work here https://github.com/apache/couchdb/pull/2309. It has only process dictionary based backend.

src/ctrace/README.md
```diff
@@ -0,0 +1,308 @@
...
+ - actions is a list which contains
+   - `report`
+ - conditions
+   - `<[condition]>`
+   - `| <[condition]> <[operator]> <[condition]>`
+ - condition:
+   - `<variable_name> <[operator]> <value>`
+     `| <[guard_function]>(<[variable_name]>)`
+ - `variable_name` - lowercase name without special characters
+ - guarg_function: one of
```
ksnavely: Typo, "guard"

src/ctrace/README.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Open tracing agent configuration
+
+```
+[tracing]
+
+thrift_format = compact ; compact | binary
+agent_host = 127.0.0.1
+agent_port = 6831
+; app_name is the value whicj would be used for
```
ksnavely: "which"

src/ctrace/README.md
```diff
@@ -0,0 +1,308 @@
...
+; `location.application` tag
+app_name = couchdb
+```
+
+
+Bellow is an example of a configuration:
```
ksnavely: "Below"

iilyak: Thank you Kyle I fixed the typos in the replacement PR https://github.com/apache/couchdb/pull/2309

iilyak:
closing in favor of https://github.com/apache/couchdb/pull/2309

PR #2302, opened 2019-11-05
Temporarily disable FreeBSD builds

kocolosk:
## Overview

Disable FreeBSD in Jenkins until we figure out the libcurl error in the test suite.

## Testing recommendations

If inclined, push this branch to a `jenkins-` prefixed branch and see if it builds successfully there.

## Related Issues or Pull Requests

See #2301

## Checklist

- [x] Code is written and works correctly
- [ ] Changes are covered by tests
- [ ] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation

Issue #2244, opened 2019-10-07
Couchdb sends more than requested range in attachments


sploders101:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

When performing a request for a partial attachment, Couchdb starts at the correct file location, but sends more data than the Content-Length header suggests, causing an error in some http clients.

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your actual CouchDB! )

## Steps to Reproduce

Upload a large text file as an attachment to a couchdb instance
Run:
```bash
nc 127.0.0.1 5984
```
and send this data, substituting your information.
```
GET /db/doc/attachment.ext HTTP/1.1
Range: bytes=1-4
Host: 127.0.0.1:5984
Connection: close


```

[NOTE]: # ( Include commands to reproduce, if possible. curl is preferred. )

## Expected Behaviour

CouchDB should send *only* the requested bytes.

[NOTE]: # ( Tell us what you expected to happen. )

## Your Environment

Fresh install

[TIP]:  # ( Include as many relevant details about your environment as possible. )
[TIP]:  # ( You can paste the output of curl http://YOUR-COUCHDB:5984/ here. )

* CouchDB Version used: 2.3.1
* Browser name and version: N/A
* Operating System and version: Linux (docker image couchdb:latest)

## Additional context

The specific type of attachment I'm using is an asar archive. I am attempting to read files from it without downloading the entire thing, and storing it to the fs/ram. It has a Content-Type of `application/octet-stream` in the database.

I am using node.js http.request function as my http client, but this is reproducible with any client that doesn't close the connection after `Content-Length` bytes.

[TIP]:  # ( Add any other context about the prbolem here. )

janl:
it doesn’t look like I can reproduce this:

```
> curl -vX PUT http://a:a@127.0.0.1:15984/a
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 15984 (#0)
* Server auth using Basic with user 'a'
> PUT /a HTTP/1.1
> Host: 127.0.0.1:15984
> Authorization: Basic YTph
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 201 Created
< Cache-Control: must-revalidate
< Content-Length: 12
< Content-Type: application/json
< Date: Fri, 25 Oct 2019 09:05:44 GMT
[...]

PR #2305, opened 2019-11-11
enable couchjs to accept HTTP 0.9 responses

dch:
Since curl 7.66.0, HTTP/0.9 is disabled by default [1]. This is present in
FreeBSD's latest ports tree as of 12.1-RELEASE, and breaks a number of
our couchjs-related tests [2], on several platforms, as reported on IRC.

As this libcurl update will propagate through the open source space-time
continuum, pro-actively un-break future couchjs libcurl.

[1]: https://github.com/curl/curl/pull/4191
[2]: https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/master/608/pipeline/#step-163-log-2062

## testing recommendations

- @janl does this resolve the build on OSX for you?
- @kocolosk to re-enable the FreeBSD ci test to see if this helps

### DIFF
--- a/src/couch/priv/couch_js/http.c
+++ b/src/couch/priv/couch_js/http.c
@@ -450,6 +450,9 @@ go(JSContext* cx, JSObject* obj, HTTPData* http, char* body, size_t bodylen)
         curl_easy_setopt(HTTP_HANDLE, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
         curl_easy_setopt(HTTP_HANDLE, CURLOPT_ERRORBUFFER, ERRBUF);
         curl_easy_setopt(HTTP_HANDLE, CURLOPT_COOKIEFILE, "");
+#if LIBCURL_VERSION_NUM >= 0x074000
+        curl_easy_setopt(HTTP_HANDLE, CURLOPT_HTTP09_ALLOWED, 1L);
+#endif
         curl_easy_setopt(HTTP_HANDLE, CURLOPT_USERAGENT,
                                             "CouchHTTP Client - Relax");
     }


dch:
In https://github.com/curl/curl, we see this option only exists in `curl > 7.64.0` 
from roughly a year ago.

How can we handle this in our build tests?

Alternatively, my understanding is that this 0.9 downgrade warning issue
occurs when there are *no* HTTP headers sent at all -- perhaps we could
more easily set an appropriate `X-Couch-Seriously: NopeNopeNope`
header than do terrible things in the build again?

```
commit 006ff62d8c51f664c167c6337f009f9f65dd8ea7
Author: Daniel Stenberg <daniel@haxx.se>
Date:   Mon Dec 17 15:46:56 2018 +0100

    http: added options for allowing HTTP/0.9 responses
    
    Added CURLOPT_HTTP09_ALLOWED and --http0.9 for this purpose.
    
    For now, both the tool and library allow HTTP/0.9 by default.
    docs/DEPRECATE.md lays out the plan for when to reverse that default: 6
    months after the 7.64.0 release. The options are added already now so
    that applications/scripts can start using them already now.
    
    Fixes #2873
    Closes #3383
```

kocolosk:
Good stuff @dch. For what it's worth my PR #2302 is still open so nothing needs to be reverted at this point.

I'm also happy to investigate the "send a dummy header" option; it seems somewhat cleaner than playing games with windows of libcurl version numbers.

kocolosk:
Pushed this branch to `jenkins-libcurl-http09` to get Jenkins to try it out on FreeBSD.

kocolosk:
Nevermind I see we already have a `jenkins-curl-09-test` branch ...

kocolosk:
The failure at https://builds.apache.org/blue/organizations/jenkins/CouchDB/detail/jenkins-curl-09-test/4/pipeline/ suggests this patch didn't do the trick, although there's unfortunately almost no detail in the logs beyond the fact that the same test failed.

kocolosk:
Turns out we have an actual bug in our range request support over in #2244 that is causing libcurl to think it's seeing HTTP/0.9 traffic. I suggest we close this issue in favor of resolving that one.

PR #2309, opened 2019-11-12
Opentracing pd

iilyak:
## Overview

This PR implements opentracing support. It mostly implements [RFC-011](https://github.com/apache/couchdb-documentation/blob/master/rfcs/011-opentracing.md).

## Testing recommendations

0. Run test suite 

```bash
$ make eunit apps=ctrace
```

1. Write connection string into fdb.cluster file
2. Start FDB server
3. Update rel/overlay/etc/default.ini similar to this:

```ini
[tracing]
enabled = true
thrift_format = compact ; compact | binary
agent_host = 127.0.0.1
agent_port = 6831
app_name = couchdb

[tracing.filters]
; Log all traces
all = (#{}) -> true
; Disable tracing POST requests to `/dbname`
database-info.read = (#{'http.method' := Method}) when Method == 'POST' -> false
; Log 50% of all view builds
view.build = (#{}) -> 0.5
```
4. Start jaeger `docker run -d -p 6831:6831/udp   -p 16686:16686   jaegertracing/all-in-one:1.14`
5. Start couchdb `dev/run --admin=adm:pass`
6. Create database `curl -X PUT -u adm:pass http://127.0.0.1:15984/test`
7. Query database `curl -X GET -u adm:pass http://127.0.0.1:15984/test`
8. Open http://localhost:16686/ in the browser and search for spans

As an alternative to Step 4, you can run the `jaeger-all-in-one` binary without arguments in a separate shell. The Jaeger project provides prebuilt binaries for download.

To generate a bunch of spans for investigation you can also run `make elixir`.

This PR is a replacement PR for https://github.com/apache/couchdb/pull/2286

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .venv
 .DS_Store
 .rebar/
+.rebar3/
 .erlfdb/
 .eunit/
 log
@@ -54,16 +55,20 @@ src/hyper/
 src/ibrowse/
 src/ioq/
 src/hqueue/
+src/jaeger_passage/
 src/jiffy/
 src/ken/
 src/khash/
+src/local/
 src/meck/
 src/mochiweb/
 src/oauth/
+src/passage/
 src/proper/
 src/rebar/
 src/smoosh/
 src/snappy/
+src/thrift_protocol/
 src/triq/
 tmp/
 
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -259,6 +259,7 @@ handle_request_int(MochiReq) ->
 
     case after_request(HttpReq2, HttpResp) of
         #httpd_resp{status = ok, response = Resp} ->
+            span_ok(HttpResp),
             {ok, Resp};
         #httpd_resp{status = aborted, reason = Reason} ->
             couch_log:error("Response abnormally terminated: ~p", [Reason]),
@@ -266,6 +267,7 @@ handle_request_int(MochiReq) ->
     end.
 
 before_request(HttpReq) ->
+    ctrace:is_enabled() andalso start_span(HttpReq),
     try
         chttpd_stats:init(),
         chttpd_plugin:before_request(HttpReq)
@@ -316,6 +318,8 @@ process_request(#httpd{mochi_req = MochiReq} = HttpReq) ->
     end.
 
 handle_req_after_auth(HandlerKey, HttpReq) ->
+    #httpd{user_ctx = #user_ctx{name = User}} = HttpReq,
+    ctrace:tag(#{user => User}),
     try
         HandlerFun = chttpd_handlers:url_handler(HandlerKey,
             fun chttpd_db:handle_request/1),
@@ -1050,16 +1054,20 @@ send_error(#httpd{} = Req, Code, ErrorStr, ReasonStr) ->
     send_error(Req, Code, [], ErrorStr, ReasonStr, []).
 
 send_error(Req, Code, Headers, ErrorStr, ReasonStr, []) ->
-    send_json(Req, Code, Headers,
+    Return = send_json(Req, Code, Headers,
         {[{<<"error">>,  ErrorStr},
-        {<<"reason">>, ReasonStr}]});
+        {<<"reason">>, ReasonStr}]}),
+    span_error(Code, ErrorStr, ReasonStr, []),
+    Return;
 send_error(Req, Code, Headers, ErrorStr, ReasonStr, Stack) ->
     log_error_with_stack_trace({ErrorStr, ReasonStr, Stack}),
-    send_json(Req, Code, [stack_trace_id(Stack) | Headers],
+    Return = send_json(Req, Code, [stack_trace_id(Stack) | Headers],
         {[{<<"error">>,  ErrorStr},
         {<<"reason">>, ReasonStr} |
         case Stack of [] -> []; _ -> [{<<"ref">>, stack_hash(Stack)}] end
-    ]}).
+    ]}),
+    span_error(Code, ErrorStr, ReasonStr, Stack),
+    Return.
 
 update_timeout_stats(<<"timeout">>, #httpd{requested_path_parts = PathParts}) ->
     update_timeout_stats(PathParts);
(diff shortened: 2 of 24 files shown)


src/ctrace/src/ctrace_action.erl
```diff
@@ -0,0 +1,60 @@
...
+        [ReporterId :: passage:tracer_id()]
+    ) -> action_fun().
+
+report([_TracerId]) ->
+    fun(_Span, Options) ->
+        passage_pd:finish_span(Options),
+        true
+    end.
+
+report_longer_than([_TracerId, Threshold]) ->
```
iilyak: I'll remove this. Since as @davisp pointed out [here](https://github.com/apache/couchdb/pull/2286#issuecomment-551164613) it is not really useful.

iilyak: removed in https://github.com/apache/couchdb/pull/2309/commits/50e7a8172719a4c0002a2e9afa73b5926bfe3c46

src/fabric/src/fabric2_fdb.erl
```diff
@@ -1364,3 +1371,21 @@ run_on_commit_fun(Tx) ->
             Fun(),
             ok
     end.
+
+start_span_fun(Fun, Options, ExtraTags) ->
+    {name, OperationName} = erlang:fun_info(Fun, name),
```
davisp: I added a handy `fun_to_op` function in ctrace which makes it super easy to put spans in throughout the code when debugging/optimizing:

https://github.com/apache/couchdb/blob/03f3054e5987bc0c1927bc417c41f3a518391f6b/src/ctrace/src/ctrace.erl#L176-L181

We might also want to make it optional so that `ctrace:with_span` accepts a `fun` and auto creates the name as that saves on the fairly repetitious `OpName = ctrace:fun_to_op/1` .

src/fabric/src/fabric2_fdb.erl
```diff
@@ -1364,3 +1371,21 @@ run_on_commit_fun(Tx) ->
...
+        component => <<"couchdb.fabric">>,
+        'db.instance' => fabric2_server:fdb_cluster(),
+        'db.namespace' => fabric2_server:fdb_directory(),
+        'db.type' => <<"fdb">>,
+        user => UserCtx#user_ctx.name,
+        nonce => get(nonce),
+        pid => self()
+    }, ExtraTags),
+    fun() ->
+        ctrace:has_span()
```
davisp: If we wrap `with_span` here we can get rid of passing these closures around. Regardless, it'll be slightly more performant to move all of the OpName/UserCtx/Tags creation inside the check for `ctrace:has_span()` so that when tracing is turned off we're not needlessly creating these extra terms on the heap.

src/fabric/src/fabric2_server.erl
```diff
@@ -128,3 +133,17 @@ fdb_directory() ->
         Dir ->
             Dir
     end.
+
+fdb_cluster() ->
```
davisp: Super minor nit, but we can dry out the fdb_directory/fdb_cluster if we just pass the `?FDB_DIRECTORY` or `?FDB_CLUSTER` to a common function.

src/couch/src/couch_util.erl
```diff
@@ -46,15 +46,16 @@
...
+    <<"^daemons$">>,
+    <<"^external$">>,
+    <<"^httpd_design_handlers$">>,
+    <<"^httpd_db_handlers$">>,
+    <<"^httpd_global_handlers$">>,
+    <<"^native_query_servers$">>,
+    <<"^os_daemons$">>,
+    <<"^query_servers$">>,
+    <<"^feature_flags$">>,
+    <<"^tracing\..*$">>
```
davisp: Adding the `\.` to this regexp means that we're allowing HTTP access to the main `tracing` section which seems like something even more important to protect than the samplers and action bits (granted action bits are dynamically compiled so also important for those).

src/fabric/src/fabric2_fdb.erl
```diff
@@ -72,24 +72,29 @@
 
 
 transactional(Fun) ->
-    do_transaction(Fun, undefined).
-
+    BeforeTx = start_span_fun(Fun, [], #{}),
```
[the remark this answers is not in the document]
iilyak: I tried using with_span it doesn't work as needed. Because we want to trace do_transaction and we don't have database name in there. We want to trace do_transaction because we want to see retries in the trace.

src/couch/src/couch_util.erl
iilyak: Good point about adding `tracing` into `BLACKLIST_CONFIG_SECTIONS` although I don't quite understand a problem with regex itself. The current regexp should block sections named `<tracing> + <dot> + <anything else>`.
```
2> re:run(<<"tracing.all-docs">>, <<"^tracing\..*$">>).
{match,[{0,16}]}
```

src/ctrace/src/ctrace.erl
```diff
@@ -0,0 +1,425 @@
...
+       Result :: term().
+
+with_span(OperationId, ExtraTags, Fun) when is_map(ExtraTags) ->
+    with_span(OperationId, [{tags, ExtraTags}], Fun);
+
+with_span(OperationId, Options, Fun)  ->
+    try
+        start_span(OperationId, Options),
+        Fun()
+    after
```
davisp: Ops, just noticed that we should log an error on exception like `passage_pd:with_span` does by default:

https://github.com/apache/couchdb/compare/prototype/fdb-layer...opentracing-davisp#diff-89ac3e12a9462a9e6c44129ae34d3a0eR122-R128

src/couch/src/couch_util.erl
davisp: Sure. I was just saying removing the `\.` would also match the `[tracing]` main section as well. Having it as two matches is fine as well.

src/chttpd/src/chttpd.erl
```diff
@@ -1228,6 +1236,110 @@ maybe_trace_fdb("true") ->
...
+        couch_log:error("Cannot set tracing action ~p:~p", [Tag, Error]),
+        {undefind, #{}}
+    end.
+
+finish_span(Fun) ->
+    Fun(),
+    ctrace:finish_span().
+
+span_ok(#httpd_resp{code = Code}) ->
+    fun() ->
```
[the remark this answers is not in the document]
iilyak: We do this optimization in [`ctrace:finish_span`](https://github.com/cloudant/couchdb/blob/opentracing-pd/src/ctrace/src/ctrace.erl#L159)
```
finish_span(Options0) ->
    case get(?ACTIONS_KEY) of
        undefined ->
            ok;
        Actions ->
            Options = add_time(Options0),
            lists:takewhile(fun(Action) ->
                Action(passage_pd:current_span(), Options)
            end, Actions)
    end.
```

The `?ACTIONS_KEY` is on the process dictionary only if there was a successful match on [`ctrace:start_span`](https://github.com/cloudant/couchdb/blob/opentracing-pd/src/ctrace/src/ctrace.erl#L139).

Are you proposing to pass list of extra tags to finish_span instead of closure? This would work. The one issue with this would be how to emit ctrace:log if we would want to. Since we don't use the logs at the moment I can make it possible to pass extra tags.

src/fabric/src/fabric2_fdb.erl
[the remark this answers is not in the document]
davisp: I don't think I agree with the premise that we specifically want to trace `do_transaction`. Using `with_span` further up in `fabric2_fdb:transactional` gives us an easy place to track the database name and covers the scope of the entire transaction. I do understand wanting to also trace transaction retries, however that's not actually happening because `BeginTx` and `AfterTx` are outside of the closure passed to `erlfdb:transactional/2`.

To track retries I'd add another span around [1] when we actually invoke the closure. And if we update `ctrace:with_span` to capture exceptions like `passage_pd` does we have the extra benefit of seeing the `reopen` exceptions as well as capturing any FoundationDB error codes in the second span.

[1] https://github.com/apache/couchdb/blob/03f3054e5987bc0c1927bc417c41f3a518391f6b/src/fabric/src/fabric2_fdb.erl#L1376

src/couch/src/couch_util.erl
iilyak: I wanted to allow modification of `tracing.samplers` since it would make it possible to enable/disable pre-configured tracing rules via http. I am going to add a second regex to block `<<"^tracing$">>`. Let me know if you prefer to block `tracing.samplers`.

src/chttpd/src/chttpd.erl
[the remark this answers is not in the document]
davisp: I was just saying that the finish_span isn't executing any logic when invoking the closure. So the whole `finish_span(span_ok(...))` is basically `(fun() -> stuff)()` which begs the question why not just execute `stuff` without wrapping in a closure.

Alternatively, if `ctrace:finish_span` accepted a closure I could see it in that case so that the work in the closure doesn't happen until after you check `?ACTIONS_KEY`.

iilyak:
@davisp I think I fixed all issues

.gitignore
```diff
@@ -54,16 +55,19 @@ src/hyper/
 src/ibrowse/
 src/ioq/
 src/hqueue/
+src/jaeger_passage/
```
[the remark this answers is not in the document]
davisp: Still not seeing an ignore for passage.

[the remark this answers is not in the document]
iilyak: 1385220afffe62e9b79fdf8ad3906db4d9235e21

PR #2324, opened 2019-11-22
Remove vestiges of view-based `_changes` feed from codebase

### LINKED ISSUE #2167: Remove vestiges of view-based `_changes` feed
Bits and pieces of this are still hiding in the codebase, and #592 and #831 are examples of where it causes confusion. We accepted a proposal to remove it in [19].

[19]:
https://lists.apache.org/thread.html/516793df0c1913c045441d0ff78339f307e2aff517d9223da44edd9e@%3Cdev.couchdb.apache.org%3E

eiri:
## Overview

This PR removes all bits and pieces of view_changes from mrview, couch_changes and fabric applications, removes `/_view_changes` end-point and removes seq, kseq and log btrees from mrview file structure. Migration code for 1.2 views was modified to support transparent migration from 2.x views.

## Testing recommendations

All the tests should pass. Sinice this is removing functionality tests for which been disabled anyway, it's hard to validate it otherwise.

Additional manual testing could be done with creation of views on CouchDB 2.x with `seq_indexed` and / or `keyseq_indexed` options set and then accessing them on this CouchDB version. Here is [a setup I've used for testing](https://gist.github.com/eiri/1e5ba550f5dbd271b0b7d03d65e66414), it could help to get up to speed with that. 

Closes #2167 

## Checklist


src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -1168,10 +965,10 @@ maybe_update_index_file(State) ->
     end.
 
 update_index_file(State) ->
```
kocolosk: Suggest upgrading the logging here to `notice` level and including more details about the index file that we're attempting to update.

eiri:
@kocolosk @davisp Thank you for the reviews! I've added details to upgrade logs and fixed a small upgrade bug, added a commit right after upgrade to make sure we have new sig in header on disk. Will to merge once travis allows and then add a note to upgrade docs on views 1.x incompatibility.

PR #2336, opened 2019-11-26
1523 bye bye 5986 

rnewson:
## Overview

This PR enhances the _node endpoint in two ways;

1) The _system, _stats, and _restart endpoints on port 5984 which only affect the local node are now available under _node.
2) The full internal interface reachable on port 5986 is available under _node. This is achieved by sending the http response of the remote node over erlang rpc.

## Testing recommendations

Try;

```
curl 'foo:bar@localhost:15984/_node/node2@127.0.0.1/_stats'
curl 'foo:bar@localhost:15984/_node/node2@127.0.0.1/_system'
curl 'foo:bar@localhost:15984/_node/node2@127.0.0.1/_restart'
curl 'foo:bar@localhost:15984/_node/node2@127.0.0.1/_all_dbs'
curl 'foo:bar@localhost:15984/_node/node2@127.0.0.1/_uuids'
curl 'foo:bar@localhost:15984/_node/node2@127.0.0.1/_config'
curl 'foo:bar@localhost:15984/_node/node2@127.0.0.1/_nodes/_changes?include_docs=true
```

https://github.com/apache/couchdb/issues/1523

## Checklist


src/chttpd/src/chttpd_node.erl
```diff
@@ -0,0 +1,272 @@
...
+-compile(tuple_calls).
+
+-export([
+    handle_node_req/1,
+    get_stats/0
+]).
+
+-include_lib("couch/include/couch_db.hrl").
+
+-import(chttpd,
```
davisp: Imports are the devil.

rnewson: they are but it's a house style for chttpd modules given how often we use those functions. 

_config is justified as it has logic around white/blacklisting not present in the couch_httpd one, and _stats has simply moved from chttpd_misc

davisp: Legacy style mebbe. I don't care enough to ask for a change though.

Ah, good call on the blacklisting. I figured there was enough of a difference that just deleting the specializations wasn't enough.

willholley:
Thinking ahead to the 3.x release, Is there any documentation update needed to go with this? 

It looks as though there is a new option under `httpd.enabled` - does that mean it's now possible to disable HTTP and enable HTTPS only?

Are we still maintaining the distinct `[chttpd]` and `[httpd]` config sections? Perhaps there's an opportunity to sanitize those for 3.x as the split seems a bit redundant if 5986 is gone.

wohali:
@rnewson @willholley we need to remove `couchup` and the documentation for it per #2324 and this PR, unless `couchup` can be rewritten to work with the new `_node` interface soon enough - which it might be. But if we're going to say 1.x -> 3.x is forbidden, then we should just say that and drop `couchup` entirely.

PR #2338, opened 2019-12-02
Allow all params to be passed via body for POST view

bessbd:
## Overview
This change should allow users to supply all params in [POST](http://docs.couchdb.org/en/stable/api/ddoc/views.html#post--db-_design-ddoc-_view-view) that can be supplied for [GET](http://docs.couchdb.org/en/stable/api/ddoc/views.html#get--db-_design-ddoc-_view-view) now. This way we could avoid the `?key="foo"` things that would probably cause a lot of pain for users.

## Testing recommendations
So far this has been tested manually and it seems to be working.
```
curl -H 'Content-Type: application/json'  -XPOST -d '{"key": "yo"}' http://127.0.0.1:15984/asd/_design/newDD/_view/newview |jq
{
  "total_rows": 2,
  "offset": 0,
  "rows": [
    {
      "id": "yo",
      "key": "yo",
      "value": {
        "fromView": "yes",
        "_id": "yo",
        "_rev": "1-ffdc431dc4d2660c2af2e85e3abf723e",
        "asd": "dsa"
      }
    }
  ]
}
```

## Checklist

- [x] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation: https://github.com/apache/couchdb-documentation/pull/459

### DIFF
--- a/src/chttpd/src/chttpd_view.erl
+++ b/src/chttpd/src/chttpd_view.erl
@@ -39,9 +39,15 @@ multi_query_view(Req, Db, DDoc, ViewName, Queries) ->
     {ok, Resp1} = chttpd:send_delayed_chunk(VAcc2#vacc.resp, "\r\n]}"),
     chttpd:end_delayed_json_response(Resp1).
 
+design_doc_post_view(Req, Props, Db, DDoc, ViewName, Keys) ->
+    Args = couch_mrview_http:parse_body_and_query(Req, Props, Keys),
+    fabric_query_view(Db, Req, DDoc, ViewName, Args).
 
 design_doc_view(Req, Db, DDoc, ViewName, Keys) ->
     Args = couch_mrview_http:parse_params(Req, Keys),
+    fabric_query_view(Db, Req, DDoc, ViewName, Args).
+
+fabric_query_view(Db, Req, DDoc, ViewName, Args) ->
     Max = chttpd:chunked_response_buffer_size(),
     VAcc = #vacc{db=Db, req=Req, threshold=Max},
     Options = [{user_ctx, Req#httpd.user_ctx}],
@@ -89,16 +95,9 @@ handle_view_req(#httpd{method='POST',
     chttpd:validate_ctype(Req, "application/json"),
     Props = couch_httpd:json_body_obj(Req),
     assert_no_queries_param(couch_mrview_util:get_view_queries(Props)),
-    case couch_mrview_util:get_view_keys(Props) of
-        Keys when is_list(Keys) ->
-            couch_stats:increment_counter([couchdb, httpd, view_reads]),
-            design_doc_view(Req, Db, DDoc, ViewName, Keys);
-        _ ->
-            throw({
-                bad_request,
-                "POST body must contain an array called `keys`"
-            })
-    end;
+    Keys = couch_mrview_util:get_view_keys(Props),
+    couch_stats:increment_counter([couchdb, httpd, view_reads]),
+    design_doc_post_view(Req, Props, Db, DDoc, ViewName, Keys);
 
 handle_view_req(Req, _Db, _DDoc) ->
     chttpd:send_method_not_allowed(Req, "GET,POST,HEAD").
--- a/src/couch_mrview/src/couch_mrview_http.erl
+++ b/src/couch_mrview/src/couch_mrview_http.erl
@@ -29,6 +29,7 @@
     parse_int/1,
     parse_pos_int/1,
     prepend_val/1,
+    parse_body_and_query/3,
     parse_params/2,
     parse_params/3,
     parse_params/4,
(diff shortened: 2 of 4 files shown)


bessbd:
This change would break backwards compatibility for a cleaner API (no query params from now on for POST). Do you think that would be acceptable? If no, can you suggest me a way to have it backwards compatible?

This was my attempt at backwards compatibility (that did not work and does not look very elegant):
```
mrargs_to_list(Rec) ->
        Fields = record_info(fields, mrargs),
        [_Tag| Values] = tuple_to_list(Rec),
        lists:zip(Fields, Values).

list_to_mrargs(Inlist) ->
    Fields = record_info(fields, mrargs),
    [Tag| Values] = tuple_to_list(#mrargs{}),
    Defaults = lists:zip(Fields, Values),
    L = lists:map(fun ({K,V}) -> proplists:get_value(K, Inlist, V) end, Defaults),
    list_to_tuple([Tag|L]).


design_doc_post_view(Req, Props, Db, DDoc, ViewName, Keys) ->
    QueryArgs = couch_mrview_http:parse_params(Req, Keys),
    BodyList = couch_mrview_http:parse_body(Props, Keys),
%%    Args = list_to_mrargs(
%%        lists:keymerge(1,
%%            lists:sort(mrargs_to_list(QueryArgs)),
%%            lists:sort(mrargs_to_list(BodyList))
%%        )
%%    ),
```

bessbd:
Once the production part is in a somewhat acceptable state, I’ll proceed with writing tests and the doc changes.

davisp:
@bessbd Thanks for the PR! The general idea here is fine but its not going to be acceptable to drop the query string parameters. Reading through this I think the best approach would be to add an option in `couch_mrview_http:parse_params/4` that disables reseting the group/group_level properties in the provided `#mrargs{}` record and then you'd just call that function twice. First for the body then the query string parameters (so that query string parameters override the body to maintain the current behavior).

Let me know if you need any pointers on that approach or find any issues with it.

bessbd:
> @bessbd Thanks for the PR! The general idea here is fine but its not going to be acceptable to drop the query string parameters. Reading through this I think the best approach would be to add an option in `couch_mrview_http:parse_params/4` that disables reseting the group/group_level properties in the provided `#mrargs{}` record and then you'd just call that function twice. First for the body then the query string parameters (so that query string parameters override the body to maintain the current behavior).
> 
> Let me know if you need any pointers on that approach or find any issues with it.

Thank you for the review and the suggestion, @davisp ! I think I'll have a commit pushed for that in about 24 hours.

src/chttpd/src/chttpd_view.erl
```diff
@@ -89,16 +95,9 @@ handle_view_req(#httpd{method='POST',
...
-        Keys when is_list(Keys) ->
-            couch_stats:increment_counter([couchdb, httpd, view_reads]),
-            design_doc_view(Req, Db, DDoc, ViewName, Keys);
-        _ ->
-            throw({
-                bad_request,
-                "POST body must contain an array called `keys`"
-            })
-    end;
+    Keys = couch_mrview_util:get_view_keys(Props),
```
davisp: We should probably remove the debug message from get_view_keys [1] now that we're saying a missing keys field is ok.

[1] https://github.com/apache/couchdb/blob/1fdfe466223380d1ee7c82e21283e161290aed87/src/couch_mrview/src/couch_mrview_util.erl#L1371

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -473,12 +474,21 @@ parse_params(Props, Keys, Args) ->
 
 parse_params(Props, Keys, #mrargs{}=Args0, Options) ->
     IsDecoded = lists:member(decoded, Options),
-    % group_level set to undefined to detect if explicitly set by user
-    Args1 = Args0#mrargs{keys=Keys, group=undefined, group_level=undefined},
+    Args1 = case lists:member(keep_group_level, Options) of
+                true ->
```
davisp: Style nit. Your case should be indented like such:

```erlang
Args1 = case lists:member(...) of
    true ->
        Args0;
    _ ->
        % group_level set to undefined ...
        Args0#mrargs{...}
end,
```

bessbd:
Thank you for the review, @davisp ! I'll push the commits for those changes you've requested soon.

bessbd:
Doc PR: https://github.com/apache/couchdb-documentation/pull/459

src/chttpd/src/chttpd_view.erl
bessbd: Fixed in bcbb0ab

src/couch_mrview/src/couch_mrview_http.erl
bessbd: Fixed in cf89a51

bessbd:
@davisp : I think I've addressed all review comments on this change. Can you please do a next / final pass?

davisp:
I restarted the failed worker on Travis. If that comes back green let me know and I'll merge this for you.

bessbd:
Thank you for the approval and the retrigger, @davisp ! It seems the build failed again with an unrelated issue. What do you think I should do so that this change could be merged?

bessbd:
Thank you for the merge, @garrensmith !

PR #2340, opened 2019-12-04
Normalize fabric2 test suite

davisp:
## Overview

Just a bit of cleanup on the fabric tests. I noticed the awesome pattern that @iilyak added to `fabric2_doc_fold_tests` that ensure that all test names are printed and have gone and applied it uniformly across the entire fabric suite.

## Testing recommendations

`make eunit apps=fabric`

Narp

## Checklist


src/fabric/test/fabric2_doc_crud_tests.erl
```diff
@@ -19,55 +19,58 @@
 -include("fabric2.hrl").
 
 
+-define(TDEF(Name), {atom_to_list(Name), fun Name/1}).
```
garrensmith: Could we move this into the couch_eunit.hrl file so we can use it everywhere. And also move the `with` function somewhere so we don't have to keep repeating it.

davisp: I contemplated something along those lines but couldn't convince myself on the best place to put it. I avoided couch_eunit.hrl because that could spider out to a number of other tests causing unneeded code churn. I'll see if a .hrl in the fabric test directory works.

davisp: Done.

PR #2343, opened 2019-12-06
Allow all params to be passed via body for POST _all_docs

bessbd:
## Overview
This change should allow users to supply all params in POST that can be supplied for GET now. This way we could avoid the ?key="foo" things that would probably cause a lot of pain for users.

As `/{db}/_design_docs` and `/{db}/_local_docs` are analogous to `_all_docs`, this change applies to all three of them.

## Testing recommendations

Just run the tests included with this PR.
Let me know if you need instructions for manual testing.

https://github.com/apache/couchdb-documentation/pull/462

## Checklist

- [x] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation: https://github.com/apache/couchdb-documentation/pull/462

### DIFF
--- a/src/chttpd/src/chttpd_show.erl
+++ b/src/chttpd/src/chttpd_show.erl
@@ -204,7 +204,7 @@ handle_view_list(Req, Db, DDoc, LName, {ViewDesignName, ViewName}, Keys) ->
     DbName = couch_db:name(Db),
     {ok, VDoc} = ddoc_cache:open(DbName, <<"_design/", ViewDesignName/binary>>),
     CB = fun list_cb/2,
-    QueryArgs = couch_mrview_http:parse_params(Req, Keys),
+    QueryArgs = couch_mrview_http:parse_body_and_query(Req, Keys),
     Options = [{user_ctx, Req#httpd.user_ctx}],
     couch_query_servers:with_ddoc_proc(DDoc, fun(QServer) ->
         Acc = #lacc{
--- a/src/couch_mrview/src/couch_mrview_http.erl
+++ b/src/couch_mrview/src/couch_mrview_http.erl
@@ -29,6 +29,7 @@
     parse_int/1,
     parse_pos_int/1,
     prepend_val/1,
+    parse_body_and_query/2,
     parse_body_and_query/3,
     parse_params/2,
     parse_params/3,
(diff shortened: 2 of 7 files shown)


bessbd:
Link to the doc PR: https://github.com/apache/couchdb-documentation/pull/462

src/chttpd/src/chttpd_show.erl
```diff
@@ -204,7 +204,7 @@ handle_view_list(Req, Db, DDoc, LName, {ViewDesignName, ViewName}, Keys) ->
     DbName = couch_db:name(Db),
     {ok, VDoc} = ddoc_cache:open(DbName, <<"_design/", ViewDesignName/binary>>),
     CB = fun list_cb/2,
-    QueryArgs = couch_mrview_http:parse_params(Req, Keys),
+    Args = couch_mrview_http:parse_body_and_query(Req, Keys),
```
davisp: I'd not change the variable name here which will shorten the diff. I assume that was done because its now more than the query string. Though we refer to view reads as queries in multiple places so there's no harm leaving it as QueryArgs.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -465,6 +466,14 @@ parse_params(Props, Keys, #mrargs{}=Args0, Options) ->
         parse_param(K, V, Acc, IsDecoded)
     end, Args1, Props).
 
+
+parse_body_and_query(#httpd{method='POST'} = Req, Keys) ->
+    Props = chttpd:json_body_obj(Req),
+    parse_body_and_query(Req, Props, Keys);
+
+parse_body_and_query(Req, Keys) ->
+    parse_params(chttpd:qs(Req), Keys, #mrargs{keys=Keys, group=undefined, group_level=undefined}, [keep_group_level]).
```
davisp: Project style is to wrap lines at 80 columns.

bessbd:
Thank you for the review @davisp ! I'll push commits for the changes you've requested soon.

src/chttpd/src/chttpd_show.erl
bessbd: Fixed in 5f163849a236b9309dbb06e7f0f62e57127207d3

src/couch_mrview/src/couch_mrview_http.erl
bessbd: Fixed in b025b2c36ff6ee4c82689e3fcd612623852c59f0

davisp:
Go ahead and squash those fixup commits and then we'll wait for CI to come back.

PR #2349, opened 2019-12-09
Prototype/fdb layer delete attachments

davisp:
## Overview

Properly delete attachment data when no longer referenced

## Testing recommendations

make eunit apps=fabric,couch_views,couch_jobs

## Checklist

### DIFF
--- a/src/fabric/src/fabric2_util.erl
+++ b/src/fabric/src/fabric2_util.erl
@@ -25,6 +25,8 @@
 
     validate_security_object/1,
 
+    hash_atts/1,
+
     dbname_ends_with/2,
 
     get_value/2,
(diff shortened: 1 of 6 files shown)


src/fabric/src/fabric2_util.erl
```diff
@@ -124,6 +126,20 @@ validate_json_list_of_strings(Member, Props) ->
     end.
 
 
+hash_atts([]) ->
+    <<>>;
+
+hash_atts(Atts) ->
+    lists:sort(fun(A, B) ->
```
iilyak: Am I reading it correctly we ignore the result of a lists:sort?

davisp: You sure are! Fixed now.

PR #2352, opened 2019-12-11
Remsh improvements

willholley:
## Overview

This adds a couple of improvements to remsh following an incident where I needed to use it in a CouchDB deployed via Helm. The CouchDB docker container and Helm chart both utilize `ERL_FLAGS` to set the Erlang cookie. Unfortunately, this interferes with `./remsh`, which adds it's own `-setcookie` parameter when calling `erl` - you end up with multiple `-setcookie` and, in the case of the Helm chart, multiple `-name` parameters, resulting in Erlang failing to connect to the remote node.

The PR firstly sets `ERL_FLAGS` to empty when invoking `erl`, to avoid conflicts with explicitly set parameters. Secondly, it attempts to automatically locate the cookie from `ERL_FLAGS` or `vm.args` which should yield a better default than `monster` in most cases.

## Testing recommendations

Testing is a little tricky as it requires a CouchDB deployed using ERL_FLAGS and/or vm.args. I tested it by building a CouchDB release / Docker container locally and passing `-e ERL_FLAGS="-setcookie somecookie"` to `docker run`.    

## Checklist

### DIFF
--- a/rel/overlay/bin/remsh
+++ b/rel/overlay/bin/remsh
@@ -23,8 +23,18 @@ BINDIR=$ROOTDIR/erts-$ERTS_VSN/bin
 PROGNAME=${0##*/}
 VERBOSE=""
 NODE="couchdb@127.0.0.1"
-COOKIE=monster
 LHOST=127.0.0.1
+VM_ARGS=$COUCHDB_BIN_DIR/../etc/vm.args
+
+# If present, extract cookie from ERL_FLAGS
+# This is used by the CouchDB Dockerfile and Helm chart
+COOKIE=$(echo "$ERL_FLAGS" | sed 's/^.*setcookie \([^ ][^ ]*\).*$/\1/g')
+if test -f "$VM_ARGS"; then
+# else attempt to extract from vm.args
+  VM_ARGS_COOKIE=$(awk '$1=="-setcookie"{print $2}' "$VM_ARGS")
+  COOKIE="${COOKIE:-$VM_ARGS_COOKIE}"
+fi
+COOKIE="${COOKIE:-monster}"
 
 printHelpAndExit() {
   echo "Usage: ${PROGNAME} [OPTION]... [-- <additional Erlang cli options>]"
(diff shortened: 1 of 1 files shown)


wohali:
@willholley @janl has slapped my hand fiercely in the past for using bashisms like this.

Can you consider rewriting it not to require bash? Note that line 1 **explicitly** is `#!/bin/sh` to highlight this choice.

wohali:
Output of `checkbashisms (1)`:

```
possible bashism in rel/overlay/bin/remsh line 32 (alternative test command ([[ foo ]] should be [ foo ])):
if [[ $ERL_FLAGS =~ $kCOOKIE_REGEX ]]; then
possible bashism in rel/overlay/bin/remsh line 33 ($BASH_SOMETHING):
  COOKIE="${BASH_REMATCH[1]}"
possible bashism in rel/overlay/bin/remsh line 33 (bash arrays, ${name[0|*|@]}):
  COOKIE="${BASH_REMATCH[1]}"
```

and I can't recall if your line 38 is a bashism or not.

rel/overlay/bin/remsh
```diff
@@ -71,6 +81,7 @@ if [ ! -z "$VERBOSE" ]; then
   set -x
 fi
 
-exec "$BINDIR/erl" -boot "$ROOTDIR/releases/$APP_VSN/start_clean" \
+# clear ERL_FLAGS to avoid conflicts with the cli parameters
```
wohali: Is it possible that `ERL_FLAGS` might have other settings you *would* want to pass into `erl`, such as `-args_file` or `-pa`? The latter seems especially useful for e.g. `recon`, `weatherreport`, etc. that might be stored out of tree.

wohali:
note: i haven't been able to test it, just did a static code review.

willholley:
thanks for the feedback @wohali. The latest changes only scrub `-name` and `-setcookie` from `ERL_FLAGS`.

PR #2357, opened 2019-12-13
Address flakey Mango Elixir tests

willholley:
## Overview

When creating a Mango index in the Elixir tests, we need to block until the database reports the index as available. This workaround was present in the python Mango tests but did not
get ported to Elixir when the partitioned Mango tests were added.

Presumably this is needed to workaround an underlying race condition between the index write and the ddoc cache.

## Testing recommendations

let's see if it reliably passes in CI...

## Checklist


garrensmith:
@willholley this seems to break the tests.

willholley:
@garrensmith oops - should be okay now. I think this may still be needed in addition #2363 given the python-based Mango tests have always run with n=1 iirc.

PR #2358, opened 2019-12-13
Update warning/error messages for Erlang versions

davisp:
## Overview

Soft drop support for Erlang 19 with a warning and be more clear in the
warnings about the broken versions of Erlang 20+.

## Checklist

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -27,25 +27,69 @@ end.
 VerList = lists:map(fun(X) -> {Int, _} = string:to_integer(X), Int end,
     string:tokens(VerString, ".")).
 
+
+DisplayMsg = fun(Msg, Args) ->
+    Base = iolist_to_binary(io_lib:format(Msg, Args)),
+    Lines = binary:split(Base, <<"\n">>, [global]),
+    MaxLen = lists:foldl(fun(Line, Acc) ->
+        max(Acc, size(Line))
+    end, 0, Lines),
+    Decoration = iolist_to_binary(["*" || _ <- lists:seq(1, MaxLen)]),
+    ReNewLined = [[L, "~n"] || L <- Lines],
+    NewLines = ["~n", Decoration, "~n", ReNewLined, Decoration, "~n~n"],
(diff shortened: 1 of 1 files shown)


rebar.config.script
```diff
@@ -27,25 +27,47 @@ end.
...
+end.
+
 NotSupported = fun(Ver) ->
-    io:fwrite("CouchDB does not support this version of Erlang (~p).~n", [Ver]),
-    io:fwrite("Check https://docs.couchdb.org/en/latest/whatsnew/index.html for the~n"),
-    io:fwrite("latest information on supported releases.~n"),
+    Msg = "This version of Erlang (~p) is not officially~n"
+        "supported by Apache CouchDB.~n"
+        "While we do not officially support this version,~n"
+        "there are also no known bugs or incompatibilities.~n",
```
wohali: Same comment here re width, and below. Perhaps this needs some callouts (`***` or `===`) and maybe a pause of a few seconds? Because otherwise it'll get scrolled off and no one will ever read it.

davisp: Oooh, good idea on pausing for a few seconds there.

rebar.config.script
```diff
@@ -27,25 +27,47 @@ end.
 VerList = lists:map(fun(X) -> {Int, _} = string:to_integer(X), Int end,
     string:tokens(VerString, ".")).
 
+
+ErlangTooOld = fun(Ver) -.
+    Msg = "This version of Erlang (~p) is too old for~n"
+        "use with Apache CouchDB.~n",
```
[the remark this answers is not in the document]
davisp: Done. Also yes. 80 columns is way too many to read all at once.

rebar.config.script
davisp: Added decoration and a pause when we're warning about 19.x. The others get the decoration but no pause given we either halt or ignore it on Travis.

davisp:
Also, I've tested this on the following Erlang versions:

* 18.3.4.9
* 19.3.6.8
* 20.3.8.15
* 21.2.2

PR #2359, opened 2019-12-13
Warn about orphan .git directories

jaydoane:
## Overview

A .git directory found in src/ usually indicates that the code in the
enclosing directory is managed from said .git directory. This can lead
to confusion, and developers opening PRs against obsolete repos when, as
happens frequently, a formerly separate repository becomes integrated
into the primary repo.

This patch changes the configure script to warn when it finds such .git
directories.

## Testing recommendations

Example output:
```
$ ./configure --dev --skip-deps
==> configuring couchdb in rel/couchdb.config
WARNING unexpected .git directory src/couch_tests/.git
WARNING unexpected .git directory src/rexi/.git
WARNING unexpected .git directory src/mem3/.git
WARNING unexpected .git directory src/smoosh/.git
WARNING unexpected .git directory src/couch_mrview/.git
WARNING unexpected .git directory src/couch/.git
WARNING unexpected .git directory src/couch_replicator/.git
WARNING unexpected .git directory src/ddoc_cache/.git
WARNING unexpected .git directory src/couch_peruser/.git
WARNING unexpected .git directory src/setup/.git
WARNING unexpected .git directory src/couch_log/.git
WARNING unexpected .git directory src/couch_epi/.git
WARNING unexpected .git directory src/mango/.git
WARNING unexpected .git directory src/chttpd/.git
WARNING unexpected .git directory src/couch_stats/.git
WARNING unexpected .git directory src/global_changes/.git
WARNING unexpected .git directory src/couch_event/.git
WARNING unexpected .git directory src/fabric/.git
WARNING unexpected .git directory src/couch_plugins/.git
WARNING unexpected .git directory src/ken/.git
WARNING unexpected .git directory src/couch_index/.git
You have configured Apache CouchDB, time to relax. Relax.
```

## Checklist

### DIFF
--- a/configure
+++ b/configure
@@ -237,4 +237,12 @@ if [ -d .git  -a $SKIP_DEPS -ne 1 ]; then
     ${REBAR} get-deps update-deps
 fi
 
+# External repos frequently become integrated with the primary repo,
+# resulting in obsolete .git directories, and possible confusion.
+# It is usually a good idea to delete these .git directories.
+for path in $(find src -name .git -type d); do
+    git ls-files --error-unmatch $(dirname $path) > /dev/null 2>&1 && \
+        echo "WARNING unexpected .git directory $path"
+done
+
 echo "You have configured Apache CouchDB, time to relax. Relax."


davisp:
I'm on board with this other than making it conditional. I'd remove the toggle and just run it regardless as it seems like good hygiene for those of us that skip around between major versions fairly often.

Though that also assumes that its not super expensive though it doesn't look like it would be.

jaydoane:
@davisp I originally _did_ write it without the conditional, but became concerned about forcing a potentially expensive`find` (which I've since improved on by adding `-type dir`) on everyone each time it was run. Maybe more worrisome though, I'm not sure `find` is available on Windows, so it seemed best to make it optional, but figured adding it to `make dev` would probably hit most use cases.

davisp:
Windows uses the `configure.ps1` command thingamajig so that shouldn't be an issue (cc @wohali). For the expensiveness I'm not overly concerned. But if you like then only adding it to `--dev` seems fine. I don't really see it useful as a standalone option cause I don't think it'd be an issue often enough to really get into muscle memory.

jaydoane:
Great, let's make it non-optional since it's simpler.

PR #2361, opened 2019-12-13
Make smoosh use UTC for scheduling window

jaydoane:
## Overview

Smoosh's exunit tests use UTC, so they sometimes fail when run in non-GMT time zones:

```
  1) test in_allowed_window returns true when to < from < now (SmooshSchedulingWindowTest)
     src/smoosh/test/exunit/scheduling_window_test.exs:53
     Assertion with == failed
     code:  assert :smoosh_utils.in_allowed_window('test_channel') == true
     left:  false
     right: true
     stacktrace:
       src/smoosh/test/exunit/scheduling_window_test.exs:59: (test)

  * test in_allowed_window returns false when from < to < now (0.5ms)
 
 2) test in_allowed_window returns false when from < to < now (SmooshSchedulingWindowTest)
     src/smoosh/test/exunit/scheduling_window_test.exs:44
     Assertion with == failed
     code:  assert :smoosh_utils.in_allowed_window('test_channel') == false
     left:  true
     right: false
     stacktrace:
       src/smoosh/test/exunit/scheduling_window_test.exs:50: (test)

  * test in_allowed_window returns true by default (0.03ms)
  * test in_allowed_window returns false when to < now < from (0.4ms)

  3) test in_allowed_window returns false when to < now < from (SmooshSchedulingWindowTest)
     src/smoosh/test/exunit/scheduling_window_test.exs:62
     Assertion with == failed
     code:  assert :smoosh_utils.in_allowed_window('test_channel') == false
     left:  true
     right: false
     stacktrace:
       src/smoosh/test/exunit/scheduling_window_test.exs:68: (test)

  * test in_allowed_window returns true when now < to < from (0.4ms)
  * test in_allowed_window returns false when now < from < to (0.5ms)
  * test in_allowed_window returns true when from < now < to (0.4ms)

  4) test in_allowed_window returns true when from < now < to (SmooshSchedulingWindowTest)
     src/smoosh/test/exunit/scheduling_window_test.exs:35
     Assertion with == failed
     code:  assert :smoosh_utils.in_allowed_window('test_channel') == true
     left:  false
     right: true
     stacktrace:
       src/smoosh/test/exunit/scheduling_window_test.exs:41: (test)
```

## Testing recommendations

```
make exunit
```
should pass regardless of when or in which time zone the tests are run.

## Checklist

### DIFF
--- a/src/smoosh/src/smoosh_utils.erl
+++ b/src/smoosh/src/smoosh_utils.erl
@@ -66,7 +66,7 @@ in_allowed_window(Channel) ->
     in_allowed_window(From, To).
 
 in_allowed_window(From, To) ->
-    {HH, MM, _} = erlang:time(),
+    {_, {HH, MM, _}} = calendar:universal_time(),
     case From < To of
     true ->
         ({HH, MM} >= From) andalso ({HH, MM} < To);
--- a/src/smoosh/test/exunit/scheduling_window_test.exs
+++ b/src/smoosh/test/exunit/scheduling_window_test.exs
@@ -1,8 +1,6 @@
 defmodule SmooshSchedulingWindowTest do
   use Couch.Test.ExUnit.Case
 
-  alias Couch.Test.Setup
-
   setup_all(context) do
     test_ctx = :test_util.start_couch([])
 


willholley:
@jaydoane this creates a subtle breaking change from the scheduled compaction settings in previous versions and seems like it could create some user confusion, particularly given the standard requirement to run compaction outside of peak hours which tend to follow local time rather than utc (e.g. adjust for daylight savings).

I wonder if it would be better to maintain compatibility with the previous settings and change the tests to use local time instead?

jaydoane:
> @jaydoane this creates a subtle breaking change from the scheduled compaction settings in [previous versions](https://github.com/apache/couchdb/blob/2.3.1/src/couch/src/couch_compaction_daemon.erl#L400) and seems like it could create some user confusion, particularly given the standard requirement to run compaction outside of peak hours which tend to follow local time rather than utc (e.g. adjust for daylight savings).
> 
> I wonder if it would be better to maintain compatibility with the previous settings and change the tests to use local time instead?

@willholley Oh drat, I was completely unaware that the old compaction demon used local time, which is a strange decision on its face, but I can see how it might have made sense, and how it could cause problems to change it. Although I could argue that a good time to change it might be now, with the introduction of smoosh. In any case, whichever way we go, the documentation should be very clear. I'm happy to change the tests instead to local time (or approve a PR if that's easier).

@kocolosk what do you think is the best approach here?

kocolosk:
Funny, I added this enhancement and copied over the use of `erlang:time()` from the old compaction daemon without even thinking about the timezone issue. Even when I was writing the documentation it seemed totally natural to include examples of suspending compactions during  business hours. Sort of crazy in hindsight.

ISO8601 says to assume local time if no offset is given, so I guess that's what we should do for servers that set a timezone to something other than UTC.

It seems like it'd also be a good idea to enhance smoosh to understand ISO8601-formatted timestamps and recommend their usage to avoid ambiguity.

jaydoane:
CouchDB emits time in ISO8601 format in several places, but I was unable to find anywhere in the code base which reads that format. For parsing ISO8601 strings with Erlang, [this library](https://github.com/erlsci/iso8601) looks like the most well-maintained of the many options coming up in a search. It uses an MIT license, so we could include it as yet another dependency.

PR #2363, opened 2019-12-15
Switch make elixir to use n = 1

garrensmith:
## Overview

Switch `make elixir` to use n = 1. This matches how the javascript tests are run and hopefully should make the elixir tests a little more consistent and less flaky.

## Testing recommendations

Tests should pass on Travis and Jenkins

## Checklist

### DIFF
--- a/test/elixir/test/reshard_basic_test.exs
+++ b/test/elixir/test/reshard_basic_test.exs
@@ -69,7 +69,7 @@ defmodule ReshardBasicTest do
     node1 = get_first_node()
 
     resp = post_job_node(db, node1)
-    assert resp.status_code == 201
+    assert resp.status_code in [201, 202]
 
     body = resp.body
     assert is_list(body)
@@ -109,7 +109,7 @@ defmodule ReshardBasicTest do
     resp = Couch.get("/#{db}/_shards")
     assert resp.status_code == 200
     shards = resp.body["shards"]
-    assert node1 not in shards["00000000-ffffffff"]
+    assert node1 not in Map.get(shards, "00000000-ffffffff", [])
     assert shards["00000000-7fffffff"] == [node1]
     assert shards["80000000-ffffffff"] == [node1]
 
@@ -132,7 +132,7 @@ defmodule ReshardBasicTest do
     node1 = get_first_node()
 
     resp = post_job_node(db, node1)
-    assert resp.status_code == 201
+    assert resp.status_code in [201, 202]
 
     body = resp.body
     assert is_list(body)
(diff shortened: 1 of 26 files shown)


test/elixir/test/reshard_basic_test.exs
```diff
@@ -109,7 +109,7 @@ defmodule ReshardBasicTest do
     resp = Couch.get("/#{db}/_shards")
     assert resp.status_code == 200
     shards = resp.body["shards"]
-    assert node1 not in shards["00000000-ffffffff"]
+    # assert node1 not in shards["00000000-ffffffff"]
```
garrensmith: @nickva I've had to comment out this line. Does this still make the test valid?

test/elixir/test/reshard_basic_test.exs
```diff
@@ -156,8 +156,8 @@ defmodule ReshardBasicTest do
     resp = Couch.get("/#{db}/_shards")
     assert resp.status_code == 200
     shards = resp.body["shards"]
-    assert node1 not in shards["00000000-7fffffff"]
-    assert node1 not in shards["80000000-ffffffff"]
+    # assert node1 not in shards["00000000-7fffffff"]
```
garrensmith: @nickva I've had to comment out these two lines. Is the test still valid?

garrensmith:
Tests passed first time on Travis 🎉

wohali:
@garrensmith temporary, or permanent change?

garrensmith:
This would be permanent


test/elixir/test/reshard_basic_test.exs
nickva: Still valid. But maybe better to replace the checks with 

```
    assert node1 not in Map.get(shards, "00000000-7fffffff", [])
    assert node1 not in Map.get(shards, "80000000-ffffffff", [])
```

What that means is before we were checking that after those ranges were split, the copies from the original node and range were gone. With other nodes in that range on node2 and node3, shards[00..-7f..] was returning a list of [node2, node3]. Then with n=1,there are no copies in the those ranges on any nodes so `shards[...]` was failing. The fix is to do a `get` with a default value of `[]`, so it would work on both n=1 and n>1

test/elixir/test/reshard_basic_test.exs
nickva: Let's replace it with

```
assert node1 not in Map.get(shards, "00000000-ffffffff", [])
```

This way it will still check that the original range was gone and work on n=1 and n>2 cases

(Added a bit of extra explanation in a similar assert below)

test/elixir/test/reshard_basic_test.exs
garrensmith: Awesome thanks @nickva. I've changed the tests.

janl:
+1

until we rewrite the tests to be specifically n>1 compatible, we should keep it as one.

going to n=3 without rewriting the tests was aspirational at best

garrensmith:
If the next release after 3.0 is with FDB, I don't think there would be much to gain by rewriting the tests to be n = 3.

wohali:
there will absolutely be a 3.x chain for a while in parallel with 4.0; whether 3.1 predates or postdates 4.0 is anyone's guess at this point.

PR #2367, opened 2019-12-16
Rework CI setup

wohali:
This commit:

* Removes Travis CI from the build (no more .travis.yml)
* Moves Jenkinsfile to `build-aux/Jenkinsfile.full` and updates the version of Erlang used in all steps to `20.3.8.24-1` (Erlang Solutions packages).
* Introduces a new `build-aux/Jenkinsfile.pr` that just builds CouchDB against the 3 most important versions of Erlang, intended for use when building PRs, using a special `debian-buster-erlang-all` image that has `kerl` builds of those 3 versions of Erlang:
  * the oldest supported
  * the version we release our binary convenience builds with
  * the latest supported
* Builds against SpiderMonkey 60 on the platforms where it is available
  (currently, only Debian buster)
* Updated README file with new, dynamic Jenkins embeddable build status badge

## TODO
1. Replace Travis mandate to merge with Jenkins mandate to merge (ASF Infra)


build-aux/Jenkinsfile.full
```diff
@@ -486,21 +457,92 @@ pipeline {
           } // post
         } // stage
 
+	/*
+	 *
+	 * This is just taking too long to run. Right now, the khash tests are timing out
+	 * on the IBM servers:
+	 *
+         * [2019-12-31T20:58:48.704Z]   khash randomized test
+         * [2019-12-31T20:59:04.869Z]     khash_test:103: randomized_test_ (State matches dict implementation)...*timed out*
```
wohali: @davisp FYI. We don't need emulated test runs to pass, but this usually passes for me locally before timeout. Is this a test we can improve at all?

janl:
+1

build-aux/Jenkinsfile.full
davisp: Its a long test by design really. The emulation is probably just slowing things down far enough to be an issue. I have a somewhat related branch [1] where I was poking at shortening some of these property based tests when we're running a PR. I was waffling between "run shorter on PR" and "only run long ones on nightly builds" which are slightly different. The PR approach I landed on wouldn't help here. But I was also meaning to ask if we might want to add a nightly "Long test" build that bumps a lot of the test sizes that we could run nightly or some such;

[1] https://github.com/apache/couchdb/commit/f290788be4beee412438580e17bf9016851c1f6f

wohali: Hey thanks Paul.

The current approach in this PR is to run the full test+packages build on all the platforms we can support. That could optionally run a longer set of tests if you like.

Right now in this PR that will run on `master` and any release branch after any change to that branch. It's not nightly, but it is also not a PR-blocking run. I think that meets our needs?

build-aux/Jenkinsfile.full
```diff
@@ -113,16 +113,17 @@ pipeline {
 
     stage('Test and Package') {
 
+      options {
+        skipDefaultCheckout()
+        timeout(time: 90, unit: "MINUTES")
+      }
+
```
kocolosk: Nice cleanup here

build-aux/Jenkinsfile.full
```diff
@@ -156,17 +157,12 @@ pipeline {
...
           agent {
             docker {
-              image 'couchdbdev/centos-6-erlang-19.3.6:latest'
+              image 'couchdbdev/centos-6-erlang-20.3.8.24-1:latest'
+              args "${DOCKER_ARGS}"
               alwaysPull true
-              label 'ubuntu'
-              // this keeps builds landing on the same host from clashing with each other
-              customWorkspace pwd() + '/centos6'
+              customWorkspace './centos6'
```
kocolosk: Is dropping the `pwd()` a stylistic change or are you actually moving the workspace to a different location than we had previously? [JENKINS-17818](https://issues.jenkins-ci.org/browse/JENKINS-17818) made me think that the `customWorkspace` is defined relative to `JENKINS_HOME`

build-aux/README.Jenkins
```diff
@@ -0,0 +1,150 @@
...
+
+End of November 2019, ASF and CloudBees reached an agreement to allow
+the Foundation to use CloudBees Core to have a farm of managed Jenkins
+masters. This allows the ASF to give larger projects their own dedicated
+Jenkins master, which can be custom configured for the project. They can
+readily manage this farm of Jenkins masters centrally, including push
+updates to all masters and their plugins. Naturally, this also reduces
+contention, as well as providing increased security for project-level
+credentials.  CouchDB is the first project to use this setup, via
+jenkins-cm1.apache.org. A CNAME for ci.couchdb.org is being established
```
kocolosk: s/is being established/has been established/

build-aux/README.Jenkins
```diff
@@ -0,0 +1,150 @@
...
+
+# Jenkins configuration
+
+All jobs on the new Jenkins master are contained in a CouchDB folder.
+Credentials for the project are placed within this folder; this is a
+unique capability of CloudBees Core at this time.
+
+## Pull Request job configuration
+
+Blue Ocean link: https://jenkins-cm1.apache.org/blue/organizations/jenkins/jenkins-cm1%2FPullRequests/activity/
```
kocolosk: Use the new CNAME for consistency?

build-aux/Jenkinsfile.full
wohali: The problem is that this single `pwd()` call actually requires a top-level agent, meaning that for a 3-wide pipeline, that actually instantiates a fourth agent *just* to calculate `pwd()`, and does not terminate for the length of the run.

I know a while back I was worried about network traffic syncing changes between agents, but I've not seen the issue in a while. As long as each build is happening in its own subdirectory, I think we're fine, and that `pwd()` is probably only returning `/home/jenkins/path/to/workspace` anyway.

wohali: @kocolosk if you want to check exactly where it's putting those builds, those IBM agents @davisp set up should have the remnants of a few builds on this PR now. With that perhaps we can clear up the problem.

kocolosk: I'm not sure @davisp added any SSH pub keys on those instances besides his own (at least, I don't know how to log in), but yes let's check. I'm worried that the current configuration might be skipping the `/path/to/workspace` piece.

kocolosk: For example, if I check the logs of a currently running job:

https://jenkins-cm1.apache.org/job/jenkins-cm1/job/FullPlatformMatrix/job/jenkins-pr-jenkinsfile/13/consoleFull

I see a line for CentOS 6 that says

> `16:35:57  Running on couchdb-ci-worker-dal-1-5 in /home/****/workspace/ormMatrix_****-pr-****file@3`

and then a few lines later one that seems to report the invocation of the `customWorkspace` command:

> `16:35:57  Running in /home/****/centos6@2`

Comparing these two suggests to me that we've exited the workspace entirely with that `customWorkspace` argument.

wohali: OK, suggest an alternative. `pwd()` sacrificing an entire agent for the lifetime of the run seems a huge waste of resources.

davisp: @kocolosk The list of SSH keys installed is here:

https://github.com/apache/couchdb-infra-cm/blob/main/roles/common/tasks/main.yml#L6-L8

I can add yours if you like.

build-aux/Jenkinsfile.full
davisp: Sorry, I don't think I was clear enough earlier. What I was asking/proposing was that we create a new type of nightly build that is something like "Nightly Soak Test" where we jack up all of the values for our property/long unit tests that runs once a day on only a single platform.

My current approach is the opposite in that we tune down some parameters on PRs only as opposed to everything except the soak test build.

Also to be clear, these long running tests are actually not entirely supposed to be "unit tests" per se. The property based testing is (in my experience) supposed to *generate* unit tests. The workflow would basically be something like, design your property, run it for many hours, any failures are added as regression tests. And the "run for hours" part is executed before major events like large PRs and releases and so on.

And alternatively, something else we could do is pull all of those tests out into a separate manual build that *only* runs the property tests and actually does run them for hours on a manual button push.

build-aux/Jenkinsfile.full
kocolosk: OK, I will see what I can do. I don't have my dev environment set up to test it right now, but I'm wondering about whether `${WORKSPACE}` can be used in the `customWorkspace` argument in lieu of `pwd()`

wohali: No good.

```
groovy.lang.MissingPropertyException: No such property: WORKSPACE for class: groovy.lang.Binding
```

kocolosk: Bummer. Thanks for checking. Will think on it a bit more.

wohali: worse, https://issues.jenkins-ci.org/browse/JENKINS-60557

wohali: An alternative is something like

```
customWorkspace "/path/to/well/known/OK/location/prbuild/${BRANCH_NAME}/centos6"
```

wohali: Tried and didn't work:
* `${env.WORKSPACE}` (resolves to `null`)
* "${JENKINS_HOME}/workspace/${ITEM_FULL_NAME}" (see https://github.com/jenkinsci/jenkins/pull/3364/files#diff-9e24c3bccc9e921f1a5cc216abc25d03R5111 and https://issues.jenkins-ci.org/browse/JENKINS-21942) - "No such property: ITEM_FULL_NAME"
* "${JENKINS_HOME}/workspace/${ITEM_FULLNAME}" "groovy.lang.MissingPropertyException: No such property: ITEM_FULLNAME" (same as above)

@kocolosk According to Andrew Bayer, ["The problem here is that we generate the custom workspace string before we're actually on the agent."](https://issues.jenkins-ci.org/browse/JENKINS-50907?focusedCommentId=353326&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-353326) And since you can have custom paths per machine for where the workspace is stored, I think this may be intractable. Our approach previously was probably working because the scripted pipeline `pwd()` command was spinning up an agent, calculating that path, and it just happened to work everywhere (since every agent had the exact same file system layout). And, prior to your change, my `/tmp/...` was also consistent across all machines.

As a reminder, what we're working around here is https://issues.jenkins-ci.org/browse/JENKINS-57454

However, we may be chasing our tails, see https://issues.jenkins-ci.org/browse/JENKINS-48319

What about changing the build steps to `cd ${ERLANG_VERSION}` or `cd ${PLATFORM}` before they continue and give up on customWorkspaces entirely?

wohali: For what it's worth, that last suggestion seems to work for now.

https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FPullRequests/detail/PR-2367/37/pipeline

If you're OK with that @kocolosk I'll resolve this. Then we're just waiting on Infra who probably are still out until next week (as, technically, I am...)

janl:
\o/

PR #2370, opened 2019-12-17
Minor tweaks for Jenkins UI

davisp:
Minor tweaks for Jenkins UI


davisp:
These changes aren't useful so abandoning it.

PR #2385, opened 2019-12-31
Fix double "make clean" again

wohali:
This is a recurrence of #1450 caused by ec416c3
(SpiderMonkey 60 PR), where a case clause in
rebar.config.script lacks a match when configure
has not yet been run yet. Not being able to run 
make clean|distclean on a new checkout breaks
package building under Debian.

/cc @jiangphcn


jiangphcn:
+1 thanks @wohali

PR #2388, opened 2020-01-03
Allow special document member _access

janl:
This is in preparation for [per-document access control](https://github.com/apache/couchdb/issues/assigned/janl) which is
going to use this field. We are adding this in 3.0 because per-
document access control will not be ready for 3.0, but added in
3.1 or later.

This commit allows a future version of CouchDB with per-document
access control to replicate with CouchDB 3.0, to make upgrades and
interoperability easy.

The per-documnet access control code is not going to store the
_access property in the document body like this patch does, but
is going to store it in an extra field inside of #doc and
assert access control for a document without having to load the
entire document body.


janl:
definitely no harm in advertising it here, will add a commit

janl:
Done.

PR #2390, opened 2020-01-04
feat: add metric counter for unindexed mango queries. Closes #1913

### LINKED ISSUE #1913: Add Metric to count unindexed mango queries
Similar motivation as #1324.

https://github.com/apache/couchdb/blob/master/src/mango/src/mango_cursor.erl#L152 would be the location.

janl:
see https://github.com/apache/couchdb/issues/1913

### DIFF
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -182,6 +182,7 @@ maybe_add_warning_int(ok, _, UserAcc) ->
    UserAcc;
 
 maybe_add_warning_int(Warning, UserFun, UserAcc) ->
+    couch_stats:increment_counter([mango, unindexed_queries]),
     Arg = {add_key, warning, Warning},
     {_Go, UserAcc0} = UserFun(Arg, UserAcc),
     UserAcc0.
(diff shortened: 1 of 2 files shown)


src/mango/src/mango_cursor.erl
```diff
@@ -182,6 +182,7 @@ maybe_add_warning_int(ok, _, UserAcc) ->
    UserAcc;
 
 maybe_add_warning_int(Warning, UserFun, UserAcc) ->
+    couch_stats:increment_counter([mango, unindexed_queries]),
```
wohali: Don't you need a `case` here to check if `UseIndexInvalidWarning` from above is actually `ok` (sourced from `NoIndexWarning`, in which case there wouldn't be a warning?

janl: the function clause in line 181 catches the `ok` case if I read things correctly. This function clause just unconditionally prints a warning, so I assumed I can just increment here without any additional logic

janl: definitely not the most obvs code

PR #2403, opened 2020-01-06
Fix JS tests on SpiderMonkey 60

davisp:
It turns out the infinite loop can now run fast enough that we exhaust
the stack space in a given couchjs process. This ends up leaving a
couchjs process that can't be used for other things as the `map_results`
buffer doesn't get reset inbetween calls.

## Testing recommendations

`./configure --dev --spidermonkey-version 60 && make && make javascript` 

On Debian Buster this fails occasionally.

## Checklist

### DIFF
--- a/share/server/state.js
+++ b/share/server/state.js
@@ -16,6 +16,7 @@ var State = {
     State.funs = [];
     State.lib = null;
     State.query_config = config || {};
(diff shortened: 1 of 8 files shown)


davisp:
Closing in favor of #2409

PR #2407, opened 2020-01-06
Make all OOM errors fatal in couchjs

davisp:
## Checklist


davisp:
Derps. We can't do this as it'd be backwards incompatible with views that would otherwise have built. With this change we'd end up stuck on a doc that caused it.

PR #2410, opened 2020-01-07
Mango metrics

willholley:
## Overview

Improve instrumentation of Mango with the following metrics:

 * `mango.query_invalid_index`: number of mango queries that generated an invalid index warning
 * `mango, docs_examined`: number of documents examined by mango queries (the same value returned by execution_stats for individual queries). This is counted on the shard level (i.e. where the work occurs) rather than at the query coordinator.
 * `mango.quorum_docs_examined`: number of documents examined by mango queries, using cluster quorum. These are always evaluated by the query coordinator.
 * `mango.results_returned`: number of rows returned by mango queries
 * `mango.query_time`: histogram of the length of time processing mango queries
 * `mango.too_many_docs_scanned`: number of mango queries that generated an index scan warning
 * `mango.evaluate_selector`: number of calls to `mango_selector:match`

## Testing recommendations

Make some Mango queries and call `_node/_local/_stats/mango` to view the metrics. Note that there is a bug in execution stats which means documents read are only counted when documents are matched (see #2412, which should ideally land before this).

 * Prior art: https://github.com/apache/couchdb/issues/1913
 * Documentation PR: https://github.com/apache/couchdb-documentation/pull/469
 * Fauxton PR: https://github.com/apache/couchdb-fauxton/pull/1245

## Checklist

### DIFF
--- a/src/couch/priv/stats_descriptions.cfg
+++ b/src/couch/priv/stats_descriptions.cfg
@@ -302,3 +302,7 @@
     {type, counter},
     {desc, <<"number of mango queries that could not use an index">>}
 ]}.
+{[mango, query_invalid_index], [
+    {type, counter},
+    {desc, <<"number of mango queries that generated an invalid index warning">>}
+]}.
--- a/src/couch/priv/stats_descriptions.cfg
+++ b/src/couch/priv/stats_descriptions.cfg
@@ -306,3 +306,7 @@
     {type, counter},
     {desc, <<"number of mango queries that generated an invalid index warning">>}
 ]}.
+{[mango, too_many_docs_scanned], [
+    {type, counter},
+    {desc, <<"number of mango queries that generated an index scan warning">>}
+]}.
--- a/src/couch/priv/stats_descriptions.cfg
+++ b/src/couch/priv/stats_descriptions.cfg
@@ -310,3 +310,23 @@
     {type, counter},
     {desc, <<"number of mango queries that generated an index scan warning">>}
 ]}.
+{[mango, docs_examined], [
+    {type, counter},
+    {desc, <<"number of documents examined by mango queries coordinated by this node">>}
+]}.
+{[mango, quorum_docs_examined], [
+    {type, counter},
+    {desc, <<"number of documents examined by mango queries, using cluster quorum">>}
+]}.
+{[mango, results_returned], [
+    {type, counter},
+    {desc, <<"number of rows returned by mango queries">>}
+]}.
+{[mango, query_time], [
+    {type, histogram},
+    {desc, <<"length of time processing a mango query">>}
+]}.
+{[mango, evaluate_selector], [
+    {type, counter},
+    {desc, <<"number of mango selector evaluations">>}
+]}.
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -166,6 +166,7 @@ invalid_index_warning(Index, Opts) ->
 invalid_index_warning_int(Index, {use_index, [DesignId]}) ->
     case filter_indexes([Index], DesignId) of
         [] ->
+            couch_stats:increment_counter([mango, query_invalid_index]),
             Reason = fmt("_design/~s was not used because it does not contain a valid index for this query.",
                 [ddoc_name(DesignId)]),
             [Reason];
@@ -175,6 +176,7 @@ invalid_index_warning_int(Index, {use_index, [DesignId]}) ->
 invalid_index_warning_int(Index, {use_index, [DesignId, ViewName]}) ->
     case filter_indexes([Index], DesignId, ViewName) of
         [] ->
+            couch_stats:increment_counter([mango, query_invalid_index]),
             Reason = fmt("_design/~s, ~s was not used because it is not a valid index for this query.",
                 [ddoc_name(DesignId), ViewName]),
             [Reason];
--- a/src/mango/src/mango_cursor.erl
+++ b/src/mango/src/mango_cursor.erl
@@ -19,7 +19,7 @@
     execute/3,
     maybe_filter_indexes_by_ddoc/2,
     remove_indexes_with_partial_filter_selector/1,
-    maybe_add_warning/3,
+    maybe_add_warning/4,
     maybe_noop_range/2
 ]).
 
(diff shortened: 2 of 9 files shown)


src/couch/priv/stats_descriptions.cfg
```diff
@@ -302,3 +302,35 @@
...
+]}.
+{[mango, results_returned], [
+    {type, counter},
+    {desc, <<"number of rows returned by mango queries">>}
+]}.
+{[mango, query_time], [
+    {type, histogram},
+    {desc, <<"length of time processing a mango query">>}
+]}.
+{[mango, too_many_docs_scanned], [
```
garrensmith: Is this a duplicate of the one on line 309?

src/mango/src/mango_cursor.erl
```diff
@@ -146,44 +146,87 @@ group_indexes_by_type(Indexes) ->
...
+    % Docs and DocsQuorum are mutually exclusive so it's safe to sum them
+    DocsScanned = Docs + DocsQuorum,
+    Ratio = index_scan_ratio(DocsScanned, ResultCount),
+    Threshold = config:get("mango", "index_scan_warning_threshold", 10),
+    index_scan_warning_int(Threshold, Ratio).
 
 
-maybe_add_warning_int(ok, _, UserAcc) ->
-   UserAcc;
+index_scan_ratio(DocsScanned, 0) ->
```
garrensmith: can you change this to `calculate_index_scan_ratio`

src/mango/src/mango_cursor.erl
```diff
@@ -146,44 +146,87 @@ group_indexes_by_type(Indexes) ->
...
+% returned, implying a lot of in-memory filtering
+index_scan_warning(#execution_stats {
+                    totalDocsExamined = Docs,
+                    totalQuorumDocsExamined = DocsQuorum,
+                    resultsReturned = ResultCount
+                }) ->
+    % Docs and DocsQuorum are mutually exclusive so it's safe to sum them
+    DocsScanned = Docs + DocsQuorum,
+    Ratio = index_scan_ratio(DocsScanned, ResultCount),
+    Threshold = config:get("mango", "index_scan_warning_threshold", 10),
```
garrensmith: There is a `config:get_integer` you can use here.

src/mango/src/mango_cursor.erl
```diff
@@ -146,44 +146,87 @@ group_indexes_by_type(Indexes) ->
...
+index_scan_ratio(DocsScanned, 0) ->
+    DocsScanned;
+index_scan_ratio(DocsScanned, ResultCount) ->
+    DocsScanned / ResultCount.
 
-maybe_add_warning_int(Warning, UserFun, UserAcc) ->
-    couch_stats:increment_counter([mango, unindexed_queries]),
-    Arg = {add_key, warning, Warning},
+
+index_scan_warning_int(Threshold, Ratio) when is_integer(Threshold), Threshold > 0, Ratio > Threshold ->
```
garrensmith: Rather make this a case statement in the `index_scan_warning`

```erlang
case Threshold > 0 andalso Ratio > Threshold of
  true -> ...error;
  false -> []
end.
```

```

File: src/mango/src/mango_cursor.erl [hunk 5]
```diff
@@ -146,44 +146,87 @@ group_indexes_by_type(Indexes) ->
...
+    Warnings = lists:append([W0, W1, W2]),
+    maybe_add_warning_int(Warnings, UserFun, UserAcc).
+
+
+maybe_add_warning_int([], _, UserAcc) ->
+   UserAcc;
+maybe_add_warning_int(Warnings, UserFun, UserAcc) ->
+    WarningStr = lists:join(<<"\n">>, Warnings),
+    Arg = {add_key, warning, WarningStr},
     {_Go, UserAcc0} = UserFun(Arg, UserAcc),
```
garrensmith: This should be UserAcc1

src/mango/src/mango_cursor.erl
```diff
@@ -146,44 +146,87 @@ group_indexes_by_type(Indexes) ->
...
+index_scan_warning_int(_, _) ->
+    [].
+
+
+maybe_add_warning(UserFun, #cursor{index = Index, opts = Opts}, Stats, UserAcc) ->
+    W0 = invalid_index_warning(Index, Opts),
+    W1 = no_index_warning(Index),
+    W2 = index_scan_warning(Stats),
+    Warnings = lists:append([W0, W1, W2]),
+    maybe_add_warning_int(Warnings, UserFun, UserAcc).
```
garrensmith: I'm kind of inclined to just make this a case statement rather than two functions.

src/mango/src/mango_cursor.erl
```diff
@@ -146,44 +146,87 @@ group_indexes_by_type(Indexes) ->
...
+
+
+maybe_add_warning_int([], _, UserAcc) ->
```
garrensmith: Can you move this function above all the functions that it calls. That makes it easier to follow.

src/mango/src/mango_execution_stats.erl
```diff
@@ -82,9 +83,11 @@ log_end(Stats) ->
 
 
 maybe_add_stats(Opts, UserFun, Stats, UserAcc) ->
+    Stats0 = log_end(Stats),
```
garrensmith: A variable with a 0 is lower that Stats. So you could make the first variable of Stats as Stats0 eg:

```
maybe_add_stats(Opts, UserFun, Stats0, UserAcc) ->
     Stats = log_end(Stats0), 
```
Is how its typically done in the CouchDB codebase

willholley:
thanks for the review @garrensmith. I think the latest push addresses your comments.

src/mango/src/mango_cursor.erl
```diff
@@ -146,46 +146,86 @@ group_indexes_by_type(Indexes) ->
...
+invalid_index_warning_int(Index, {use_index, [DesignId]}) ->
+    case filter_indexes([Index], DesignId) of
+        [] ->
+            couch_stats:increment_counter([mango, query_invalid_index]),
+            Reason = fmt("_design/~s was not used because it does not contain a valid index for this query.",
+                [ddoc_name(DesignId)]),
+            [Reason];
+        _ ->
+            []
+    end;
```
garrensmith: Sorry, super nitpicky, but can you add a space between each function here.

src/mango/src/mango_cursor.erl
```diff
@@ -146,46 +146,86 @@ group_indexes_by_type(Indexes) ->
...
+% In this scenario, Mango will ignore the index hint and auto-select an index.
+invalid_index_warning(Index, Opts) ->
+    UseIndex = lists:keyfind(use_index, 1, Opts),
+    invalid_index_warning_int(Index, UseIndex).
 
 
-maybe_add_warning_int(ok, _, UserAcc) ->
-   UserAcc;
+invalid_index_warning_int(Index, {use_index, [DesignId]}) ->
+    case filter_indexes([Index], DesignId) of
```
garrensmith: This is fine, but you can, if you want, also write this as:

```erlang
   filter_indexes([Index], DesignId) /= [] -> []; true ->
      couch_stats:increment_counter([mango, query_invalid_index]),
            Reason = fmt("_design/~s was not used because it does not contain a valid index for this query.",
                [ddoc_name(DesignId)]),
            [Reason];
        end,
 ```
Sometimes its nice to use an if statement like that when there really is only 1 branch

File: src/mango/src/mango_cursor.erl [hunk 11]
```diff
@@ -146,46 +146,86 @@ group_indexes_by_type(Indexes) ->
...
-    {_Go, UserAcc0} = UserFun(Arg, UserAcc),
-    UserAcc0.
+
+calculate_index_scan_ratio(DocsScanned, 0) ->
+    DocsScanned;
+calculate_index_scan_ratio(DocsScanned, ResultCount) ->
+    DocsScanned / ResultCount.
+
+
+maybe_add_warning(UserFun, #cursor{index = Index, opts = Opts}, Stats, UserAcc) ->
```
garrensmith: Can you move this to the top so it follows emilio https://github.com/cloudant-labs/emilio/blob/master/priv/documentation/420/description.md

wohali:
Looks nice. If you want this in 3.0, now's the time to merge it.

PR #2411, opened 2020-01-07
Bypass authentication check for /_up

### LINKED ISSUE #823: Healthcheck endpoint authentication should be optional

janl:
Add config variable chttpd.require_valid_user_except_for_up defaulting
to false.

This will allow various automated health check systems to hit /_up
without having to provide a username/password pair when the
chttpd.require_valid_user config setting is true. Apparently, many
of these health check providers do not even allow supplying creds
for such a purpose...

Closes #823

## Checklist


janl:
supersedes #856

@rnewson how do you feel bout his approach?

janl:
I usually squash those on merge to master.

janl:
afaict, the only merge commit that made it recently is from @kocolosk

PR #2422, opened 2020-01-08
Jenkins: update binary platform matrix

wohali:
This PR drops Debian jessie, adds Debian buster, and adds CentOS 8 to the binary platform build matrix on `master`.

### DIFF
--- a/build-aux/Jenkinsfile.full
+++ b/build-aux/Jenkinsfile.full
@@ -578,10 +578,10 @@ pipeline {
           echo 'Building Debian repo...'
           sh '''
             git clone https://github.com/apache/couchdb-pkg
-            cp js/debian-jessie/*.deb pkgs/jessie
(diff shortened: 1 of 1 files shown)


wohali:
Thanks @kocolosk . At some point we should replace the cache of .js packages on that server with direct downloads from our bintray (`http://dl.bintray.com/apache/couchdb-deb/` and `-rpm`)...or at the point when SM60 is supported everywhere upstream, we can drop the entire thing and simplify the process considerably.

jiangphcn:
thanks @wohali for consideration for SM60-related package distribution.

wohali:
@jiangphcn this is still SM1.8.5, I am not yet committing to shipping SM60 for all platforms. But we'll see ;)

jiangphcn:
totally understand @wohali . We can work together to make this happen in the future :-)

PR #2438, opened 2020-01-10
Reset a view shard if the signature is wrong

rnewson:
We encountered a case_clause error when reading the header for a .view
file as the response was {ok, {Sig, nil}} where Sig is neither the
expected sig or the pre-upgrade sig (though surely the pre-1.2 goop is
not firing anymore).

We now log this specific issue and then proceed as if we found no
valid header.

### DIFF
--- a/src/couch_mrview/src/couch_mrview_index.erl
+++ b/src/couch_mrview/src/couch_mrview_index.erl
@@ -127,6 +127,12 @@ open(Db, State0) ->
                     NewSt = couch_mrview_util:init_state(Db, Fd, State, Header),
                     ensure_local_purge_doc(Db, NewSt),
                     {ok, NewSt};
+                {ok, {WrongSig, _}} ->
+                    couch_log:error("~s has the wrong signature: expected: ~p but got ~p",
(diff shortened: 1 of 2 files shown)


src/couch_mrview/src/couch_mrview_index.erl
```diff
@@ -127,6 +127,12 @@ open(Db, State0) ->
                     NewSt = couch_mrview_util:init_state(Db, Fd, State, Header),
                     ensure_local_purge_doc(Db, NewSt),
                     {ok, NewSt};
+                {ok, {WrongSig, _}} ->
```
rnewson: I'm tempted to generalize further to `{ok, WeirdHeader}` or somesuch.

davisp: I'd rather not generalize here as I don't like that we're getting `{ok, SomethingBad}` and I'd rather force our hand if it comes to that rather than just resetting and papering over something likely serious. The headers are all md5'ed so its got to be programmer error of some sort at that level.

PR #2453, opened 2020-01-14
Enable building on Power (ppc64le) architecture

davisp:
Update Jenkinsfile.full

PR #2455, opened 2020-01-14
Enable building on Power (ppc64le) architecture

davisp:
Reopening #2453 with branch name that will also test the full build.


davisp:
Huh. It even reused the old build cause I didn't change the commit sha. Neat!

PR #2460, opened 2020-01-16
Change map indexes to be stored in one row

garrensmith:
## Overview

Changes map indexes to store the original key and value in a single
FDB row.

## Testing recommendations

All tests should pass

## Checklist


garrensmith:
I've added the map index key/value limits. If a document exceeds the key/value limit for a view in a design doc. It will be removed from that view. But if there are multiple views in a design doc, it will still be indexed in the other views in that design doc.

garrensmith:
I've removed the key limiting code and just left the switch to 1 fdb row for 1 map row. Since this has already received a +1. I'm merging.

PR #2296, opened 2019-11-02
Fix/1977

AGrzes:
## Overview

Implementation of suggestion from https://github.com/apache/couchdb/issues/1977 to set up single node automatically when `[couchdb] standalone=true` is set.

I hope I found a proper place to implement it.

## Testing recommendations

I found no fully automatic tests for `setup` module so I updated prepared manual test script and updated `dev/run` script to facilitate testing.

## Related Issues or Pull Requests

Fixes #1977

## Checklist

- [x] Code is written and works correctly
- [x] Changes are covered by tests
- [x] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation

PR #2463, opened 2020-01-16
Fix ini file setting for #2296

wohali:
Settings in `default.ini` should be the defaults. Also a little text cleanup for clarity.

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -69,8 +69,9 @@ default_engine = couch
 ; is 10GiB. A value of 0 or less will disable partition size checks.
 ;max_partition_size = 10737418240
 ;
-; Start node in single_node mode so default databases are created immediately.
-;single_node = true
+; When true, system databases _users and _replicator are created immediately
+; on startup if not present.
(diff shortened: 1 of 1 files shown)


rel/overlay/etc/default.ini
```diff
@@ -69,8 +69,9 @@ default_engine = couch
 ; is 10GiB. A value of 0 or less will disable partition size checks.
 ;max_partition_size = 10737418240
 ;
-; Start node in single_node mode so default databases are created immediately.
-;single_node = true
+; When true, system databases _users and _replicator are created immediately
```
willholley: minor nit - `_global_changes` is also counted as a system database (perhaps questionably) so will get created automatically.

wohali: @janl do you want to remove `_global_changes` from the setup wizard for 3.0.0?

let me know and we can decide if we fix the message here, or `src/setup/erl/setup.src`. I can see an argument both ways.

wohali: Have moved this change out to another PR.

PR #2468, opened 2020-01-17
[Jenkins] Try to cleanup hung beam.smp (FreeBSD)

wohali:
Minimal but it might help until #2437 is fixed.

### DIFF
--- a/build-aux/Jenkinsfile.full
+++ b/build-aux/Jenkinsfile.full
@@ -152,7 +152,8 @@ pipeline {
               junit '**/.eunit/*.xml, **/_build/*/lib/couchdbtest/*.xml, **/src/mango/nosetests.xml, **/test/javascript/junit.xml'
             }
             cleanup {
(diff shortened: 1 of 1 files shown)


kxepal:
Shouldn't beam process be killed by stage shutdown? Looks like we're doing Jenkins job here for him - timeout configure should kill the process correctly. At least that's what I observe in own environment.

wohali:
It's not working, when the build fails beam.smp processes are being left behind. We're not seeing this on other platforms because we run inside Docker and Jenkins reaps the container, but we're seeing it on workstations, such as @nickva 's. It just seems to be 100% repeatable whenever Jenkins runs on FreeBSD now.

kxepal:
That's strange. Thanks! Looks like process somehow being detached from root one so Jenkins not able to kill whole tree. I'll try to reproduce this behaviour - FreeBSD is my primary system for now.

PR #2475, opened 2020-01-20
Add POST /_node/<node>/_config/_reload

willholley:
## Overview

Using the backend port 5986 it was possible to reload config from disk
using the _config/_reload endpoint. This ports it to the `/_node` API on the frontend port.

## Testing recommendations

Launch CouchDB e.g. using `./dev/run`. Change the `local.ini` somehow and run:

```
curl localhost:15984/_node/_local/_config/_reload -XPOST
```

Verify the change using 

```
curl localhost:15984/_node/_local/_config`
```

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_node.erl
+++ b/src/chttpd/src/chttpd_node.erl
@@ -46,6 +46,16 @@ handle_node_req(#httpd{method='GET', path_parts=[_, Node, <<"_config">>]}=Req) -
     send_json(Req, 200, {KVs});
 handle_node_req(#httpd{path_parts=[_, _Node, <<"_config">>]}=Req) ->
     send_method_not_allowed(Req, "GET");
+% POST /_node/$node/_config/_reload - Flushes unpersisted config values from RAM
+handle_node_req(#httpd{method='POST', path_parts=[_, Node, <<"_config">>, <<"_reload">>]}=Req) ->
+    case call_node(Node, config, reload, []) of
+        ok ->
+            send_json(Req, 200, {[{ok, true}]});
+        {error, Reason} ->
+            chttpd:send_error(Req, {bad_request, Reason})
+    end;
+handle_node_req(#httpd{path_parts=[_, _Node, <<"_config">>, <<"_reload">>]}=Req) ->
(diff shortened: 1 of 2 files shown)


src/chttpd/src/chttpd_node.erl
```diff
@@ -46,6 +46,14 @@ handle_node_req(#httpd{method='GET', path_parts=[_, Node, <<"_config">>]}=Req) -
     send_json(Req, 200, {KVs});
 handle_node_req(#httpd{path_parts=[_, _Node, <<"_config">>]}=Req) ->
     send_method_not_allowed(Req, "GET");
+% POST /_node/$node/_config/_reload - Flushes unpersisted config values from RAM
+handle_node_req(#httpd{method='POST', path_parts=[_, Node, <<"_config">>, <<"_reload">>]}=Req) ->
```
jaydoane: Do we need to send an explicit `405` for non-POST requests to this endpoint? I tried this:
```
$ curl -u adm:pass -H 'Content-Type: application/json' localhost:15984/_node/_local/_config/_reload -XGET | jq .
{}
```
which doesn't seem right?

test/elixir/test/config_test.exs
```diff
@@ -174,4 +174,11 @@ defmodule ConfigTest do
       set_config(context, section, "wohali", "rules", 403)
     end)
   end
+
+  test "Reload config", context do
+    url = "#{context[:config_url]}/_reload"
+    resp = Couch.post(url)
+
+    assert resp.status_code == 200
```
jaydoane: It might be nice to actually demo changing a config value in e.g. a .ini file and see it change with the reload, but I'm not sure if there's a simple way to make that filesystem modification via elixir test.

willholley: yes - I couldn't see a good way to test this, though that test should perhaps exist for `couch_config` rather than this endpoint specifically.

willholley: the admin case is interesting - I wonder if that's to do with couchdb_config not triggering `config_event` notifications on reload.

PR #2486, opened 2020-01-25
quote strings for all text values

tonysun83:
## Overview

When a user issues a range query $lt, $lte, $gt, $gte for text indexes,
the query is translated into a MIN, MAX range query against clouseau.
If not quoted, an error occurs:
```
{"error":"text_search_error","reason":"Cannot parse
'(a_3astring:[\"\" TO string\\ containing\\ space})'..}
```
This is because the string is broken up into 3 tokens which the parser
cannot parse. If we add quotes to the string, the the range query works
correctly.

## Testing recommendations

Tests added to mango test suites, but requires clouseau to be setup. Text tests have already
been run and passes.

## Checklist

### DIFF
--- a/src/mango/src/mango_selector_text.erl
+++ b/src/mango/src/mango_selector_text.erl
@@ -366,7 +366,8 @@ value_str(Value) when is_binary(Value) ->
         true ->
             <<"\"", Value/binary, "\"">>;
         false ->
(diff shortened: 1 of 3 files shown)


wohali:
Hrm, why did this merge without triggering the `python-black` failure that required #2505 to land?

Issue #2423, opened 2020-01-08
couchjs segfault with SpiderMonkey 60


wohali:
Starting to get random failures on Jenkins, only for JS tests, and sporadically (but failing an entire PR attempt in the process):

```
[2020-01-08T02:30:35.272Z] test/javascript/tests/replicator_db_bad_rep_id.js              fail

[2020-01-08T02:58:01.029Z] test/javascript/tests/view_errors.js           fail
```

These were both in the history of PR-2411:

https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FPullRequests/detail/PR-2411/11/pipeline
https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FPullRequests/detail/PR-2411/12/pipeline

We also are having trouble contacting a worker:

```
[2020-01-08T02:54:54.891Z] couchdb-ci-worker-dal-1-8 was marked offline:
                Connection was broken: java.util.concurrent.TimeoutException: Ping started
                at 1578451854888 hasn't completed by 1578452094889
```

Don't know if it's related.

/cc @davisp

wohali:
ALso seeing:

```
Remote call on JNLP4-connect connection from 76.9a.30a9.ip4.static.sl-reverse.com/169.48.154.118:23453 failed
```

on occasion, which is causing nodes to be marked offline temporarily. Something's amiss....

davisp:
For the JNLP4-connect thing, I saw that on couchdb-ci-worker-dal-1-5 yesterday and it cleared itself up after awhile. No idea what's causing that.

For the failed without a stack trace I've also seen that happen to `design_docs.js` as well. And I have seen it once or twice locally. No idea what's causing it either.

davisp:
That is to say, for the JS tests I think there's something awkward going on that's not test specific necessarily. Will be good to try and keep track of the failed runs so we can gather logs and so on.

wohali:
I've updated the title to be more reflective of what we're seeing (the JNLP issue is being worked separately).

For reference, we're seeing this issue in:
* Pull Request Jenkins builds
* Full Platform Jenkins builds, but only on Debian Buster
* [This build](https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FFullPlatformMatrix/detail/jenkins-add-arm/8/pipeline) which is a PR to add arm64v8 support to the Full Platform Jenkins builds (Debian Buster)

[...]

PR #2487, opened 2020-01-25
Revert default priority change for couch_server

### LINKED ISSUE #2437: Hung beam.smp sitting at 100% CPU

wohali:
## Overview

In a59540132 we introduced a default increase in process priority for `couch_server` to high.

While tracking down #2437 , I managed to git bisect the problem to this commit.

This lead me to this thread on erlang-questions:

http://erlang.org/pipermail/erlang-questions/2018-June/095677.html

Notably:

> a single high-priority process in a tight loop stops all other processes, including those of max priority.

This problem surfaces on Erlang 21 and 22 specifically - and is exactly what we're seeing.

While we could do a specific version check, I think it would be healthier for now to disable this specific optimization by default, and only re-enable once we can track down what's going on in 21/22 that is causing scheduler collapse in this fashion.

## Testing recommendations

`while true; do make eunit apps=mem3; done` should run for hours without a hang. Most commonly the hang occurs in the reshard tests (during which various `couchjs` processes are spawned.)

Closes #2437 and may help with #2423 - needs more testing.

## Checklist

### DIFF
--- a/src/couch/src/couch_util.erl
+++ b/src/couch/src/couch_util.erl
@@ -692,7 +692,7 @@ set_mqd_off_heap(Module) ->
 
 
 set_process_priority(Module, Level) ->
-    case config:get_boolean("process_priority", atom_to_list(Module), true) of
+    case config:get_boolean("process_priority", atom_to_list(Module), false) of
         true ->
             process_flag(priority, Level),
             ok;


wohali:
6 hours later, a595401 is still hung but b02599e is not. I'll let it run overnight but I think this is it.

Thanks for the +1!

PR #2491, opened 2020-01-26
bump ets_lru, khash revs

### LINKED ISSUE #2391: update rebar.config with latest module tags
ioq e.g. is missing at least one commit

wohali:
Closes #2391 except for bumping Fauxton and docs.

@willholley can you tell me if we need the last couple of changes in Fauxton in 3.0.0 or not?

@davisp Do we need the changes to jiffy in CouchDB? We're fairly far back on that dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,6 +111,7 @@ src/couch_tests/ebin/
 src/global_changes/ebin/
 src/mango/ebin/
 src/mango/test/*.pyc
+src/mango/nosetests.xml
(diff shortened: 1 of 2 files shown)


willholley:
@wohali - yes we do need to release / bump Fauxton. If we can get
https://github.com/apache/couchdb-fauxton/pull/1246 in that would be a
bonus but I can't look at until tomorrow.

PR #2492, opened 2020-01-26
Jenkins: bump Erlang versions

wohali:
Docker images have updated versions of Erlang - so Jenkinsfile must follow.

Unfortunately this breaks PRs until this PR is merged, because now `couchdbdev/debian-buster-erlang-all` has different versions of Erlang in it than `Jenkinsfile.pr` does.

We can possibly factor some of this out and avoid the PR-blocking change by having the `-all` image have aliases like `/usr/local/kerl/{low,medium,high}/activate` to the three versions.

The `Jenkinsfile.full` version needs to be explicit so that packages are built appropriately for downstream.

At one point @kocolosk wanted to use tagging to improve build times and avoid the forced pull every time, but my experience was that when tags move, docker doesn't always pick it up, leading us back to having to merge a PR every time we want or need to update the Erlang in our containers.

### DIFF
--- a/build-aux/Jenkinsfile.full
+++ b/build-aux/Jenkinsfile.full
@@ -81,7 +81,7 @@ pipeline {
       agent {
         docker {
           label 'docker'
-          image 'couchdbdev/debian-stretch-erlang-20.3.8.24-1:latest'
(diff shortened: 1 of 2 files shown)


wohali:
merging, as the world is blocked until this hits.

PR #2504, opened 2020-01-31
fix: allow spaces in startup paths

janl:
this fixes the Mac binary build

### DIFF
--- a/rel/files/couchdb.in
+++ b/rel/files/couchdb.in
@@ -14,15 +14,15 @@
 
 canonical_readlink ()
   {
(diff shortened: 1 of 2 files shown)


wohali:
@big-r81 Does this change still work for starting CouchDB from symlinks?

janl:
I did test the exact sequence from the original PR that @big-r81 and that worked just fine :)

janl:
@wohali starting CouchDB from not the system service is rather rare on windows, would not worry too much about it

PR #2512, opened 2020-02-01
Makefile parity; remove [un]install tgts

wohali:
In progress - not yet tested - attempt to normalise `Makefile` and `Makefile.win` ahead of our 3.0.0 release.

Time permitting, Monday I'll get my Windows environment set up again, with the latest build chain and SM60, and make sure this does what we expect.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,8 +147,6 @@ fauxton: share/www
 .PHONY: check
 # target: check - Test everything
 check: all
-	# @$(MAKE) test-cluster-with-quorum
-	# @$(MAKE) test-cluster-without-quorum
 	@$(MAKE) python-black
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -432,9 +429,9 @@ endif
 
 .PHONY: install
 # target: install- install CouchDB :)
-install:
+install: release
 	@echo
-	@echo "Notice: There is no 'make install' command for CouchDB 2.x."
+	@echo "Notice: There is no 'make install' command for CouchDB 2.x+."
```
janl: maybe add a line that explains “we’ve run 'make release' here for you”

wohali: will do on final pass

PR #2036, opened 2019-05-22
Exunit

iilyak:
## Overview

Eunit testing framework has a number of deficiencies especially in the cases when mocking is involved. This PR introduces Elixir based ExUnit into our testing stack. This work uses notion of testing adapter which allows us to write test suite once and reuse it to test using different interfaces:
- fabric
- clustered
- backdoor

The PR doesn't include tests themselves. The tests would be included in separate PRs.

Here is one example of how the test might look:
```
defmodule Couch.Test.CRUD do
  use ExUnit.Case
  alias Couch.Test.Adapter
  alias Couch.Test.Utils, as: Utils

  alias Couch.Test.Setup

  require Record

  test_groups = [
    "using Clustered API": Adapter.Clustered,
    "using Backdoor API": Adapter.Backdoor,
    "using Fabric API": Adapter.Fabric,
  ]

  for {describe, adapter} <- test_groups do
    describe "Database CRUD #{describe}" do
      @describetag setup: %Setup{}
        |> Setup.Start.new([:chttpd])
        |> Setup.Adapter.new(adapter)
        |> Setup.Admin.new(user: "adm", password: "pass")
        |> Setup.Login.new(user: "adm", password: "pass")
      test "Create", %{setup: setup} do
        db_name = Utils.random_name("db")
        setup_ctx = setup |> Setup.run()
        assert {:ok, resp} = Adapter.create_db(Setup.get(setup_ctx, :adapter), db_name)
        assert resp.body["ok"]
      end
    end
  end
end
```

## Testing recommendations

Use `make exunit` to run all tests. 

You can also use mix directly as follows:
```
BUILDDIR=`pwd` ERL_LIBS=`pwd`/src MIX_ENV=test mix test --trace
```

## Related Issues or Pull Requests

N/A

## Checklist

- [x] Code is written and works correctly;
- [x] Changes are covered by tests;
- [x] Documentation reflects the changes;

PR #2519, opened 2020-02-04
Fix b3 header

iilyak:
## Overview

The B3 header parsing introduced in https://github.com/apache/couchdb/pull/2309 is broken. When b3 header is set the chttpd request is crashing with the following message:
```
[error] 2020-02-03T21:25:39.049590Z dbcore@172.30.138.101 <0.23004.0> -------- CRASH REPORT Process  (<0.23004.0>) with 0 neighbors crashed with reason: bad argument in call to binary:split("C10B5B9C9F06385D04BD6B3A57971D8C-1AEA26A4FE496E13-0000000000000000", <<"-">>, [global]) at 
   chttpd:get_trace_headers/1(line:1305) 
   <= chttpd:root_span_options/1(line:1271) 
   <= chttpd:start_span/1(line:1253) 
   <= chttpd:before_request/1(line:270) 
   <= chttpd:handle_request_int/1(line:243) 
   <= mochiweb_http:headers/6(line:128) 
   <= proc_lib:init_p_do_apply/3(line:247); 
initial_call: {mochiweb_acceptor,init,['Argument__1','Argument__2',...]}, 
ancestors: [chttpd,chttpd_sup,<0.452.0>], 
message_queue_len: 0, 
messages: [], 
links: [<0.455.0>,#Port<0.16978>], 
dictionary: [{dont_log_request,true},{ctrace_is_enabled,true},{nonce,"7f8a471..."},...], 
trap_exit: false, 
status: running, 
heap_size: 2586, 
stack_size: 27, 
reductions: 4211 
```

The reason for the crash is obvious. The `binary:split/3` is used in the place where we have string as an argument. According to the code in [mochiweb_headers.erl](https://github.com/apache/couchdb-mochiweb/blob/master/src/mochiweb_headers.erl#L196) the headers are always strings. This PR switches from `binary:split/3` to `string:split/3`. In order to make sure the change is working as expected the test is written. There was a need to adjust some testing plumbing to make test pass. This adjustment was done in the same PR since it is the very first test of this kind and testing adjustment in separate PR would be quite tricky. The adjustment to test plumbing consists of two changes:
- Update httpotion to 3.1.3 --- cherry pick from master 
- Support setting base_url in Couch test helper --- this is needed in the context of unit tests when http port is dynamic

## Testing recommendations

1. checkout PR
2. run test `make exunit apps=chttpd` - the test should pass
3. revert `3b9f04db97` locally and run test again - the test should timeout 

* https://github.com/apache/couchdb/pull/2309

## Checklist

### DIFF
--- a/src/chttpd/test/exunit/tracing_test.exs
+++ b/src/chttpd/test/exunit/tracing_test.exs
@@ -1,0 +1,101 @@
+defmodule Couch.Test.OpenTracing do
+  use Couch.Test.ExUnit.Case
+  alias Couch.Test.Setup
+  alias Couch.Test.Setup.Step
+  alias Couch.Test.Utils
+  import Couch.DBTest, only: [retry_until: 1]
+
+  defp create_admin(user_name, password) do
+    hashed = String.to_charlist(:couch_passwords.hash_admin_password(password))
+    :config.set('admins', String.to_charlist(user_name), hashed, false)
+  end
+
+  defp base_url() do
+    addr = :config.get('chttpd', 'bind_address', '127.0.0.1')
+    port = :mochiweb_socket_server.get(:chttpd, :port)
+    "http://#{addr}:#{port}"
+  end
+
+  setup_all context do
+    test_ctx = :test_util.start_couch([:chttpd])
+    :ok = create_admin("adm", "pass")
+
+    Map.merge(context, %{
+      base_url: base_url(),
+      user: "adm",
+      pass: "pass"
+    })
+  end
+
+  setup context do
+    db_name = Utils.random_name("db")
+    session = Couch.login(context.base_url, context.user, context.pass)
+
+    on_exit(fn ->
+      delete_db(session, db_name)
+    end)
+
+    create_db(session, db_name)
+
+    Map.merge(context, %{
+      db_name: db_name,
+      session: session
+    })
+  end
+
+  def create_db(session, db_name, opts \\ []) do
+    retry_until(fn ->
+      resp = Couch.Session.put(session, "/#{db_name}", opts)
+      assert resp.status_code in [201, 202]
+      assert resp.body == %{"ok" => true}
+      {:ok, resp}
+    end)
+  end
+
+  def delete_db(session, db_name) do
+    retry_until(fn ->
+      resp = Couch.Session.delete(session, "/#{db_name}")
+      assert resp.status_code in [200, 202, 404]
+      {:ok, resp}
+    end)
+  end
+
+  def create_doc(session, db_name, body) do
+    retry_until(fn ->
+      resp = Couch.Session.post(session, "/#{db_name}", body: body)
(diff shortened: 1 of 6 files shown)


src/chttpd/test/exunit/tracing_test.exs
```diff
@@ -0,0 +1,101 @@
...
+  use Couch.Test.ExUnit.Case
+  alias Couch.Test.Setup
+  alias Couch.Test.Setup.Step
+  alias Couch.Test.Utils
+  import Couch.DBTest, only: [retry_until: 1]
+
+  defp create_admin(user_name, password) do
+    hashed = String.to_charlist(:couch_passwords.hash_admin_password(password))
+    :config.set('admins', String.to_charlist(user_name), hashed, false)
+  end
```
chewbranca: It seems like the `create_admin` function along with some of the other function below would be useful to have in the shared Couch library. Any particular reason to not put these functions into the `couch.ex` module?

iilyak: Few concerns:
- I tried that before [1] and [2]
  - the complain was: we are developing our own framework which people need to learn.
- Adding `create_admin` to `Couch` exibit a chicken and egg problem. `Couch` implements HTTP interface. We cannot use HTTP interface without adding admin first. In context of eunit tests we always cheated and created admin via config manipulation. 
- We could consider `create_admin` an exception and still add it to `Couch` inspite of it not being HTTP based interface. However it will introduce a confusion. People use `Couch` module from elixir integration tests and the function wouldn't work in the context of an integration test.


When I was writing this test I considered using `Step.User.new(:admin, roles: [:server_admin])`. However since pluguble adapters [1] didn't make it to the framework and [2] doesn't have fabric2 support, so I had to do it differently. I followed the advise of the committers who believe that DRY is bad for tests and they should be as self contained as possible.

---
[1]: [#2036](https://github.com/apache/couchdb/pull/2036) - plugable adaptors based approach
[2]: [#2039](https://github.com/apache/couchdb/pull/2039) - fabric only interface

ksnavely:
Thanks for this fix Ilya

src/chttpd/test/exunit/tracing_test.exs
chewbranca: Ahhhh ok, thanks for the detailed explanation @iilyak 👍. It's outside the scope of this PR, but I wonder if it would make sense to add a tag like `@tag :with_admin` like we've done with dbs and what not [1], to inject an admin user context into the test and then we can abstract away whether the admin user was created through config or the http interface. As for framework vs anti DRY, initially with the Elixir suite we tried to avoid making a heavy framework, which I think is worthwhile, but at the same time there's common activities which should be easy to take care of, like creating dbs/docs/users/ddocs/etc.

[1] https://github.com/apache/couchdb/blob/master/test/elixir/test/security_validation_test.exs#L151-L153

PR #2522, opened 2020-02-05
Strip credentials from the traces we send to open tracing server.

iilyak:
## Overview

Previously the `db.instance` tag contained FDB connection string as is.
Which included password. This change masks the password so the `db.instance`
would look like `adm:****@127.0.0.1:4689`.

## Testing recommendations

1. checkout PR 
2. update rel/overlay/etc/default.ini
  ```
  [tracing]
  enabled = true
  app_name = couchdb
  protocol = http
  endpoint = http://127.0.0.1:14268

  [tracing.filters]
  all = (#{}) -> true
  ```
3. start jaeger `docker run -d -p 14268:14268/tcp -p 6831:6831/udp -p 16686:16686 jaegertracing/all-in-one:1.14`
4. start couchdb `make && dev/run --admin=adm:pass`
5. issue requests 
  ```
  curl -u "adm:pass" -X PUT http://localhost:15984/test
  ```
6. Go to jaeger UI http://127.0.0.1:16686/search and verify that db.instance tag has masked password

## Checklist

### DIFF
--- a/src/fabric/src/fabric2_server.erl
+++ b/src/fabric/src/fabric2_server.erl
@@ -85,7 +85,7 @@ init(_) ->
             ClusterFileStr = config:get("erlfdb", "cluster_file", ?CLUSTER_FILE),
             {ok, ConnectionStr} = file:read_file(ClusterFileStr),
             DbHandle = erlfdb:open(iolist_to_binary(ClusterFileStr)),
-            {string:trim(ConnectionStr), DbHandle}
+            {strip_creds(string:trim(ConnectionStr)), DbHandle}
     end,
     application:set_env(fabric, ?FDB_CLUSTER, Cluster),
     application:set_env(fabric, db, Db),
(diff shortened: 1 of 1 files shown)


eiri:
This doesn't strike me as a proper fix. I don't know what admin creds are doing in fdb connection  string in a first place, but you are altering behaviour of server module's `init` to addresscan auxiliary tool's problem. I'll defer decision to @davisp who has more solid idea of what's going on here, but to me this looks rather backwards.

iilyak:
You are right. The connection string is `description:ID@IP:PORT`
* The `description` is a logical description of the database using alphanumeric characters (a-z, A-Z, 0-9) and underscores.
* The `ID` is an arbitrary value containing alphanumeric characters (a-z, A-Z, 0-9). We recommend using a random eight-character identifier (such as the output of mktemp -u XXXXXXXX). Note that the ID will change when coordinators change

So it should be fine. I got confused because my content string is:
```
cat /usr/local/etc/foundationdb/fdb.cluster
adm:pass@127.0.0.1:4689
```
Therefore when I saw `db.instance` value equal to `adm:pass@127.0.0.1:4689` I friked out.

davisp:
I think that's from me originally thinking it was a user/pass pair based on the docs not realizing the password looking thing was just a token. Might be useful to change it to something like "test-cluster:token".

PR #2538, opened 2020-02-11
Encode startkey/endkey for all_docs

garrensmith:
## Overview

Encodes the startkey/endkey so that if a startkey is not binary it will return the expected results.
## Checklist

### DIFF
--- a/src/fabric/src/fabric2_util.erl
+++ b/src/fabric/src/fabric2_util.erl
@@ -33,7 +33,9 @@
     get_value/3,
     to_hex/1,
     from_hex/1,
-    uuid/0
+    uuid/0,
+
+    all_docs_encode/1
 ]).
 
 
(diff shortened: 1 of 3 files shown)


src/fabric/src/fabric2_util.erl
```diff
@@ -235,3 +237,12 @@ hex_to_nibble(N) ->
 
 uuid() ->
     to_hex(crypto:strong_rand_bytes(16)).
+
+
+all_docs_encode(null) -> <<>>;
```
davisp: This function name is a bit vague. How about `encode_all_docs_key` maybe?

test/elixir/test/all_docs_test.exs
```diff
@@ -41,11 +41,9 @@ defmodule AllDocsTest do
     assert resp["total_rows"] == length(rows)
 
     # Check _all_docs offset
-    retry_until(fn ->
```
davisp: I assume the removal of retry_until is because fdb is more consistent than old clustering code? We'll want to dedent the inner block though right? mix format should take care of that I think?

garrensmith: That is right

PR #2539, opened 2020-02-11
Implement flattened doc storage

eiri:
## Overview

Implementation of "flattened" doc storage format to [Document Storage RFC](https://github.com/apache/couchdb-documentation/blob/master/rfcs/004-document-storage.md)

## Testing recommendations

```
$ make eunit apps=couch suites=couch_util tests=json_explode_test_,json_implode_test_

$ make eunit apps=couch_jobs,couch_views,fabric
```


eiri:
@davisp : Hi Paul, I'm almost done with this, the only thing I'm still wrestling is a limiter on a depth of flattening JSON. Can you please take a look at `fabric2_fdb` part if I'm missing something or off the mark on how it suppose to be done? Thank you.

davisp:
I read through this and am not seeing where you're limiting depth at all. In terms of implementing it, if I'm not mistaken that's just extending your json_explode function with an extra parameter that tracks where you are in the JSON object.

Although, I don't think depth is the correct metric here? It should be based on the length of the key path that's constructed so that its guaranteed to fit within FoundationDB's key size requirements, right? Same idea, just have to count differently.

davisp:
I don't see anything technical/algorithmic that's gonna help here. You've already enumerated the major first ways I can think of that we'd want to add thresholds. Beyond that its just a definitions game of how things behave when those thresholds are exceeded.

One new thing that does occur to me with the depth limit though. The exploded storage model was always thought of as a path to getting sub-document updates to be more efficient. Assuming that when we hit a certain depth we just store anything further down in the value, it strikes me as something we'll need to ponder extra hard when defining that subdoc update stuff. The transition at that threshold seems like it'd be an easy off-by-one territory.

eiri:
@davisp good call on complications for subdoc update! To be honest I'm confused how that suppose to work even without limiter, I mean we have revision as a part of keys and we are not keeping sub-part of old revision docs around. So even for subdoc update it'll be "read the whole range or keypath-values, update some, insert all back". This doesn't strike me as massivly superior to just reading doc body as a whole and patching it with diff send as PATCH's payload.

eiri:
Closing as obsolete.

PR #2566, opened 2020-02-18
Sync Makefile with master

eiri:
## Overview

Bring in Makefile from `master` with all the latest goodies

## Testing recommendations

`$ make && make eunit apps=couch_jobs,couch_views,fabric` should still work


eiri:
@nickva yeah, I see elixir tests are failing on `prototype/db-layer` in the same way.

PR #2576, opened 2020-02-19
feat(breaking): make _all_dbs admin-only by default

janl:


rel/overlay/etc/default.ini
```diff
@@ -135,8 +135,8 @@ max_db_number_for_dbs_info_req = 100
 ; uncomment the next line to enable proxy authentication
 ; authentication_handlers = {chttpd_auth, proxy_authentication_handler}, {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, default_authentication_handler}
 
-; prevent non-admins from accessing /_all_dbs
-;admin_only_all_dbs = false
+; allow non-admins to access /_all_dbs
```
wohali: I would keep the original comment here. The meaning of the flag hasn't changed, just that it's now true by default.

janl: done

PR #2597, opened 2020-02-25
Add support for X-Couch-Trace header

iilyak:
## Overview

Sometime it is useful to be able to trigger open tracing using single header in the request. 
For such cases this PR introduces `X-Couch-Trace` header. This header combined with the following config in `default.ini` would cause tracing of the request. 

```
[tracing.filters]
all = (#{external := External}) when External == true -> true
```

*Note*: 
  - We check for existence of the header and don't care about its value
  - This feature could allow malicious user to launch resource exhaustion attack on tracing infrastructure (elastic search backend of jaeger). You are absolutely safe if you don't use `(#{external := External}) when External == true`.

## Testing recommendations

1. update rel/overlay/etc/default.ini as follows
```
index 592445c28..69091eab3 100644
@@ -572,15 +572,15 @@ min_priority = 2.0
 ;   - jaeger.thrift over udp
 ;   - jaeger.thrift over http
 ; ## Common settings
-; enabled = false ; true | false
-; app_name = couchdb ; value to use for the `location.application` tag
-; protocol = udp ; udp | http - which reporter to use
+enabled = true ; true | false
+app_name = couchdb ; value to use for the `location.application` tag
+protocol = http ; udp | http - which reporter to use
 ; ## jaeger.thrift over udp reporter
 ; thrift_format = compact ; compact | binary
 ; agent_host = 127.0.0.1
 ; agent_port = 6831
 ; ## jaeger.thrift over udp reporter
-; endpoint = http://127.0.0.1:14268
+endpoint = http://127.0.0.1:14268
 [tracing.filters]
 ;
@@ -603,4 +603,4 @@ min_priority = 2.0
 ; corresponding operation name key configured. Thus, users can easily
 ; log every generated trace by including the following:
 ;
-; all = (#{}) -> true
+all = (#{external := External}) when External == true -> true
```
2. start jaeger in docker container
```
docker run -d -p 6831:6831/udp -p 16686:16686 -p 14268:14268 jaegertracing/all-in-one:1.14
```
3. Start couchdb dev/run --admin=adm:pass
4. Create database curl -X PUT -u adm:pass http://127.0.0.1:15984/test
5. Query database curl -X GET -u adm:pass http://127.0.0.1:15984/test/_all_docs -H 'X-Couch-Trace: true'
6. Open http://localhost:16686/ in the browser and search for spans (you should find some traces)

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -1251,6 +1251,9 @@ start_span(Req) ->
         [_ | _] -> filename:join(PathParts)
     end,
     {IsExternalSpan, RootOptions} = root_span_options(MochiReq),
+
+    CouchTrace = header_value(Req, "X-Couch-Trace") /= undefined,
+
     Tags = maps:merge(#{
         peer => Peer,
         'http.method' => Method,
@@ -1259,7 +1262,7 @@ start_span(Req) ->
         path_parts => Path,
         'span.kind' => <<"server">>,
         component => <<"couchdb.chttpd">>,
-        external => IsExternalSpan
+        external => IsExternalSpan orelse CouchTrace
     }, ExtraTags),
 
     ctrace:start_span(OperationName, [


tsloughter:
Why a custom header? W3C has a trace header https://www.w3.org/TR/trace-context/ and it is being supported by many languages and vendors in the OpenTracing/OpenCensus successor https://opentelemetry.io/, including Erlang/Elixir https://github.com/open-telemetry/opentelemetry-erlang-api/

iilyak:
Currently open tracing implementation in CouchDB supports B3 headers (see https://github.com/apache/couchdb/tree/prototype/fdb-layer/src/ctrace#b3-propagation). The B3 headers were invented by zipkin and ended up in some opentracing solutions.

Unfortunately semantics of B3 headers doesn't allow expressing the following:
* I want to trace this request but I don't have RootSpan please create it for me

I was not aware that there is a concurrent W3C standard. It is probably quite new. Here is what I found on Jaeger page:
```
In the future Jaeger will support the upcoming W3C Trace-Context specification.....
```

I briefly looked into W3C standard this morning. It doesn't look like there is a specific header which can be used to enable tracing without specifying root span at the same time. Although I admit my 15 min reading of the spec could miss it.

tsloughter:
Ah ok, now I get what you are doing.

For B3 I think they actually have a "debug" header that you can send without id's that is meant to force a trace. But w3c traceheader I don't think has added such a feature yet.

Probably no reason to depend on it for this simple case of forcing a trace with no root from the client.

But keep an eye on opentelemetry :), the spec and many of the clients (hopefully including the Erlang/Elixir library) will be going into beta in March. Can't tell what you use currently for tracing and reporting traces -- something custom and built-in?

iilyak:
> But keep an eye on opentelemetry :), the spec and many of the clients (hopefully including the Erlang/Elixir library) will be going into beta in March. 

Thank you this is definitely good to know.

PR #2636, opened 2020-03-04
Clean up view size limit enforcement

davisp:
## Overview

Clean up view size limit enforcement

Both a performance and style cleanup. This reduces the number of ets
calls by roughly `2 * 2 * $num_docs * $num_views`. The first factor of
two is because this avoids re-calculating the size twice for both the id
and map indexes. The second factor of two is because we have to lookup
two different settings.

Stylistically this also moves the logic out of the fdb modules. Keeping
key/value structure and update logic is already pretty complicated so we
should avoid mixing external application logic into those modules as
much as possible for our sanity.

The behavior is slightly changed vs the original patch as well.
Originally only the view that contained the errant key or value was
affected. However, pre-existing behavior where a document fails to be
mapped correctly resulted in it being excluded from all view indexes
defined in the design document. That behavior has been restored.

## Testing recommendations

make check-fdb

## Checklist

### DIFF
--- a/src/couch_views/src/couch_views_indexer.erl
+++ b/src/couch_views/src/couch_views_indexer.erl
@@ -29,6 +29,8 @@
 % TODO:
 %  * Handle timeouts of transaction and other errors
 
+-define(KEY_SIZE_LIMIT, 8000).
+-define(VALUE_SIZE_LIMIT, 64000).
 
 spawn_link() ->
     proc_lib:spawn_link(?MODULE, init, []).
@@ -297,8 +299,13 @@ write_docs(TxDb, Mrst, Docs, State) ->
         last_seq := LastSeq
     } = State,
 
-    lists:foreach(fun(Doc) ->
-        couch_views_fdb:write_doc(TxDb, Sig, Views, Doc)
+    ViewIds = [View#mrview.id_num || View <- Views],
+    KeyLimit = key_size_limit(),
+    ValLimit = value_size_limit(),
+
+    lists:foreach(fun(Doc0) ->
+        Doc1 = calculate_kv_sizes(Mrst, Doc0, KeyLimit, ValLimit),
+        couch_views_fdb:write_doc(TxDb, Sig, ViewIds, Doc1)
     end, Docs),
 
     couch_views_fdb:set_update_seq(TxDb, Sig, LastSeq).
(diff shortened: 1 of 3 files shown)


src/couch_views/src/couch_views_indexer.erl
```diff
@@ -29,6 +29,8 @@
 % TODO:
 %  * Handle timeouts of transaction and other errors
 
+-define(KEY_SIZE_LIMIT, 8000).
```
iilyak: Maybe rename it to `DEFAULT_KEY_SIZE` and `DEFAULT_VALUE_SIZE_LIMIT`

src/couch_views/src/couch_views_indexer.erl
```diff
@@ -297,8 +299,13 @@ write_docs(TxDb, Mrst, Docs, State) ->
...
 
-    lists:foreach(fun(Doc) ->
-        couch_views_fdb:write_doc(TxDb, Sig, Views, Doc)
+    ViewIds = [View#mrview.id_num || View <- Views],
+    KeyLimit = key_size_limit(),
+    ValLimit = value_size_limit(),
+
+    lists:foreach(fun(Doc0) ->
+        Doc1 = calculate_kv_sizes(Mrst, Doc0, KeyLimit, ValLimit),
+        couch_views_fdb:write_doc(TxDb, Sig, ViewIds, Doc1)
```
garrensmith: I prefer that we pass in the Views here rather than the ViewId's. The extra loop to extract the ViewIds seems unnecessary

garrensmith:
> +1. Not sure what's up with Jenkins, but `make check-fdb` passes locally.

I don't think Jenkins is configured yet to run FDB

src/couch_views/src/couch_views_indexer.erl
davisp: Done.

src/couch_views/src/couch_views_indexer.erl
davisp: Started to try moving this but I didn't like how it moved extra logic into the fdb module. Given that the loop is the number of views on a design document the performance cost is insignificant so leaving the logic for pulling out the view id here seems better.

Issue #2643, opened 2020-03-09
v3.0 Unable to disable read-only _users database with the "users_db_security_editable" option


danielbarbarito:
## Description

When I set "users_db_security_editable" to true, I still get an error message when editing the permissions in _users.

## Steps to Reproduce

1. Go to the config page (http://localhost:5984/_utils/#_config/couchdb@localhost)
2. Set "users_db_security_editable" to true
3. Go to the permissions in the _users database (http://localhost:5984/_utils/#/database/_users/permissions)
4. Remove the _admin role
5. See the "Could not update permissions - reason: Error: You can't edit the security object of the user database" at the top of the page

## Expected Behaviour

I expect to be able to remove the _admin role from the members. I want to be able to allow anybody to sign up to my app without exposing the admin password to the user.

## Your Environment

* CouchDB version used: 3.0
* Browser name and version: Chrome 79.0.3945.130 
* Operating system and version: Mac OS

wohali:
Hi @danielbarbarito thanks for the report.

This looks like a bug in `default.ini`, because the setting has been put into the `[purge] ` section when it should be in the `[couchdb]` section. Your PR is changing things the wrong way; this was never meant to be in `purge`.

Can you please try adding the following text to your `local.ini` file, then restarting CouchDB and test if you can edit the `_users` DB permissions?

```ini
[couchdb]
users_db_security_editable = true
```

If that works, would you submit a new PR that moves the setting line into the `[couchdb]` section of the `default.ini` section?

danielbarbarito:
@wohali That worked! Just submitted a new PR.

rnewson:
PR is merged.

PR #2645, opened 2020-03-09
Move "users_db_security_editable" to the correct location in the ini file

danielbarbarito:
## Overview
This PR makes puts the "users_db_security_editable" configuration option in the correct location in the ini file.

## Related Issues
This PR fixes issue #2643

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -73,6 +73,9 @@ default_engine = couch
 ; on startup if not present.
 ;single_node = false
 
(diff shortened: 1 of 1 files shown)


rel/overlay/etc/default.ini
```diff
@@ -73,6 +73,9 @@ default_engine = couch
 ; on startup if not present.
 ;single_node = false
```
rnewson: the convention here is to use `;` for a blank line between commented out sections, can you update pls.

PR #2661, opened 2020-03-13
Enhance alg check

rnewson:
## Overview

The "alg" check can now take list of algorithms that are supported,
which must be from the valid list of algorithms.

## Testing recommendations

The new functionality is covered by eunit tests.

https://github.com/apache/couchdb/pull/2648/

## Checklist

### DIFF
--- a/src/jwtf/src/jwtf.erl
+++ b/src/jwtf/src/jwtf.erl
@@ -139,10 +139,11 @@ validate_alg(Props, Checks) ->
     case {Required, Alg} of
         {undefined, _} ->
             ok;
-        {true, undefined} ->
+        {Required, undefined} when Required /= undefined ->
             throw({bad_request, <<"Missing alg header parameter">>});
-        {true, Alg} ->
-            case lists:member(Alg, valid_algorithms()) of
+        {Required, Alg} when Required == true; is_list(Required) ->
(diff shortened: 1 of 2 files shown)


src/jwtf/src/jwtf.erl
```diff
@@ -139,10 +139,11 @@ validate_alg(Props, Checks) ->
     case {Required, Alg} of
         {undefined, _} ->
             ok;
-        {true, undefined} ->
+        {Required, undefined} when Required /= undefined ->
             throw({bad_request, <<"Missing alg header parameter">>});
-        {true, Alg} ->
-            case lists:member(Alg, valid_algorithms()) of
+        {Required, Alg} when Required == true; is_list(Required) ->
+            AllowedAlg = if Required == true -> true; true -> lists:member(Alg, Required) end,
```
jaydoane: This logic is a bit awkward, but I assume it's being done for backward compatibility.

rnewson: hm, nope? it's because you can say `[<<"alg">>]` to insist that there _is_ an alg property and `[{<<"alg">>, [<<"FOO">>, <<"BAR">>]}]` to insist that alg is present and one of the two options. This is the tidiest I could make that logic.

PR #2666, opened 2020-03-16
soft-deletion for database

jiangphcn:
## Overview

Instead of automatically and immediately removing data and index in database after a delete operation, soft-deletion allows to restore the deleted data back to original state due to a “fat finger”or undesired delete operation, up to defined periods, such as 48 hours.

In CouchDB 3.0, soft-deletion of database is implemented in [1]. The .couch file is renamed with the .<timestamp>.deleted.couch file after soft-deletion is enabled, and such file can be changed back to .couch for the purpose of restore. If restore is not needed and some specified period passed, the .<timestamp>.deleted.couch file can be deleted to achieve deletion of database permanently.

In CouchDB 4.0, with the introduction of FoundationDB, the data model and storage is changed. In order to support soft-deletion, we propose below solution and then implement them.

## Testing recommendations

make check skip_deps+=couch_epi apps=fabric tests=crud_test_
```
Test database CRUD operations
  fabric2_db_crud_tests:36: crud_test_ (create_db)...[0.032 s] ok
  fabric2_db_crud_tests:37: crud_test_ (open_db)...[0.020 s] ok
  fabric2_db_crud_tests:38: crud_test_ (delete_db)...[0.027 s] ok
  fabric2_db_crud_tests:39: crud_test_ (recreate_db)...[0.064 s] ok
  fabric2_db_crud_tests:40: crud_test_ (undelete_db)...[0.038 s] ok
  fabric2_db_crud_tests:41: crud_test_ (remove_deleted_db)...[0.037 s] ok
  fabric2_db_crud_tests:42: crud_test_ (old_db_handle)...[0.154 s] ok
  fabric2_db_crud_tests:43: crud_test_ (list_dbs)...[0.023 s] ok
  fabric2_db_crud_tests:44: crud_test_ (list_dbs_user_fun)...[0.014 s] ok
  fabric2_db_crud_tests:45: crud_test_ (list_dbs_user_fun_partial)...ok
  fabric2_db_crud_tests:46: crud_test_ (list_deleted_dbs)...[0.029 s] ok
  fabric2_db_crud_tests:47: crud_test_ (list_deleted_dbs_user_fun)...[0.027 s] ok
  fabric2_db_crud_tests:48: crud_test_ (list_deleted_dbs_user_fun_partial)...ok
  fabric2_db_crud_tests:49: crud_test_ (list_dbs_info)...[0.049 s] ok
  fabric2_db_crud_tests:50: crud_test_ (list_dbs_info_partial)...ok
  fabric2_db_crud_tests:51: crud_test_ (list_dbs_tx_too_old)...[0.078 s] ok
  fabric2_db_crud_tests:52: crud_test_ (list_dbs_info_tx_too_old)...[0.757 s] ok
  fabric2_db_crud_tests:53: crud_test_ (list_deleted_dbs_info)...[0.060 s] ok
  fabric2_db_crud_tests:54: crud_test_ (get_info_wait_retry_on_tx_too_old)...[0.098 s] ok
  fabric2_db_crud_tests:55: crud_test_ (get_info_wait_retry_on_tx_abort)...[0.072 s] ok
  [done in 2.674 s]
  All 20 tests passed.
```

make check skip_deps+=couch_epi apps=chttpd tests=deleted_dbs_test_
```
chttpd deleted dbs tests
  chttpd_deleted_dbs_test:72: should_return_error_for_unsupported_method...[0.107 s] ok
  chttpd_deleted_dbs_test:82: should_list_deleted_dbs...[0.052 s] ok
  chttpd_deleted_dbs_test:95: should_list_deleted_dbs_info...[0.022 s] ok
  chttpd_deleted_dbs_test:105: should_undelete_db...[0.040 s] ok
  chttpd_deleted_dbs_test:130: should_remove_deleted_db...[0.040 s] ok
  chttpd_deleted_dbs_test:145: should_undelete_db_to_target_db...[0.040 s] ok
  chttpd_deleted_dbs_test:172: should_not_undelete_db_to_existing_db...[0.034 s] ok
  [done in 0.357 s]
  All 7 tests passed.
```

https://github.com/apache/couchdb/pull/2707

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_misc.erl
+++ b/src/chttpd/src/chttpd_misc.erl
@@ -15,6 +15,7 @@
 -export([
     handle_all_dbs_req/1,
     handle_dbs_info_req/1,
+    handle_deleted_dbs_req/1,
     handle_favicon_req/1,
     handle_favicon_req/2,
     handle_replicate_req/1,
@@ -164,35 +165,7 @@ all_dbs_callback({error, Reason}, #vacc{resp=Resp0}=Acc) ->
 
 handle_dbs_info_req(#httpd{method = 'GET'} = Req) ->
     ok = chttpd:verify_is_server_admin(Req),
-
-    #mrargs{
-        start_key = StartKey,
-        end_key = EndKey,
-        direction = Dir,
-        limit = Limit,
-        skip = Skip
-    } = couch_mrview_http:parse_params(Req, undefined),
-
-    Options = [
-        {start_key, StartKey},
-        {end_key, EndKey},
-        {dir, Dir},
-        {limit, Limit},
-        {skip, Skip}
-    ],
-
-    % TODO: Figure out if we can't calculate a valid
-    % ETag for this request. \xFFmetadataVersion won't
-    % work as we don't bump versions on size changes
-
-    {ok, Resp} = chttpd:start_delayed_json_response(Req, 200, []),
-    Callback = fun dbs_info_callback/2,
-    Acc = #vacc{req = Req, resp = Resp},
-    {ok, Resp} = fabric2_db:list_dbs_info(Callback, Acc, Options),
-    case is_record(Resp, vacc) of
-        true -> {ok, Resp#vacc.resp};
-        _ -> {ok, Resp}
-    end;
+    send_db_infos(Req, list_dbs_info);
 handle_dbs_info_req(#httpd{method='POST', user_ctx=UserCtx}=Req) ->
     chttpd:validate_ctype(Req, "application/json"),
     Props = chttpd:json_body_obj(Req),
@@ -225,6 +198,92 @@ handle_dbs_info_req(#httpd{method='POST', user_ctx=UserCtx}=Req) ->
     chttpd:end_json_response(Resp);
 handle_dbs_info_req(Req) ->
     send_method_not_allowed(Req, "GET,HEAD,POST").
+
+handle_deleted_dbs_req(#httpd{method='GET', path_parts=[_]}=Req) ->
+    ok = chttpd:verify_is_server_admin(Req),
+    send_db_infos(Req, list_deleted_dbs_info);
+handle_deleted_dbs_req(#httpd{method='POST', user_ctx=Ctx, path_parts=[_]}=Req) ->
+    couch_httpd:verify_is_server_admin(Req),
+    chttpd:validate_ctype(Req, "application/json"),
+    GetJSON = fun(Key, Props, Default) ->
+        case couch_util:get_value(Key, Props) of
+            undefined when Default == error ->
+                Fmt = "POST body must include `~s` parameter.",
+                Msg = io_lib:format(Fmt, [Key]),
+                throw({bad_request, iolist_to_binary(Msg)});
+            undefined ->
+                Default;
+            Value ->
+                Value
+        end
+    end,
+    {BodyProps} = chttpd:json_body_obj(Req),
+    {UndeleteProps} = GetJSON(<<"undelete">>, BodyProps, error),
+    DbName = GetJSON(<<"source">>, UndeleteProps, error),
+    TimeStamp = GetJSON(<<"timestamp">>, UndeleteProps, error),
+    TgtDbName = GetJSON(<<"target">>, UndeleteProps, DbName),
+    case fabric2_db:undelete(DbName, TgtDbName, TimeStamp, [{user_ctx, Ctx}]) of
+        ok ->
+            send_json(Req, 200, {[{ok, true}]});
+        {error, file_exists} ->
+            chttpd:send_error(Req, file_exists);
+        {error, not_found} ->
+            chttpd:send_error(Req, not_found);
+        Error ->
+            throw(Error)
+    end;
+handle_deleted_dbs_req(#httpd{path_parts = PP}=Req) when length(PP) == 1 ->
+    send_method_not_allowed(Req, "GET,HEAD,POST");
+handle_deleted_dbs_req(#httpd{method='DELETE', user_ctx=Ctx, path_parts=[_, DbName]}=Req) ->
+    couch_httpd:verify_is_server_admin(Req),
+    TS = case ?JSON_DECODE(couch_httpd:qs_value(Req, "timestamp", "null")) of
+        null ->
+            throw({bad_request, "`timestamp` parameter is not provided."});
+        TS0 ->
+            TS0
+    end,
+    case fabric2_db:delete(DbName, [{user_ctx, Ctx}, {deleted_at, TS}]) of
+        ok ->
+            send_json(Req, 200, {[{ok, true}]});
+        {error, not_found} ->
+            chttpd:send_error(Req, not_found);
+        Error ->
+            throw(Error)
+    end;
+handle_deleted_dbs_req(#httpd{path_parts = PP}=Req) when length(PP) == 2 ->
+    send_method_not_allowed(Req, "HEAD,DELETE");
+handle_deleted_dbs_req(Req) ->
+    chttpd:send_error(Req, not_found).
+
+send_db_infos(Req, ListFunctionName) ->
+    #mrargs{
+        start_key = StartKey,
+        end_key = EndKey,
+        direction = Dir,
+        limit = Limit,
+        skip = Skip
+    } = couch_mrview_http:parse_params(Req, undefined),
+
+    Options = [
+        {start_key, StartKey},
+        {end_key, EndKey},
+        {dir, Dir},
+        {limit, Limit},
+        {skip, Skip}
+    ],
+
+    % TODO: Figure out if we can't calculate a valid
+    % ETag for this request. \xFFmetadataVersion won't
+    % work as we don't bump versions on size changes
+
+    {ok, Resp1} = chttpd:start_delayed_json_response(Req, 200, []),
+    Callback = fun dbs_info_callback/2,
+    Acc = #vacc{req = Req, resp = Resp1},
+    {ok, Resp2} = fabric2_db:ListFunctionName(Callback, Acc, Options),
+    case is_record(Resp2, vacc) of
+        true -> {ok, Resp2#vacc.resp};
+        _ -> {ok, Resp2}
+    end.
 
 dbs_info_callback({meta, _Meta}, #vacc{resp = Resp0} = Acc) ->
     {ok, Resp1} = chttpd:send_delayed_chunk(Resp0, "["),
--- a/src/chttpd/test/eunit/chttpd_deleted_dbs_test.erl
+++ b/src/chttpd/test/eunit/chttpd_deleted_dbs_test.erl
@@ -1,0 +1,234 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(chttpd_deleted_dbs_test).
+
+-include_lib("couch/include/couch_eunit.hrl").
+-include_lib("couch/include/couch_db.hrl").
+
+-define(USER, "chttpd_db_test_admin").
+-define(PASS, "pass").
+-define(AUTH, {basic_auth, {?USER, ?PASS}}).
+-define(CONTENT_JSON, {"Content-Type", "application/json"}).
+
+
+setup() ->
+    Hashed = couch_passwords:hash_admin_password(?PASS),
+    ok = config:set("admins", ?USER, ?b2l(Hashed), _Persist=false),
+    Addr = config:get("chttpd", "bind_address", "127.0.0.1"),
+    Port = mochiweb_socket_server:get(chttpd, port),
+    lists:concat(["http://", Addr, ":", Port, "/"]).
+
+
+teardown(_Url) ->
+    ok = config:delete("couchdb", "enable_database_recovery", false),
+    ok = config:delete("admins", ?USER, _Persist=false).
+
+
+create_db(Url) ->
+    {ok, Status, _, _} = http(put, Url, ""),
+    ?assert(Status =:= 201 orelse Status =:= 202).
+
+
+delete_db(Url) ->
+    {ok, 200, _, _} = http(delete, Url).
+
+
+deleted_dbs_test_() ->
+    {
+        "chttpd deleted dbs tests",
+        {
+            setup,
+            fun chttpd_test_util:start_couch/0,
+            fun chttpd_test_util:stop_couch/1,
+            {
+                foreach,
+                fun setup/0,
+                fun teardown/1,
+                [
+                    fun should_return_error_for_unsupported_method/1,
+                    fun should_list_deleted_dbs/1,
+                    fun should_list_deleted_dbs_info/1,
+                    fun should_undelete_db/1,
+                    fun should_remove_deleted_db/1,
+                    fun should_undelete_db_to_target_db/1,
+                    fun should_not_undelete_db_to_existing_db/1
+                ]
+            }
+        }
+    }.
+
+
+should_return_error_for_unsupported_method(Url) ->
+    ?_test(begin
+        {ok, Code, _, Body} = http(delete, mk_url(Url)),
+
+        ?assertEqual(405, Code),
+        ?assertEqual(<<"method_not_allowed">>, get_json(<<"error">>, Body))
+    end).
+
+
+should_list_deleted_dbs(Url) ->
+    ?_test(begin
+        DbName1 = create_and_delete_db(Url),
+        DbName2 = create_and_delete_db(Url),
+        {ok, Code, _, Body} = http(get, mk_url(Url)),
+        DeletedDbs = get_db_names(Body),
+
+        ?assertEqual(200, Code),
+        ?assertEqual(true, lists:member(DbName1, DeletedDbs)),
+        ?assertEqual(true, lists:member(DbName2, DeletedDbs))
+    end).
+
+
+should_list_deleted_dbs_info(Url) ->
+    ?_test(begin
+         DbName = create_and_delete_db(Url),
+         {ok, _, _, Body} = http(get, mk_url(Url, DbName)),
+         [{Props}] = jiffy:decode(Body),
+
+         ?assertEqual(DbName, couch_util:get_value(<<"db_name">>, Props))
+     end).
+
+
+should_undelete_db(Url) ->
+    ?_test(begin
+        DbName = create_and_delete_db(Url),
+        {ok, _, _, ResultBody} = http(get, mk_url(Url, DbName)),
+        [{Props}] = jiffy:decode(ResultBody),
+        TimeStamp = couch_util:get_value(<<"timestamp">>, Props),
+
+        ErlJSON = {[
+            {undelete, {[
+                {source, DbName},
+                {timestamp, TimeStamp}
+            ]}}
+        ]},
+
+        {ok, Code1, _, _} = http(get, Url ++ DbName),
+        ?assertEqual(404, Code1),
+
+        {ok, Code2, _, _} = http(post, mk_url(Url), ErlJSON),
+        ?assertEqual(200, Code2),
+
+        {ok, Code3, _, _} = http(get, Url ++ DbName),
+        ?assertEqual(200, Code3)
+    end).
+
+
+should_remove_deleted_db(Url) ->
+    ?_test(begin
+        DbName = create_and_delete_db(Url),
+        {ok, _, _, Body1} = http(get, mk_url(Url, DbName)),
+        [{Props}] = jiffy:decode(Body1),
+        TimeStamp = couch_util:get_value(<<"timestamp">>, Props),
+
+        {ok, Code, _, _} = http(delete, mk_url(Url, DbName, TimeStamp)),
+        ?assertEqual(200, Code),
+
+        {ok, _, _, Body2} = http(get, mk_url(Url, DbName)),
+        ?assertEqual([], jiffy:decode(Body2))
+    end).
+
+
+should_undelete_db_to_target_db(Url) ->
+    ?_test(begin
+        DbName = create_and_delete_db(Url),
+        {ok, _, _, Body} = http(get, mk_url(Url, DbName)),
+        [{Props}] = jiffy:decode(Body),
+        TimeStamp = couch_util:get_value(<<"timestamp">>, Props),
+
+        NewDbName = ?tempdb(),
+        ErlJSON = {[
+            {undelete, {[
+                {source, DbName},
+                {timestamp, TimeStamp},
+                {target, NewDbName}
+            ]}}
+        ]},
+
+        {ok, Code1, _, _} = http(get, Url ++ NewDbName),
+        ?assertEqual(404, Code1),
+
+        {ok, Code2, _, _} = http(post, mk_url(Url), ErlJSON),
+        ?assertEqual(200, Code2),
+
+        {ok, Code3, _, _} = http(get, Url ++ NewDbName),
+        ?assertEqual(200, Code3)
+    end).
+
+
+should_not_undelete_db_to_existing_db(Url) ->
+    ?_test(begin
+        DbName = create_and_delete_db(Url),
+        {ok, _, _, ResultBody} = http(get, mk_url(Url, DbName)),
+        [{Props}] = jiffy:decode(ResultBody),
+        TimeStamp = couch_util:get_value(<<"timestamp">>, Props),
+
+        NewDbName = ?tempdb(),
+        create_db(Url ++ NewDbName),
+        ErlJSON = {[
+            {undelete, {[
+                {source, DbName},
+                {timestamp, TimeStamp},
+                {target, NewDbName}
+            ]}}
+        ]},
+        {ok, Code2, _, ResultBody2} = http(post, mk_url(Url), ErlJSON),
+        ?assertEqual(412, Code2),
+        ?assertEqual(<<"file_exists">>, get_json(<<"error">>, ResultBody2))
+    end).
+
+
+create_and_delete_db(BaseUrl) ->
+    DbName = ?tempdb(),
+    DbUrl = BaseUrl ++ DbName,
+    create_db(DbUrl),
+    ok = config:set("couchdb", "enable_database_recovery", "true", false),
+    delete_db(DbUrl),
+    DbName.
+
+
+http(Verb, Url) ->
+    Headers = [?CONTENT_JSON, ?AUTH],
+    test_request:Verb(Url, Headers).
+
+
+http(Verb, Url, Body) ->
+    Headers = [?CONTENT_JSON, ?AUTH],
+    test_request:Verb(Url, Headers, jiffy:encode(Body)).
+
+
+mk_url(Url) ->
+    Url ++ "/_deleted_dbs".
+
+
+mk_url(Url, DbName) ->
+    Url ++ "/_deleted_dbs?key=\"" ++ ?b2l(DbName) ++ "\"".
+
+
+mk_url(Url, DbName, TimeStamp) ->
+    Url ++ "/_deleted_dbs/" ++ ?b2l(DbName) ++ "?timestamp=\"" ++
+        ?b2l(TimeStamp) ++ "\"".
+
+
+get_json(Key, Body) ->
+    {Props} = jiffy:decode(Body),
+    couch_util:get_value(Key, Props).
+
+
+get_db_names(Body) ->
+    RevDbNames = lists:foldl(fun({DbInfo}, Acc) ->
+        DbName = couch_util:get_value(<<"db_name">>, DbInfo),
+        [DbName | Acc]
+    end, [], jiffy:decode(Body)),
+    lists:reverse(RevDbNames).
--- a/src/fabric/src/fabric2_db.erl
+++ b/src/fabric/src/fabric2_db.erl
@@ -17,6 +17,7 @@
     create/2,
     open/2,
     delete/2,
+    undelete/4,
 
     list_dbs/0,
     list_dbs/1,
@@ -25,6 +26,10 @@
     list_dbs_info/0,
     list_dbs_info/1,
     list_dbs_info/3,
+
+    list_deleted_dbs_info/0,
+    list_deleted_dbs_info/1,
+    list_deleted_dbs_info/3,
 
     check_is_admin/1,
     check_is_member/1,
@@ -202,12 +207,30 @@ delete(DbName, Options) ->
     % Delete doesn't check user_ctx, that's done at the HTTP API level
     % here we just care to get the `database_does_not_exist` error thrown
     Options1 = lists:keystore(user_ctx, 1, Options, ?ADMIN_CTX),
-    {ok, Db} = open(DbName, Options1),
-    Resp = fabric2_fdb:transactional(Db, fun(TxDb) ->
-        fabric2_fdb:delete(TxDb)
-    end),
-    if Resp /= ok -> Resp; true ->
-        fabric2_server:remove(DbName)
+    case lists:keyfind(deleted_at, 1, Options1) of
+        {deleted_at, TimeStamp} ->
+            fabric2_fdb:transactional(DbName, Options1, fun(TxDb) ->
+                fabric2_fdb:remove_deleted_db(TxDb, TimeStamp)
+            end);
+        false ->
+            {ok, Db} = open(DbName, Options1),
+            Resp = fabric2_fdb:transactional(Db, fun(TxDb) ->
+                fabric2_fdb:delete(TxDb)
+            end),
+            if Resp /= ok -> Resp; true ->
+                fabric2_server:remove(DbName)
+            end
+    end.
+
+
+undelete(DbName, TgtDbName, TimeStamp, Options) ->
+    case validate_dbname(TgtDbName) of
+        ok ->
+            fabric2_fdb:transactional(DbName, Options, fun(TxDb) ->
+                fabric2_fdb:undelete(TxDb, TgtDbName, TimeStamp)
+            end);
+        Error ->
+            Error
     end.
 
 
@@ -276,6 +299,87 @@ list_dbs_info(UserFun, UserAcc0, Options) ->
                     Options
                 ),
             UserAcc3 = drain_all_info_futures(FinalFutureQ, UserFun, UserAcc2),
+            {ok, maybe_stop(UserFun(complete, UserAcc3))}
+        catch throw:{stop, FinalUserAcc} ->
+            {ok, FinalUserAcc}
+        end
+    end).
+
+
+list_deleted_dbs_info() ->
+    list_deleted_dbs_info([]).
+
+
+list_deleted_dbs_info(Options) ->
+    Callback = fun(Value, Acc) ->
+        NewAcc = case Value of
+            {meta, _} -> Acc;
+            {row, DbInfo} -> [DbInfo | Acc];
+            complete -> Acc
+        end,
+        {ok, NewAcc}
+    end,
+    {ok, DbInfos} = list_deleted_dbs_info(Callback, [], Options),
+    {ok, lists:reverse(DbInfos)}.
+
+
+list_deleted_dbs_info(UserFun, UserAcc0, Options0) ->
+    Dir = fabric2_util:get_value(dir, Options0, fwd),
+    StartKey0 = fabric2_util:get_value(start_key, Options0),
+    EndKey0 = fabric2_util:get_value(end_key, Options0),
+
+    {FirstBinary, LastBinary} = case Dir of
+        fwd -> {<<>>, <<255>>};
+        rev -> {<<255>>, <<>>}
+    end,
+
+    StartKey1 = case StartKey0 of
+        undefined ->
+            {FirstBinary};
+        DbName0 when is_binary(DbName0) ->
+            {DbName0, FirstBinary};
+        [DbName0, TimeStamp0] when is_binary(DbName0), is_binary(TimeStamp0) ->
+            {DbName0, TimeStamp0};
+        BadStartKey ->
+            erlang:error({invalid_start_key, BadStartKey})
+    end,
+    EndKey1 = case EndKey0 of
+        undefined ->
+            {LastBinary};
+        DbName1 when is_binary(DbName1) ->
+            {DbName1, LastBinary};
+        [DbName1, TimeStamp1] when is_binary(DbName1), is_binary(TimeStamp1) ->
+            {DbName1, TimeStamp1};
+        BadEndKey ->
+            erlang:error({invalid_end_key, BadEndKey})
+    end,
+
+    Options1 = Options0 -- [{start_key, StartKey0}, {end_key, EndKey0}],
+    Options2 = [
+        {start_key, StartKey1},
+        {end_key, EndKey1},
+        {wrap_keys, false}
+    ] ++ Options1,
+
+    FoldFun = fun(DbName, TimeStamp, InfoFuture, {FutureQ, Count, Acc}) ->
+        NewFutureQ = queue:in({DbName, TimeStamp, InfoFuture}, FutureQ),
+        drain_deleted_info_futures(NewFutureQ, Count + 1, UserFun, Acc)
+    end,
+    fabric2_fdb:transactional(fun(Tx) ->
+        try
+            UserAcc1 = maybe_stop(UserFun({meta, []}, UserAcc0)),
+            InitAcc = {queue:new(), 0, UserAcc1},
+            {FinalFutureQ, _, UserAcc2} = fabric2_fdb:list_deleted_dbs_info(
+                    Tx,
+                    FoldFun,
+                    InitAcc,
+                    Options2
+                ),
+            UserAcc3 = drain_all_deleted_info_futures(
+                    FinalFutureQ,
+                    UserFun,
+                    UserAcc2
+                ),
             {ok, maybe_stop(UserFun(complete, UserAcc3))}
         catch throw:{stop, FinalUserAcc} ->
             {ok, FinalUserAcc}
@@ -1059,6 +1163,37 @@ drain_all_info_futures(FutureQ, UserFun, Acc) ->
             DbInfo = make_db_info(DbName, InfoProps),
             NewAcc = maybe_stop(UserFun({row, DbInfo}, Acc)),
             drain_all_info_futures(RestQ, UserFun, NewAcc);
+        {empty, _} ->
+            Acc
+    end.
+
+
+drain_deleted_info_futures(FutureQ, Count, _UserFun, Acc) when Count < 100 ->
+    {FutureQ, Count, Acc};
+
+drain_deleted_info_futures(FutureQ, Count, UserFun, Acc) when Count >= 100 ->
+    {{value, {DbName, TimeStamp, Future}}, RestQ} = queue:out(FutureQ),
+    BaseProps = fabric2_fdb:get_info_wait(Future),
+    DeletedProps = BaseProps ++ [
+        {deleted, true},
+        {timestamp, TimeStamp}
+    ],
+    DbInfo = make_db_info(DbName, DeletedProps),
+    NewAcc = maybe_stop(UserFun({row, DbInfo}, Acc)),
+    {RestQ, Count - 1, NewAcc}.
+
+
+drain_all_deleted_info_futures(FutureQ, UserFun, Acc) ->
+    case queue:out(FutureQ) of
+        {{value, {DbName, TimeStamp, Future}}, RestQ} ->
+            BaseProps = fabric2_fdb:get_info_wait(Future),
+            DeletedProps = BaseProps ++ [
+                {deleted, true},
+                {timestamp, TimeStamp}
+            ],
+            DbInfo = make_db_info(DbName, DeletedProps),
+            NewAcc = maybe_stop(UserFun({row, DbInfo}, Acc)),
+            drain_all_deleted_info_futures(RestQ, UserFun, NewAcc);
         {empty, _} ->
             Acc
     end.
--- a/src/fabric/src/fabric2_fdb.erl
+++ b/src/fabric/src/fabric2_fdb.erl
@@ -22,12 +22,15 @@
     open/2,
     ensure_current/1,
     delete/1,
+    undelete/3,
+    remove_deleted_db/2,
     exists/1,
 
     get_dir/1,
 
     list_dbs/4,
     list_dbs_info/4,
+    list_deleted_dbs_info/4,
 
     get_info/1,
     get_info_future/2,
@@ -340,18 +343,70 @@ reopen(#{} = OldDb) ->
 
 
 delete(#{} = Db) ->
+    DoRecovery = fabric2_util:do_recovery(),
+    case DoRecovery of
+        true -> soft_delete_db(Db);
+        false -> hard_delete_db(Db)
+    end.
+
+
+undelete(#{} = Db0, TgtDbName, TimeStamp) ->
     #{
         name := DbName,
         tx := Tx,
-        layer_prefix := LayerPrefix,
-        db_prefix := DbPrefix
-    } = ensure_current(Db),
+        layer_prefix := LayerPrefix
+    } = ensure_current(Db0, false),
+    DbKey = erlfdb_tuple:pack({?ALL_DBS, TgtDbName}, LayerPrefix),
+    case erlfdb:wait(erlfdb:get(Tx, DbKey)) of
+        Bin when is_binary(Bin) ->
+            file_exists;
+        not_found ->
+            DeletedDbTupleKey = {
+                ?DELETED_DBS,
+                DbName,
+                TimeStamp
+            },
+            DeleteDbKey = erlfdb_tuple:pack(DeletedDbTupleKey, LayerPrefix),
+            case erlfdb:wait(erlfdb:get(Tx, DeleteDbKey)) of
+                not_found ->
+                    not_found;
+                DbPrefix ->
+                    erlfdb:set(Tx, DbKey, DbPrefix),
+                    erlfdb:clear(Tx, DeleteDbKey),
+                    bump_db_version(#{
+                        tx => Tx,
+                        db_prefix => DbPrefix
+                    }),
+                    ok
+            end
+    end.
 
-    DbKey = erlfdb_tuple:pack({?ALL_DBS, DbName}, LayerPrefix),
-    erlfdb:clear(Tx, DbKey),
-    erlfdb:clear_range_startswith(Tx, DbPrefix),
-    bump_metadata_version(Tx),
-    ok.
+
+remove_deleted_db(#{} = Db0, TimeStamp) ->
+    #{
+        name := DbName,
+        tx := Tx,
+        layer_prefix := LayerPrefix
+    } = ensure_current(Db0, false),
+
+    DeletedDbTupleKey = {
+        ?DELETED_DBS,
+        DbName,
+        TimeStamp
+    },
+    DeletedDbKey = erlfdb_tuple:pack(DeletedDbTupleKey, LayerPrefix),
+    case erlfdb:wait(erlfdb:get(Tx, DeletedDbKey)) of
+        not_found ->
+            not_found;
+        DbPrefix ->
+            erlfdb:clear(Tx, DeletedDbKey),
+            erlfdb:clear_range_startswith(Tx, DbPrefix),
+            bump_db_version(#{
+                tx => Tx,
+                db_prefix => DbPrefix
+            }),
+            ok
+    end.
 
 
 exists(#{name := DbName} = Db) when is_binary(DbName) ->
@@ -398,6 +453,20 @@ list_dbs_info(Tx, Callback, AccIn, Options0) ->
         {DbName} = erlfdb_tuple:unpack(DbNameKey, Prefix),
         InfoFuture = get_info_future(Tx, DbPrefix),
         Callback(DbName, InfoFuture, Acc)
+    end, AccIn, Options).
+
+
+list_deleted_dbs_info(Tx, Callback, AccIn, Options0) ->
+    Options = case fabric2_util:get_value(restart_tx, Options0) of
+        undefined -> [{restart_tx, true} | Options0];
+        _AlreadySet -> Options0
+    end,
+    LayerPrefix = get_dir(Tx),
+    Prefix = erlfdb_tuple:pack({?DELETED_DBS}, LayerPrefix),
+    fold_range({tx, Tx}, Prefix, fun({DbKey, DbPrefix}, Acc) ->
+        {DbName, TimeStamp} = erlfdb_tuple:unpack(DbKey, Prefix),
+        InfoFuture = get_info_future(Tx, DbPrefix),
+        Callback(DbName, TimeStamp, InfoFuture, Acc)
     end, AccIn, Options).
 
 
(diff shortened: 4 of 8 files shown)


src/fabric/src/fabric2_fdb.erl
```diff
@@ -314,12 +317,65 @@ delete(#{} = Db) ->
     } = ensure_current(Db),
 
     DbKey = erlfdb_tuple:pack({?ALL_DBS, DbName}, LayerPrefix),
-    erlfdb:clear(Tx, DbKey),
-    erlfdb:clear_range_startswith(Tx, DbPrefix),
+    DoRecovery = fabric_util:do_recovery(),
+    case DoRecovery of
+        true ->
+            {Mega, Secs, _} = os:timestamp(),
+            NowSecs = Mega * 1000000 + Secs,
```
nickva: This could be `erlang:system_time(second).` I think?

jiangphcn: yes, thanks @nickva for comments.

I changed this locally with 
```
            {{Y, Mon, D}, {H, Min, S}} = calendar:universal_time(),
            NowSecs = list_to_binary(lists:flatten(
                io_lib:format("~w~2.10.0B~2.10.0B."
                ++ "~2.10.0B~2.10.0B~2.10.0B",
                    [Y, Mon, D, H, Min, S])
            )),
            DeletedDbKey = erlfdb_tuple:pack({?DELETED_DBS, DbName, NowSecs},
                LayerPrefix),
```
and will push new commit today with other changes.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -172,7 +174,8 @@ create(#{} = Db0, Options) ->
     % Eventually DbPrefix will be HCA allocated. For now
     % we're just using the DbName so that debugging is easier.
     DbKey = erlfdb_tuple:pack({?ALL_DBS, DbName}, LayerPrefix),
-    DbPrefix = erlfdb_tuple:pack({?DBS, DbName}, LayerPrefix),
+    #{db_prefix_allocator := DBPrefixAllocator} = erlfdb_directory:root(),
```
nickva: Maybe we'd want an allocator instance under the LayerPrefix and not use the main directory one just to keep things isolated. Without reading the hca code too closely, I assume that's where it would store its counters. We should do a  quick benchmark at least to see if we see considerable contention when making lots of new databases in parallel. I suspect we won't but it's worth checking it out perhaps...

src/fabric/src/fabric2_fdb.erl
```diff
@@ -172,7 +174,8 @@ create(#{} = Db0, Options) ->
...
+    DbPrefix = erlfdb_hca:allocate(DBPrefixAllocator, Tx),
```
nickva: We want to generate small prefixes but I  we'd also want to keep the data under the `LayerPrefix` subspace. This might become important for backups and such or any places where we do range reads over the whole "CouchDB" instances sharing the same FDB cluster for example.

Then it might be something like 

```
AllocPrefix = erlfdb_hca:allocate(Allocator, Tx),
DbPrefix = erlfdb_tuple:pack({?DBS, AllocPrefix}, LayerPrefix)
```

src/fabric/src/fabric2_fdb.erl
```diff
@@ -314,12 +317,65 @@ delete(#{} = Db) ->
...
+    #{
+        name := DbName,
+        tx := Tx,
+        layer_prefix := LayerPrefix
+    } = ensure_current(Db0, false),
+
+    DeletedDbKey =  erlfdb_tuple:pack({?DELETED_DBS, DbName}, LayerPrefix),
+    DeletedDbs = erlfdb:wait(erlfdb:get_range_startswith(Tx, DeletedDbKey)),
+    lists:foldl(fun({DbKey, DbPrefix}, Acc) ->
+        DBInfo = get_info_wait(get_info_future(Tx, DbPrefix)),
```
nickva: I think we forgot to include `DBInfo` in the results?

src/fabric/src/fabric2_fdb.erl
```diff
@@ -314,12 +317,65 @@ delete(#{} = Db) ->
...
+        {?DELETED_DBS, DbName, DeletedTS} =
+            erlfdb_tuple:unpack(DbKey, LayerPrefix),
+        [DeletedTS| Acc]
+    end, [], DeletedDbs).
+
+
+restore(#{} = Db0, [DeletedTS|_]) ->
```
nickva: Do we expect any trailing path items, wondering about the `|_` part ?

src/fabric/src/fabric_util.erl
```diff
@@ -23,6 +23,8 @@
 -export([validate_all_docs_args/2, validate_args/3]).
 -export([upgrade_mrargs/1]).
 -export([worker_ranges/1]).
+-export([do_recovery/0]).
```
nickva: Let's avoid using `fabric_util` and use `fabric2_util` for consistency. However, in this case if we only check the config value once from one module, let's keep it in that module as a local function for now.

Wonder if `recovery` or `soft_deletion` is a better name... The discussion so far has used the `soft_deletion` term, maybe we should use that as the config value? The section should probably be `[fabric]` just to keep consistent with other FDB related settings.

src/fabric/src/fabric2_fdb.erl
jiangphcn: No, trailing path items will not be used. Just think not too strict here. But I can change to be strict by removing `|_`

src/fabric/src/fabric2_fdb.erl
jiangphcn: `DbPrefixAllocator` is defined in https://github.com/cloudant-labs/couchdb-erlfdb/pull/16/files#diff-cf5acaebd8465ae3276370bbf8d7aaf1R380-R383 with different `DbPrefix`. It will be different from `NodeIdAllocator`.

src/fabric/src/fabric2_fdb.erl
nickva: Makes sense. There is a balance here between using a non-standard format for time and the timestamp being an identifier of an db instance we want to restore. That is on one hand it might be nice to use https://en.wikipedia.org/wiki/ISO_8601 format (`2007-03-01T13:00:00Z` for ex.), however since we'd be passing that in the URL path we'd have to deal with `:` being a reserved character and encode it and all that.

src/fabric/src/fabric2_fdb.erl
nickva: We should probably not mix CouchDB internals with the erlfdb library. At first glance I would expect us not to touch the erlfdb library with this work. `Db` in `erlfdb`-land is an FDB database but here we are dealing with CouchDb `Db` instances so it already gets confusing.

Would this work `erlfdb_hca:create(erlfdb_tuple:pack({?HCA}, LayerPrefix))` to create an allocator instance for our layer prefix only. We are not going for full on FDB directories and nodes just need a short and unique prefix.

src/fabric/src/fabric_util.erl
nickva: Ah if it is an already setting that's fine. It'd still move the check for it to the module were we are reading the setting (fabric2_fdb).

nickva:
Thinking about this some more I wonder if it would make sense to separate enabling HCA from soft-deletion. Those are rather independent features. So for example, we'd have a PR to enable HCA. We'd test make sure that works. Then, as a separate piece of work, we'd enable soft-deletion. Would that be reasonable?

src/fabric/src/fabric2_fdb.erl
jiangphcn: It looks to be fine to use ISO_8601 format, i.e. using `:` in URL.
```
curl -u foo:bar  http://127.0.0.1:15984/db01 -X GET
{"update_seq":"000000000000000000000000","cluster":{"n":0,"q":0,"r":0,"w":0},"compact_running":false,"data_size":0,"db_name":"db01","disk_format_version":0,"disk_size":0,"instance_start_time":"0","purge_seq":0,"doc_del_count":0,"doc_count":0,"sizes":{"external":2,"views":0}}
curl -u foo:bar  http://127.0.0.1:15984/db01 -X DELETE
{"ok":true}
curl -u foo:bar  http://127.0.0.1:15984/db01/_deleted_dbs_info
"[\"2020-03-20T11:42:54Z\"]"
curl -u foo:bar  http://127.0.0.1:15984/db01?recover=2020-03-20T11:42:54Z -X PUT
{"ok":true}
```

src/fabric/src/fabric_util.erl
jiangphcn: Sure, and moved this to `fabric2_util`

jiangphcn:
> Thinking about this some more I wonder if it would make sense to separate enabling HCA from soft-deletion. Those are rather independent features. So for example, we'd have a PR to enable HCA. We'd test that make sure that works. Then, as a separate piece of work, we'd enable soft-deletion. Would that be reasonable?

Yes, it is totally reasonable. We can have 2 PRs, i.e. one is for enabling HCA, and another is for soft-deletion. The only point I have is where to call ` DbPrefixAllocator = erlfdb_hca:create(?ERLFDB_EXTEND(DbId, <<"hca">>)),`. In https://github.com/cloudant-labs/couchdb-erlfdb/pull/16/files#diff-cf5acaebd8465ae3276370bbf8d7aaf1R383, it was supposed that this can be called once or limited times, but after moving to `fabric2_fdb.erl `,  I stil try to find right place to call it.

src/fabric/src/fabric2_fdb.erl
jiangphcn: hey @nickva, with the suggestion from @mikerhodes, the timestamp is specified in payload of POST instead of URL
```
  POST /{db}/_restore

  { "sourceTS": "<deletedTS>", "destination": "animaldb"}
```
Thus, it is probably safe to use ISO_8601 format, like `2007-03-01T13:00:00Z` where `:` exists in payload.

src/chttpd/src/chttpd_db.erl
```diff
@@ -79,6 +79,8 @@ handle_request(#httpd{path_parts=[DbName|RestParts],method=Method}=Req)->
          end;
     {_, []} ->
         do_db_req(Req, fun db_req/2);
+    {_, [<<"_restore">>|_]} ->
```
davisp: This needs to go into the handlers list [1] and not be hard coded at this level.

[1] https://github.com/apache/couchdb/blob/prototype/fdb-layer/src/chttpd/src/chttpd_httpd_handlers.erl#L20-L33

src/chttpd/src/chttpd_misc.erl
```diff
@@ -146,6 +147,65 @@ handle_all_dbs_req(#httpd{method='GET'}=Req) ->
...
+    send_method_not_allowed(Req, "GET,HEAD").
+
+deleted_dbs_info_req(#httpd{user_ctx=Ctx}=Req, DbName) ->
+    couch_httpd:verify_is_server_admin(Req),
+    case fabric2_db:deleted_dbs_info(DbName, [{user_ctx, Ctx}]) of
+        {ok, Result} ->
+            {ok, Resp} = chttpd:start_json_response(Req, 200),
+            send_chunk(Resp, "["),
+            lists:foldl(fun({Timestamp, Info}, AccSeparator) ->
+                Json = ?JSON_ENCODE({[{key, DbName}, {timestamp, Timestamp},
```
davisp: Lists of key/value pairs that span multiple lines should have one pair per line to improve readability. Also, do we need to consider that this is a different structure than most of our list operations like this? We generally have `key`, `value`, `id`, `doc`, and `error`. I'd think moving `{Info}` to `value` would be straight forward. For `Timestamp` I've got nothing super awesome beyond putting it inside Info in the deleted_dbs_info implementation rather than returning `{Timestamp, Info}`.

src/fabric/src/fabric2_db.erl
```diff
@@ -213,6 +219,26 @@ delete(DbName, Options) ->
     end.
 
 
+deleted_dbs_info(DbName, Options) ->
+    Options1 = lists:keystore(user_ctx, 1, Options, ?ADMIN_CTX),
```
davisp: This should be handled by chttpd.

src/chttpd/src/chttpd_db.erl
davisp: Also, `_undelete` might be a good replacement given the mailing list freedback.

jiangphcn:
thanks Paul @davisp for your review and comments. Could you please check 3d53721 I tried to use to address your comments?

davisp:
Also for the `POST /_deleted_dbs/_undelete` I'll also note that we do have a set of system database names that start with an underscore (`_dbs/_nodes/_users`). Obviously we don't have an `_undelete` database but it is technically mixing that API endpoint into the valid range of database names.

ricellis:
> Obviously we don't have an `_undelete` database but it is technically mixing that API endpoint into the valid range of database names.

Agree that is probably better avoided.

I was thinking that maybe we could keep the functionality in the `_deleted_dbs` namespace by `POST /_deleted_dbs` with something like this:
```json
{
  "undelete": {
    "source": "source_database_name",
    "source_timestamp": "2020-03-24T12:00:00",
    "target": "target_database_name"
  }
}
```
but ultimately I think that probably ends up more opaque than just using differently named endpoint  `POST /_undelete_db` so I'm OK with `_undelete_db`.

@mikerhodes as the initial proposer of the `_deleted_dbs/_restore` pattern are you OK with `POST /_undelete_db`?

Unrelated to the endpoint name the example shows no TZ data in the ISO-8601 timestamp, is that an accurate representation of the output/input we're expecting?

jiangphcn:
> Unrelated to the endpoint name the example shows no TZ data in the ISO-8601 timestamp, is that an accurate representation of the output/input we're expecting?

There are 2 different formats for ISO-8601 timestamp, `1994-11-05T08:15:30-05:00` and `1994-11-05T13:15:30Z`. They all correspond to November 5, 1994, 8:15:30 am, US Eastern Standard Time. The detail can be found at https://www.w3.org/TR/NOTE-datetime. We are using `1994-11-05T13:15:30Z` where TZ information exists as well. 

We can add more description in document to state the format we used here.

mikerhodes:
> Also for the `POST /_deleted_dbs/_undelete` I'll also note that we do have a set of system database names that start with an underscore (`_dbs/_nodes/_users`). Obviously we don't have an `_undelete` database but it is technically mixing that API endpoint into the valid range of database names.

Isn't this the same for adding a `/_undelete_db` into the `/` namespace? In both cases you're basically creating a new reserved name in the list of databases, although there is certainly a lot of prior art for this in `/` so it wouldn't be as surprising and likely to be forgotten later.

While I prefer keeping new `_` endpoints in `/` to a minimum, I can live with having a few new endpoints in the `/` namespace. Before we make the final call, one additional option which keeps things contained to a single new entry in `/` while removing the dbname vs. action ambiguity @davisp mentioned, and keeping things somewhat RESTful, would be to extract the `dbs` bit into a further sub-namespace:

1.    `GET   /_deleted_dbs/dbs`
1.    `GET   /_deleted_dbs/dbs/DbName`
1.    `POST /_deleted_dbs/undelete`

Alternatively you could have `/_deleted/dbs` (etc.) if you wanted to avoid the `dbs` repetition, though I feel that `_deleted` is a bit generic for my tastes.

jiangphcn:
Thanks for more comments. 

It looks to me that `_deleted_dbs` is more clear than `_deleted`.  So the current design becomes

```
    GET /_deleted_dbs/dbs
    GET /_deleted_dbs/dbs/DbName
    POST /_deleted_dbs/undelete
```
Any more comments?

src/fabric/src/fabric2_fdb.erl
```diff
@@ -346,12 +350,68 @@ delete(#{} = Db) ->
...
     DbKey = erlfdb_tuple:pack({?ALL_DBS, DbName}, LayerPrefix),
-    erlfdb:clear(Tx, DbKey),
-    erlfdb:clear_range_startswith(Tx, DbPrefix),
+    DoRecovery = fabric2_util:do_recovery(),
+    case DoRecovery of
+        true ->
+            Timestamp = list_to_binary(fabric2_util:iso8601_timestamp()),
+            DeletedDbKey = erlfdb_tuple:pack({?DELETED_DBS, DbName, Timestamp},
+                LayerPrefix),
+            erlfdb:set(Tx, DeletedDbKey, DbPrefix),
```
davisp: It just occurred to me, we need to read the `DeletedDbKey` and make sure that it comes back `not_found` or else we end up losing references to deleted databases if users create/delete databases too quickly.

jiangphcn: thanks @davisp for bringing up this corner case. I added some lines to make check. Also I made change on API naming. Could you please review again?

davisp:
So, I apparently somehow managed to write an entire comment yesterday and then not post it (or I posted it to a random unrelated ticket that I can't find...)

I really don't like the `_deleted_dbs/dbs` approach. I'd vote for:

* `GET /_deleted_dbs`
* `GET /_deleted_dbs/DbName`
* `POST /_deleted_dbs` with payload:

```json
{
  "undelete": {
    "source": "source_database_name",
    "source_timestamp": "2020-03-24T12:00:00",
    "target": "target_database_name"
  }
}
```

And then I also realized that we really need a "actually delete the deleted thing" API so one or both of:

* `DELETE /_deleted_dbs/DbName?timestamp=2020-03-24T12:00:00`
* `POST /_deleted_dbs` with payload:

```json
{
  "delete": {
    "source": "source_database_name",
    "source_timestamp": "2020-03-24T12:00:00"
  }
}
```

src/fabric/src/fabric2_fdb.erl
```diff
@@ -356,8 +356,13 @@ delete(#{} = Db) ->
...
             DeletedDbKey = erlfdb_tuple:pack({?DELETED_DBS, DbName, Timestamp},
                 LayerPrefix),
-            erlfdb:set(Tx, DeletedDbKey, DbPrefix),
-            erlfdb:clear(Tx, DbKey);
+            case erlfdb:wait(erlfdb:get(Tx, DeletedDbKey)) of
+                not_found ->
+                    erlfdb:set(Tx, DeletedDbKey, DbPrefix),
+                    erlfdb:clear(Tx, DbKey);
+                _Val ->
+                    error(database_was_deleted)
```
davisp: Lets qualify this and also include a better error message:

```erlang
erlang:error({deleted_database_exists, DbName})
```

jiangphcn:
thanks again @davisp for your suggestion. It totally makes sense with

```

    GET /_deleted_dbs
    GET /_deleted_dbs/DbName
    POST /_deleted_dbs with payload:

{
  "undelete": {
    "source": "source_database_name",
    "source_timestamp": "2020-03-24T12:00:00",
    "target": "target_database_name"
  }
}
```
I will wait until today to see whether there is more comment from @mikerhodes and @ricellis. Otherwise, I will change with above naming convention. 

Also, it is nice to have concept of `"actually delete the deleted thing"`.

mikerhodes:
I am good with @davisp's suggestion in https://github.com/apache/couchdb/pull/2666#issuecomment-605109017. I've also thumbs-up'd it to doubly make the point 😄

jiangphcn:
thanks @mikerhodes 
hey Paul @davisp, based on new API design, I changed codes with 
65a88d0 . Could you please review again?

src/fabric/src/fabric2_db.erl
```diff
@@ -17,11 +17,18 @@
     create/2,
     open/2,
     delete/2,
+    deleted_dbs_info/2,
+    undelete/4,
+    delete_deleted/3,
```
davisp: This API is getting a bit too specific I think. How about if we pass a `deleted_at` parameter to `delete/2`. So something like `fabric2_db:delete(DbName, [{deleted_at, <<"2020-03-29T12:00:00">>}])`.

For the `fabric2_fdb` API, if the logic gets too funky with combining into a single function then I'd name it `fabric2_fdb:remove_deleted_db(DbName, TimeStamp)`.

Also, in this export group lets move `deleted_dbs_info/2` down below `list_dbs_info/1,2,3`.

src/fabric/src/fabric2_db.erl
```diff
@@ -178,8 +185,7 @@ create(DbName, Options) ->
 
 
 open(DbName, Options) ->
-    UUID = fabric2_util:get_value(uuid, Options),
-    case fabric2_server:fetch(DbName, UUID) of
```
davisp: You'll need to rebase this PR.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -23,12 +23,17 @@
     ensure_current/1,
     delete/1,
     exists/1,
+    deleted_dbs_info/1,
```
davisp: Need to move this export as well.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -346,12 +351,97 @@ delete(#{} = Db) ->
     } = ensure_current(Db),
 
     DbKey = erlfdb_tuple:pack({?ALL_DBS, DbName}, LayerPrefix),
-    erlfdb:clear(Tx, DbKey),
-    erlfdb:clear_range_startswith(Tx, DbPrefix),
+    DoRecovery = fabric2_util:do_recovery(),
```
davisp: I think splitting this up into three helpers would be the cleanest API. Something like this:

```erlang
delete(Db, Options) ->
    DoRecovery = fabric2_util:do_recovery(),
    DeletedAt = fabric2_util:get_value(deleted_at, Options),

    case {DoRecovery, DeletedAt} of
        {true, undefined} -> soft_delete_db(Db);
        {false, undefined} -> hard_delete_db(Db);
        {_, TimeStamp} -> remove_deleted_db(Db, Timestamp)
    end.


soft_delete_db(Db) ->
    #{
        name := DbName,
        tx := Tx,
        layer_prefix := LayerPrefix
    } = ensure_current(Db),

    DbKey = erlfdb_tuple:pack({?ALL_DBS, DbName}, LayerPrefix)
    Timestamp = list_to_binary(fabric2_util:iso8601_timestamp()),
    DeletedDbKeyTuple = {?DELETED_DBS, DbName, Timestamp},
    DeletedDbKey = erlfdb_tuple:pack(DeletedDbKeyTuple, LayerPrefix),
    case erlfdb:wait(erlfdb:get(Tx, DeletedDbKey)) of
        not_found ->
            erlfdb:set(Tx, DeletedDbKey, DbPrefix),
            erlfdb:clear(Tx, DbKey);
        _Val ->
            erlang:error({deleted_database_exists, DbName})
    end,
    bump_metadata_version(Tx),
    ok.


hard_delete_db(Db) ->
    #{
        name := DbName,
        tx := Tx,
        layer_prefix := LayerPrefix,
        db_prefix := DbPrefix
    } = ensure_current(Db),

    DbKey = erlfdb_tuple:pack({?ALL_DBS, DbName}, LayerPrefix)

    erlfdb:clear(Tx, DbKey),
    erlfdb:clear_range_startswith(Tx, DbPrefix),
    bump_metadata_version(Tx),
    ok.


remove_deleted_db(Db, Timestamp) ->
    #{
        name := DbName,
        tx := Tx,
        layer_prefix := LayerPrefix
    } = ensure_current(Db0, false),

    DeletedDbTupleKey = {?DELETED_DBS, DbName, TimeStamp},
    DeletedDbKey =  erlfdb_tuple:pack(DeletedDbTupleKey, LayerPrefix),
    case erlfdb:wait(erlfdb:get(Tx, DeletedDbKey)) of
        not_found ->
            erlang:error({not_found, missing_timestamp});
        DbPrefix ->
            erlfdb:clear(Tx, DeleteDbKey),
            erlfdb:clear_range_startswith(Tx, DbPrefix)
    end,
    ok.
```

Though the three helpers get moved down into the helper function section.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -346,12 +351,97 @@ delete(#{} = Db) ->
...
 
 
+deleted_dbs_info(#{} = Db0) ->
+    #{
+        name := DbName,
+        tx := Tx,
+        layer_prefix := LayerPrefix
+    } = ensure_current(Db0, false),
+
+    DeletedDbKey =  erlfdb_tuple:pack({?DELETED_DBS, DbName}, LayerPrefix),
```
davisp: Extra space after the`=` operator.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -346,12 +351,97 @@ delete(#{} = Db) ->
...
+            end
+    end.
+
+
+delete_deleted(#{} = Db0, TimeStamp) ->
+    #{
+        name := DbName,
+        tx := Tx,
+        layer_prefix := LayerPrefix
+    } = ensure_current(Db0),
```
davisp: I'm pretty sure this needs to be `ensure_current(Db0, false)`. I'm surprised it works without disabling the db version check.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -346,12 +351,97 @@ delete(#{} = Db) ->
...
+
+    DeletedDbTupleKey = {
+        ?DELETED_DBS,
+        DbName,
+        TimeStamp
+    },
+    DeleteDbKey =  erlfdb_tuple:pack(DeletedDbTupleKey, LayerPrefix),
```
davisp: `DeletedDbKey` to match the variable name in all other places. And there's an extra space after the `=` operator.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -346,12 +351,97 @@ delete(#{} = Db) ->
...
+    case erlfdb:wait(erlfdb:get(Tx, DeleteDbKey)) of
+        not_found ->
+            erlang:error({not_found});
+        DbPrefix ->
+            erlfdb:clear(Tx, DeleteDbKey),
+            erlfdb:clear_range_startswith(Tx, DbPrefix),
+            bump_metadata_version(Tx)
```
davisp: I don't think we need to bump the metadata version here since we're not affecting `_all_dbs`? What do you think, @nickva?

src/fabric/src/fabric2_fdb.erl
```diff
@@ -346,12 +351,97 @@ delete(#{} = Db) ->
...
+            end;
+        false ->
+            erlfdb:clear(Tx, DbKey),
+            erlfdb:clear_range_startswith(Tx, DbPrefix)
+    end,
     bump_metadata_version(Tx),
     ok.
 
 
+deleted_dbs_info(#{} = Db0) ->
```
davisp: Make sure we move this implementation to match the exports order.

src/fabric/test/fabric2_db_crud_tests.erl
```diff
@@ -139,6 +146,44 @@ recreate_db(_) ->
...
+undelete_db(_) ->
+    DbName = ?tempdb(),
+    ?assertError(database_does_not_exist, fabric2_db:delete(DbName, [])),
+
+    ?assertMatch({ok, _}, fabric2_db:create(DbName, [])),
+    ?assertEqual(true, ets:member(fabric2_server, DbName)),
+
+    ok = config:set("couchdb", "enable_database_recovery", "true", false),
+    ?assertEqual(ok, fabric2_db:delete(DbName, [])),
+    ok = config:delete("couchdb", "enable_database_recovery", false),
```
davisp: We can remove these calls undoing the config change now that we have the call in the cleanup (unless you have a test that's making assertions with and without it on of course).

src/fabric/test/fabric2_db_crud_tests.erl
```diff
@@ -139,6 +146,44 @@ recreate_db(_) ->
...
+    ?assertEqual(false, ets:member(fabric2_server, DbName)),
+
+    {ok, [{Timestamp, _Info}]} = fabric2_db:deleted_dbs_info(DbName, []),
```
davisp: For both undelete and delete_deleted we should add tests that assert the not_found error if we give it a bad dbname as well as a valid dbname but invalid timestamp.

src/chttpd/src/chttpd_misc.erl
```diff
@@ -146,6 +147,143 @@ handle_all_dbs_req(#httpd{method='GET'}=Req) ->
...
+            throw({bad_request,
+                <<"POST body must include `source_timestamp` parameter.">>});
+        TimeStamp0 ->
+            TimeStamp0
+    end,
+    case fabric2_db:delete_deleted(DbName, TimeStamp, [{user_ctx, Ctx}]) of
+        ok ->
+            send_json(Req, 200, {[{ok, true}]});
+        {error, file_exists} ->
+            chttpd:send_error(Req, file_exists);
```
davisp: I don't think file_exists is the right error here. I should be not_found, shouldn't it?

src/chttpd/src/chttpd_misc.erl
```diff
@@ -146,6 +147,143 @@ handle_all_dbs_req(#httpd{method='GET'}=Req) ->
...
+    end,
+    TgtDbName = case couch_util:get_value(<<"target">>, JsonProps) of
+        undefined ->  DbName;
+        TgtDbName0 -> TgtDbName0
+    end,
+    case fabric2_db:undelete(DbName, TgtDbName, TimeStamp, [{user_ctx, Ctx}]) of
+        ok ->
+            send_json(Req, 200, {[{ok, true}]});
+        {error, file_exists} ->
+            chttpd:send_error(Req, file_exists);
```
davisp: file_exists for the target database existing is good. But we'll also need a not_found when the source dbname or timestmap are incorrect.

davisp: Also now that I think harder, these should be exceptions instead of returned values. chttpd will catch them and do the proper send_error logic (assuming that we match one of the existing error patterns or add any new ones). So something like:

```erlang
ok = fabric2_db:undelete(DbName, TgtDbName, TimeStamp, [{user_ctx, Ctx}]),
send_json(Req, 200, {[{ok, true}]}).
```

src/chttpd/src/chttpd_misc.erl
```diff
@@ -146,6 +147,143 @@ handle_all_dbs_req(#httpd{method='GET'}=Req) ->
 handle_all_dbs_req(Req) ->
     send_method_not_allowed(Req, "GET,HEAD").
 
+handle_deleted_dbs_req(#httpd{path_parts=[<<"_deleted_dbs">>]}=Req) ->
```
davisp: The combination of function names here is a bit difficult. I'd change this around to move function implementations to this top level function. Something like:

```erlang
handle_deleted_dbs_req(#httpd{method = 'GET', path_parts = [_]}) ->
    ... body of deleted_dbs_req with method='GET' ...;
handle_deleted_dbs_req(#httpd{method = 'GET', path_parts = [_, DbName]) ->
    ... body of deleted_dbs_info_req ...;
handle_deleted_dbs_req(#httpd{method = 'POST', path_parts = [_]}) ->
    ... body of deleted_dbs_post_req ...;
handle_deleted_dbs_req(#httpd{path_parts = PP}) when length(PP) =< 2 ->
    send_method_not_allowed("GET,HEAD,POST");
handle_deleted_dbs_req(Req) ->
    chttpd:send_error({error, not_found}).
```

And to be clear, I'd leave handle_undelete/delete_deleted as separate functions (though rename delete_deleted to remove_deleted or similar to match fabric2_db API names.

davisp: Also note you don't have to re-assert `_deleted_dbs` in the URL path as that's defined in the URL handlers bits.

src/fabric/src/fabric2_fdb.erl
nickva: I think we're ok not bumping it here if we guarantee (let's have a test for it) that any soft deleted db handle will reopen. For that we need to bump the db version when we soft-delete and ensure that to hard deletion (snapshot removal) always goes through the soft-deletion path first.

davisp: Ooooh, good call on db versions.

We should call `bump_db_version` after soft deletion and undeletion so that any inflight requests will fail properly.

I don't think hard deletion or snapshot deletion require that check since clearing the key would create a write conflict, no need to change it just to delete it.

src/chttpd/src/chttpd_misc.erl
jiangphcn: Similar to `create_db_req` https://github.com/apache/couchdb/blob/prototype/fdb-layer/src/chttpd/src/chttpd_db.erl#L386-L387, we use returned values.

src/fabric/src/fabric2_fdb.erl
jiangphcn: I split `delete` into `soft_delete_db/1` and `hard_delete_db/1`. For `remove_deleted_db`, it is a separate call from `fabric2_db:delete/2`.

```
delete(#{} = Db) ->
    DoRecovery = fabric2_util:do_recovery(),
    case DoRecovery of
        true -> soft_delete_db(Db);
        false -> hard_delete_db(Db)
    end.
```

src/fabric/src/fabric2_fdb.erl
nickva: Thinking more about tests I think we'd want at least for these scenarios:

Cases where old db handle can't read the data anymore:
  - db soft deleted
  - db hard deleted
  - db soft deleted and re-created
  - db soft deleted and undeleted
  - db hard deleted and re-created 
  
Indexing stops when:
   - db soft deleted
   - db hard deleted
   - db soft deleted and re-created
   - db soft deleted and undeleted
   - db hard deleted and re-created

Obviously indexing will stop if the first set of tests pass, however we want to make sure the way indexing detects that db is now deleted hasn't changed and it reacts the same as before. Look for db re-created or rename tests in recent commits and I think we could modify those to take soft deletion into consideration.

If we are using timestamps with 1 second resolution also make sure we handle cases were we soft-delete/undelete multiple times in the same second, and also tests where we do it across a few different seconds.

nickva: Actually I was wrong about the `db soft deleted and undeleted`.  Thanks, @jiangphcn for pointing it out. It seems in that case the old handle will be usable because:

1) we bumped the db version only
2) old handle would notice metadata and db version being out of date
3) it would automatically re-open
4) it would continue to be used

We could assert that in the test perhaps then. Good catch @jiangphcn !

jiangphcn: Great suggestion @nickva on test case. I added it in https://github.com/apache/couchdb/pull/2666/commits/b1d2b08afb80ca9f8d8f4b9ca37ad65cd54b3352#diff-5c0948df6382504d34bd12ee0d4485d0R197-R250.

jiangphcn:
@davisp and @nickva could you please check whether there are other comments left?

src/fabric/src/fabric2_fdb.erl
davisp: Oh, for after creation. For that "use after deletion" we should make sure that between deletion and undeletion that the handle isn't usable as well.

src/chttpd/test/eunit/chttpd_deleted_dbs_test.erl
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }.
+
+
+should_return_error_for_deleted_dbs(Url) ->
+    ?_test(begin
+        create_and_delete_db(Url),
+        NewDoc = "{\"keys\": [\"db1\"]}",
+        {ok, Code, _, ResultBody} = test_request:post(Url ++ "/_deleted_dbs/",
+            [?CONTENT_JSON, ?AUTH], NewDoc),
```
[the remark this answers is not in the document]
jiangphcn: Originally, `POST` is not supported for `/_deleted_dbs` endpoint, so I added one test to check unsupported method. Now I use `DELETE` instead.

src/chttpd/test/eunit/chttpd_deleted_dbs_test.erl
```diff
@@ -0,0 +1,206 @@
...
+
+should_list_deleted_dbs(Url) ->
+    ?_test(begin
+        DbName1 = create_and_delete_db(Url),
+        DbName2 = create_and_delete_db(Url),
+        {ok, _, _, ResultBody} = test_request:get(Url ++ "/_deleted_dbs/",
+            [?CONTENT_JSON, ?AUTH]),
+        BodyJson = jiffy:decode(ResultBody),
+        ?assertEqual(true, lists:member(DbName1, BodyJson)),
+        ?assertEqual(true, lists:member(DbName2, BodyJson))
```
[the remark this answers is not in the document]
jiangphcn: Okay, I added DBInfo for `GET _deleted_dbs` endpoint, like we did for `_deleted_dbs/DbName` endpoint.  We turn on database recovery here https://github.com/apache/couchdb/pull/2666/files#diff-3fdd5966871338636ff72ba49490942fR223

jiangphcn:
Just for simplification,  the _deleted_dbs wasn't returning DbInfo blobs directly in the past. In order to get DbInfo blobs, need to call `GET _deleted_dbs/{db}/`. But with new commit 031cb85, DbInfo Blobs is added as well for `GET _deleted_dbs`.

src/fabric/src/fabric2_db.erl
```diff
@@ -204,12 +211,37 @@ delete(DbName, Options) ->
...
+            Resp = fabric2_fdb:transactional(Db, fun(TxDb) ->
+                fabric2_fdb:delete(TxDb)
+            end),
+            if Resp /= ok -> Resp; true ->
+                fabric2_server:remove(DbName)
+            end
+    end.
+
+
+deleted_dbs_info(DbName, Options) ->
```
davisp, line 230: This should be moved to after list_dbs_info so that it matches the order in the exports list.

src/fabric/src/fabric2_db.erl
```diff
@@ -1025,6 +1101,51 @@ drain_all_info_futures(FutureQ, UserFun, Acc) ->
...
+            {compact_running, false},
+            {data_size, 0},
+            {db_name, DbName},
+            {timestamp, TimeStamp},
+            {disk_format_version, 0},
+            {disk_size, 0},
+            {instance_start_time, <<"0">>},
+            {purge_seq, 0}
+            ]}
+        }
```
davisp, line 1119: Super minor nit, but this should be a single line `]}}` at two indents.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -365,6 +364,65 @@ exists(#{name := DbName} = Db) when is_binary(DbName) ->
...
+        not_found ->
+            DeletedDbTupleKey = {
+                ?DELETED_DBS,
+                DbName,
+                TimeStamp
+            },
+            DeleteDbKey = erlfdb_tuple:pack(DeletedDbTupleKey, LayerPrefix),
+            case erlfdb:wait(erlfdb:get(Tx, DeleteDbKey)) of
+                not_found ->
+                    erlang:error({not_found});
```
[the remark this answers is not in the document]
davisp: You resolved this but you don't appear to have addressed the issue.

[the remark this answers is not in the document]
davisp: Also a note that if you pick the exceptions that we shouldn't be raising `{not_found}` as a 1-tuple. It should be either `not_found` or `{not_found, missing}` to match the existing exceptions in other places.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -1172,6 +1261,45 @@ check_db_version(#{} = Db, CheckDbVersion) ->
...
+    Timestamp = list_to_binary(fabric2_util:iso8601_timestamp()),
+    DeletedDbKeyTuple = {?DELETED_DBS, DbName, Timestamp},
+    DeletedDbKey = erlfdb_tuple:pack(DeletedDbKeyTuple, LayerPrefix),
+    case erlfdb:wait(erlfdb:get(Tx, DeletedDbKey)) of
+        not_found ->
+            erlfdb:set(Tx, DeletedDbKey, DbPrefix),
+            erlfdb:clear(Tx, DbKey),
+            bump_db_version(Db);
+        _Val ->
+            erlang:error({deleted_database_exists, DbName})
```
davisp, line 1282: This error message isn't quite right since we can have multiple deleted databases, just not at the same timestamp. How about `{deletion_frequency_exceeded, DbName}`?

src/fabric/src/fabric2_util.erl
```diff
@@ -40,7 +40,9 @@
     encode_all_doc_key/1,
 
     pmap/2,
-    pmap/3
+    pmap/3,
+    iso8601_timestamp/0,
+    do_recovery/0
```
davisp, line 45: Super minor nit but lest move these above `pmap/N` into their own section. Don't forget to move the implementations as well.

davisp:
Also, a couple quick thoughts on the HTTP API. Looking at it a bit harder I realized that we've basically duplicated some functionality because these two requests should be functionally equivalent:

```
GET /_deleted_dbs/DbName`
GET /_deleted_dbs?key=DbName`
```

And I could see a convincing argument that the second form would be more "natural" within the existing APIs for views and so on.

I've also heard some concerns that having the `undelete` vs `delete` semantics of our `POST /_deleted_dbs` setup makes auditing traffic hard/impossible without introspecting the request body which seems like a legitimate critique. How about if we tweaked things to be like such:

* `GET /_deleted_dbs` - List deleted dbs infos as currently.
* `POST /_deleted_dbs` - Only accepts the `undelete` JSON body
* `DELETE /_deleted_dbs/DbName?timestamp=2020-01-01T00:00:00` - Remove deleted db

/cc @mikerhodes @ricellis @nickva

davisp:
I wasn't super fond of the chttpd tests for deleted dbs but not in a way specific enough to review very well. So I just took a quick pass through reformatting/styling etc on what looks better to me. Feel free to completely ignore this though as its purely style related. But in case you're interested, this is how I'd change things so that there's a lot less noise in that test module:

https://gist.github.com/davisp/257ea1631d443031f5c3995b70093bbe

davisp:
Also, something I noticed while playing with the tests is that the `GET /_deleted_dbs/DbName` is not very ergonomic. Due to how things are implemented, if we don't have any deleted snapshots we still return a `200` response with an empty array `[]` as the body. I would have expected a `404` in that situation given that the `DbName` is part of the URL. I realize that's purely due to how things are implemented, but it suggests to me that the `GET /_deleted_dbs?key=DbName` approach is better since we'd be more likely to understand that a `200` and empty array returned is expected since its more like a view query where results might or might not exist.

jaydoane:
> I've also heard some concerns that having the `undelete` vs `delete` semantics of our `POST /_deleted_dbs` setup makes auditing traffic hard/impossible without introspecting the request body which seems like a legitimate critique. How about if we tweaked things to be like such:
> 
>     * `GET /_deleted_dbs` - List deleted dbs infos as currently.
> 
>     * `POST /_deleted_dbs` - Only accepts the `undelete` JSON body
> 
>     * `DELETE /_deleted_dbs/DbName?timestamp=2020-01-01T00:00:00` - Remove deleted db

I very much like this particular API, since it's trivial to derive  three separate action names from method and path, unlike earlier versions. The corresponding names (eliding the ubiquitous `cloudantnosqldb.` prefix), I'm considering:

- `account-deleted-dbs.read`
- `account-deleted-dbs.undelete`
- `account-deleted-dbs.delete`

mikerhodes:
- The main problem I can see for `GET /_deleted_dbs?key=DbName` is that we lose the mirror with including the database name like we have for `DELETE /_deleted_dbs/DbName?timestamp=2020-01-01T00:00:00`.
- I also wondered whether the `timestamp` is required for the `DELETE`? If so, passing it as part of the path would be more natural (which I'm aware was one of the original suggestions I wasn't so keen on :grin:). 
- If we look to `_all_dbs` as a model (I pondered this as we're looking at databases and the API is much simpler to copy), this isn't correct:

     ```
     GET /_deleted_dbs/DbName
     GET /_deleted_dbs?key=DbName
     ```

    Because these return different values:

     ```
     GET /DbName
     GET /_all_dbs?key=DbName
     ```

     So there's a question here as to whether `/_deleted_dbs` is more like a `_all_dbs` or `_all_docs`? Are we talking databases or some other type of entity, that `/_deleted_dbs` is more like talking about a set of metadata about deleted databases rather than teh databases themselves.


A quick proposal with responses based on `_all_dbs` that is as REST as I could make it, with all the lessons from the discussion above tried to be taken onboard:

https://gist.github.com/mikerhodes/39c66576eee468a5aeaa23c5715aec9a

I think this could be trivially adapted to make the responses more `_all_docs`/view like if we wanted using the same paths.

I guess we have `_dbs_info` as a guide too, though that doesn't tell us what `/_deleted_dbs/DbName` should be...

jiangphcn:
Thanks @davisp and @mikerhodes for your comments. Today I have one commit eefd115 which essentially implemented below API spec.

```
GET /_deleted_dbs - List deleted dbs infos
GET /_deleted_dbs?key=dbName - List specified deleted db infos
POST /_deleted_dbs - Only accepts the undelete JSON body to *restore* the deleted db
DELETE /_deleted_dbs/DbName?timestamp=2020-01-01T00:00:00 - Remove deleted db
```
From the functional set exposed to users, I think that it is similar to what @mikerhodes drafted in [1]. The difference is the API spec. 

From user case and implementation's point of view, there are several points I want to mention.
1. GET /_deleted_dbs - List deleted dbs infos
We directly return dbs info instead of the deleted dbname here. The same case is for `GET /_deleted_dbs?key=dbName`. This is mainly following the approach in `_dbs_info`. I think that most likely user is not only interested in dbname, but also in some detailed information about deleted database. 
```
  {
    "key": "db01",
    "timestamp": "2020-04-03T12:09:15Z",
    "value": {
      "update_seq": "000000000000000000000000",
      "doc_del_count": 0,
      "doc_count": 0,
      "sizes": {
        "external": 2,
        "views": 0
      }
    }
```
User can gradually use the progressive approach like mentioned in [1] to get the detailed information of the deleted db.
```
GET /_deleted_dbs
GET /_deleted_dbs/DbName
GET /_deleted_dbs/DbName/2020-01-01T00:00:00
```
However, looks that user most like should review information of the deletion time, doc number, disk size to decide to restore it or permanently delete it.

2. with regard to 
```
GET /_deleted_dbs/DbName
GET /_deleted_dbs?key=DbName
```
In theory, we can implement both of them to get information of deleted db. @davisp also found one situation where we have to return 200 and [] if the specified db is not in deleted_db list when selecting `GET /_deleted_dbs/DbName`. This is not expected error code. So we switch to `GET /_deleted_dbs?key=DbName`.

Hope this can give some explanation how we are getting to the current status of API spec.  

[1] https://gist.github.com/mikerhodes/39c66576eee468a5aeaa23c5715aec9a

src/fabric/src/fabric2_fdb.erl
```diff
@@ -365,6 +364,65 @@ exists(#{name := DbName} = Db) when is_binary(DbName) ->
...
+
+    DeletedDbTupleKey = {
+        ?DELETED_DBS,
+        DbName,
+        TimeStamp
+    },
+    DeletedDbKey = erlfdb_tuple:pack(DeletedDbTupleKey, LayerPrefix),
+    case erlfdb:wait(erlfdb:get(Tx, DeletedDbKey)) of
+        not_found ->
+            erlang:error({not_found});
```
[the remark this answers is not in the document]
jiangphcn: I made change locally and didn't `git push` for this. The change was reflected in eefd115 as well.

ricellis:
> `POST /_deleted_dbs` - Only accepts the undelete JSON body to *restore* the deleted db

Can I see an example of the expected body now? If `undelete` is the only "action" for `POST` is there any reason to call it out separately?

jiangphcn:
> POST /_deleted_dbs - Only accepts the undelete JSON body to restore the deleted db

> Can I see an example of the expected body now? If undelete is the only "action" for POST is there any reason to call it out separately?

Yes, below is one example.
{
        "undelete": {
                "source_timestamp": "2020-04-01T07:13:42Z",
                "source": "db01"
        }
}

ricellis:
OK, so now the `POST` can only do an undelete couldn't the body just be:
```
{
"source_timestamp": "2020-04-01T07:13:42Z",
"source": "db01",
}
```
That also removes the awkward naming around `undelete` that I know was bothering some people.

Also given
`DELETE /_deleted_dbs/{dbName}`
shouldn't we also
`POST /_deleted_dbs/{dbName}`
since they are operating on the same resource?

Then the body would be just (or also with an optional `target`):
```
{
"source_timestamp": "2020-04-01T07:13:42Z"
}
```

Although I think the parameter name in the body should match that used on the `DELETE` request as well; either both `timestamp` or both `source_timestamp`.

It seems like it would be nice to use paths consistently throughout as [@mikerhodes proposed](https://gist.github.com/mikerhodes/39c66576eee468a5aeaa23c5715aec9a), but I can see the rationale for preferring to use `key=dbname` to be consistent with the view-like APIs.

davisp:
So, apparently I'm the only one that's been assuming that this is a mirror of the `GET _dbs_info` API end point and working from that perspective.

@mikerhodes I don't think the `_all_dbs` approach is very useful. When would you ever want just a list of database names that have soft deletions or just a list of timestamps? Its RESTy in all the bad ways that lead us to adding another API endpoint that dumps all of that information in a single response like we already had to do for `GET|POST /_dbs_info`.

@ricellis If we stay with the `POST _deleted_dbs` then I don't think its a good idea to remove the `"undelete"` key as the API loses its self-documentation aspect. Plus we'd be hamstrung adding other operations in the future (not that I can think of any). If we want to revert to `POST _undelete` then I'd be fine dropping the JSON key.

Consistency in the `"timestamp"` key is a good idea. I'd vote for `"timestamp"` since there's no target timestamp to disambiguate and `DELETE /_deleted_dbs/DbName?source_timestamp=foo` reads awkwardly.

@jiangphcn I'm can't decide whether you're suggesting we support the `GET /_deleted_dbs/DbName/Timestamp` API or not. I don't think we should add alternative APIs for things that can be accomplished with a single view based API so I would vote against adding them. But you might also saying we don't need them, in which case we're in agreement and you can ignore me.

jiangphcn:
hey @davisp 

I think that you are *not* the only one that's been assuming that this is a mirror of the GET _dbs_info API end point and working from that perspective.  From https://github.com/apache/couchdb/pull/2666/#issuecomment-608411228, you can see that I am for that and implemented with this approach.

In addition, I don't support the suggestion about `the GET /_deleted_dbs/DbName/Timestamp API`. As i mentioned before, it looks that user most likely should review information of the deletion time, doc number, disk size to decide to restore it or permanently delete it. It is convenient to return DbInfo one time instead of using the progressive approach like mentioned in [1] to get the detailed information of the deleted db.
```
GET /_deleted_dbs
GET /_deleted_dbs/DbName
GET /_deleted_dbs/DbName/2020-01-01T00:00:00
```

mikerhodes:
@davisp @jiangphcn Apologies if I missed that we were mirroring `_dbs_info` in our API. That makes things a bit clearer around use of `key`. Flipping between internal structures (`DbInfo`) and external APIs (`/_dbs_info`) probably got in the way of my understanding.

I didn't particularly think `_all_dbs` was that useful, but it seemed like we were not following any particular pattern so I just picked the one that required me to type fewest characters to demonstrate the idea (an idea which doesn't make much sense now I understand the motivation behind the older API, which I'm not sure anyone really spelled out in the mailing list or here before now, but I probably missed something somewhere 😞 ).

Anyway, I think we've reached a decent stopping point here (given that I think mostly I've just got everyone to justify an already good solution tbh rather than moving us forward), so let's move forward and not make @jiangphcn implement yet another thing 😉 Nice work all in all.

jiangphcn:
thanks @davisp and @mikerhodes for comments over weekend. I come with up one new commit 
527d281 where mainly covers the suggestion from Paul about code refactor. One thing is related to combine implementation of `/_deleted_dbs` and `/_deleted_dbs?key`. This mainly follows the way of `_dbs_info` while there is slight difference for `/_deleted_dbs?key` because the key in key/value pair for `_deleted_dbs` is not only `dbname`, but also includes `timestamp`. So i introduce the way of https://github.com/apache/couchdb/pull/2666/commits/527d2817e296a30439c6a5337dc0794adf2eeec0#diff-20d0112d14281f71e2bd09f75da80f27R286-R295 to find dbinfo for specified key.

jiangphcn:
## Soft delete

Instead of automatically and immediately removing the data and index in the database after a delete operation, soft delete restores the deleted data back to the original state after an accidental delete or undesired delete operation, up to a defined period, such as 48 hours.

### `GET /_deleted_dbs`

Send a `GET` request to find a list of all the deleted databases in the {{site.data.keyword.cloudant_short_notm}} instance. 

*Request Headers*:

```sh    	
Accept –

    application/json
	text/plain
```
{: codeblock}

| Query parameters | Description |
|------------------|-------------|     	
| `descending` (boolean) | Return the databases in descending order by key. Default is false. |
| `endkey` (json) | Stop returning databases when the specified key is reached. |
| `end_key` (json) | Alias for `endkey` parameter. |
| `key` (json) | Return the databases for the specified key. It can be database name or database name + timestamp.|
| `limit` (number) | Limit the number of the returned databases to the specified number. |
| `skip` (number) | Skip this number of databases before starting to return the results. Default is 0. |
| `startkey` (json) | Return databases starting with the specified key. |
| `start_key` (json) | Alias for `startkey`. |

*Response Headers*:
```sh     	
Content-Type –
    application/json
    text/plain; charset=utf-8
```
{: codeblock}

| Code | Description |	
|------|-------------|
| `200 OK` | Request completed successfully. |
| `400 Bad Request` | Bad Request. Invalid payload in request. |
| `401 Unauthorized` | CouchDB Server Administrator privileges required. |
| `404 NotFound` | Invalid deleted database or timestamp. |

See the following example request:

```sh
GET /_deleted_dbs HTTP/1.1
Accept: application/json
Host: localhost:5984
```
{: codeblock}

See the following example response: 

```sh
HTTP/1.1 200 OK
Content-Length: 52
Date: Tue, 23 Mar 2020 06:57:48 GMT
Server: CouchDB (Erlang/OTP)

[
  {
    "cluster": {
      "n": 0,
      "q": 0,
      "r": 0,
      "w": 0
    },
    "compact_running": false,
    "data_size": 193164408719,
    "db_name": "invoices",
    "disk_format_version": 0,
    "disk_size": 46114703224,
    "instance_start_time": "0",
    "purge_seq": 0,
    "update_seq": "981...uUQ",
    "doc_del_count": 5564,
    "doc_count": 9818541,
    "sizes": {
      "external": 193164408719,
      "views": 34961621142
    },
    "deleted": true,
    "timestamp": "2020-04-07T23:19:12Z"
  },
  {
    "cluster": {
      "n": 0,
      "q": 0,
      "r": 0,
      "w": 0
    },
    "compact_running": false,
    "data_size": 0,
    "db_name": "contacts",
    "disk_format_version": 0,
    "disk_size": 0,
    "instance_start_time": "0",
    "purge_seq": 0,
    "update_seq": "000000000000000000000000",
    "doc_del_count": 0,
    "doc_count": 0,
    "sizes": {
      "external": 2,
      "views": 0
    },
    "deleted": true,
    "timestamp": "2020-04-07T23:19:02Z"
  },
]


```
{: codeblock}

### `POST /_deleted_dbs`
{: #post-_deleted_dbs_undelete}

Send a `POST` request to restore a deleted database.

Parameters:
    None

Request Headers:

```sh
Accept –
    application/json
    text/plain

{
    "undelete": {
        "timestamp" : "2020-03-23T02:26:35Z",
        "source" : "db01",
        "target" : "db02"
    }
}
-  "timestamp" - timestamp when database was deleted
-  "target" could be optional and default to the source database name.
```
{: codeblock}

Response Headers:
```sh
Content-Type –
    application/json
    text/plain; charset=utf-8
```
{: codeblock}

| Response JSON object | Description |
|----------------------|-------------|
| `ok` (boolean) | Operation status. Available in case of success. |
| `error` (string) | Error type. Available if response code is 4xx. |
| `reason` (string) | Error description. Available if response code is 4xx. |

| Code | Message |
|--------------|---------|
| `200 Undeleted` | Database undeleted successfully. |
| `400 Bad Request` | Bad Request. Invalid payload in request. |
| `401 Unauthorized` | CouchDB Server Administrator privileges required. |
| `404 NotFound` | Invalid deleted timestamp. |
| `412 Precondition Failed` | Database already exists. |

### `DELETE /_deleted_dbs/{db}`
{: #get-_deleted_dbs-db}

Send a `DELETE` request to permanently delete the database instance which was soft-deleted with the specified timestamp.

| Parameters | Description |
|------------|-------------|
| `timestamp` | Timestamp when database was deleted |

Request Headers:
```sh
Content-Type –
    application/json
```
{: codeblock}

Response Headers:
```sh
Content-Type –
    application/json
    text/plain; charset=utf-8
```
{: codeblock}

| Response JSON object | Description |
|----------------------|-------------|
| `ok` (boolean) | Operation status. Available in case of success. |
| `error` (string) | Error type. Available if response code is 4xx. |
| `reason` (string) | Error description. Available if response code is 4xx. |

| Code | Message | 
|--------------|---------|
| `200 Deleted` | Database permanently deleted successfully. |
| `400 Bad Request` | Bad Request. Invalid payload in request. |
| `401 Unauthorized` | CouchDB Server Administrator privileges required. |
| `404 NotFound` | Invalid deleted database or timestamp. |

PR #2668, opened 2020-03-16
Throw if an unknown check is passed to jwtf:decode

rnewson:
## Overview

Throw if an unknown check is passed to jwtf:decode

## Testing recommendations

N/A

https://github.com/apache/couchdb/pull/2648

## Checklist

### DIFF
--- a/src/jwtf/src/jwtf.erl
+++ b/src/jwtf/src/jwtf.erl
@@ -35,6 +35,16 @@
     {<<"HS384">>, {hmac, sha384}},
     {<<"HS512">>, {hmac, sha512}}]).
 
+-define(CHECKS, [
+    alg,
+    exp,
+    iat,
+    iss,
+    kid,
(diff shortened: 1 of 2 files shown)


src/jwtf/src/jwtf.erl
```diff
@@ -112,6 +113,15 @@ validate(Header0, Payload0, Signature, Checks, KS) ->
     Key = key(Header1, Checks, KS),
     verify(Alg, Header0, Payload0, Signature, Key).
 
+validate_checks(Checks) when is_list(Checks) ->
+    ValidChecks = [alg, exp, iat, iss, kid, nbf, sig, typ],
```
jaydoane: I would prefer to see these defined at the top of the file along with `ALGS`.

rnewson: definitely.

PR #2672, opened 2020-03-16
View is partitioned if db and view are partitioned

rnewson:
## Overview

We've seen a crash if DbPartitioned is false and ViewPartitioned is
true, which is obviously nonsense. The effect of the `nocase` is the
termination of the couch_index_server gen_server, which is a serious
amplification of a small (user-initiated) oddity.

## Testing recommendations

Difficult to induce. The protection that prevents us creating a ddoc with partitioned:true if the
db is partitioned:false must have a gap (perhaps ?new_edits=false / replicator path?)

N/A

## Checklist


rnewson:
A more plausible path is that the shard was created by a create_if_missing call that did not set the partitioned:true property. So db creation was fine, but this db was in maintenance mode, and so got its .couch file via mem3 internals.

rnewson:
pretty sure that's it. node was in mm, so did not participate in the fabric rpc call that created the shard with the partitioned setting. data was then added to the db and mem3 internal replication pushed to the mm'ed node. That does a create_if_missing without setting the partitioned flag.

Oops.

rnewson:
(this PR will not fix the root cause just identified, it only prevents couch_index_server crashing when it does).

rnewson:
(This PR must not be merged until a fix for the root cause is added to it, presumably in mem3).

davisp:
For the underlying bit, get_or_create should probably check the dbs db for creation options. If its not got the doc replicated yet then freak out and not create anything etc etc.

chewbranca:
Here's a WIP PR for addressing the underlying db creation issues: https://github.com/apache/couchdb/pull/2690

wohali:
Shouldn't this and #2690 target the `3.x` branch, from which our next release will be cut?

PR #2685, opened 2020-03-20
[WIP] Implement per database encryption for primary data

eiri:
# Overview

This PR adds an option for enabling encryption for primary data. The encryption done with aes-256-gcm algorithm per document with document keys (Data Encryption Keys) derived from an individual database's key (Key Encryption Key). KEK supplied at time of the database creation and stored along with the rest of its configuration in encrypted "wrapped" form.

The encoding and decoding of documents handled in a dedicated module that also acts as a cache of "unwrapped" KEKs. The general management of KEKs implemented through epi plugin interface to allow a possibility for implementation of interfaces to alternative Key Management Services, e.g. Hashicorps Vault or SecretHub.

The default epi plugin generates KEK at a databae creation time and encrypts it with a Master Encryption Key specified in CouchDB configuration.

## Testing recommendations

Make target `make check-fdb` should pass

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -707,3 +707,18 @@ opts = #{budget => 100, target => 2500, window => 60000, sensitivity => 1000}
 
 [encryption]
 enabled = false
+;
+; To generate master key and initialization vector run the following command
+; ("secret" is an example passphrase here, use something else)
+;
+; `openssl enc -aes-256-ctr -k secret -P -md sha1`
+; output:
+;   salt=FC0D0243C5126FB5
+;   key=9B43A7711CDDE41FE065FC03A14BBD6A177CBD6A4B474A05DEC9798C79B98045
+;   iv =5B3C6478BBC698AAA8CA5BA51DB8FF95
+; Put key in "key.dat" file excluding "key" characters and a carriage return
+; Put iv in "iv.dat" file excluding "iv" characters and a carriage return
+;
+; Keep both files as read-only and owned by couch process.
+; key_file = /var/secured/mount/location/key.dat
+; iv_file = /var/secured/mount/location/iv.dat
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -708,17 +708,10 @@ opts = #{budget => 100, target => 2500, window => 60000, sensitivity => 1000}
 [encryption]
 enabled = false
 ;
-; To generate master key and initialization vector run the following command
-; ("secret" is an example passphrase here, use something else)
+; A key provider is an external script, that on call prints master key
+; in its binary form to standartd output. The master key must be 128, 192
+; or 256 (preferred) bites long. The master key cached in memory
+; for the length of CouchDB run, i.e. it key provider script executed
+; only once.
 ;
-; `openssl enc -aes-256-ctr -k secret -P -md sha1`
-; output:
-;   salt=FC0D0243C5126FB5
-;   key=9B43A7711CDDE41FE065FC03A14BBD6A177CBD6A4B474A05DEC9798C79B98045
-;   iv =5B3C6478BBC698AAA8CA5BA51DB8FF95
-; Put key in "key.dat" file excluding "key" characters and a carriage return
-; Put iv in "iv.dat" file excluding "iv" characters and a carriage return
-;
-; Keep both files as read-only and owned by couch process.
-; key_file = /var/secured/mount/location/key.dat
-; iv_file = /var/secured/mount/location/iv.dat
+;key_provider = /usr/local/bin/decrypt_master_key
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -704,3 +704,6 @@ compaction = false
 [couch_rate.views]
 limiter = couch_rate_limiter
 opts = #{budget => 100, target => 2500, window => 60000, sensitivity => 1000}
+
+[encryption]
+enabled = false
--- a/src/fabric/src/fabric2_encryption.erl
+++ b/src/fabric/src/fabric2_encryption.erl
@@ -83,10 +83,11 @@ terminate(_, _St) ->
 handle_call({encode, DbName, DocId, DocRev, DocBody}, From, St) ->
     #{
         iid := InstanceId,
+        cache := Cache,
         waiters := Waiters
     } = St,
 
-    {ok, KEK} = get_kek(DbName),
+    {ok, KEK} = get_kek(Cache, DbName),
     {Pid, _Ref} = erlang:spawn_monitor(?MODULE,
         do_encode, [KEK, InstanceId, DbName, DocId, DocRev, DocBody]),
 
@@ -98,10 +99,11 @@ handle_call({encode, DbName, DocId, DocRev, DocBody}, From, St) ->
 handle_call({decode, DbName, DocId, DocRev, Encoded}, From, St) ->
     #{
         iid := InstanceId,
+        cache := Cache,
         waiters := Waiters
     } = St,
 
-    {ok, KEK} = get_kek(DbName),
+    {ok, KEK} = get_kek(Cache, DbName),
     {Pid, _Ref} = erlang:spawn_monitor(?MODULE,
         do_decode, [KEK, InstanceId, DbName, DocId, DocRev, Encoded]),
 
@@ -138,8 +140,10 @@ code_change(_OldVsn, St, _Extra) ->
 
 init_st() ->
     FdbDirs = fabric2_server:fdb_directory(),
+    Cache = ets:new(?MODULE, [set, private, compressed]),
     {ok, #{
         iid => iolist_to_binary(FdbDirs),
+        cache => Cache,
         waiters => dict:new()
     }}.
 
@@ -187,6 +191,12 @@ get_dek(KEK, DocId, DocRev) when bit_size(KEK) == 256 ->
     {ok, DEK}.
 
 
-get_kek(DbName) ->
-    KEK = crypto:hash(sha256, DbName),
-    {ok, KEK}.
+get_kek(Cache, DbName) ->
+    case ets:lookup(Cache, DbName) of
+        [{DbName, KEK}] ->
+            {ok, KEK};
+        [] ->
+            KEK = crypto:hash(sha256, DbName),
+            true = ets:insert(Cache, {DbName, KEK}),
+            {ok, KEK}
+    end.
--- a/src/fabric/src/fabric2_encryption.erl
+++ b/src/fabric/src/fabric2_encryption.erl
@@ -17,8 +17,9 @@
 
 -export([
     start_link/0,
-    encode/4,
-    decode/4
+    get_wrapped_kek/1,
+    encode/5,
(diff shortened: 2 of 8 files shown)


eiri:
@davisp @rnewson This PR is still work in progress, but can you please take a look if general shape of things seems fine to you?

rel/overlay/etc/default.ini
```diff
@@ -699,3 +699,21 @@ compaction = false
...
+; output:
+;   salt=FC0D0243C5126FB5
+;   key=9B43A7711CDDE41FE065FC03A14BBD6A177CBD6A4B474A05DEC9798C79B98045
+;   iv =5B3C6478BBC698AAA8CA5BA51DB8FF95
+; Put key in "key.dat" file excluding "key" characters and a carriage return
+; Put iv in "iv.dat" file excluding "iv" characters and a carriage return
+;
+; Keep both files as read-only and owned by couch process.
+; key_file = /var/secured/mount/location/key.dat
+; iv_file = /var/secured/mount/location/iv.dat
```
rnewson: 1) not sure why these are external and not inline given the other files are readable by the same user.
2) storing an IV here makes no immediate sense to me so I worry we have a crypto problem to come.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -808,10 +816,11 @@ write_doc(#{} = Db0, Doc, NewWinner0, OldWinner, ToUpdate, ToRemove) ->
             ok
     end,
 
-    % Update database size
+    % Update database size and db's update counter
     AddSize = sum_add_rev_sizes([NewWinner | ToUpdate]),
     RemSize = sum_rem_rev_sizes(ToRemove),
     incr_stat(Db, <<"sizes">>, <<"external">>, AddSize - RemSize),
+    incr_stat(Db, <<"update_count">>, 1),
```
rnewson: This increment will only happen if the transaction succeeds.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -1328,7 +1339,20 @@ doc_to_fdb(Db, #doc{} = Doc) ->
...
     DiskAtts = lists:map(fun couch_att:to_disk_term/1, Atts),
 
-    Value = term_to_binary({Body, DiskAtts, Deleted}, [{minor_version, 1}]),
+    ValueTerm = case WrappedKEK of
+        false ->
+            {Body, DiskAtts, Deleted};
+        _ ->
+            UpdateCounter = get_stat(Db, <<"update_count">>),
+            BinBody = term_to_binary(Body,
+                [{compressed, 0}, {minor_version, 1}]),
```
rnewson: please compress the body.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -1328,7 +1339,20 @@ doc_to_fdb(Db, #doc{} = Doc) ->
...
+            {ok, Encoded} = fabric2_encryption:encode(
```
rnewson: `encode` is not an acceptable synonym for `encrypt`.

src/fabric/src/fabric2_fdb.erl
```diff
@@ -1328,7 +1339,20 @@ doc_to_fdb(Db, #doc{} = Doc) ->
 
     DiskAtts = lists:map(fun couch_att:to_disk_term/1, Atts),
 
-    Value = term_to_binary({Body, DiskAtts, Deleted}, [{minor_version, 1}]),
+    ValueTerm = case WrappedKEK of
+        false ->
+            {Body, DiskAtts, Deleted};
+        _ ->
+            UpdateCounter = get_stat(Db, <<"update_count">>),
```
rnewson: that we fetch this stat and use it without guaranteeing it is unique is a serious problem. the update happens elsewhere.

src/fabric/src/fabric2_encryption.erl
```diff
@@ -0,0 +1,254 @@
...
+    gen_server:call(?MODULE,
+        {decode, WrappedKEK, DbName, DocId, UpdateCounter, DocBody}).
+
+
+
+init(_) ->
+    process_flag(sensitive, true),
+    process_flag(trap_exit, true),
+
+    case init_st() of
```
rnewson: this seems elaborate given the implementation of init_st().

src/fabric/src/fabric2_encryption.erl
```diff
@@ -0,0 +1,254 @@
...
+    end.
+
+
+encode(WrappedKEK, DbName, DocId, UpdateCounter, DocBody)
+    when is_binary(WrappedKEK),
+         is_binary(DbName),
+         is_binary(DocId),
+         is_integer(UpdateCounter), UpdateCounter > 0,
+         is_binary(DocBody) ->
+    gen_server:call(?MODULE,
```
rnewson: doing all the crypto work in a single gen_server sounds like a recipe for performance issues, even if we do spawn off worker processes there.

could we instead a) set the sensitive flag on self(), b) fetch the unwrapped key c) perform the operation in process d) unset sensitive flag?

src/fabric/src/fabric2_encryption.erl
```diff
@@ -0,0 +1,254 @@
...
+
+
+get_wrapped_kek(DbName) when is_binary(DbName) ->
+    case config:get_boolean("encryption", "enabled", false) of
+        true -> gen_server:call(?MODULE, {get_wrapped_kek, DbName});
+        false -> {ok, false}
+    end.
+
+
+encode(WrappedKEK, DbName, DocId, UpdateCounter, DocBody)
```
rnewson: this function performs a cryptographic encryption not a mere encoding. rename to `encrypt` for clarity.

src/fabric/src/fabric2_encryption.erl
```diff
@@ -0,0 +1,254 @@
...
+    when is_binary(WrappedKEK),
+         is_binary(DbName),
+         is_binary(DocId),
+         is_integer(UpdateCounter), UpdateCounter > 0,
+         is_binary(DocBody) ->
+    gen_server:call(?MODULE,
+        {encode, WrappedKEK, DbName, DocId, UpdateCounter, DocBody}).
+
+
+decode(WrappedKEK, DbName, DocId, UpdateCounter, DocBody)
```
rnewson: `decrypt`

src/fabric/src/fabric2_encryption.erl
```diff
@@ -0,0 +1,254 @@
...
+            proc_lib:init_ack(Error)
+    end.
+
+
+terminate(_, _St) ->
+    ok.
+
+
+handle_call({get_wrapped_kek, DbName}, _From, #{cache := Cache} = St) ->
+    {ok, KEK, WrappedKEK} = fabric2_encryption_plugin:get_wrapped_kek(DbName),
```
rnewson: can we avoid doing slow external work within the gen_server loop?

src/fabric/src/fabric2_encryption.erl
```diff
@@ -0,0 +1,254 @@
...
+         is_integer(UpdateCounter), UpdateCounter > 0,
+         is_binary(DocBody) ->
+    gen_server:call(?MODULE,
+        {decode, WrappedKEK, DbName, DocId, UpdateCounter, DocBody}).
+
+
+
+init(_) ->
+    process_flag(sensitive, true),
+    process_flag(trap_exit, true),
```
rnewson: let's try to avoid building another supervisor of our own.

src/fabric/src/fabric2_encryption_provider.erl
```diff
@@ -0,0 +1,69 @@
...
+            Enc = crypto:stream_init(aes_ctr, MEK, IV),
+            {_, WrappedKEK} = crypto:stream_encrypt(Enc, KEK),
+            {ok, KEK, WrappedKEK};
+        {error, Error} ->
+            {error, Error}
+    end.
+
+
+unwrap_kek(WrappedKEK) ->
+    case get_mek_iv() of
```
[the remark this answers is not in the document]
rnewson: (we can use the key wrapping and unwrapping code from aegis instead)

rel/overlay/etc/default.ini
eiri: Config is writable, my idea is that key will be kept in read-only file with an option to have that file on a separate encrypted mount point.

How do you suggest to store IV?

src/fabric/src/fabric2_fdb.erl
eiri: stats are using atomic operations, if the counter got updated since the start of this transaction it'll get aborted here and retried, so counter guaranteed to be unique in a single transaction.

I have a question though: using read-write on atomic ops kills its performance benefits - can we use uuid instead of update counter, is it cryptographically solid enough? that would allow us to avoid "external" monitonic dependancy all together

src/fabric/src/fabric2_encryption_provider.erl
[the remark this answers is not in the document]
eiri: This is Master key though, not databases' Key encryption key, it's not touching data, it only used to encrypt generated per db KEKs.  I can change `aes_ctr` encryption of KEK I'm using here for aegis wrapping, but it'll still be the same key for wrap/unwrap operation.

src/fabric/src/fabric2_encryption.erl
eiri: This is because of KEK cache. Its erts table set to `protected`, so only gen server can read it, fetching unwrapped key from worker would require to change ets to `private`, but then any proc could read from it, there are no option to limit this, unfortunally.

rel/overlay/etc/default.ini
rnewson: there shouldn't _be_ one.

src/fabric/src/fabric2_fdb.erl
rnewson: That would not have a guarantee of uniqueness, which is the only reason we have "update counter" instead of "rev" from the original proposal.

src/fabric/src/fabric2_encryption.erl
rnewson: you mean `private` there. a `protected` table can be read by all processes.

in the current code you hand the unwrapped key to another process (`spawn_monitor`). It doesn't appear to be more protected in that process than if we did the work in the calling process.

src/fabric/src/fabric2_encryption_provider.erl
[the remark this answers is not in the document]
rnewson: we must use a standard key wrapping technique, not one of your own invention.

src/fabric/src/fabric2_encryption.erl
eiri: yes, `private`, thank you. it's more protected in sense that I can't remsh and read unwrapped key from ets. if you think this is not a concern then I'll make worker to do the unwrap fetch.

davisp:
I'm gonna leave all the encryption related stuff to @rnewson and others that know better. My only concern was already called out about having a single `gen_server` handling all of that work. Though looks like there's been changes there already.

eiri:
Closing in favour of https://github.com/apache/couchdb/compare/prototype/fdb-layer...aegis

PR #2687, opened 2020-03-20
Jwt enhancements

rnewson:
## Overview

This enhances the JWT authentication handler in the following ways;

* Multiple keys are allowed, identified by the "kid" attribute in the token.
* A default key (called "_default") is used if there is no "kid" attribute.
* Support for EC keys.
* Support for RSA keys.

## Testing recommendations

Test coverage is pretty good but real world testing is a good idea too.

N/A

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -141,12 +141,24 @@ max_db_number_for_dbs_info_req = 100
 ; admin_only_all_dbs = true
 
 ;[jwt_auth]
-; Symmetric secret to be used when checking JWT token signatures
-; secret =
 ; List of claims to validate
 ; required_claims = exp
 ; List of algorithms to accept during checks
 ; allowed_algorithms = HS256
+;
+; [jwt_keys]
+; Configure at least one key here if using the JWT auth handler.
+; If your JWT tokens do not include a "kid" attribute, use "_default"
+; as the config key, otherwise use the kid as the config key.
+; Examples
+; _default = aGVsbG8=
+; foo = aGVsbG8=
+; The config values can represent symmetric and asymmetrics keys.
+; For symmetrics keys, the value is base64 encoded;
+; _default = aGVsbG8= # base64-encoded form of "hello"
+; For asymmetric keys, the value is the PEM encoding of the public
+; key with newlines replaced with the escape sequence \n.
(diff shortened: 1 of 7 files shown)


rel/overlay/etc/default.ini
```diff
@@ -141,12 +141,16 @@ max_db_number_for_dbs_info_req = 100
...
 ;[jwt_auth]
-; Symmetric secret to be used when checking JWT token signatures
-; secret =
 ; List of claims to validate
 ; required_claims = exp
 ; List of algorithms to accept during checks
 ; allowed_algorithms = HS256
+;
+; [jwt_keys]
+; Default secret to be used when checking JWT token signatures
```
atrauzzi: Maybe just add a note here for people like me who only want to configure a single symmetric like "put your single symmetric key here"?

wohali:
@atrauzzi FYI we'll also need an approval from a committer before this can land. (That shouldn't be me.)

atrauzzi:
I noticed that @rnewson added me for review, but yes, of course! 🙂

test/elixir/test/jwtauth_test.exs
```diff
@@ -3,26 +3,110 @@ defmodule JwtAuthTest do
...
-    run_on_modified_server(server_config, fn ->
-      test_fun()
-    end)
+    run_on_modified_server(server_config, fn -> test_fun("ES256", private_key) end)
+    run_on_modified_server(server_config, fn -> test_fun("ES384", private_key) end)
+    run_on_modified_server(server_config, fn -> test_fun("ES512", private_key) end)
   end
 
-  def test_fun() do
+  def test_fun(alg, key) do
```
dottorblaster: Are we sure we can't have a better name than `test_fun`?

rel/overlay/etc/default.ini
rnewson: I'll expand the text. I also think the symmetric key will need to be base64 encoded in the config, so we can allow binary values. So far we've assumed text is ok.

rnewson: I've addressed both things, please see if the updated text is more helpful.

atrauzzi: Hmm, not sure how I feel about base64'ing the key in config. I've yet to encounter any configuration format that would insist on this and it could frustrate deployments for some people.

test/elixir/test/jwtauth_test.exs
rnewson: nothing great springs to mind. We can update this after if inspiration strikes.

PR #2690, opened 2020-03-20
Fix create db options on secondary shard creation

chewbranca:
## Overview

This is a continuation of the work in https://github.com/apache/couchdb/pull/2672 to fix issues around secondary shard creation when nodes re-enter a cluster after a database was created while the node was down (or potentially in maintenance mode).

This PR introduces logic to handle the following four scenarios while opening a db in context of `fabric_rpc` and `mem3_rpc`:

  1) the db already exists
    - just return the db as normal
  2) the db does not exist locally, but the dbs db doc with db create options is present
    - load the db create options and create the local shard
  3) the db does not exist locally but it's not a sharded database
    - just create the local db as expected
  4) the db does not exist locally, it's a sharded database, and the dbs db doc is not present
    - throw an error and prevent creating the db with incomplete settings

This PR changes the behavior of scenario 2) by loading the appropriate config and creating the db with the desired parameters, and of scenario 4) by no longer allowing the db to be created inappropriately.

For scenario three, I've taken the approach of not loading config as a non sharded db won't have a dbs db doc. In particular here, we want to properly handle updates to internal dbs like `_dbs` and also `sys_dbs` like `_users`.

Another open item here is the merging of the db create options and the provided options. Right now, I'm just doing `DbOpts ++ Opts` basically, but I wrote a functions `opts_merge/{2,3}` that will merge two proplists, prioritizing the first list of items, and supports atoms. It's... a bit gnarly, and I'm not convinced we should use it. So we need to figure out what to do with the options lists here, but I _think_ we might be ok just combining the two.

While in the general neighborhood, I took the opportunity to de-duplicate some of the shards_db config lookups as well. I've isolated those changes to a dedicated commit.

## Testing recommendations

This needs more testing. I've got a WIP commit for adding a test suite here that is not yet complete. I figured it was worthwhile to get this PR out for feedback while continuing to work on the tests.

https://github.com/apache/couchdb/pull/2672

## Checklist

### DIFF
--- a/src/fabric/test/eunit/fabric_rpc_tests.erl
+++ b/src/fabric/test/eunit/fabric_rpc_tests.erl
@@ -1,0 +1,181 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(fabric_rpc_tests).
+
+
+-include_lib("couch/include/couch_eunit.hrl").
+-include_lib("couch/include/couch_db.hrl").
+
+
+-define(TDEF(A), {A, fun A/1}).
+
+
+main_test_() ->
+    {
+        setup,
+        spawn,
+        fun setup_all/0,
+        fun teardown_all/1,
+        [
+            {
+                foreach,
+                fun setup_no_db_or_config/0,
+                fun teardown_db/1,
+                lists:map(fun wrap/1, [
+                    ?TDEF(t_no_config_non_shard_db_create_succeeds)
+                ])
+            },
+            {
+                foreach,
+                fun setup_shard/0,
+                fun teardown_noop/1,
+                lists:map(fun wrap/1, [
+                    ?TDEF(t_no_db),
+                    ?TDEF(t_no_config_db_create_fails_for_shard),
+                    ?TDEF(t_no_config_db_create_fails_for_shard_rpc)
+                ])
+            },
+            {
+                foreach,
+                fun setup_shard/0,
+                fun teardown_db/1,
+                lists:map(fun wrap/1, [
+                    ?TDEF(t_db_create_with_config)
+                ])
+            }
+
+        ]
+    }.
+
+
+setup_all() ->
+    test_util:start_couch([rexi, mem3, fabric]).
+
+
+teardown_all(Ctx) ->
+    test_util:stop_couch(Ctx).
+
+
+setup_no_db_or_config() ->
+    ?tempdb().
+
+
+setup_shard() ->
+    ?tempshard().
+
+
+teardown_noop(_DbName) ->
+    ok.
+
+teardown_db(DbName) ->
+    ok = couch_server:delete(DbName, []).
+
+
+wrap({Name, Fun}) ->
+    fun(Arg) ->
+        {timeout, 60, {atom_to_list(Name), fun() ->
+            process_flag(trap_exit, true),
+            Fun(Arg)
+        end}}
+    end.
+
+
+t_no_db(DbName) ->
+    ?assertEqual({not_found, no_db_file}, couch_db:open_int(DbName, [?ADMIN_CTX])).
+
+
+t_no_config_non_shard_db_create_succeeds(DbName) ->
+    ?assertEqual({not_found, no_db_file}, couch_db:open_int(DbName, [?ADMIN_CTX])),
+    ?assertEqual(DbName, mem3:dbname(DbName)),
+    ?assertMatch({ok, _}, mem3_util:get_or_create_db(DbName, [?ADMIN_CTX])).
+
+
+t_no_config_db_create_fails_for_shard(DbName) ->
+    ?assertEqual({not_found, no_db_file}, couch_db:open_int(DbName, [?ADMIN_CTX])),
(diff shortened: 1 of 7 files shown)


chewbranca:
BTW... I believe this issue also extends to PSE databases created while a node was out of rotation. Basically anything that relies on database creation config options, which is PSE and PQ (anything else?).

src/fabric/test/eunit/fabric_rpc_tests.erl
```diff
@@ -0,0 +1,96 @@
...
+teardown_noop(_DbName) ->
+    ok.
+
+teardown_db(DbName) ->
+    ok = couch_server:delete(DbName, []).
+
+
+wrap({Name, Fun}) ->
+    fun(Arg) ->
+        {timeout, 60, {atom_to_list(Name), fun() ->
```
davisp: These tests need a 60s timeout?

src/fabric/test/eunit/fabric_rpc_tests.erl
```diff
@@ -0,0 +1,96 @@
...
+            process_flag(trap_exit, true),
```
davisp: Do we need to drain messages here?

davisp: Well, after Fun runs.

src/mem3/src/mem3_util.erl
```diff
@@ -509,6 +507,59 @@ sort_ranges_fun({B1, _}, {B2, _}) ->
...
+    case couch_db:open_int(DbName, Options) of
+        {ok, _} = OkDb ->
+            OkDb;
+        {not_found, no_db_file} ->
+            try
+                DbOpts = case mem3:dbname(DbName) of
+                    DbName  -> [];
+                    MDbName -> mem3_shards:opts_for_db(MDbName)
+                end,
+                %% Options1 = opts_merge(DbOpts, Options)
```
davisp: Commented out code?

src/mem3/src/mem3_util.erl
```diff
@@ -509,6 +507,59 @@ sort_ranges_fun({B1, _}, {B2, _}) ->
...
+opts_merge(L1, L2) ->
+    lists:reverse(opts_merge(lists:sort(L1), lists:sort(L2, []))).
+
+opts_merge([], [], Acc) ->
+    Acc;
+opts_merge(L1, [], Acc) ->
+    L1 ++ Acc;
+opts_merge([], R1, Acc) ->
+    R1 ++ Acc;
+opts_merge([A1|R1], [A2|R2], Acc) when is_atom(A1) orelse is_atom(A2) ->
```
davisp: The complexity here for what's going on seems a bit out of balance. Why not just fold new options and lists:keystore them into the existing option accumulator?

src/fabric/test/eunit/fabric_rpc_tests.erl
chewbranca: @davisp probably not necessary, but I cripped this test module from https://github.com/apache/couchdb/blob/master/src/fabric/test/eunit/fabric_rpc_purge_tests.erl

src/mem3/src/mem3_util.erl
chewbranca: Because I wasn't sure how that would work with atom options like `sys_db`? I don't want to include that `opts_merge` code, but we need a merge function that prioritizes items in the first list and can handle atom label options. If you've got a chunk of code to do that then I'll happily use it.

src/mem3/src/mem3_util.erl
chewbranca: Yeah, I mentioned that in the PR notes. I'm not a fan of this function, so I left it commented out and just merged the lists in the hopes of finding a better approach.

src/mem3/src/mem3_util.erl
davisp:
```erlang
merge_opts([], Opts) ->
    Opts;
merge_opts([{K, V} | RestToAdd], Opts) ->
    merge_opts(RestToAdd, lists:keystore(K, 1, Opts, {K, V}));
merge_opts([K | RestToAdd], Opts) when is_atom(K) ->
    merge_opts(RestToAdd, [K | Opts]).

File: src/fabric/test/eunit/fabric_rpc_tests.erl (hunk 1 above)
Comment (davisp) [reply src/fabric/test/eunit/fabric_rpc_tests.erl]: Fine.

File: src/mem3/src/mem3_util.erl (hunk 3 above)
Comment (davisp) [reply src/mem3/src/mem3_util.erl]: Ah, didn't realize there was commentary in the comments area. 😃

File: src/fabric/test/eunit/fabric_rpc_tests.erl [hunk 5]
```diff
@@ -94,3 +103,68 @@ t_no_config_db_create_fails_for_shard(DbName) ->
...
+            [<<"add">>,<<"00000000-7fffffff">>, <<"node2@127.0.0.1">>],
+            [<<"add">>,<<"00000000-7fffffff">>, <<"node3@127.0.0.1">>],
+            [<<"add">>,<<"80000000-ffffffff">>, <<"node1@127.0.0.1">>],
+            [<<"add">>,<<"80000000-ffffffff">>, <<"node2@127.0.0.1">>],
+            [<<"add">>,<<"80000000-ffffffff">>, <<"node3@127.0.0.1">>]
+        ]},
+        {<<"by_node">>, {[
+            {<<"node1@127.0.0.1">>, [<<"00000000-7fffffff">>, <<"80000000-ffffffff">>]},
+           {<<"node2@127.0.0.1">>, [<<"00000000-7fffffff">>, <<"80000000-ffffffff">>]},
+           {<<"node3@127.0.0.1">>, [<<"00000000-7fffffff">>, <<"80000000-ffffffff">>]}
```
davisp: Missing a space on these two lines.

wohali:
Shouldn't this and https://github.com/apache/couchdb/pull/2672 target the `3.x` branch, from which our next release will be cut?

rnewson:
yes. :)

PR #2706, opened 2020-03-23
add info endpoint for fdb stored views

tonysun83:
## Overview

This is a WIP to add the `_info` endpoint for views stored in fdb. Currently, I am unsure of the backwards compatibility for some of the fields returned by traditional couchdb views.

Info can be divided into 3 categories: 

1) Status of the build - In traditional views, we looked at
    **a) the update seq of the view**
    b) committed seq
    c) number of clients waiting to commit
    d) number of clients waiting
    e) if the indexer is still running

We can determine **a)** by calling `couch_views_fdb:get_update_seq/2` (note it's a binary instead of couch_mrview's `update_seq` values which are integers).  In traditional views, **b)-d)** refers the number of workers in a queue that are waiting to start indexing and then committing. I'm not sure how this transfers over to fdb as workers are part of the couch_jobs framework and commits occur for very transaction. For e), I looked at `couch_jobs` to see if crafting the `job_id` can give us the state of the job (confirmed with @nickva )

2) Size of the Index - Disk size does not seem applicable. For data size, `couch_view` has a `couch_views_fdb:get_kv_size/2` function for each view built. I don't think the subcategories of size are applicable either: `file`, `external`, `active`. 

Compaction - Not Applicable.

## Testing recommendations

Added a new `_info` endpoint test.

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -368,7 +368,7 @@ bad_action_req(#httpd{path_parts=[_, _, Name|FileNameParts]}=Req, Db, _DDoc) ->
 
 handle_design_info_req(#httpd{method='GET'}=Req, Db, #doc{} = DDoc) ->
     [_, _, Name, _] = Req#httpd.path_parts,
-    {ok, GroupInfoList} = fabric:get_view_group_info(Db, DDoc),
+    {ok, GroupInfoList} = couch_views:get_views_info(Db, DDoc),
     send_json(Req, 200, {[
         {name,  Name},
         {view_index, {GroupInfoList}}
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -368,7 +368,7 @@ bad_action_req(#httpd{path_parts=[_, _, Name|FileNameParts]}=Req, Db, _DDoc) ->
 
 handle_design_info_req(#httpd{method='GET'}=Req, Db, #doc{} = DDoc) ->
     [_, _, Name, _] = Req#httpd.path_parts,
-    {ok, GroupInfoList} = couch_views:get_views_info(Db, DDoc),
+    {ok, GroupInfoList} = couch_views:get_info(Db, DDoc),
     send_json(Req, 200, {[
         {name,  Name},
         {view_index, {GroupInfoList}}
--- a/src/couch_views/src/couch_views.erl
+++ b/src/couch_views/src/couch_views.erl
@@ -205,6 +205,7 @@ view_cmp(SK, SKD, EK, EKD) ->
     PackedEK = erlfdb_tuple:pack({BinEK, EKD}),
     PackedSK =< PackedEK.
 
+
 get_update_options(#mrst{design_opts = Opts}) ->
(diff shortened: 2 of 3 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -368,7 +368,7 @@ bad_action_req(#httpd{path_parts=[_, _, Name|FileNameParts]}=Req, Db, _DDoc) ->
 
 handle_design_info_req(#httpd{method='GET'}=Req, Db, #doc{} = DDoc) ->
     [_, _, Name, _] = Req#httpd.path_parts,
-    {ok, GroupInfoList} = fabric:get_view_group_info(Db, DDoc),
+    {ok, GroupInfoList} = couch_views:get_views_info(Db, DDoc),
```
davisp: `get_views_info` is a bit awkward since we're getting a singular info object back. How about just `get_info` instead?

tonysun83:
I decided to use `couch_mrview_index:get/2` even though we aren't supporting `Partitioned`. It's saves additional code that's going to be very similar and partitioned will only be returned on the odd chance it's incorrectly specified for indexing. It won't break anything but instead just return the value. The new layout:

```
{
    "name": "foo",
    "view_index": {
        "language": "javascript",
        "signature": "b7bf9f8b5adefa4b23c4cf60448d8fad",
        "sizes": {
            "active": 0
        },
        "update_seq": "",
        "updater_running": false,
        "data_size": 0,
        "update_options": []
    }
}
```

src/couch_views/test/couch_views_info_test.erl
```diff
@@ -0,0 +1,134 @@
...
+    }.
+
+sig_is_binary({_, Info}) ->
+    ?_assert(is_binary(prop(signature, Info))).
+
+language_is_js({_, Info}) ->
+    ?_assertEqual(<<"javascript">>, prop(language, Info)).
+
+data_size_is_non_neg_int({_, Info}) ->
+    ?_assert(check_non_neg_int(data_size, Info)).
```
davisp: This should fail, right?

src/couch_views/src/couch_views.erl
```diff
@@ -73,6 +74,43 @@ build_indices(#{} = Db, DDocs) when is_list(DDocs) ->
         end
     end, DDocs).
 
+get_info(Db, DDoc) ->
```
davisp: Getting into the super nitty gritty on style, but its two empty lines between functions.

src/couch_views/test/couch_views_info_test.erl
tonysun83: It's just checking that it's >= 0 right?

```
  Views index info test
    couch_views_info_test:75: sig_is_binary...ok
    couch_views_info_test:78: language_is_js...ok
    couch_views_info_test:90: update_seq_is_binary...ok
    couch_views_info_test:87: updater_running_is_boolean...ok
    couch_views_info_test:81: data_size_is_non_neg_int...ok
    couch_views_info_test:84: active_size_is_non_neg_int...ok
    couch_views_info_test:94: update_opts_is_bin_list...ok
```

davisp: Ah, I should have looked harder. You need to remove `data_size` as that regresses our API changes.

PR #2731, opened 2020-03-30
All docs raw collation

garrensmith:
## Overview
Change `_all_docs` to do raw collation instead of view collation. 

## Testing recommendations

`make check-fdb` should pass.

## Checklist

### DIFF
--- a/src/fabric/src/fabric2_util.erl
+++ b/src/fabric/src/fabric2_util.erl
@@ -294,13 +294,9 @@ uuid() ->
     to_hex(crypto:strong_rand_bytes(16)).
 
 
-encode_all_doc_key(null) -> <<>>;
-encode_all_doc_key(true) -> <<>>;
-encode_all_doc_key(false) -> <<>>;
-encode_all_doc_key(N) when is_number(N) -> <<>>;
 encode_all_doc_key(B) when is_binary(B) -> B;
-encode_all_doc_key(L) when is_list(L) -> <<255>>;
-encode_all_doc_key({O}) when is_list(O) -> <<255>>.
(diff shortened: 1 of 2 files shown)


src/fabric/src/fabric2_util.erl
```diff
@@ -294,13 +294,9 @@ uuid() ->
...
 
-encode_all_doc_key(null) -> <<>>;
-encode_all_doc_key(true) -> <<>>;
-encode_all_doc_key(false) -> <<>>;
-encode_all_doc_key(N) when is_number(N) -> <<>>;
 encode_all_doc_key(B) when is_binary(B) -> B;
-encode_all_doc_key(L) when is_list(L) -> <<255>>;
-encode_all_doc_key({O}) when is_list(O) -> <<255>>.
+encode_all_doc_key(Term) when Term < <<>> -> <<>>;
+encode_all_doc_key(_) -> <<255>>.
```
iilyak: I don't understand why we are always returning `<<255>>` here. If we are ignoring the term itself we would have 
`start_key == end_key == <<255>>` here https://github.com/apache/couchdb/blob/prototype/fdb-layer/src/chttpd/src/chttpd_db.erl#L952:L960. I would understand if we would have:
```
encode_all_doc_key(Term) when Term < <<>> -> <<>>;
encode_all_doc_key(B) when is_binary(B) -> B;
encode_all_doc_key(_) -> <<255>>.
```

iilyak: I didn't notice the `encode_all_doc_key(B) when is_binary(B) -> B;` clause in the diff. Now it is clear.

PR #2732, opened 2020-03-30
Enhance JWT controls (again)

rnewson:
## Overview

This PR;

* fixes the broken required_claims property (due to my misunderstanding of `proplists:get_keys`).
* Removes the enhancement to the `alg` check as it is superceded by the binding of algorithm type and key value in jwtf_keystore.
* Changes the default required_claims to include nbf and exp. Administrators can remove either or both restriction if they choose.

## Testing recommendations

All changes covered by unit and/or integration tests in this PR.

N/A

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -143,8 +143,6 @@ max_db_number_for_dbs_info_req = 100
 ;[jwt_auth]
 ; List of claims to validate
 ; required_claims = exp
-; List of algorithms to accept during checks
-; allowed_algorithms = HS256
 ;
 ; [jwt_keys]
 ; Configure at least one key here if using the JWT auth handler.
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -142,7 +142,7 @@ max_db_number_for_dbs_info_req = 100
 
 ;[jwt_auth]
 ; List of claims to validate
-; required_claims = exp
+; required_claims =
 ;
 ; [jwt_keys]
 ; Configure at least one key here if using the JWT auth handler.
--- a/src/jwtf/src/jwtf.erl
+++ b/src/jwtf/src/jwtf.erl
@@ -159,11 +159,11 @@ validate_typ(Props, Checks) ->
     Required = prop(typ, Checks),
     TYP = prop(<<"typ">>, Props),
     case {Required, TYP} of
-        {undefined, _} ->
+        {undefined, undefined} ->
             ok;
         {true, undefined} ->
             throw({bad_request, <<"Missing typ header parameter">>});
-        {true, <<"JWT">>} ->
+        {_, <<"JWT">>} ->
             ok;
         {true, _} ->
             throw({bad_request, <<"Invalid typ header parameter">>})
@@ -174,11 +174,11 @@ validate_alg(Props, Checks) ->
     Required = prop(alg, Checks),
     Alg = prop(<<"alg">>, Props),
     case {Required, Alg} of
-        {undefined, _} ->
+        {undefined, undefined} ->
             ok;
         {true, undefined} ->
             throw({bad_request, <<"Missing alg header parameter">>});
-        {true, Alg} ->
+        {_, Alg} ->
             case lists:member(Alg, valid_algorithms()) of
                 true ->
                     ok;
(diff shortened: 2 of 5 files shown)


rel/overlay/etc/default.ini
```diff
@@ -142,7 +142,7 @@ max_db_number_for_dbs_info_req = 100
 
 ;[jwt_auth]
 ; List of claims to validate
-; required_claims = exp
+; required_claims = exp,nbf
```
jaydoane: I found this [answer on SO](https://stackoverflow.com/questions/43291659/usage-of-nbf-in-json-web-tokens) which talks about some possible use of `nbf`. How are we justifying this requirement?

src/jwtf/src/jwtf.erl
```diff
@@ -123,15 +123,33 @@ validate(Header0, Payload0, Signature, Checks, KS) ->
...
             ok;
         UnknownChecks ->
             error({unknown_checks, UnknownChecks})
     end.
 
+
+valid_check(Check) when is_atom(Check) ->
+    lists:member(Check, ?CHECKS);
+
+valid_check({Check, _}) when is_atom(Check) ->
```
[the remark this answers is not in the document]
rnewson: for example, the `iss` check has a value, you have to say `{iss, <<"name of issuer">>}` in the `Check` param, and then jwtf will reject tokens with no `iss` property or with one where the value doesn't match.

rel/overlay/etc/default.ini
rnewson: it's akin to the "Not Valid Before" and "Not Valid After" fields that all TLS certificates comes with, and recognises that tracking the "true" time is difficult, and that being able to influence clocks could lead to a security issue.

We check for expiration under the reasonable expectation that clocks typically move forward. If we don't check the nbf then there's an infinite window of time where the token is valid. If we check bnf then it's a specific timeframe, which is harder to target.

An alternative approach here that I considered, and would like your views on;

1) We validate every field present in the token, regardless of server setting. So, if the token includes `nbf`, `exp` or both, we'll validate them, rejecting them if they don't hold with respect to the servers clock.
2) The required_claims is then _only_ about whether we require those claims or not.
3) The default list would then be empty. Not insisting on exp or nbf but also not accepting a token with an nbf in the future or an exp in the past as valid.

jaydoane: I would prefer that we require a minimum of `exp` via default.ini, so you'd have to explicitly remove it if you wanted an arguably less secure system.

rnewson: that was my original feeling for `exp,nbf` so I will add another commit tomorrow. Noting that before this PR the default of `exp` was ignored anyway.

rnewson: ah, you know, I think I will advocate for not including either -- because of replication. To use JWT with replication, you can set the "headers" directly and include your token, but this is no use if it expires. So, the token issuer can issue short-lived tokens for users (setting exp and/or nbf as they see fit) and we'll enforce them just because they're present. That same token issuer can issue a token without exp for use in replication tasks, though the only way to then revoke them is to change the secrets the server trusts...

atrauzzi: I think enforcing anything outside of the spec could surprise or get in the way of some legitimate design that somebody intends for.

My thought is to have the bare minimum out of a zero-configuration setup.  As in, if I run CouchDB and only enable JWTs, they should work.  This is more in the spirit of being easy to get started with and being able to dial in desired configuration incrementally.

That all said, I am open to some degree of caution, without forcing people to implement things that aren't applicable to their system (`nbf` being one potential example of such).

rnewson: we're discussing the defaults, @atrauzzi. Admins can change these settings. "Easy to get started" is my goal as long as it comports with "secure by default". What is your concrete suggestion for the required_claims default?

atrauzzi: Secure by default would be to require an `exp` claim in my opinion. :smile:

rnewson: An administrator that wishes to use JWT and replication would have to disable it, though.

atrauzzi: Ah, right!

Well, let's make sure the docs include something about changing the secret as a way to mitigate compromised tokens then.  Just so that people have some evidence that this scenario was considered.

rnewson: right. and of course you can use different secrets for the different key uses, so you could rotate your "replication" secret independently.

jaydoane: > ah, you know, I think I will advocate for not including either -- because of replication. To use JWT with replication, you can set the "headers" directly and include your token, but this is no use if it expires. So, the token issuer can issue short-lived tokens for users (setting exp and/or nbf as they see fit) and we'll enforce them just because they're present. That same token issuer can issue a token without exp for use in replication tasks, though the only way to then revoke them is to change the secrets the server trusts...

I suspect someone using token based replication is also going to know how to edit default.ini, so would still vote +1 to include `exp`. FWIW, I think the replication token should still expire, even if it's set years in advance. An eternal token seems no better than a password.

PR #2759, opened 2020-04-04
The ddoc `list` call got timeout even though `os_process_timeout` is adjusted

iblislin:
I'm using the ddoc list function with the native Erlang query server.
It keeps `timeout`ing in a time-consuming list function, even though my `os_process_timeout` is set to `80000` .

I have no idea about the couchdb internal, and my Erlang is quite rusty,
but this simple patch can solve my problem. 

Hope this patch can help for identifying the actually issue.


iblis17:
The log is here
```
[error] 2020-04-04T08:58:10.500532Z node1@127.0.0.1 <0.698.0> a72702703a req_err(4053083524) timeout : The request could not be
 processed in a reasonable amount of time.                                                                                     
    [<<"gen_server:call/2 L215">>,<<"couch_query_servers:proc_prompt/2 L520">>,<<"couch_mrview_show:send_list_row/2 L279">>,<<"
fabric_view:maybe_send_row/1 L132">>,<<"rexi_utils:process_mailbox/6 L55">>,<<"fabric_view_map:go/6 L81">>,<<"fabric_view_map:g
o/8 L44">>,<<"couch_query_servers:with_ddoc_proc/2 L514">>]                                                                    
```

bessbd:
Hi @iblis17,

Thank you for this patch!
Do you think you could include a test with this change?

PR #2768, opened 2020-04-08
Clean up existing expiry key on update

jaydoane:
## Overview

When an existing key is inserted with different timestamps, the primary
key is the same but the primary value is different from the existing one.
Currently, this results in a new expiry key being inserted, but the old
one is not deleted, and lingers until it is removed by the inexorable
advance of time via the `remove_expired` server messages.

This checks whether there's already primary key for the inserted key,
and if so, cleans up the existing expiry key before proceeding with the
insert.

## Testing recommendations

```
make eunit apps=couch_expiring_cache
```

## Checklist

### DIFF
--- a/src/couch_expiring_cache/src/couch_expiring_cache_fdb.erl
+++ b/src/couch_expiring_cache/src/couch_expiring_cache_fdb.erl
@@ -13,6 +13,7 @@
 -module(couch_expiring_cache_fdb).
 
 -export([
+    get_range_to/3,
     insert/6,
     lookup/3,
     clear_all/1,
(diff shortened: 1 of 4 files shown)


eiri:
My only note here is that a name `get_entries/3` is a bit generic and doesn't match `clear_range/3`. I'd rename them both to something lilke `get_range_from/3` and `clear_range_from/3` to indicate that they are expecting TS on input and work on ranges, but this is your call.

jaydoane:
> My only note here is that a name `get_entries/3` is a bit generic and doesn't match `clear_range/3`. I'd rename them both to something lilke `get_range_from/3` and `clear_range_from/3` to indicate that they are expecting TS on input and work on ranges, but this is your call.

I think you're right that the names could be improved, however since the ranges are all _to_ an end time (rather than from one), it seems like better names might be `get_range_to` and `clear_range_to`, so I'll change them.

PR #2771, opened 2020-04-09
Fix couch rate limiter bug

davisp:
View updates can have batches with zero writes


davisp:
@iilyak @nickva Can y'all look at this again. I don't think the math works out trying to protect the division by zero. I'm pretty sure that's basically just equivalent resetting to 1 doc per batch since it effectively ended up being whatever the read/write ratio was.

Instead, I just detect when it would have thrown badarith and used the previous batch size.

PR #2778, opened 2020-04-10
Enable configurable binary chunk size

jaydoane:
## Overview

Currently, the size of binary chunks used for values is fixed at the FDB imposed limit of 100kB, although they recommend using 10KB [1], (also note they subtly change units).

This makes that value configurable, allowing e.g. benchmarks to compare performance of runs with varying chunk size. The cost is a ~10µs config lookup penalty each time data needs to be
chunked.

[1] https://www.foundationdb.org/files/record-layer-paper.pdf

## Testing recommendations

`make check` should pass. This code is trivial, and gets run every time a db is created or opened, so it will get coverage from the normal test suite.

## Checklist


jaydoane:
> Rather than carry that option around in the database map I'd just modify `chunkify_binary/1` to load it directly. So you'd have something like:
> 
> ```erlang
> chunkify_binary(Data) ->
>     chunkify_data(Data, binary_chunk_size()).
> ```

I considered doing that first, but was concerned with the ~10µs cost for each chunk that gets written, rather than just once when the db is opened/created. Is the idea that most data won't get chunked at all, so we shouldn't penalize the normal use case for a more obscure one?

jaydoane:
@davisp pushed a fixup with your requested changes

davisp:
@jaydoane I just saw that @iilyak merged an emilio PR this morning. You may want to force push this branch again to see if the CI complains.

jaydoane:
@davisp I did not see any additional checks after force pushing.

Issue #2756, opened 2020-04-03
String.toLowerCase() produces error on strings with Umlaut.


Swaser:
When using the .toLowerCase() (or the .toLocalLowerCase()) function on strings containing Umlauts, the design document view will not produce any results.

## Description

See above.

## Steps to Reproduce

1. Create a database and add any document to it.
1. Creata a design document with a view and use the following code
```
function (doc) {
  emit('Ä'.toLowerCase(),1);
}
```

## Expected Behaviour

A view will be created with at least one entry (as many as there were documents) with `key: 'ä'` and `value: 1`.

## Your Environment

* CouchDB version used: 3.0.0 dockerized and pulled from dockerhub (couchdb:latest)
* Browser name and version: Firefox 74.0
* Operating system and version: docker on Xubuntu 19.10

## Additional Context

[TIP]:  # ( Add any other context about the problem here. )

florianb:
Trying to test the example leads to the following error on my instance:

```
couchdb_1  | [info] 2020-04-03T16:48:09.950145Z nonode@nohost <0.5379.0> -------- Starting index update for db: shards/80000000-ffffffff/test.1585932129 idx: _design/test
couchdb_1  | [error] 2020-04-03T16:48:09.954051Z nonode@nohost emulator -------- Error in process <0.5387.0> with exit value:
couchdb_1  | {{nocatch,{invalid_json,{5,invalid_string}}},[{couch_util,json_decode,1,[{file,"src/couch_util.erl"},{line,505}]},{couch_mrview_updater,merge_results,3,[{file,"src/couch_mrview_updater.erl"},{line,245}]},{lists,foldl,3,[{file,"lists.erl"},{line,1263}]},{couch_mrview_updater,merge_results,4,[{file,"src/couch_mrview_updater.erl"},{line,238}]},{couch_mrview_updater,accumulate_writes,3,[{file,"src/couch_mrview_updater.erl"},{line,215}]},{couch_mrview_updater,write_results,2,[{file,"src/couch_mrview_updater.erl"},{line,179}]}]}
```

This applies to an used Umlaut in **any** string-literal in the view. I am short on time, couldn't test it with different Unicode-Characters or functions.
I use Fauxton to update the view.

---
Tested on hub.docker.com provided `couchdb:latest` 3.0.0.

wohali:
Hi there,

Unfortunately, this is a known problem with CouchDB + JavaScript + string mutation going way back. We need to document this better.

[...]

Issue #2779, opened 2020-04-12
Broken unicode in CouchDB 3.0.0


kimaiavd:
## Description

After upgrading to CouchDB 3 we had a big problem with JS query server that now incorrectly works with unicode characters.

## Steps to Reproduce

Create view with following map function:

`function (doc) { log(doc.string_field.length); }`

Put document with field `string_field` with any unicode characters. We're have text fields with Hebrew and Russian. Now you can see in console incorrect string length. Thus `substring` and `slice` also don't work.

For example:

`{"string_field": "мама мыла раму"}`

## Expected Behaviour

Correct behaviour of `length`, `substring` and `slice`.

## Your Environment

CouchDB 3.0 in single node mode or CouchDB 3.0 in cluster mode with 3 nodes. Installed from official docker image on 64bit host system.

* CouchDB version used: CouchDB 3.0, git_sha: `03a77db6c `
* Operating system and version: Official docker image - `couchdb:3`

wohali:
Duplicate of #2756 . Thanks for the additional background.

PR #2786, opened 2020-04-14
Fix couchjs utf8 conversions

davisp:
## Overview

This fixes UTF-8 conversions in the SpiderMonkey 60 implementation

## Testing recommendations

I should add tests

#2756

## Checklist

- [N/A] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [N/A] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation


src/couch/priv/couch_js/60/main.cpp
```diff
@@ -473,6 +473,7 @@ main(int argc, const char* argv[])
         // Compile and run
         JS::CompileOptions options(cx);
         options.setFileAndLine(args->scripts[i], 1);
+        options.setUTF8(true);
```
wohali, line 476: Good catch!

In SM68 (I'm working on this PR right now, hope to have it done today) this conversion is done explicitly via the (`JS::SourceText` class)[https://searchfox.org/mozilla-central/source/js/public/SourceText.h#29-44] and associated JS functions to convert to UTF-8 or UTF-16 "code units."

src/couch/priv/couch_js/60/util.cpp
```diff
@@ -231,7 +212,7 @@ couch_readline(JSContext* cx, FILE* fp)
     bytes = tmp;
     byteslen = used;
 
-    str = string_to_js(cx, std::string(tmp));
+    str = string_to_js(cx, std::string(tmp, byteslen));
```
wohali, line 215: Why not replace this with the direct call to `dec_string` and drop `string_to_js`? Unless you think it'd be useful over in `utf8.cpp`/`http.cpp` (for the short time left that we care about those)?

I wonder if we can't replace some of this with something native now.

At the very least, `dec_char` has an equivalent in [`Utf8ToOneUcs4Char`](https://searchfox.org/mozilla-central/source/js/src/vm/CharacterEncoding.cpp#178) (in `js/CharacterEncoding.h`.

This is used in [`Decode`](https://searchfox.org/mozilla-central/source/js/src/builtin/String.cpp#4219) but unfortunately this and its wrappers are not public anywhere. On the whole this doesn't look too different to our `dec_charbuf` or `dec_string`, but it would be useful to compare and ensure our code is current vs. what Firefox does.

I'll take this on shortly for the SM68 port if you want to punt on it for this PR.

src/couch/test/eunit/couch_js_tests.erl
```diff
@@ -33,15 +22,101 @@ couch_js_test_() ->
...
+    Doc = {[
+        {<<"value">>, <<16#C3, 16#84>>}
+    ]},
+    Result = couch_query_servers:proc_prompt(Proc, [<<"map_doc">>, Doc]),
+    ?assertEqual([[[<<16#C3, 16#A4>>, <<16#C3, 16#9C>>]]], Result).
+
+
+should_replace_broken_utf16() ->
+    % This test reverse the surrogate pair of
+    % the Boom emoji U+1F4A5
```
wohali, line 90: :boom:

src/couch/test/eunit/couch_js_tests.erl
```diff
@@ -33,15 +22,101 @@ couch_js_test_() ->
...
+      "  emit(doc.value, \"", 16#C3, 16#9C, "\");\n"
+      "}\n"
+    >>,
+    Proc = couch_query_servers:get_os_process(<<"javascript">>),
+    true = couch_query_servers:proc_prompt(Proc, [<<"add_fun">>, Src]),
+    Doc = {[
+        {<<"value">>, <<16#C3, 16#84>>}
+    ]},
+    Result = couch_query_servers:proc_prompt(Proc, [<<"map_doc">>, Doc]),
+    ?assertEqual([[[<<16#C3, 16#84>>, <<16#C3, 16#9C>>]]], Result).
```
wohali, line 69: Do you also want to test any of the other problems found in #2779 ?

jiangphcn:
thanks @davisp 

I saw below error in https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FPullRequests/detail/PR-2786/3/pipeline

```
[2020-04-14T22:40:43.754Z] test/javascript/tests/invalid_docids.js                        ported to elixir


[2020-04-14T22:40:44.105Z] test/javascript/tests/jsonp.js                                 double free or corruption (out)
```
looks that we expected `test/javascript/tests/jsonp.js                                 ported to elixir`.  Let me re-trigger CI.

src/couch/priv/couch_js/60/util.cpp
```diff
@@ -30,36 +32,32 @@ js_to_string(JSContext* cx, JS::HandleValue val)
...
 
-JSString*
-string_to_js(JSContext* cx, const std::string& s)
-{
-    JSString* ret = JS_NewStringCopyN(cx, s.c_str(), s.size());
-    if(ret != nullptr) {
-        return ret;
+    if(ret) {
+        // JS_NewUCString took ownership on succses
+        mozilla::Unused << utf16.release();
```
wohali, line 57: Not so sure about this, and @jiangphcn saw a double free in CI output.

src/couch/test/eunit/couch_js_tests.erl
wohali: FYI this fails for me locally in my SM68 work, when I copy over just your test. The others succeed.

src/couch/priv/couch_js/60/util.cpp
davisp: double free in CI output with this? Or from the other thing we talked about?

davisp: I pulled that from SpiderNode to be clear. I'm not 100% on the magic of that other than the ownership makes sense to me.

src/couch/test/eunit/couch_js_tests.erl
davisp: I'll look at adding those, yeah. They seemed reasonable enough on the face of it.

src/couch/test/eunit/couch_js_tests.erl
davisp: Can you paste the failure? Might be obvious if I can see the binary mismatch.

src/couch/priv/couch_js/60/util.cpp
davisp: I've gone through and repalced and removed utf8.{h,c} so this isn't just a pass through shim any more.

src/couch/priv/couch_js/60/main.cpp
davisp: heh, yeah, I just stumbled over it. And also tested to make sure it's actually necessary and effective. Hooray for scouring the source and headers for SpiderMonkey.

src/couch/priv/couch_js/60/util.cpp
wohali: Great, marking as resolved. I'll pull over your changes into my SM68 PR once you've merged this, then.

src/couch/priv/couch_js/60/util.cpp
wohali: Looking at SearchFox suggests that `mozilla::Unused` is only to suppress GCC warnings about unused return values:

https://searchfox.org/mozilla-central/source/mfbt/Unused.cpp
https://searchfox.org/mozilla-central/source/mfbt/Unused.h#17-24

I see evidence of it in the current tree for `js.cpp`:

https://searchfox.org/mozilla-central/source/js/src/shell/js.cpp#6041

so this is probably OK.

However, this might be a more elegant pattern:

https://searchfox.org/mozilla-central/source/js/src/vm/CompilationAndEvaluation.cpp#146-152

Note the `.get()` directly in the ctor for `JS:UniqueTwoBytesChars`. You could do this directly on line 46 and avoid the `.reset()` call. (I only used `.reset()` in one place to try and preserve the wacky `goto` hijinx in my PR, but I think in the end I removed the `goto` and stopped trying to work around the `goto` crossing the initialization error.

Double free might be related, unsure. Has it happened again?

wohali:
SM sure has come a long way since 1.8.5! I'm really happy to see these improvements in our code, and suspect they will pay off in the long term.

src/couch/test/eunit/couch_js_tests.erl
wohali:
```
module 'couch_js_tests'
  Test couchjs
    couch_js_tests: couch_js_test_...[0.052 s] ok
    couch_js_tests: couch_js_test_...[0.049 s] ok
    couch_js_tests: couch_js_test_...[0.043 s] ok
    couch_js_tests: couch_js_test_...*failed*
in function couch_util:json_decode/2 (src/couch_util.erl, line 508)
in call from couch_js_tests:should_replace_broken_utf16/0 (test/eunit/couch_js_tests.erl, line 101)
**throw:{invalid_json,{11,invalid_string}}
  output:<<"">>

    couch_js_tests: couch_js_test_...out of memory
[1.280 s] ok
```

Note I haven't yet pulled over your removal-of-UTF8 changes, so it might be a bug somewhere in the `enc_/dec_` functions I haven't bothered to diagnose.

src/couch/priv/couch_js/60/util.cpp
```diff
@@ -30,36 +33,32 @@ js_to_string(JSContext* cx, JS::HandleValue val)
...
-    JS::UniqueChars chars(JS_EncodeString(cx, str));
-    if(!chars) {
-        JS_ClearPendingException(cx);
-        fprintf(stderr, "Error converting  to string.\n");
-        exit(3);
+    JS::UTF8Chars utf8(raw.c_str(), raw.size());
+    JS::UniqueTwoByteChars utf16;
+    size_t len;
+
+    utf16.reset(JS::UTF8CharsToNewTwoByteCharsZ(cx, utf8, &len).get());
```
wohali, line 49: :+1: :+1:

src/couch/priv/couch_js/60/http.cpp
```diff
@@ -89,461 +91,276 @@ http_uri(JSContext* cx, JSObject* req, couch_args* args, JS::Value* uri_val)
...
-{
-    return;
-}
-
-
 // Map some of the string function names to things which exist on Windows
 #ifdef XP_WIN
 #define strcasecmp _strcmpi
 #define strncasecmp _strnicmp
 #define snprintf _snprintf
```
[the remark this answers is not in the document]
davisp: I undid the moderniziation so have to keep these.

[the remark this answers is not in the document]
davisp: Oooh, but I can remove snprintf. Doing that.

src/couch/test/eunit/couch_js_tests.erl
davisp: There's a smell off-by-one error in that the old code would expect to see a trailing `\0` null byte that no longer goes. I've had to tweak things a few times to make sure that I've got it right.

src/couch/priv/couch_js/60/util.cpp
davisp: Ah, I should comment on that. Yeah, the JS/Mozilla headers have a lot of declarations about forcing errors and warnings so that's specifically to silence the warning (which makes sense cause without knowing where that pointer went it looks like a leak).

I stole the pattern from SpiderNode who've been using it unchanged for multiple years. For the second version where its like three levels deep in the constructor I did it this way to avoid that style eyesore. I had to read it like five times to even get an idea of what they were doing.

I'll add a comment about the Unused bit being used to silence the compiler.

src/couch/test/eunit/couch_js_tests.erl
davisp: I've added cases from #2779.

davisp: Link might've been handy: https://github.com/apache/couchdb/pull/2786/files#diff-8957ae687446904d27579a9d199344e0R109

wohali:
As discussed on IRC, I'm going to squash and merge this, then port my changes over to my #2788 PR, which needs one final thoughtful review due to the Realm vs. Container change in SM68.

Once both have landed we'll backport this one to `3.0.x`, and both to `3.x`.

PR #2788, opened 2020-04-15
SpiderMonkey 68 support

### LINKED ISSUE #2770: Support SpiderMonkey 68
## Summary

Ubuntu focal (20.04) comes out in 2 weeks. That distribution no longer carries libmozjs-60, substituting libmozjs-68 instead. This will also be the case for Debian bullseye (11.0).

Unless we want to go back to shipping our own SM packages, our best option is to carry forward support to SM68.

## Possible Solution

I've started work on this locally and will push a PR once it's ready. There's significant enough changes that it's not just a `rebar.config.script` tweak, unfortunately.

wohali:
## Overview

PR is WIP, pending additional changes from #2786 when it's ready to go.

Also, global `s/,/.` on the `rebar.config.script` files I touched. That part is pre-approved by @rnewson .

## Testing recommendations

Set up a dev environment under Ubuntu 20.04 (Focal Fossa), then `make check`.

Here's one way:

```
docker pull couchdbdev/ubuntu-focal-erlang-20.3.8.25-1
docker run -it -w /home/jenkins couchdbdev/ubuntu-focal-erlang-20.3.8.25-1
git clone https://github.com/apache/couchdb/
cd couchdb && git checkout feat/spidermonkey-68
./configure --disable-fauxton --disable-docs --spidermonkey-version 68
make check
```

Fixes #2770 .

## Checklist


wohali:
Some sources that really helped me pull this together, for the archives:

* The include headers (of course)
* https://searchfox.org/ (search all of Mozilla code at once, finds both implementation details for SM as well as some use thereof)
* https://github.com/GNOME/gjs (GNOME's SM bindings, the reason why libmozjs is still packaged separately for Linux by the packagers)
* https://groups.google.com/forum/#!topic/mozilla.dev.tech.js-engine.internals/4crv5DNVuQk Discussion of the Container/Realm changes in SM post 60
* http://thbecker.net/articles/rvalue_references/section_01.html on `std::move()` and rvalues in C++

src/couch/priv/couch_js/68/utf8.cpp
```diff
@@ -0,0 +1,310 @@
...
+    
+    if(!dec_charbuf(bytes, byteslen, NULL, &charslen)) return NULL;
+
+    JS::UniqueTwoByteChars chars(js_pod_malloc<char16_t>(charslen + 1));
+    if(!chars) return NULL;
+    chars.get()[charslen] = 0;
+
+    if(!dec_charbuf(bytes, byteslen, chars.get(), &charslen)) goto error;
+
+    str = JS_NewUCString(cx, std::move(chars), charslen - 1);
```
wohali, line 299: The requirement here is the use of `js_pod_malloc` on L293, to prevent reaping of the memory when going out of the JS realm (this happens with the http bindings), combined with `std::move` here, necessary per the header file which actually deletes the ctor if you refuse to use it.

Drop either of those decisions and you get either double `free()`s or segfaults.

src/couch/priv/couch_js/68/utf8.cpp
```diff
@@ -0,0 +1,310 @@
...
+    if (!rawChars.initTwoByte(cx, str))
+        return NULL;
+
+    src = rawChars.twoByteRange().begin().get();
+    srclen = JS_GetStringLength(str);
+
+    if(!enc_charbuf(src, srclen, NULL, &byteslen)) goto error;
+    
+    //bytes = static_cast<char *>(JS_malloc(cx, (byteslen) + 1));
+    bytes = js_pod_malloc<char>(byteslen + 1);
```
wohali, line 154: Just like `dec_string` this has to be `js_pod_malloc`, not `JS_alloc` to prevent problems outside the JS runtime.

rebar.config.script
```diff
@@ -162,18 +162,18 @@ DepDescs = [
...
-WithProper = lists:keyfind(with_proper, 1, CouchConfig) == {with_proper, true},
+WithProper = lists:keyfind(with_proper, 1, CouchConfig) == {with_proper, true}.
 
 OptionalDeps = case WithProper of
     true ->
         [{proper, {url, "https://github.com/proper-testing/proper"}, {tag, "v1.3"}}];
     false ->
         []
-end,
+end.
```
davisp, line 174: Why did you change all of the statement separators?

src/couch/priv/couch_js/68/utf8.cpp
davisp: Where did you see the bit about js_pod_malloc vs JS_alloc? Or was that just trial and error?

wohali: Trial & error plus multiple places where this is done over in `js/src/shell/js.cpp`.

rebar.config.script
wohali: Because I burnt a day on trying to understand why I had a bug that jumped around when I changed `,` to `.`. Bob agreed we should pick one and stick to it. It's unrelated, I can move to a separate commit, but I don't see why we shouldn't do it.

davisp: Ah, that's more or less what i assumed. I'm fine not breaking it out into a separate commit.

src/couch/priv/couch_js/68/utf8.cpp
wohali: @davisp FYI, if you leave off the `std::move()` here, compile fails with this informative error - probably intentional on Mozilla's part:

```
Compiling /home/jenkins/couchdb/src/couch/priv/couch_js/68/utf8.cpp
/home/jenkins/couchdb/src/couch/priv/couch_js/68/utf8.cpp: In function 'JSString* dec_string(JSContext*, const char*, size_t)':
/home/jenkins/couchdb/src/couch/priv/couch_js/68/utf8.cpp:299:49: error: use of deleted function 'mozilla::UniquePtr<T [], D>::UniquePtr(const mozilla::UniquePtr<T [], D>&) [with T = char16_t; D = JS::FreePolicy]'
  299 |     str = JS_NewUCString(cx, chars, charslen - 1);
      |                                                 ^
In file included from /usr/include/mozjs-68/mozilla/Span.h:31,
                 from /usr/include/mozjs-68/mozilla/Range.h:12,
                 from /usr/include/mozjs-68/jsapi.h:15,
                 from /home/jenkins/couchdb/src/couch/priv/couch_js/68/utf8.cpp:13:
/usr/include/mozjs-68/mozilla/UniquePtr.h:455:3: note: declared here
  455 |   UniquePtr(const UniquePtr& aOther) = delete;  // construct using std::move()!
      |   ^~~~~~~~~
In file included from /home/jenkins/couchdb/src/couch/priv/couch_js/68/utf8.cpp:13:
/usr/include/mozjs-68/jsapi.h:2355:70: note:   initializing argument 2 of 'JSString* JS_NewUCString(JSContext*, JS::UniqueTwoByteChars, size_t)'
 2355 |                                               JS::UniqueTwoByteChars chars,
      |                                               ~~~~~~~~~~~~~~~~~~~~~~~^~~~~
ERROR: compile failed while processing /home/jenkins/couchdb/src/couch: rebar_abort
```

Looks to me like they both intentionally deleted the function it could have used, _and_ put that helpful comment there to help you out.

That convinced me, finally, that `std::move()` here is necessary and sufficient.

src/couch/priv/couch_js/68/util.cpp
```diff
@@ -0,0 +1,368 @@
...
+    // Shrink the buffer to the actual data size
+    tmp = static_cast<char*>(JS_realloc(cx, bytes, byteslen, used));
+    if(!tmp) {
+        JS_free(cx, bytes);
+        return NULL;
+    }
+    bytes = tmp;
+    byteslen = used;
+
+    str = string_to_js(cx, std::string(tmp, byteslen - 1));
```
wohali, line 265: @davisp As we discussed...this is tricky to get right.

Without this, all JS tests fail with:

```
error: JSON.parse: unexpected non-whitespace character after JSON data at line 1 column 302 of the JSON data
  
  Loop@/home/jenkins/couchdb/share/server/main.js:8406:16
  @/home/jenkins/couchdb/share/server/main.js:8433:1
  @/home/jenkins/couchdb/share/server/main.js:8434:3
```

At least I can confirm `couch_error` is working right, indentation and all! :upside_down_face:

wohali:
@davisp So this now passes `make elixir` and `make javascript`, but fails on `make eunit` with the following log output:

```
module 'couch_js_tests'
  Test couchjs
    couch_js_tests: couch_js_test_...[0.066 s] ok
    couch_js_tests: couch_js_test_...[0.063 s] ok
    couch_js_tests: couch_js_test_...[0.063 s] ok
    couch_js_tests: couch_js_test_...*failed*
in function couch_util:json_decode/2 (src/couch_util.erl, line 508)
in call from couch_js_tests:should_replace_broken_utf16/0 (test/eunit/couch_js_tests.erl, line 102)
**throw:{invalid_json,{11,invalid_string}}
  output:<<"">>

    couch_js_tests: couch_js_test_...<<208,188,208,176,208,188,208,176,32,208,188,209,139,208,187,208,176,32,209,128,208,176,208,188,209,131>>
[[[<<108,101,110,103,116,104>>,14]],[[<<115,117,98,115,116,114,105,110,103>>,<<208,188,209,139,208,187,208,176>>]],[[<<115,108,105,99,101>>,<<208,188,208,176,208,188,208,176>>]]]
[0.069 s] ok
    couch_js_tests: couch_js_test_...out of memory
[1.258 s] ok
```

Can you help?

wohali:
Finally, we need to review https://groups.google.com/forum/#!topic/mozilla.dev.tech.js-engine.internals/4crv5DNVuQk to ensure that our use of Realms vs. Containers holds up in this SM68 world.

src/couch/priv/couch_js/68/util.cpp
wohali: I'll note though that this is counter to something @kocolosk was working on to ensure flattening of all errors in `couch.log` to a single line, though I imagine we can "fix that in post" as they say.

davisp: This goes to stderr so it'd be in `/etc/sv/couchdb/logs/main/current` as opposed to what gets shipped out over rsyslog (assuming runit and that runit isn't configured for rsyslog).

davisp: Also that "invalid whitespace after JSON data" is that `std::string` constructor issue I was telling you about. Its because we've included the null byte in the length we gave the constructor instead of the pointer-only constructor that stops at the first null byte.

davisp:
Ah derps. I'll add a commit here to remove that debug logging. I'll dig into that test failure and the Compartment/Realm duality as well. I'm on ops so might have some idle time to think hard.

src/couch/priv/couch_js/68/util.cpp
wohali: So are you saying that I should leave it this way for SM68, or...?

davisp: I was saying I think we should not specify a length in the constructor, sorry.

wohali: Removed the length and it works fine.

wohali:
Length from the ctor removed, and L140 debugging output also removed.

Still need to fix the failing test case and confirm the Realm/Compartment partitioning.

davisp:
So a few updates. We didn't end up needing to use the js_pod_alloc thing so I didn't investigate it too much other than to say it appears related to the new zone concept that's vaguely referenced in places.

The Realm vs. Compartment situation is correct. The definitions are that a Compartment is a security layer and may have 1 or more Realms. This split was made so that a browser can share a single compartment for all iFrame's for a given remote host without incurring the cost of wrapping all shared objects. When we create new global objects we properly set the `setNewCompartmentAndZone()` option so that we have separate compartments for each sandbox.

The `std::move` discussion was resolved once I realized that the API of JS_NewUCString was changed from accepting a `char16_t*` to a `JS::UniqueTwoByteChars`. The version we have in 60 is correct when we extract the pointer and then tell the `UniquePtr` to not destroy its resource on success. In 68 the use of `std::move` is appropriate since we're transferring ownership correct (and no reset() is required, I'll make a note of this on the PR).

The failing test case is a bit of an awkward spot. The change that's happened is that in 68 there is now something magical that is automatically converting broken surrogate pairs into their equivalent `\uXXXX` escape codes. This happens before the UTF-8 encoding step somewhere and there does not appear to be any sort of configuration knob at least in the code path of converting strings. I've not been able to find references to when or how the change was made to see if we can sidestep it. I'm at a bit of a loss on how we might want to handle this change. The way I see it we can do one of a few things:

1. Figure out how to make SpiderMonkey stop doing this
2. Patch couch_query_servers and friends to return these as errors that map rows are discarded and/or reported to the client
3. Patch jiffy to do `0xFFFD` character replacement on "broken" `\uXXXX` escape sequences
4. Patch jiffy to have an option to not decode `\uXXXX` escape sequences and then do... something?
5. Light it all on fire

I'll keep poking at that test case as it appears to be the last major thing left.

src/couch/priv/couch_js/68/util.cpp
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    JSString* ret = JS_NewUCString(cx, std::move(utf16), len);
+
+    if(ret) {
+        // JS_NewUCString took ownership on success. We shift
+        // the resulting pointer into Unused to silence the
+        // compiler warning.
+        mozilla::Unused << utf16.release();
+    }
```
davisp, line 86: This can be removed and we can just `return JS_NewUCString(cx, std::move(utf16), len);` now. The reason for this is because now that `JS_NewUCString` accepts the `JS::UniqueTwoByteChars` smart pointer that the lifetime is guaranteed by `JS_NewUCString`.

wohali: Thanks, that was my hunch but your confirmation gives me permission to actually do it.

wohali:
I have cherry-picked @davisp 's fix for `couch_query_servers` that resolves the broken Unicode escapes.

All tests pass here and in CI.

ronag:
Any plans to apply this to https://github.com/apache/couchdb-docker?

wohali:
It should be in the `3.1.1` image already. We're not backporting to 1.x/2.x.

PR #2792, opened 2020-04-17
Make aegis into app and add key cache server

eiri:
## Overview

This is first PR which adds a none-blocking key cache. It uses shim for key manager which I will add in second PR and missing LRU parts which I'll add in third PR.

From point of functinality separation `aegis.erl` consist of public API and all crypto implementation, `aegis_key_cache.erl` is a private keys cache that manages workers and waiting lists and acts as a bridge to a Key Management system that interfaced through `aegis_key_manager.erl`

## Details

Key cache made to not block the main loop and it `spawn` or `spawn_monitor` all the incoming requests. In addition to ets cache it maintains three `dicts` - one for openers that requested new wrapped key or rewrap on existing key, second acting as a reverse index for those processes to make sure we have only one key request process per wrapped key in flight and third as a collection of waiting workloads to start once "their" key unwrapping finished and key inservet in the cache.

I'm including a link on a flow diagram I was using to plan how it works, I hope it helps:

https://bl.ocks.org/eiri/raw/5c166c1fef95e41835fd02df81f83c90/

## Testing

```
./configure --dev --key-manager=aegis_key_manager
make
make eunit apps=aegis
```


src/aegis/src/aegis_key_cache.erl
```diff
@@ -0,0 +1,306 @@
...
+
+    NewSt1 = case dict:take(WrappedKey, Unwrappers) of
+        {Ref, Unwrappers1} ->
+            ok = insert(Cache, WrappedKey, DbKey),
+            St#{unwrappers := Unwrappers1};
+        error when IsOpener ->
+            ok = insert(Cache, WrappedKey, DbKey),
+            St;
+        error ->
+            %% FIXME! it might be new wrapped key != old wrapped key
```
rnewson, line 137: ??

src/aegis/src/aegis_key_cache.erl
```diff
@@ -0,0 +1,306 @@
...
+                #{
+                    from := From,
+                    action := Action,
+                    args := Args
+                } = Waiter,
+                erlang:spawn(?MODULE, Action, [From, DbKey | Args])
+            end, WaitList),
+            NewSt1#{waiters := Waiters1};
+        error ->
+            NewSt1
```
rnewson, line 155: defensive programming.

src/aegis/src/aegis_key_cache.erl
```diff
@@ -0,0 +1,306 @@
...
+    end.
+
+
+do_decrypt(From, DbKey, #{uuid := UUID}, Key, Value) ->
+    process_flag(sensitive, true),
+    try
+        aegis:decrypt(DbKey, UUID, Key, Value)
+    of
+        Resp ->
+            gen_server:reply(From, Resp)
```
rnewson, line 220: So you reply to the client directly, cool.

src/aegis/src/aegis_key_manager.erl
```diff
@@ -0,0 +1,37 @@
...
+-module(aegis_key_manager).
+
+
+-export([
+    key_wrap/1,
+    key_unwrap/1
+]).
+
+
+-define(ROOT_KEY, <<1:256>>).
```
rnewson, line 22: this can't be here for GA.

src/couch/rebar.config.script
```diff
@@ -92,15 +92,6 @@ MD5Config = case lists:keyfind(erlang_md5, 1, CouchConfig) of
         []
 end,
 
-AegisConfig = case lists:keyfind(crypto_module, 1, CouchConfig) of
```
rnewson, line 95: why did this need to move?

src/aegis/src/aegis.erl
```diff
@@ -61,23 +60,25 @@ open(#{} = Db, Options) ->
...
     <<>>;
 
 encrypt(#{} = Db, Key, Value) when is_binary(Key), is_binary(Value) ->
-    #{
-        uuid := UUID,
-        aegis := DbKey
-    } = Db,
+    gen_server:call(?CACHE, {encrypt, Db, Key, Value}).
 
+encrypt(DbKey, UUID, Key, Value) ->
```
rnewson, line 81: do not like this. I'd rather move this code into aegis_key_cache, though I think aegis_server would be a better name, since it does more than cache keys.

src/aegis/src/aegis.erl
```diff
@@ -99,11 +100,9 @@ decrypt(#{} = _Db, _Key, <<>>) ->
     <<>>;
 
 decrypt(#{} = Db, Key, Value) when is_binary(Key), is_binary(Value) ->
-    #{
-        uuid := UUID,
-        aegis := DbKey
-    } = Db,
+    gen_server:call(?CACHE, {decrypt, Db, Key, Value}).
 
+decrypt(DbKey, UUID, Key, Value) ->
```
rnewson, line 105: ditto

src/aegis/src/aegis_key_cache.erl
```diff
@@ -0,0 +1,306 @@
...
+    end, ok, Openers),
+    ok.
+
+
+handle_call({get_wrapped_key, Db}, From, #{openers := Openers} = St) ->
+    {_Pid, Ref} = erlang:spawn_monitor(?MODULE, get_wrapped_key, [Db]),
+    Openers1 = dict:store(Ref, From, Openers),
+    {noreply, St#{openers := Openers1}, ?TIMEOUT};
+
+handle_call({maybe_rewrap_key, #{aegis := WrappedKey} = Db}, From, St) ->
```
rnewson, line 91: this is incomplete, right? nothing is currently rotating keys.

src/aegis/src/aegis_key_cache.erl
```diff
@@ -0,0 +1,306 @@
...
+code_change(_OldVsn, St, _Extra) ->
+    {ok, St}.
+
+
+%% workers functions
+
+get_wrapped_key(#{} = Db) ->
+    process_flag(sensitive, true),
+    try
+        ?AEGIS_KEY_MANAGER:key_wrap(Db)
```
rnewson, line 178: where's the custom behaviour declaration for an aegis key manager?

src/aegis/src/aegis_key_cache.erl
```diff
@@ -0,0 +1,306 @@
...
+    catch
+        _:Error ->
+            exit({key, {error, WrappedKey, Error}})
+    end.
+
+
+do_encrypt(From, DbKey, #{uuid := UUID}, Key, Value) ->
+    process_flag(sensitive, true),
+    try
+        aegis:encrypt(DbKey, UUID, Key, Value)
```
rnewson, line 204: the actual encryption should happen here (move code from aegis.erl)

src/aegis/src/aegis_key_cache.erl
```diff
@@ -0,0 +1,306 @@
...
+    } = St,
+
+    Reply = {error, Error},
+
+    NewSt = case dict:take(WrappedKey, Waiters) of
+        {WaitList, Waiters1} ->
+            [ gen_server:reply(From, Reply) || #{from := From} <- WaitList ],
+            St#{waiters := Waiters1};
+        error ->
+            St
```
rnewson, line 278: defensive programming

src/aegis/src/aegis_key_cache.erl
```diff
@@ -0,0 +1,306 @@
...
+    end,
+    maybe_reply(NewSt, Ref, Reply);
+
+maybe_reply(#{openers := Openers} = St, Ref, Resp) ->
+    case dict:take(Ref, Openers) of
+        {From, Openers1} ->
+            gen_server:reply(From, Resp),
+            St#{openers := Openers1};
+        error ->
+            St
```
rnewson, line 288: defensive programming

src/aegis/src/aegis.erl
```diff
@@ -14,44 +14,43 @@
...
         db_prefix := DbPrefix
     } = Db,
 
-    % Generate new key
-    DbKey = crypto:strong_rand_bytes(32),
-
-    % protect it with root key
-    WrappedKey = aegis_keywrap:key_wrap(?ROOT_KEY, DbKey),
+    % Fetch unwrapped key
+    WrappedKey = gen_server:call(?CACHE, {get_wrapped_key, Db}),
```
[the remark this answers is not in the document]
rnewson: also this should be more like `generate_key` as all key manager implementations are required to generate a new key per database.

src/aegis/src/aegis.erl
```diff
@@ -14,44 +14,43 @@
...
     decrypt/2,
     decrypt/3,
+    decrypt/4,
     encrypt/3,
+    encrypt/4,
     wrap_fold_fun/2
 ]).
 
-create(#{} = Db, Options) ->
+create(#{} = Db, _Options) ->
```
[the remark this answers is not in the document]
eiri: Sure. How this option should be defined, based on parameter in default.ini or something else?

src/couch/rebar.config.script
eiri: Macro set in confg `src/couch/rebar.config.script` defined only for modules in `couch` and aegis can't compile when it set here, failing with `macro ?AEGIS_KEY_MANAGER undefined`.

src/aegis/src/aegis_key_manager.erl
eiri: Yes, of course, I'll remove it once we discuss how default key manger should look like.

eiri:
@rnewson I believe I addressed all the comments, except how to handle `Option` in `aegis:create/2, `aegis:open/2` and how the default key manager should looks like, for which I need your advice. Can you take another look when you'll have time, please?

src/aegis/src/aegis.erl
```diff
@@ -36,58 +33,40 @@ create(#{} = Db, Options) ->
         db_prefix := DbPrefix
     } = Db,
 
-    % Generate new key
-    DbKey = crypto:strong_rand_bytes(32),
+    {ok, AegisConfig} = aegis_server:generate_key(Db, Options),
```
rnewson, line 36: `aegis_server:generate_key` to `aegis_server:init_db`?

src/aegis/src/aegis.erl
```diff
@@ -36,58 +33,40 @@ create(#{} = Db, Options) ->
...
 
-    % protect it with root key
-    WrappedKey = aegis_keywrap:key_wrap(?ROOT_KEY, DbKey),
-
-    % And store it
     FDBKey = erlfdb_tuple:pack(?WRAPPED_KEY, DbPrefix),
-    ok = erlfdb:set(Tx, FDBKey, WrappedKey),
+    Packed = erlfdb_tuple:pack({AegisConfig}),
```
rnewson, line 39: will this work for arbitrary AegisConfig values or is there an undocument constraint on its form?

src/aegis/src/aegis.erl
```diff
@@ -36,58 +33,40 @@ create(#{} = Db, Options) ->
...
     } = Db,
 
-    % Fetch wrapped key
     FDBKey = erlfdb_tuple:pack(?WRAPPED_KEY, DbPrefix),
-    WrappedKey = erlfdb:wait(erlfdb:get(Tx, FDBKey)),
-
-    % Unwrap it
-    DbKey = aegis_keywrap:key_unwrap(?ROOT_KEY, WrappedKey),
+    Packed = erlfdb:wait(erlfdb:get(Tx, FDBKey)),
+    {AegisConfig} = erlfdb_tuple:unpack(Packed),
```
rnewson, line 55: Perhaps it's the job of the key manager to do this? Not all key managers might need to read or write to erlfdb, so perhaps they should do their own persistence, and just return a value that aegis.erl puts in the the Db meta map, but does not save it or transform it at all.

src/aegis/src/aegis_example_key_manager.erl
```diff
@@ -0,0 +1,55 @@
...
+
+init() ->
+    case config:get("aegis_example_key_manager", "key_provider") of
+        undefined ->
+            erlang:error(misconfigured_key_provider);
+        KeyProvider ->
+            PassPhrase = os:cmd(KeyProvider),
+            {ok, RootKey} = couch_passwords:pbkdf2(
+                iolist_to_binary(PassPhrase), <<0:256>>, 10000, 32),
+            <<(binary_to_integer(RootKey, 16)):256>>
```
rnewson, line 35: even as an example not applying salt is a problem, I'm also confused by the construction of line 35, is it a reluctance to modify couch_passwords.erl? We really want sha256 involved in the derivation not sha1. Should I take on that piece?

src/aegis/src/aegis_key_manager.erl
```diff
@@ -0,0 +1,115 @@
...
+
+
+-callback init() -> key_manager_state().
+
+
+-callback generate_key(
+    St :: key_manager_state(),
+    Db :: #{},
+    DbOptions :: list()) ->
+        {ok, key(), aegis_config()} | {ok, false}.
```
rnewson, line 32: `| false` would be clearer?

src/aegis/src/aegis.erl
rnewson: e.g,

```
{ok, AegisConfig} = aegis_server:open_db(Db, Options),
Db#{
    aegis = AegisConfig
}
```

in aegis_server, we call the aegis key manager with the `Db` and `Options` and our one will read the appropriate value from fdb (containing the wrapped key).

At the moment we have a mixture of aegis and the key manager jointly responsible for a piece of data. The key manager isn't free to return what it likes, because aegis will try to pack it into a tuple, which will fail for many inputs.

src/aegis/src/aegis_server.erl
```diff
@@ -0,0 +1,338 @@
...
+
+maybe_spawn_worker(St, From, Action, Db, Key, Value) ->
+    #{
+        cache := Cache,
+        waiters := Waiters
+    } = St,
+
+    #{aegis := AegisConfig} = Db,
+
+    case lookup(Cache, AegisConfig) of
```
rnewson, line 292: using the entire config as the key in ets table highlights this opaque-but-not-really problem.

eiri: Agree. Let's use database's UUID as a key.

src/aegis/src/aegis_example_key_manager.erl
eiri: This is because `couch_passwords:pbkdf2/4` returns value in hex format, so this line converts it back.

I agree that modifying this in `couch_passwords.erl` is much cleaner. Unfortunately I'm not familiar with that part of codebase, so can I take you up on the offer, please?

src/aegis/src/aegis.erl
eiri: `erlfdb_tuple:pack/1,2` only accepts tuples (deep tuples) with elements been boolean, int, float or binary. This wrapping here to accommodate possible `false` from key manager.

src/aegis/src/aegis.erl
eiri: Very good point, let me try to do this to confirm there are no hidden constrains like breaking of original process boundary on erlfdb's transaction.

What do you think if we make `key_manager` to serialize `AegisConfig` value it's getting from `?AEGIS_KEY_MANAGER` callback with `term_to_binary`, so from the callback's perspective there will be no limitations on `AegisConfig` format?

src/aegis/src/aegis_example_key_manager.erl
rnewson: agreed on call to delete this module and come back to it after GA

src/aegis/src/aegis_server.erl
```diff
@@ -0,0 +1,271 @@
...
+
+do_open_db(#{} = Db) ->
+    #{
+        uuid := UUID,
+        user_ctx := UserCtx,
+        db_options := Options0
+    } = Db,
+
+    %% put back elements removed in fabric2_fdb:open/2
+    Options = [{uuid, UUID}, {user_ctx, UserCtx} | Options0],
```
rnewson, line 206: This doesn't appear to be necessary. both items are present by the time aegis:open_db is called. If it _is_ necessary, we should change the far end (fabric2_fdb:open) rather than depend on the current behaviour of its implementation.

eiri: This is subtle, this method not called by `aegis_server:open_db/2`, it called by `encrypt/decrypt/2` when they are re-retriving db key - they don't have `Options` available and have to restore it.

I can rename this method to something like `do_reopen_db/1` to clarify the intent and use-case.

src/aegis/src/aegis_server.erl
```diff
@@ -0,0 +1,271 @@
...
+    erlang:spawn(fun() ->
+        process_flag(sensitive, true),
+        try
+            do_encrypt(DbKey, Db, Key, Value)
+        of
+            Resp ->
+                gen_server:reply(From, Resp)
+        catch
+            _:Error ->
+                gen_server:reply(From, {error, Error})
```
rnewson, line 152: this needs to throw on the receiving end, yes? encrypt is expecting a binary() return and we get badargs if we sent a tuple

PR #2799, opened 2020-04-21
Add fold_docs for DocId list

garrensmith:
## Overview
Adds a fold_docs function that will fetch the docs for a list of
Doc Id's in parallel. This is used for _all_docs?keys=["id1", "id2"].

## Testing recommendations

All existing tests should pass

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -902,9 +902,10 @@ send_all_docs_keys(Db, #mrargs{} = Args, VAcc0) ->
         _ -> Args#mrargs.doc_options
     end,
     IncludeDocs = Args#mrargs.include_docs,
-    lists:foldl(fun(DocId, Acc) ->
-        OpenOpts = [deleted | DocOpts],
-        Row0 = case fabric2_db:open_doc(Db, DocId, OpenOpts) of
+    OpenOpts = [deleted | DocOpts],
+
+    CB = fun(DocId, Doc, Acc) ->
+        Row0 = case Doc of
             {not_found, missing} ->
                 #view_row{key = DocId};
             {ok, #doc{deleted = true, revs = Revs}} ->
@@ -938,9 +939,10 @@ send_all_docs_keys(Db, #mrargs{} = Args, VAcc0) ->
                 }
         end,
         Row1 = fabric_view:transform_row(Row0),
-        {ok, NewAcc} = view_cb(Row1, Acc),
-        NewAcc
-    end, VAcc1, Keys).
+        view_cb(Row1, Acc)
+    end,
+    {ok, VAcc2} = fabric2_db:fold_docs(Db, Keys, CB, VAcc1, OpenOpts),
+    VAcc2.
 
 
 apply_args_to_keylist(Args, Keys0) ->
--- a/src/fabric/src/fabric2_db.erl
+++ b/src/fabric/src/fabric2_db.erl
@@ -110,6 +110,7 @@
 
     fold_docs/3,
     fold_docs/4,
+    fold_docs/5,
     fold_design_docs/4,
     fold_local_docs/4,
     fold_changes/4,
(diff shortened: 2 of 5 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -902,9 +902,10 @@ send_all_docs_keys(Db, #mrargs{} = Args, VAcc0) ->
         _ -> Args#mrargs.doc_options
     end,
     IncludeDocs = Args#mrargs.include_docs,
-    lists:foldl(fun(DocId, Acc) ->
-        OpenOpts = [deleted | DocOpts],
-        Row0 = case fabric2_db:open_doc(Db, DocId, OpenOpts) of
+    OpenOpts = [deleted | DocOpts],
+
+    CB = fun({doc, DocId, Doc}, Acc) ->
```
davisp, line 907: This `{doc, DocId, Doc}` seems a bit awkward. Maybe better as `fun(DocId, Doc, Acc) -> ... end`?

src/fabric/src/fabric2_db.erl
```diff
@@ -969,6 +970,145 @@ fold_docs(Db, UserFun, UserAcc0, Options) ->
...
+fold_docs(Db, DocIds, UserFun, UserAcc0, Options) ->
+    fabric2_fdb:transactional(Db, fun(TxDb) ->
+        try
+            NeedsTreeOpts = [revs_info, conflicts, deleted_conflicts],
+            NeedsTree = (Options -- NeedsTreeOpts /= Options),
+
+            FetchRevs = case NeedsTree of
+                true ->
+                    fun(DocId) ->
+                        fabric2_fdb:get_all_revs_future(TxDb, DocId)
```
davisp, line 982: I don't think you've included an implementation of `get_all_revs_future` in fabric2_fdb unless I'm really blind. I'd probably add it as part of the refactoring commit.

src/fabric/src/fabric2_db.erl
```diff
@@ -969,6 +970,145 @@ fold_docs(Db, UserFun, UserAcc0, Options) ->
...
+            } = FinalAcc3,
+            {ok, FinalUserAcc}
+
+        catch throw:{stop, StopUserAcc} ->
+            {ok, StopUserAcc}
+        end
+    end).
+
+
+drain_fold_docs_revs_futures(_TxDb, #{revs_count := C} = Acc) when C < 100 ->
```
davisp, line 1026: All of these support functions should be moved to the end of the module. I'd put them just after `drain_all_deleted_info_futures`.

test/elixir/test/all_docs_test.exs
```diff
@@ -242,13 +262,17 @@ defmodule AllDocsTest do
     resp = Couch.post(
       "/#{db_name}/_all_docs",
       body: %{
-        :keys => [1, 2],
-        :limit => 1
+        :keys => ["1", "2"],
```
davisp, line 265: Why did we have to change the keys to strings here? Is there an API change here that we should make a note of?

davisp: I notice the previous test is still using integer keys so I assume those still work. The addition of testing include_docs is nice, but I'd prefer to not change unrelated bits if possible.

davisp:
Also, it looks like we're only using `get_winning_revs_wait` in four spots. We should probably change those to `get_revs_wait` instead of keeping the `get_winning_revs_wait` alias function. That'll help keep cruft to a minimum rather than letting it accumulate slowly.

src/chttpd/src/chttpd_db.erl
garrensmith: Fixed

src/fabric/src/fabric2_db.erl
garrensmith: Good spot. Fixed

src/fabric/src/fabric2_db.erl
garrensmith: Fixed

test/elixir/test/all_docs_test.exs
garrensmith: The original test was just returning `%{"error" => "not_found", "key" => 1}` which is not quite what we should be testing. We should be testing we return the document for that key. So I've added in the test to check that we return an error for the incorrect key and a proper test to prove we return keys. No functionality changed. Just improving the tests.

davisp: Ah! Sounds good. That certainly wasn't testing what I thought it was testing by reading it.

PR #2800, opened 2020-04-21
Distinguish updates

iilyak:
## Overview

This PR allow us to distinguish between new doc and update of the existent doc for tracing.

## Testing recommendations

1. update rel/overlay/etc/default.ini as follows
```
rel/overlay/etc/default.ini
[tracing]
enabled = true ; true | false
app_name = couchdb ; value to use for the `location.application` tag
protocol = http ; udp | http - which reporter to use
endpoint = http://127.0.0.1:14268
 
[tracing.filters]
all = (#{}) -> true
```
2. start jaeger in docker container
```
docker run -d -p 6831:6831/udp -p 16686:16686 -p 14268:14268 jaegertracing/all-in-one:1.14
```
3. Start couchdb `dev/run --admin=adm:pass --no-join`
4. Create database curl -X PUT -u adm:pass http://127.0.0.1:15984/test
5. Create document `curl -X PUT -u adm:pass http://127.0.0.1:15984/test -H Content-Type:application/json  -d '{"_id":"doc1"}'`
6. Update document `curl -X PUT -u adm:pass "http://127.0.0.1:15984/test/doc1?rev=8-4f98b201f7a6f147a952fe70651f0a95" -H Content-Type:application/json  -d '{"_id":"doc1","val": 4}'`
7. Update document `curl -X PUT -u adm:pass "http://127.0.0.1:15984/test/doc1" -H Content-Type:application/json  -d '{"_id":"doc1","dfd": 4, "_rev": "11-fdcab8b70c87eb64255f504a5d983ab0"}'`
8. Open http://localhost:16686/ in the browser and search for spans (you should find some traces)
   - the first event should have action equal to `db.doc.write`
   - the second event should have action equal to `db.doc.update`
   - the third event should have action equal to `db.doc.update`

N/A

## Checklist


src/chttpd/src/chttpd_httpd_handlers.erl
```diff
@@ -328,8 +338,14 @@ handler_info('GET', [Db, <<"_design">>, Name | Path], _) ->
         'attachment.name' => filename:join(Path)
     }};
 
-handler_info('PUT', [Db, <<"_design">>, Name | Path], _) ->
-    {'db.design.doc.attachment.write', #{
+handler_info('PUT', [Db, <<"_design">>, Name | Path], Req) ->
+    ActionName = case chttpd:qs_value(Req, "rev") of
```
chewbranca, line 342: So I thought about just checking the rev like this, but the problem here is that for PUT and POST, the rev can be embed'ed in the request body and is not guaranteed to be in the query string. I looked at `chttpd:qs_value` and that appears to just be looking at the query string params, so I don't think this is currently sufficient to properly tag these as write vs update.

jaydoane: I think `_bulk_docs` may have similar issues, where you can both add new docs and update existing docs in the same request, and would presumably need to inspect each element of the `docs` field to know what's actually going on?

iilyak: @chewbranca updated

src/chttpd/src/chttpd_httpd_handlers.erl
```diff
@@ -233,11 +233,21 @@ handler_info('POST', [Db, <<"_compact">>], _) ->
...
+        undefined ->
+            {'db.design.doc.write', #{'db.name' => Db, 'design.id' => Name}};
+        _ ->
+            {'db.design.doc.update', #{'db.name' => Db, 'design.id' => Name}}
+    end;
 
-handler_info('PUT', [Db, <<"_design">>, Name], _) ->
-    {'db.design.doc.write', #{'db.name' => Db, 'design.id' => Name}};
+handler_info('PUT', [Db, <<"_design">>, Name], Req) ->
+    case chttpd:qs_value(Req, "rev") of
```
chewbranca, line 245: Should these (ddoc PUT and POST) be updated as well with `has_rev`?

iilyak: missed that

src/chttpd/src/chttpd_httpd_handlers.erl
iilyak: @jaydoane This is getting into expensive call territory. It would force us to traverse docs multiple times.

jaydoane: @iilyak I think it's just another example of `_bulk_docs` being a lousy API, and I agree it's expensive. However, each request in principle can do a lot of work, so the upfront cost _might_ be worth it for correctness? Dunno, tbh.

PR #2825, opened 2020-04-26
Improve robustness of couch expiring cache test

jaydoane:
## Overview

In its current incarnation, the so-called "simple lifecycle" test is
prone to numerous failures in the CI system [1], doubtless because it's
riddled with race conditions. The original author makes many assumptions
about how quickly an (actual, unmocked) FDB instance will respond to a
request.

The primary goal is to stop failing CI builds, while other
considerations include: keeping the run time of the test as low as
possible, keeping the code coverage high, and documenting the known
races.

Specifically:
- Increase the `stale` and `expired` times by a factor of 5 to decrease
sensitivity to poor FDB performance.
- Change default timer from `erlang:system_time/1` to `os:timestamp` on
the assumption that the latter is less prone to warping [2].
- Decrease the period of the cache server reaper by half to increase
accuracy of eviction time.
- Inline and modify the `test_util:wait` code to make the timer
explicit, and emphasize that `timer:delay/1` only works with millisecond
resolution.
- Don't fail the test if it can't get a fresh lookup immediately after
insertion, but let it continue on to the next race, at least to the
point of expiration and deletion, which continue to be asserted.
- Factor `Timeout` and `Interval` to allow declarations near the other
hard-coded parameters.
- Move cache server `Opts` into `setup/0` and eliminate `start_link/0`.
- Double the overall test timeout to 20 seconds.

This has soaked for hundreds of runs on a 5 year old laptop, but the
real test is the CI system.

Should this test continue to fail CI builds, additional improvements
could include mocking the timer and/or FDB layer to eliminate the
variability of an integrated system.

[1] https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FPullRequests/detail/PR-2813/10/pipeline
[2] http://erlang.org/doc/apps/erts/time_correction.html#terminology

## Testing recommendations

```sh
#!/bin/bash -ex

for run in {1..100}
do
    echo $run
    make eunit apps=couch_expiring_cache
done
```

## Checklist

### DIFF
--- a/src/couch_expiring_cache/src/couch_expiring_cache_server.erl
+++ b/src/couch_expiring_cache/src/couch_expiring_cache_server.erl
@@ -17,6 +17,7 @@
 -callback start_link() -> {ok, pid()} | ignore | {error, term()}.
 
 -export([
+    now_ts/0,
     start_link/2
 ]).
 
@@ -80,10 +81,10 @@ handle_info(remove_expired, St) ->
         largest_elapsed := LargestElapsed
     } = St,
 
-    NowTS = erlang:system_time(?TIME_UNIT),
+    NowTS = now_ts(),
     OldestTS = max(OldestTS0,
         couch_expiring_cache_fdb:clear_range_to(Name, NowTS, BatchSize)),
-    Elapsed = erlang:system_time(?TIME_UNIT) - NowTS,
+    Elapsed = now_ts() - NowTS,
 
     {noreply, St#{
         timer_ref := schedule_remove_expired(Period, MaxJitter),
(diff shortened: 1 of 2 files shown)


jaydoane:
I ran CI 20 times, and not once did `couch_expiring_cache_tests` fail the build, although others did. Just to call out the failures, they were:
```
fabric2_index_tests:29: with (ignore_deleted_dbs)...*timed out*
```
and
```
  * test send error for failed indexing (55075.0ms)
[2020-04-27T06:49:11.934Z] 
[2020-04-27T06:49:11.934Z]   1) test send error for failed indexing (ViewMapTest)
[2020-04-27T06:49:11.934Z]      test/elixir/test/map_test.exs:506
[2020-04-27T06:49:11.934Z]      ** (KeyError) key :status_code not found in: %HTTPotion.ErrorResponse{message: "req_timedout"}
[2020-04-27T06:49:11.934Z]      code: assert resp.status_code == 500
[2020-04-27T06:49:11.934Z]      stacktrace:
[2020-04-27T06:49:11.934Z]        test/elixir/test/map_test.exs:533: (test)
[2020-04-27T06:49:11.934Z] 
```
the latter test failed twice

PR #2826, opened 2020-04-27
Add native encryption support

rnewson:
A new application, aegis, is introduced to provide strong at-rest
protection of CouchDB data (where possible).

Currently we encrypt the following values (if enabled):

1. Document content
2. Attachment content
3. Index values

Things not encrypted:

1. _all_docs
2. _changes
3. doc id
4. doc rev
5. Index keys
6. All other metadata

Co-Authored-By: Eric Avdey <eiri@apache.org>
Co-Authored-By: Robert Samuel Newson <rnewson@apache.org>

### DIFF
--- a/configure
+++ b/configure
@@ -96,6 +96,24 @@ parse_opts() {
                 continue
                 ;;
 
+            --key-manager)
+                if [ -n "$2" ]; then
+                    eval AEGIS_KEY_MANAGER=$2
+                    shift 2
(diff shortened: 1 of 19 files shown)


rnewson:
Eric and I have been collaborating and reviewing this extensively for a while now, lots of commits and branches (all in couchdb repo history), so this is already reviewed work.

PR #2870, opened 2020-05-05
Pagination API

iilyak:
## Overview

This PR implements a pagination API proposed in [RFC](https://github.com/apache/couchdb-documentation/pull/530).

## Testing recommendations

Running `make check` should be sufficient.

You can also test manually using:

```
# run test
make eunit apps=couch_mrview tests=check_completion_test
make exunit tests=src/chttpd/test/exunit/pagination_test.exs
# create database
curl -u adm:pass -X PUT "http://127.0.0.1:15984/test"
# create documents
curl -u adm:pass -X PUT "http://127.0.0.1:15984/test/foo" -d '{"value": "foo"}' | jq '.'
curl -u adm:pass -X PUT "http://127.0.0.1:15984/test/bar" -d '{"value": "bar"}' | jq '.'
curl -u adm:pass -X PUT "http://127.0.0.1:15984/test/baz" -d '{"value": "baz"}' | jq '.'
# query without pagination API
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?limit=1" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?keys=%5B%22foo%22%2C%20%22bar%22%5D" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?limit=1&discending=true" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?limit=1&start_key=%22bar%22" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?limit=1&startkey_docid=%22doc1%22" | jq '.'
# query using pagination API
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?page_size=1" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?bookmark={bookmark}" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?keys=%5B%22foo%22%2C%20%22bar%22%5D&page_size=1" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?bookmark={bookmark}" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?page_size=1&discending=true" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?bookmark={bookmark}" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?page_size=1&start_key=%22bar%22" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?bookmark={bookmark}" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?page_size=1&startkey_docid=%22doc1%22" | jq '.'
curl -u adm:pass "http://127.0.0.1:15984/test/_all_docs?bookmark={bookmark}" | jq '.'
```

- [RFC](https://github.com/apache/couchdb-documentation/pull/530)

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -831,10 +831,10 @@ multi_all_docs_view(Req, Db, OP, Queries) ->
         prepend = "\r\n"
     },
     VAcc1 = lists:foldl(fun
-        (#mrargs{keys = undefined} = Args, Acc0) ->
-            send_all_docs(Db, Args, Acc0);
-        (#mrargs{keys = Keys} = Args, Acc0) when is_list(Keys) ->
-            Acc1 = send_all_docs_keys(Db, Args, Acc0),
+        (#mrargs{keys = undefined} = ArgsIn, Acc0) ->
+            send_all_docs(Db, ArgsIn, Acc0);
+        (#mrargs{keys = Keys} = ArgsIn, Acc0) when is_list(Keys) ->
+            Acc1 = send_all_docs_keys(Db, ArgsIn, Acc0),
             {ok, Acc2} = view_cb(complete, Acc1),
             Acc2
     end, VAcc0, ArgQueries),
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -812,14 +812,20 @@ db_req(#httpd{path_parts=[_, DocId | FileNameParts]}=Req, Db) ->
     db_attachment_req(Req, Db, DocId, FileNameParts).
 
 multi_all_docs_view(Req, Db, OP, Queries) ->
-    Args0 = couch_mrview_http:parse_params(Req, undefined),
+    Args = couch_views_http:parse_params(Req, undefined),
+    case couch_views_util:is_paginated(Args) of
+        false ->
+            stream_multi_all_docs_view(Req, Db, OP, Args, Queries);
+        true ->
+            paginate_multi_all_docs_view(Req, Db, OP, Args, Queries)
+    end.
+
+
+stream_multi_all_docs_view(Req, Db, OP, Args0, Queries) ->
     Args1 = Args0#mrargs{view_type=map},
-    ArgQueries = lists:map(fun({Query}) ->
-        QueryArg1 = couch_mrview_http:parse_params(Query, undefined,
-            Args1, [decoded]),
-        QueryArgs2 = couch_views_util:validate_args(QueryArg1),
-        set_namespace(OP, QueryArgs2)
-    end, Queries),
+    ArgQueries = chttpd_view:parse_queries(Req, Args1, Queries, fun(QArgs) ->
+        set_namespace(OP, QArgs)
+    end),
     Max = chttpd:chunked_response_buffer_size(),
     First = "{\"results\":[",
     {ok, Resp0} = chttpd:start_delayed_json_response(Req, 200, [], First),
@@ -842,8 +848,34 @@ multi_all_docs_view(Req, Db, OP, Queries) ->
     chttpd:end_delayed_json_response(Resp1).
 
 
+paginate_multi_all_docs_view(Req, Db, OP, Args0, Queries) ->
+    Args1 = Args0#mrargs{view_type=map},
+    ArgQueries = chttpd_view:parse_queries(Req, Args1, Queries, fun(QArgs) ->
+        set_namespace(OP, QArgs)
+    end),
+    KeyFun = fun({Props}) -> couch_util:get_value(id, Props) end,
+    #mrargs{page_size = PageSize} = Args0,
+    #httpd{path_parts = Parts} = Req,
+    UpdateSeq = fabric2_db:get_update_seq(Db),
+    EtagTerm = {Parts, UpdateSeq, Args0},
+    Response = couch_views_http:paginated(
+        Req, EtagTerm, PageSize, ArgQueries, KeyFun,
+        fun(Args) ->
+            all_docs_paginated_cb(Db, Args)
+        end),
+    chttpd:send_json(Req, Response).
+
+
 all_docs_view(Req, Db, Keys, OP) ->
-    Args0 = couch_mrview_http:parse_body_and_query(Req, Keys),
+    Args = couch_views_http:parse_body_and_query(Req, Keys),
+    case couch_views_util:is_paginated(Args) of
+        false ->
+            stream_all_docs_view(Req, Db, Args, OP);
+        true ->
+            paginate_all_docs_view(Req, Db, Args, OP)
+    end.
+
+stream_all_docs_view(Req, Db, Args0, OP) ->
     Args1 = Args0#mrargs{view_type=map},
     Args2 = couch_views_util:validate_args(Args1),
     Args3 = set_namespace(OP, Args2),
@@ -864,14 +896,45 @@ all_docs_view(Req, Db, Keys, OP) ->
     end.
 
 
+paginate_all_docs_view(Req, Db, Args0, OP) ->
+    Args1 = Args0#mrargs{view_type=map},
+    Args2 = chttpd_view:validate_args(Req, Args1),
+    Args3 = set_namespace(OP, Args2),
+    KeyFun = fun({Props}) -> couch_util:get_value(id, Props) end,
+    #httpd{path_parts = Parts} = Req,
+    UpdateSeq = fabric2_db:get_update_seq(Db),
+    EtagTerm = {Parts, UpdateSeq, Args3},
+    Response = couch_views_http:paginated(
+        Req, EtagTerm, Args3, KeyFun,
+        fun(Args) ->
+            all_docs_paginated_cb(Db, Args)
+        end),
+    chttpd:send_json(Req, Response).
+
+
+all_docs_paginated_cb(Db, Args) ->
+    #vacc{meta=MetaMap, buffer=Items} = case Args#mrargs.keys of
+        undefined ->
+            send_all_docs(Db, Args, #vacc{paginated=true});
+        Keys when is_list(Keys) ->
+            send_all_docs_keys(Db, Args, #vacc{paginated=true})
+    end,
+    {MetaMap, Items}.
+
+
 send_all_docs(Db, #mrargs{keys = undefined} = Args, VAcc0) ->
     Opts0 = fabric2_util:all_docs_view_opts(Args),
-    Opts = Opts0 ++ [{restart_tx, true}],
-    NS = couch_util:get_value(namespace, Opts),
+    NS = couch_util:get_value(namespace, Opts0),
     FoldFun = case NS of
         <<"_all_docs">> -> fold_docs;
         <<"_design">> -> fold_design_docs;
         <<"_local">> -> fold_local_docs
+    end,
+    Opts = case couch_views_util:is_paginated(Args) of
+        false ->
+            Opts0 ++ [{restart_tx, true}];
+        true ->
+            Opts0
     end,
     ViewCb = fun view_cb/2,
     Acc = {iter, Db, Args, VAcc0},
@@ -980,25 +1043,15 @@ view_cb({row, Row}, {iter, Db, Args, VAcc}) ->
             Row
     end,
     chttpd_stats:incr_rows(),
-    {Go, NewVAcc} = couch_mrview_http:view_cb({row, NewRow}, VAcc),
+    {Go, NewVAcc} = couch_views_http:view_cb({row, NewRow}, VAcc),
     {Go, {iter, Db, Args, NewVAcc}};
 
 view_cb(Msg, {iter, Db, Args, VAcc}) ->
-    {Go, NewVAcc} = couch_mrview_http:view_cb(Msg, VAcc),
+    {Go, NewVAcc} = couch_views_http:view_cb(Msg, VAcc),
     {Go, {iter, Db, Args, NewVAcc}};
 
-view_cb({row, Row} = Msg, Acc) ->
-    case lists:keymember(doc, 1, Row) of
-        true ->
-            chttpd_stats:incr_reads();
-        false ->
-            ok
-    end,
-    chttpd_stats:incr_rows(),
-    couch_mrview_http:view_cb(Msg, Acc);
-
 view_cb(Msg, Acc) ->
-    couch_mrview_http:view_cb(Msg, Acc).
+    couch_views_http:view_cb(Msg, Acc).
 
 db_doc_req(#httpd{method='DELETE'}=Req, Db, DocId) ->
     % check for the existence of the doc to handle the 404 case.
--- a/src/chttpd/test/exunit/pagination_test.exs
+++ b/src/chttpd/test/exunit/pagination_test.exs
@@ -68,6 +68,52 @@ defmodule Couch.Test.Pagination do
     %{view_name: "all", ddoc_id: ddoc_id}
   end
 
+  defp all_docs(context) do
+    assert Map.has_key?(context, :page_size), "Please define '@describetag page_size: 4'"
+
+    assert Map.has_key?(context, :descending),
+           "Please define '@describetag descending: false'"
+
+    resp =
+      Couch.Session.get(context.session, "/#{context.db_name}/_all_docs",
+        query: %{page_size: context.page_size, descending: context.descending}
+      )
+
+    assert resp.status_code == 200, "got error #{inspect(resp.body)}"
+
+    %{
+      response: resp.body
+    }
+  end
+
+  defp paginate(context) do
+    if Map.has_key?(context.response, "next") do
+      bookmark = context.response["next"]
+      pages = Map.get(context, :pages, [context.response])
+      assert length(pages) < div(context.n_docs, context.page_size) + 1
+
+      resp =
+        Couch.Session.get(context.session, "/#{context.db_name}/_all_docs",
+          query: %{bookmark: bookmark}
+        )
+
+      context =
+        Map.merge(context, %{
+          pages: [resp.body | pages],
+          response: resp.body
+        })
+
+      paginate(context)
+    else
+      context =
+        Map.update(context, :pages, [], fn acc ->
+          Enum.reverse(acc)
+        end)
+
+      context
+    end
+  end
+
   def create_db(session, db_name, opts \\ []) do
     retry_until(fn ->
       resp = Couch.Session.put(session, "/#{db_name}", opts)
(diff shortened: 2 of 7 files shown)


iilyak:
The `previous` page is not yet implemented.

iilyak:
I'll rename `bookmark` key into `next` (as per RFC) in one of the commits to follow. Also `prev` wouldn't be implemented in this PR and would be added latter.

src/chttpd/src/chttpd_db.erl
```diff
@@ -839,58 +843,81 @@ multi_all_docs_view(Req, Db, OP, Queries) ->
...
+    IsPaginated = couch_mrview_http:is_paginated(Args),
+    {ViewCb, Opts, Acc1} = case IsPaginated of
+        false ->
+            {
+                fun streaming_cb/2,
+                Opts0 ++ [{restart_tx, true}],
+                {iter, Db, Args, Acc0}
+            };
+        true ->
+            {fun paginated_cb/2, Opts0, {iter, Db, Args, Acc0}}
```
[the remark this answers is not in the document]
iilyak: the formatting style for Elixir code is automatic and done via `mix format <file>` in order to satisfy credo linter. I'll try and see if `mix format` would override my change.

src/chttpd/test/exunit/pagination_test.exs
```diff
@@ -0,0 +1,255 @@
+defmodule Couch.Test.Pagination do
+  use ExUnit.Case
+  import Couch.DBTest, only: [retry_until: 1]
+  alias Couch.DBTest, as: Utils
```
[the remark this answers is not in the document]
iilyak: I was not sure how compete our test coverage in the aspects I wanted to make sure work after refactoring. I can remove legacy API tests if you prefer.

iilyak:
Not implemented parts of the spec are:
- `first`/`previous` bookamarks are missing
- return 400 when we reach max limit for streaming replies (It is unclear how to implement it yet)

src/chttpd/test/exunit/pagination_test.exs
```diff
@@ -657,4 +657,278 @@ defmodule Couch.Test.Pagination do
...
+              "/#{ctx.db_name}/_design/#{ctx.ddoc_id}/_view/#{ctx.view_name}",
+              query: %{page_size: ctx.page_size, descending: ctx.descending}
+            )
+
+          assert resp.status_code == 200
+          body = resp.body
+          assert body["total_rows"] == length(body["rows"])
+        end
+
+        test "total_rows matches the requested page_size", ctx do
```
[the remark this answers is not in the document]
iilyak: How strong do you feel about it?

src/chttpd/src/chttpd_db.erl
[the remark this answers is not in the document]
iilyak: fixed

src/chttpd/src/chttpd_db.erl
```diff
@@ -839,58 +843,81 @@ multi_all_docs_view(Req, Db, OP, Queries) ->
...
+    IsPaginated = couch_mrview_http:is_paginated(Args),
     Meta = case Args#mrargs.update_seq of
         true ->
             UpdateSeq = fabric2_db:get_update_seq(Db),
             [{update_seq, UpdateSeq}];
         false ->
             []
     end ++ [{total, TotalRows}, {offset, null}],
-    {ok, VAcc1} = streaming_cb({meta, Meta}, VAcc0),
+    {ok, Acc1} = case IsPaginated of
```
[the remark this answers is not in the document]
iilyak: fixed

iilyak:
I am not sure where to open a documentation update PR. We don't have FDB releated integration branch on https://github.com/apache/couchdb-documentation

garrensmith:
@iilyak I think we can do the documentation PR later once we closer to CouchDB 4.x release

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -380,22 +389,49 @@ view_cb({meta, Meta}, #vacc{meta_sent=false, row_sent=false}=Acc) ->
...
+            {total, _Value} ->
+                %% We set total_rows elsewere
+                Acc;
+            {Key, null} ->
+                maps:put(
+                    list_to_binary(atom_to_list(Key)), <<"null">>, Acc);
+            {Key, Value} ->
+                maps:put(list_to_binary(atom_to_list(Key)), Value, Acc)
+            end
+        end, #{}, Meta),
```
davisp, line 431: Indents are wrong on these two lines.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -380,22 +389,49 @@ view_cb({meta, Meta}, #vacc{meta_sent=false, row_sent=false}=Acc) ->
...
+    MetaMap = lists:foldl(fun(MetaData, Acc) ->
+        case MetaData of
+            {_Key, undefined} ->
+                Acc;
+            {total, _Value} ->
+                %% We set total_rows elsewere
+                Acc;
+            {Key, null} ->
+                maps:put(
+                    list_to_binary(atom_to_list(Key)), <<"null">>, Acc);
```
davisp, line 427: You're converting a `null` to `"null"` here which seems wrong. That is to say, you've gone from the `null` terminal type, to a string with the letters null.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -467,7 +503,13 @@ parse_params(Props, Keys) ->
 parse_params(Props, Keys, Args) ->
     parse_params(Props, Keys, Args, []).
 
+parse_params([{"bookmark", Bookmark}], _Keys, #mrargs{}, _Options) ->
+    bookmark_decode(Bookmark);
+
 parse_params(Props, Keys, #mrargs{}=Args0, Options) ->
+    Bookmark = couch_util:get_value("bookmark", Props, nil),
+    Bookmark == nil
+        orelse throw({error, "Cannot use bookmark with other options"}),
```
davisp, line 512: This is awfully corner case-ish. How about rewriting with our more standard:

```erlang
if Bookmark == nil -> ok; true ->
    throw({error, ...})
end
```

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -709,3 +750,169 @@ maybe_set_page_size(#mrargs{page_size = undefined} = Args, MaxPageSize) ->
...
+                {Acc, Idx + 1};
+            {[], Default} ->
+                {setelement(Idx, Acc, Default), Idx + 1};
+            {_, _} when Idx == #mrargs.bookmark ->
+                {setelement(Idx, Acc, true), Idx + 1};
+            {Value, _} ->
+                {setelement(Idx, Acc, Value), Idx + 1}
+        end
+    end, {#mrargs{}, 1}, [mrargs | tuple_to_list(Term)]),
+    Args.
```
davisp, line 806: First things first, this definitely needs to have a versioned field. Even if we don't offer upgrade code (which is fine cause I think we'd all agree that storing bookmarks is not an intended use case) we'll still want to reject with a smarter error message rather than breaking weirdly or worse, responding with data that's misinterpreted.

I'm also not sold on the tuple-to-record implicit conversion. Given that an `mrargs` record is 24 fields, I'd think that most cases would likely be a bit smaller if we encoded the index for non-default values. There's some examples of this sort of approach in `fabric.erl` in the `kl_to_record` function. Though instead of atoms I'd spit out the actual field index so there's no lookup on decode.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -709,3 +750,169 @@ maybe_set_page_size(#mrargs{page_size = undefined} = Args, MaxPageSize) ->
...
+    case check_completion(Limit, Items) of
+        {Rows, nil} ->
+            maps:merge(Response, #{
+                rows => Rows,
+                total_rows => length(Rows)
+             });
+        {Rows, Next} ->
+            NextKey = KeyFun(Next),
+            is_binary(NextKey)
+                orelse throw("Provided KeyFun should return binary"),
```
davisp, line 872: Another place. I'd rather use our normal `if` construct for this rather than requiring everyone to remember the subtle specifics of `orelse`.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -709,3 +750,169 @@ maybe_set_page_size(#mrargs{page_size = undefined} = Args, MaxPageSize) ->
...
+
+maybe_add_bookmark(PageSize, Args0, Response, Items, KeyFun) ->
+    #mrargs{page_size = Limit} = Args0,
+    Args = Args0#mrargs{page_size = PageSize},
+    case check_completion(Limit, Items) of
+        {Rows, nil} ->
+            maps:merge(Response, #{
+                rows => Rows,
+                total_rows => length(Rows)
+             });
```
davisp, line 868: You've got an extra space here.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -709,3 +750,169 @@ maybe_set_page_size(#mrargs{page_size = undefined} = Args, MaxPageSize) ->
...
+                    NewKeys = lists:foldl(fun(Row, Acc) ->
+                        gb_sets:del_element(KeyFun(Row), Acc)
+                    end, gb_sets:from_list(Keys), Rows),
+                    bookmark_encode(Args#mrargs{keys=gb_sets:to_list(NewKeys)})
+                end,
+            maps:merge(Response, #{
+                rows => Rows,
+                next => Bookmark,
+                total_rows => length(Rows)
+             })
```
davisp, line 886: Extra space here it looks like as well.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -709,3 +750,169 @@ maybe_set_page_size(#mrargs{page_size = undefined} = Args, MaxPageSize) ->
...
+            is_binary(NextKey)
+                orelse throw("Provided KeyFun should return binary"),
+            Bookmark = case Args#mrargs.keys of
+                undefined ->
+                    bookmark_encode(Args#mrargs{start_key=NextKey});
+                Keys when is_list(Keys) ->
+                    NewKeys = lists:foldl(fun(Row, Acc) ->
+                        gb_sets:del_element(KeyFun(Row), Acc)
+                    end, gb_sets:from_list(Keys), Rows),
+                    bookmark_encode(Args#mrargs{keys=gb_sets:to_list(NewKeys)})
```
davisp, line 880: A couple questions here. First, why are we bookmarking pages of POST'ed keys? Seems like that's more of a 400 error sort of situation. Paging through a list of keys that the client has in RAM seems like a bit of a waste. I guess I could be convinced if it makes client libraries easier to implement but I'm really not seeing how that'd be the case off the top of my head.

Second, the gb_sets approach seems not right here. First, I'm not sure I get why you've gone with gb_sets here over sets or ordsets even. At first I thought it might retain ordering of the input list but that appears to not be the case:

```erlang
2> gb_sets:to_list(gb_sets:from_list([a, b, c, d])).
[a,b,c,d]
3> gb_sets:to_list(gb_sets:from_list([a, c, d, b])).
[a,b,c,d]
```

Secondly, this seems to be also changing the API contract as you're deleting keys from a set. My memory is a bit hazy, but I'm pretty sure if a user provides a docid/key twice then we return that row twice. And we retain their ordering (subject to query string parameters like descending/skip/limit).

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -709,3 +750,169 @@ maybe_set_page_size(#mrargs{page_size = undefined} = Args, MaxPageSize) ->
...
+
+
+paginated(Req, EtagTerm, PageSize, QueriesArgs, KeyFun, Fun) when is_list(QueriesArgs) ->
+    Etag = couch_httpd:make_etag(EtagTerm),
+    chttpd:etag_respond(Req, Etag, fun() ->
+        Results = paginated(PageSize, QueriesArgs, KeyFun, Fun),
+        #{results => Results}
+    end).
+
+paginated(PageSize, QueriesArgs, KeyFun, Fun) when is_list(QueriesArgs) ->
```
davisp: These three functions using the same name makes things surprisingly difficult to comprehend without having to stare at the arguments and work back why things would be different.

davisp: If I'm reading this right, in the case of a multi-query HTTP response you're returning a separate bookmark for each individual query, but applying a page limit to across all queries. Assuming I've got that correct, I'm confused how a client would make use of that. Is it expected that we'd POST back all the bookmarks? That seems somewhat odd but I guess the way that our multi-query returns individual responses maybe its the least worst thing?

Also using foldr here seems awkward. If I submit a list of 5 queries that need to be paged, I'd very much expect to page through queries in the order I provided. Doing it in reverse seems less than useful.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -732,6 +732,8 @@ max_page_size(#httpd{path_parts=[_Db, <<"_design">>, _DDocName, <<"_view">> | _]
 parse_queries(Req, #mrargs{page_size = PageSize} = Args0, Queries)
         when is_integer(PageSize) ->
     MaxPageSize = max_page_size(Req),
+    PageSize > length(Queries)
+        orelse throw({query_parse_error, <<"Provided number of queries is more than given page_size">>}),
```
davisp, line 754: Another place to avoid the `orelse` cleverness.

src/chttpd/src/chttpd_db.erl
```diff
@@ -991,23 +1022,12 @@ streaming_cb(Msg0, Acc0) ->
...
-
-paginated_cb(complete, {iter, Db, Args, {Meta, Acc}}) ->
-    {ok, {iter, Db, Args, {Meta, lists:reverse(Acc)}}};
-
-paginated_cb({row, Row}, {Meta, Acc}) ->
-    {ok, {Meta, [couch_mrview_http:row_to_obj(Row) | Acc]}};
-
-paginated_cb({meta, Meta}, {_Meta, Acc}) ->
-    {ok, {Meta, Acc}};
+paginated_cb(Msg, #vacc{} = VAcc) ->
```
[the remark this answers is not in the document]
iilyak: Unit tests were passing on each commit separately. Unless I broke this property on rebase while doing cleanup.

src/couch_mrview/src/couch_mrview_http.erl
iilyak: Would you propose to include keys into a bookmark as is and copy to bookmark for each next page?

src/couch_mrview/src/couch_mrview_http.erl
iilyak: > you're returning a separate bookmark for each individual query, but applying a page limit to across all queries.

Yes I am. Otherwise the client could bypass the limit by using huge number of queries. Currently (in this PR) we limit number of queries in POST body to 2000. Even with this setting if we would apply individual limits the user would get 2000^2=4000000 rows in a single response. An alternative approach is to set individual limits to be `2000/n`. Where `n` is number of queries in the POST body. The downside of this approach is: it makes reusing of bookmarks in separate _all_docs requests less efficient. Currently client can issue POST, get a bunch of bookmarks (each bookmark has `page_size=2000`) and call GET with each bookmark.

> Is it expected that we'd POST back all the bookmarks?

It is possibility in case when client use a certain combination of parameters. For example when the client set `page_size` equal to number of queries in queries. However these bookmarks can be used in individual calls using GET.

src/chttpd/test/exunit/pagination_test.exs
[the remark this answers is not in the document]
garrensmith: I'm happy for you to leave them for now.

src/chttpd/test/exunit/pagination_test.exs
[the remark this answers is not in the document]
garrensmith: I don't mind its up to you.

src/couch_mrview/src/couch_mrview_http.erl
iilyak: > First things first, this definitely needs to have a versioned field. 

Agree

> I'm also not sold on the tuple-to-record implicit conversion. Given that an mrargs record is 24 fields, I'd think that most cases would likely be a bit smaller if we encoded the index for non-default values.

I am explicitly converting default values to `[]` which consumes 1 byte per field (see http://erlang.org/doc/apps/erts/erl_ext_dist.html#nil_ext). This is the smallest possible representation I am aware of. The `kl_to_record` is worse for number of fields greater than 3.

```
18> Nil = term_to_binary({1,2,3,4,[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]}, [compressed, {minor_version, 2}]).
<<131,80,0,0,0,30,120,156,203,144,72,100,76,100,74,100,78,
  100,201,194,2,0,140,196,10,87>>
19> size(Nil).
26
20> TList = term_to_binary([{1, 1},{2,2},{3,3},{4,4}], [compressed, {minor_version, 2}]).
<<131,108,0,0,0,4,104,2,97,1,97,1,104,2,97,2,97,2,104,2,
  97,3,97,3,104,2,97,4,97,...>>
21> size(TList).
31
```

iilyak: The only smaller encoding I can think of is a 24bit bit mask (3 bytes) in front of a binary which would have `1` for non default fields. Something like:
```
<<Mask:24, (term_to_binary([1,2,3]))/binary>>
```

iilyak:
```
23> Fields = term_to_binary([1,2,3], [compressed, {minor_version, 2}]).
<<131,107,0,3,1,2,3>>
24> <<12:24, Fields/binary>>.
<<0,0,12,131,107,0,3,1,2,3>>
```

src/chttpd/src/chttpd_db.erl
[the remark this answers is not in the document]
davisp: Cool. I really only noticed cause I was reviewing each commit individually.

src/couch_mrview/src/couch_mrview_http.erl
davisp: The difference was in eliding the value if it was set to the default so you'd have an extra byte per field, but my assumption was that in most cases a user would be using fewer than half the fields available. Ah, and not wrapping in tuples as well. So it'd be:

```erlang
6> Bin = term_to_binary([1, 1, 2, 2, 3, 3, 4, 4], [compressed, {minor_version, 2}]).
<<131,107,0,8,1,1,2,2,3,3,4,4>>
7> size(Bin).
12
```

However, I think your mask idea is even better. I can see a decode algorithm pretty clearly by converting the mask to a list of integers and then zipping against the list of values. The opposite for encoding seems like it'd be a straightforward reversal of that.

src/couch_mrview/src/couch_mrview_http.erl
davisp: I was more thinking that we'd reject a paginated request with keys. But other than that I'd just drop them and include directly. Otherwise we've got two different result sets depending on paginated vs not.

src/couch_mrview/src/couch_mrview_http.erl
davisp: I'm struggling to envision how a client would use this API in a more efficient or ergonomic way than just issue each query individually. My personal expectation ignoring the current implementation would be that I a pagination response would return result sets in the order I specified them up to one page at a time. And getting the next page would just be a matter of passing a single bookmark.

One oddity of that approach is that I'm not sure what the response format would look like after you paged through the initial result sets. Maybe it'd return an empty response or something that indicates "already consumed" or some such?

Alternatively as you suggest maybe spreading the page limit across all queries with the `2000/n` approach? But that still doesn't cover what happens if any of the queries result sets are exhausted before all of the others.

iilyak: You can find some additional background info in the RFC https://github.com/apache/couchdb-documentation/pull/530/files#diff-062d330b3595274c2461c38a43f29a11R135

src/chttpd/src/chttpd_db.erl
```diff
@@ -864,19 +896,50 @@ all_docs_view(Req, Db, Keys, OP) ->
...
+        false ->
+            Opts0 ++ [{restart_tx, true}];
+        true ->
+            Opts0
+    end,
     ViewCb = fun view_cb/2,
-    Acc = {iter, Db, Args, VAcc0},
-    {ok, {iter, _, _, VAcc1}} = fabric2_db:FoldFun(Db, ViewCb, Acc, Opts),
-    VAcc1.
+    Acc1 = {iter, Db, Args, VAcc0},
```
iilyak, line 940: It looks like this chunk can be removed.

src/couch_views/src/couch_views_http.erl
```diff
@@ -0,0 +1,272 @@
...
+    parse_params(Props, Keys, Args, []).
+
+
+parse_params([{"bookmark", Bookmark}], _Keys, #mrargs{}, _Options) ->
+    bookmark_decode(Bookmark);
+
+parse_params(Props, Keys, #mrargs{}=Args, Options) ->
+    case couch_util:get_value("bookmark", Props, nil) of
+        nil ->
+            ok;
```
iilyak, line 62: Need to write test for this case.

src/couch_views/src/couch_views_http.erl
```diff
@@ -0,0 +1,272 @@
...
+        true ->
+            throw({bad_request, "Cannot use bookmark with other options"})
```
iilyak, line 64: Needs test

src/couch_views/src/couch_views_http.erl
```diff
@@ -0,0 +1,272 @@
...
+                {[Value | Acc], <<((1 bsl Idx) bor MI):N>>, Idx + 1}
+        end
+    end, {[], <<0:N>>, 1}, tuple_to_list(Args0)),
+    Binary = term_to_binary(lists:reverse(Term), [compressed, {minor_version, 2}]),
+    couch_util:encodeBase64Url(<<?BOOKMARK_VSN, Mask/binary, Binary/binary>>).
+
+
+bookmark_decode(Bookmark) ->
+    try
+        N = ((record_info(size, mrargs) - 1) div 8 + 1) bsl 3,
```
iilyak, line 208: Should we just harcode the size of the mrargs and rely on CI to catch the case when mrargs is updated but mask size is not?

src/couch_views/src/couch_views_http.erl
```diff
@@ -0,0 +1,272 @@
...
+
+parse_params(Props, Keys, Args) ->
+    parse_params(Props, Keys, Args, []).
+
+
+parse_params([{"bookmark", Bookmark}], _Keys, #mrargs{}, _Options) ->
+    bookmark_decode(Bookmark);
+
+parse_params(Props, Keys, #mrargs{}=Args, Options) ->
+    case couch_util:get_value("bookmark", Props, nil) of
```
iilyak, line 60: Actually I think it can be removed. Because this check is enforced by `parse_params([{"bookmark", Bookmark}]` clause.

iilyak: never mind we need it.

src/couch_views/src/couch_views_http.erl
```diff
@@ -0,0 +1,272 @@
...
+        nil ->
+            ok;
+        true ->
```
iilyak, line 63: This should be `_`

src/couch_views/src/couch_views_http.erl
iilyak: I miss compile time computations in Erlang a lot.

PR #2875, opened 2020-05-07
Convert aegis key cach to LRU with hard expiration time

eiri:
## Overview

This converts aegis's key cache into LRU with hard expiration time.

The cache evicts the least recently used entries when number of entries in the cache exceeds "cache_limit" parameter with default set to 100000.

Additionally keys removed once their lifetime exceeds "cache_max_age_sec" parameter with default set to 30 minutes.

## Testing recommendations

` make eunit apps=aegis suites=aegis_server_test tests=lru_cache_with_expiration_test_`

## Checklist

### DIFF
--- a/src/aegis/src/aegis_server.erl
+++ b/src/aegis/src/aegis_server.erl
@@ -44,9 +44,13 @@
 -define(KEY_CHECK, aegis_key_check).
 -define(INIT_TIMEOUT, 60000).
 -define(TIMEOUT, 10000).
+-define(CACHE_LIMIT, 100000).
+-define(CACHE_MAX_AGE_SEC, 1800).
+-define(CACHE_EXPIRATION_CHECK_SEC, 10).
+-define(LAST_ACCESSED_INACTIVITY_SEC, 10).
 
 
--record(entry, {uuid, encryption_key}).
+-record(entry, {uuid, encryption_key, counter, last_accessed, expires_at}).
 
 
 start_link() ->
@@ -84,7 +88,7 @@ encrypt(#{} = Db, Key, Value) when is_binary(Key), is_binary(Value) ->
         uuid := UUID
     } = Db,
 
-    case ets:member(?KEY_CHECK, UUID) of
+    case is_key_fresh(UUID) of
         true ->
             case gen_server:call(?MODULE, {encrypt, Db, Key, Value}) of
                 CipherText when is_binary(CipherText) ->
@@ -109,7 +113,7 @@ decrypt(#{} = Db, Key, Value) when is_binary(Key), is_binary(Value) ->
         uuid := UUID
     } = Db,
 
-    case ets:member(?KEY_CHECK, UUID) of
+    case is_key_fresh(UUID) of
         true ->
             case gen_server:call(?MODULE, {decrypt, Db, Key, Value}) of
                 PlainText when is_binary(PlainText) ->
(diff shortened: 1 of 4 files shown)


src/aegis/src/aegis_server.erl
```diff
@@ -197,10 +205,21 @@ handle_call(_Msg, _From, St) ->
...
+    {noreply, NewSt};
+
+
 handle_cast(_Msg, St) ->
     {noreply, St}.
 
 
+handle_info(maybe_remove_expired, St) ->
+    remove_expired_entries(St),
+    CheckInterval = timer:seconds(expiration_check_interval()),
```
rnewson, line 219: there's a cleaner api for time unit conversion. is this done for compatibility reasons? (it's not clear from the funciton name that it does seconds to milliseconds)

src/aegis/src/aegis_server.erl
```diff
@@ -120,7 +128,7 @@ decrypt(#{} = Db, Key, Value) when is_binary(Key), is_binary(Value) ->
                 {error, Reason} ->
                     erlang:error(Reason)
             end;
-        false ->
+        _ ->
             process_flag(sensitive, true),
```
rnewson, line 132: shouldn't we evict the entry here?

src/aegis/src/aegis_server.erl
```diff
@@ -109,8 +115,10 @@ decrypt(#{} = Db, Key, Value) when is_binary(Key), is_binary(Value) ->
         uuid := UUID
     } = Db,
 
-    case ets:member(?KEY_CHECK, UUID) of
-        true ->
+    Now = fabric2_util:now(sec),
+
+    case ets:lookup(?KEY_CHECK, UUID) of
+        [{UUID, ExpiresAt}] when ExpiresAt >= Now ->
```
rnewson, line 121: this pattern is reproduced twice in the module, can we extract it?

src/aegis/src/aegis_server.erl
```diff
@@ -259,17 +278,125 @@ do_decrypt(DbKey, #{uuid := UUID}, Key, Value) ->
...
+        expires_at = ExpiresAt
+    },
+
     true = ets:insert(Cache, Entry),
-    true = ets:insert(?KEY_CHECK, {UUID, true}),
-    ok.
+    true = ets:insert_new(ByAccess, Entry),
+    true = ets:insert(?KEY_CHECK, {UUID, ExpiresAt}),
+
+    maybe_evict_old_entries(St),
```
rnewson, line 303: how can this ever be more than 1 item? "old" here is misleading as it implies time is a factor, but we're really evicted the least recently used item to make way for this one. It seems we could make that clearer by checking ets:size here and removing the (single) oldest item if it goes over.

src/aegis/src/aegis_server.erl
```diff
@@ -95,7 +101,7 @@ encrypt(#{} = Db, Key, Value) when is_binary(Key), is_binary(Value) ->
                 {error, Reason} ->
                     erlang:error(Reason)
             end;
-        false ->
+        _ ->
```
rnewson, line 104: Now that I look at it, we'll be exposing users to a high latency period every 30 minutes that they _don't_ revoke access to their key, which we expect to be the case almost every time.

Like in epep, we had "fresh" "stale" and "expired" to cover this. A fresh item is used directly, as your first clause does. An "expired" one is not used (gets removed from cache if present, expensive external call must happen again). but a "stale" entry _is_ used and also kicks off a check that we still have access (if we do, nothing happens, if we don't, we evict our entry).

src/aegis/src/aegis_server.erl
eiri: Good call! I've added it in `insert_key` handler called from `do_open_db` to avoid second gen_call.

src/aegis/src/aegis_server.erl
eiri: Unlike epep we can't refresh keys async, because refreshing requires reading a wrapped key from fdb, i.e. it must be done in the same process as the transaction. This means that we are going to have this high latency regardless if we removing key and fetching it on the next read or fetching it proactively once it's "stale".

The only way to avoid this is to add another server process to aegis that'll do refresh of the stale keys in its own separate transactions.

src/aegis/src/aegis_server.erl
```diff
@@ -276,6 +279,15 @@ do_decrypt(DbKey, #{uuid := UUID}, Key, Value) ->
     end.
 
 
+is_key_present(UUID) ->
```
rnewson, line 282: this returns false if the key is present but expired. `is_key_fresh`?

rnewson:
we've agreed to revisit the eviction pattern in a new PR. Would like to avoid evicting only to immediately repopulate on the next request but needs thought.

PR #2885, opened 2020-05-14
Add aegis info to db_info

eiri:
## Overview

This adds a function `get_db_info/1` to aegis key manager behaviour that allows to supply db_info with aegis specific information

### DIFF
--- a/src/aegis/src/aegis.erl
+++ b/src/aegis/src/aegis.erl
@@ -20,6 +20,7 @@
 -export([
     init_db/2,
     open_db/1,
+    get_db_info/1,
 
     decrypt/2,
     decrypt/3,
@@ -37,6 +38,10 @@ open_db(#{} = Db) ->
     Db#{
         is_encrypted => aegis_server:open_db(Db)
     }.
+
+
+get_db_info(#{} = Db) ->
+    ?AEGIS_KEY_MANAGER:get_db_info(Db).
 
 
 encrypt(#{} = _Db, _Key, <<>>) ->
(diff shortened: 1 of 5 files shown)


eiri:
@rnewson Can you take a look, please, if you have time?

In "real" key manger this could be used with something like this:

```erlang
get_db_info(#{} = Db) ->
    #{
        tx := Tx,
        db_prefix := DbPrefix
    } = Db,

    {Start, End} = erlfdb_tuple:range({?WRAPPED_KEY}, DbPrefix),
    Range = erlfdb:get_range(Tx, Start, End),

    lists:foldl(fun({K, V}, Acc) ->
        case erlfdb_tuple:unpack(K, DbPrefix) of
            {?WRAPPED_KEY, <<"wrapped_key">>} -> Acc;
            {?WRAPPED_KEY, Key} -> [{Key, V} | Acc]
        end
    end, [{encrypted, true}], Range).
```

I have two things I'm not sure about:
1. Do we want to set info from aegis key manager into special namespace in db_info, something like this:
```json
  "disk_size": 0,
  "instance_start_time": "0",
  "purge_seq": 0,
  "aegis": {
    "encrypted": true,
    "crk": ...
  }
  ...
```
Right now this is totally up to key manager.

2. Do we want to make callback `get_db_info/1` optional on `aegis_key_manger` behaviour? Then `aegis:get_db_info/1` can just check if it's exported and return `[]` otherwise, so nothing will be added to db_info.

eiri:
@rnewson ok, I'll wait with the merge then until we add aegis key manager example module we discussed before.

Issue #2886, opened 2020-05-14
Error when marking the `iss` claim as required


mauroporras:
## Description

I tried to add the `iss` claim as required, as indicated in the docs: https://docs.couchdb.org/en/stable/config/auth.html#jwt_auth/required_claims, but it's throwing `Invalid iss claim`.

## Steps to Reproduce

I put this in my `local.ini` file:

```
[jwt_auth]
; List of claims to validate
required_claims = iss
;
[jwt_keys]
; Configure at least one key here if using the JWT auth handler.
; If your JWT tokens do not include a "kid" attribute, use "_default"
; as the config key, otherwise use the kid as the config key.
; Examples
hmac:_default = aGVsbG8=
hmac:foo = Zm9v
; The config values can represent symmetric and asymmetrics keys.
; For symmetrics keys, the value is base64 encoded;
; hmac:_default = aGVsbG8= # base64-encoded form of "hello"
; For asymmetric keys, the value is the PEM encoding of the public
; key with newlines replaced with the escape sequence \n.
; rsa:foo = -----BEGIN PUBLIC KEY-----\nMIIBIjAN...IDAQAB\n-----END PUBLIC KEY-----\n
; ec:bar = -----BEGIN PUBLIC KEY-----\nMHYwEAYHK...AzztRs\n-----END PUBLIC KEY-----\n
```

## Expected Behaviour

A `Bad Request` error if the `iss` claim is missing.

## Your Environment

* CouchDB version used: 3.1.0
* Browser name and version: Google Chrome Version 81.0.4044.138 (Official Build) (64-bit)
* Operating system and version: Ubuntu 20.04, installed from the Apache CouchDB package repository.

## Additional Context

Stack trace:

```
[error] 2020-05-14T14:42:13.663783Z couchdb@127.0.0.1 <0.15084.6> b05a2bae1a req_err(1032618818) unknown_error : badarg
    [<<"erlang:list_to_existing_atom/1">>,<<"couch_httpd_auth:-get_configured_claims/0-lc$^0/1-0-/1 L216">>,<<"couch_httpd_auth:jwt_authentication_handler/1 L194">>,<<"chttpd:authenticate_request/2 L531">>,<<"chttpd:process_request/1 L303">>,<<"chttpd:handle_request_int/1 L243">>,<<"mochiweb_http:headers/6 L150">>,<<"proc_lib:init_p_do_apply/3 L247">>]
[...]

PR #2888, opened 2020-05-14
allow configurability of JWT claims that require a value

rnewson:
e.g;

[jwt]
required_claims = {iss, "hello"}

fixes https://github.com/apache/couchdb/issues/2886

### DIFF
--- a/src/couch/src/couch_httpd_auth.erl
+++ b/src/couch/src/couch_httpd_auth.erl
@@ -209,12 +209,21 @@ jwt_authentication_handler(Req) ->
 
 get_configured_claims() ->
     Claims = config:get("jwt_auth", "required_claims", ""),
-    case re:split(Claims, "\s*,\s*", [{return, list}]) of
-        [[]] ->
-            []; %% if required_claims is the empty string.
-        List ->
-            [list_to_existing_atom(C) || C <- List]
+    Re = "((?<key1>[a-z]+)|{(?<key2>[a-z]+)\s*,\s*\"(?<val>[^\"]+)\"})",
+    case re:run(Claims, Re, [global, {capture,  [key1, key2, val], binary}]) of
+        nomatch when Claims /= "" ->
+            couch_log:error("[jwt_auth] required_claims is set to an invalid value.", []),
+            throw({misconfigured_server, <<"JWT is not configured correctly">>});
+        nomatch ->
+            [];
+        {match, Matches} ->
+            lists:map(fun to_claim/1, Matches)
     end.
+
+to_claim([Key, <<>>, <<>>]) ->
+    binary_to_atom(Key, latin1);
+to_claim([<<>>, Key, Value]) ->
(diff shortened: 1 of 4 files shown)


src/couch/src/couch_httpd_auth.erl
```diff
@@ -209,11 +209,12 @@ jwt_authentication_handler(Req) ->
...
-    case re:split(Claims, "\s*,\s*", [{return, list}]) of
-        [[]] ->
-            []; %% if required_claims is the empty string.
-        List ->
-            [list_to_existing_atom(C) || C <- List]
+    case Claims of
+        "" ->
+            [];
+        Claims ->
+            {ok, Parsed} = couch_util:parse_term("[" ++ Claims ++ "]"),
```
davisp, line 216: Not sure how much you care, but we don't usually mutate the value that we pass to parse_term (i.e., we require useres to add the square brackets in the config. Not sure if  we care or not?

rnewson: the problem is 3.1.0 released with it being a comma-separate list of strings.

rnewson: I switched to a better format.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -209,13 +209,19 @@ jwt_authentication_handler(Req) ->
...
 get_configured_claims() ->
     Claims = config:get("jwt_auth", "required_claims", ""),
-    case re:split(Claims, "\s*,\s*", [{return, list}]) of
-        [[]] ->
-            []; %% if required_claims is the empty string.
-        List ->
-            [list_to_existing_atom(C) || C <- List]
+    Re = "((?<key1>[a-z]+)|{(?<key2>[a-z]+)\s*,\s*\"(?<val>[^\"]+)\"})",
+    case re:run(Claims, Re, [global, {capture,  [key1, key2, val], binary}]) of
+        nomatch ->
```
davisp, line 214: Seems like it'd be a good idea here to have a clause that notices when we are unable to match a configured value that's non-empty. Just something like `nomatch when Claims /= "" -> log thing`.

janl: Why not mention claims in the http response? Seems odd to require folks to look in two places.

rnewson: the user gets the http response, the admin gets the log message (who might also be the user). I didn't want to point at the specific misconfiguration to the user (who in this situation could be anyone, they might not have valid credentials). But I did want to mention that it's specific to JWT in case the user can use a different method, and so the admin has something more meaningful to report to user@ / github / etc.

janl: fair, thanks for explaining!

gbshhennsi:
Hey guys, in which version of couchdb will you release this fix or when will I be able to use it?

PR #2899, opened 2020-05-21
make jwtf_keystore compatible with erlang 19

rnewson:


janl:
except that `string:split` is also R20+… oops.

Issue #2906, opened 2020-05-25
couchjs -V should also return version of SM compiled


sklassen:
## Summary

couchjs returns couchdb version but not the version of spidermonkey. 

## Desired Behaviour

Add SM version to the output of couchjs -V

## Possible Solution

Add the version number (by hand?) to the help.h file.

PR #2911, opened 2020-05-27
2906 couchjs sm version

sklassen:
## Overview

couchjs returns couchdb version but not the version of spidermonkey used at compile time

Since each version has it's own help.h I just added the directories version number to the end of the first line eg "(SpiderMonkey 60)"

## Testing recommendations

I have compiled the code can ran couchjs -V

This addresses issue #2906 

## Checklist

- [X] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation


janl:
this is good!

IIRC the 60 side can also build against 68, so we need to be a little bit smarter there. Or we are more lazy and make it `60 or later`, not sure that would help ;)

wohali:
@janl Nope, we ended up having to fork, so the `60/` directory only builds SM60 now. `68/` does 68.

janl:
wait., when I looked there was no diff for /68, did this get added behind my back or did I just miss this earlier? Apologies!

PR #2953, opened 2020-06-17
Make restricted partition search parameters return bad request

bessbd:
## Overview
According to https://docs.couchdb.org/en/master/ddocs/search.html there
are parameters for searches that are not allowed for partitioned queries.

Those restrictions were not enforced, thus making the software and docs
inconsistent.

This PR adds them to validation so that the behavior matches the one
described in the docs.

## Testing recommendations

Run the tests included with the PR or perform them manually.

## Checklist

- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation - Not Applicable


bessbd:
Thank you for the review, @rnewson and thank you for the merge, @garrensmith !

PR #2972, opened 2020-06-29
Start all required deps automatically in test

iilyak:
## Overview

The `couch_epi` test suite is flaky. The result of the tests depends on the order of applications eunit runs.
The failures happen when `crypto` app is not running prior to starting couch_epi test suite. This PR changes the way we start `couch_epi` to insure its dependencies are automatically started. 

## Testing recommendations

There is no good way to reproduce the issue. We just run full test suite to make sure there are no new regressions.

```
make check
```

N/A

## Checklist


dottorblaster:
I think it's a good idea, actually I've seen this happening a ton of times in other projects. Looks like a nice fix, I'll approve it if the CI is green

PR #2983, opened 2020-07-06
New cname for couchdb-vm2, see INFRA-20435

wohali:
Because of ASF Infra changes, couchdb-vm2.apache.org is going away and will be replaced by couchdb-vm before August 1st 2020.

This PR adds redirection through a new CNAME, logs.couchdb.org, for the logfile uploader.

It also uses repo-nightly.couchdb.org instead of couchdb-vm2.a.o.

This requires @janl to get the new CNAME in before this can land.

This is a cherrypick of #2982 .


wohali:
The machine is ready now and just needs the CNAMEs.

PR #3024, opened 2020-07-22
Pass fdb key to the encode_fun

rnewson:
Pass fdb key to the encode_fun

This commit also removes the get/wait/future split as it made it much
harder to pass the fdb key down. Since those functions only call each
other, there is no loss of functionality.


rnewson:
With this feature, we can add aegis support with an encode_fun like this;

```
    EncodeFun = fun
        (encode, Key, Term) ->
            Bin = term_to_binary(Term, [compressed, {minor_version, 2}]),
            aegis:encrypt(TxDb, Key, Bin);
        (decode, Key, Ciphertext) ->
            Bin = aegis:decrypt(TxDb, Key, Ciphertext),
            binary_to_term(Bin, [safe])
    end,```

PR #2627, opened 2020-03-02
Remove sensitive data from log files

iilyak:
## Overview

There are cases where sensitive data can leak into a log file. This PR makes a first stab at the problem. The sensitive data can end up in the log when process implementing `gen_server` or `gen_event` behavior crashes. The erlang implementation of a `gen_server` support a `format_status/2` callback to remove sensitive data from `state` term. However the last message (`LastMsg`) recveived by the process is still logged. In order to solve this problem this PR modifies `couch_log_formatter` to remove `LastMsg` in case when formatted state is a tupple with arity 2 and second element of a tuple is map containing `sensitive` key with value `true`. This allows us to remove sensitive information from both `LastMsg` and `State` via `format_status/2` callback. Here is an example:
```
format_status(_Opt, [_PDict, State]) ->
    #state{
        url = Url,
        proxy_url = ProxyURL,
        limit = Limit
    } = State,
    {[
        {url, couch_util:url_strip_password(Url)},
        {proxy_url, ProxyURL},
        {limit, Limit}
    ], #{sensitive => true}}.
```

## Testing recommendations

1. Make sure all dependent PRs are pulled
2. Run `make eunit`
3. Run `make elixir`

## Related Issues or Pull Requests

This PR depends on the following:

* https://github.com/apache/couchdb-config/pull/30
* https://github.com/apache/couchdb-ibrowse/pull/3
* https://github.com/apache/couchdb-mochiweb/compare/master...iilyak:implement-format-status?expand=1 <-- I need an advise which base branch to use. 

## Checklist

- [x] Code is written and works correctly
- [x] Changes are covered by tests
- [ ] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation

PR #3031, opened 2020-07-23
Clean up logs

iilyak:
## Overview

When OTP processes crash OTP logs the current state of the process and last message it received in the log file.
However this feature can lead to leakage of sensitive date into log files. Preventing this from happening requires fixing both sources of leakage. The leakage via state term could be prevented via usage of `format_status/2` callback intended to reformat state before passing it to logger. The leakage via last message is harder to prevent because there is no good way to detect sensitive data in arbitrary term. The safest approach is to remove the last message term before it is included into log message. However in some cases last message can be a useful debugging tool that's why instead of removing it altogether I've added `[log]-> strip_last_msg = false` configuration option. The extra configuration option wouldn't cause an extra latency thanks to an optimization in couch_log application.

This PR removes sensitive data from state terms. It also tries to make state size predictable by replacing complex terms with size of the term (for example for lists we use `length(List)` and for dict we use `dict:size(Dict)`).  
This PR is not sufficient to remove all sensitive data from logs, see other parts of the solution in the following independent PRs:

* https://github.com/apache/couchdb-config/pull/30 - this would require a follow up PR to pass sensitive flag
* https://github.com/apache/couchdb-ibrowse/pull/4 - similar PR for ibrowse

## Testing recommendations

Regression testing can be done using `make check`

* This is an alternative to https://github.com/apache/couchdb/pull/2627

## Checklist


src/setup/src/setup_httpd.erl
```diff
@@ -173,3 +173,8 @@ get_body(Req) ->
         couch_log:notice("Body Fail: ~p~n", [Else]),
         couch_httpd:send_error(Req, 400, <<"bad_request">>, <<"Missing JSON body'">>)
     end.
+
+remove_sensitive(KVList0) ->
+    KVList1 = lists:keyreplace(<<"username">>, 1, KVList0, {<<"username">>, <<"****">>}),
+    KVList2 = lists:keyreplace(<<"password">>, 1, KVList1, {<<"password">>, <<"****">>}),
+    KVList2.
```
tonysun83, line 180: is this redundant or just a stylist preference?

src/couch_log/src/couch_log_config.erl
```diff
@@ -97,4 +98,10 @@ transform(max_message_size, SizeStr) ->
...
         16000
-    end.
\ No newline at end of file
+    end;
+
+transform(strip_last_msg, "false") ->
+    false;
+
+transform(strip_last_msg, _) ->
+    true.
```
tonysun83, line 107: add an  end of line here

src/couch_mrview/src/couch_mrview_index.erl
```diff
@@ -324,3 +325,14 @@ update_local_purge_doc(Db, State, PSeq) ->
             BaseDoc
     end,
     couch_db:update_doc(Db, Doc, []).
+
+format_status(_Opt, [_PDict, State]) ->
```
tonysun83, line 329: If we're doing this for traditional views as well, do we need to modify this section as well: https://github.com/apache/couchdb/blob/master/src/dreyfus/src/dreyfus_index.erl#L37?

src/setup/src/setup_httpd.erl
iilyak: Just a style. Can change if needed.

src/couch/src/couch_work_queue.erl
```diff
@@ -186,3 +186,18 @@ code_change(_OldVsn, State, _Extra) ->
...
+        work_waiters = Waiters
+    } = Queue,
+    Scrubbed = Queue#q{
+        queue = queue:len(Q),
+        blocked = length(Blocked),
+        work_waiters = length(Waiters)
+    },
+    [{data, [{"State",
+        ?record_to_keyval(q, Scrubbed)
+    }]}].
```
tonysun83, line 203: whitespace

tonysun83:
I think shortening the complex terms is the right approach. Similar to what I wrote in the other PR about more meaningful output,  I was thinking if providing a length(Term) is necessary. It'll just be a bunch of integers in the output that don't have much meaning. Some other options
1) Leaving it nil or blank (like what you did for `mrviews`
2) Only show first X number or characters  (a little bit more work)
3) Prepend it with some descriptor '{<name of field>_size, length(Term)}

src/couch_mrview/src/couch_mrview_index.erl
iilyak: done

tonysun83:
+1, please squash and merge, good work!

iilyak:
`couch_jobs_tests:210: type_timeouts_and_server...*failed*` - is known to be flaky.

PR #3057, opened 2020-08-05
fixup: Build couch_js for redhat linux

jiangphcn:
## Overview

When building couch_js in RHEL, there is one error occurring with "undefined reference to symbol '_ZTVN10__cxxabiv117__class_type_infoE@@CXXABI_1.3'". This commit is to adjust binding library to address this issue.

```
Compiling priv/couch_ejson_compare/couch_ejson_compare.c
/usr/bin/ld: priv/couch_js/60/main.o: undefined reference to symbol '_ZTVN10__cxxabiv117__class_type_infoE@@CXXABI_1.3'
//usr/lib64/libstdc++.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
ERROR: sh(cc priv/couch_js/60/http.o priv/couch_js/60/main.o priv/couch_js/60/utf8.o priv/couch_js/60/util.o -L/usr/local/lib -std=c++14 -lmozjs-60 -lm   -L"/usr/local/lib/erlang/lib/erl_interface-3.10.2.2/lib" -lerl_interface -lei -o priv/couchjs)
failed with return code 1 and the following output:
/usr/bin/ld: priv/couch_js/60/main.o: undefined reference to symbol '_ZTVN10__cxxabiv117__class_type_infoE@@CXXABI_1.3'
//usr/lib64/libstdc++.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```
## Testing recommendations

make, and make fdb-check in RHEL

## Checklist

### DIFF
--- a/src/couch/rebar.config.script
+++ b/src/couch/rebar.config.script
@@ -121,7 +121,7 @@ end,
     {unix, _} when SMVsn == "60" ->
         {
             "-DXP_UNIX -I/usr/include/mozjs-60 -I/usr/local/include/mozjs-60 -std=c++14",
-            "-L/usr/local/lib -std=c++14 -lmozjs-60 -lm"
+            "-L/usr/local/lib -std=c++14 -lmozjs-60 -lm -lstdc++"
         }
 end.
 


src/couch/rebar.config.script
```diff
@@ -118,6 +118,11 @@ end,
             "-DXP_UNIX -I/usr/include/mozjs-60 -I/usr/local/include/mozjs-60 -std=c++14",
             "-L/usr/local/lib -lmozjs-60 -lm -std=c++14 -lc++"
         };
+    {unix, linux} when SMVsn == "60" ->
```
wohali, line 121: @jiangphcn thanks for this, what happens if we just add `-lstdc++` to the `{unix, _}` case? Does it break other Linux builds? Or, is there a reason the `-std=c++14` line won't work on RHEL?

jiangphcn: @wohali thanks for your question and suggestion. I tested with Centos and Debian,  and it will not break if we just add -lstdc++ to the {unix, _} case. I change accordingly. Could you please check again?

PR #3058, opened 2020-08-05
Ebtree fixes

davisp:
Fixes for ebtree

### DIFF
--- a/src/ebtree/src/ebtree.erl
+++ b/src/ebtree/src/ebtree.erl
@@ -25,6 +25,7 @@
      fold/4,
      fold/5,
      reduce/4,
+     reduce/5,
      full_reduce/2,
      group_reduce/7,
      group_reduce/8,
--- a/src/ebtree/src/ebtree.erl
+++ b/src/ebtree/src/ebtree.erl
@@ -283,6 +283,9 @@ reduce(Db, #tree{} = Tree, StartKey, EndKey, Options) ->
     do_reduce(Tree, MapValues, ReduceValues).
 
 
+do_reduce(#tree{} = Tree, [], []) ->
+    reduce_values(Tree, [], false);
+
 do_reduce(#tree{} = Tree, [], ReduceValues) when is_list(ReduceValues) ->
     reduce_values(Tree, ReduceValues, true);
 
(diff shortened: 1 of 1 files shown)


src/ebtree/src/ebtree.erl
```diff
@@ -411,6 +419,9 @@ range(Db, #tree{} = Tree, StartKey, EndKey, AccFun, Acc0) ->
     end).
 
 
+range(_Tx, #tree{}, #node{level = 0, members = []}, _StartKey, _EndKey, _AccFun, Acc0) ->
```
rnewson, line 422: when will a leaf node have zero members? That... should never happen.

src/ebtree/src/ebtree.erl
```diff
@@ -282,6 +283,9 @@ reduce(Db, #tree{} = Tree, StartKey, EndKey, Options) ->
     do_reduce(Tree, MapValues, ReduceValues).
 
 
+do_reduce(#tree{} = Tree, [], []) ->
+    reduce_values(Tree, [], false);
```
rnewson, line 287: shouldn't that be `true`? the lack of map values is the signal we're working with reductions.

src/ebtree/src/ebtree.erl
davisp: When the tree is empty like it says in the commit message.

src/ebtree/src/ebtree.erl
davisp: We're not working with reductions though. We're reducing an empty list of key/value pairs. Plainly the empty list from the second argument is the same empty list being passed to reduce_values.

Also, I think most folks writing reduce functions would do better to think of it as "reduce empty list of key/values" vs "rereduce empty list of reduce values that might need to be guarded against for the normal case when you actually have reduce values".

src/ebtree/src/ebtree.erl
rnewson: got it. can you add a test that shows the problem pls.

src/ebtree/src/ebtree.erl
rnewson: kk

src/ebtree/src/ebtree.erl
rnewson: also tighten this to match only the root node (instead of `level=0`)

PR #3062, opened 2020-08-06
Prototype/fdb layer ebtree enhance

rnewson:
Improve test coverage (to 97%) and tighten the collation of members.

### DIFF
--- a/src/ebtree/src/ebtree.erl
+++ b/src/ebtree/src/ebtree.erl
@@ -1111,6 +1111,25 @@ collate_validation_test() ->
     ?assertError(invalid_collation_result, collate(Tree, 1, 2)).
 
 
+order_is_preserved_test() ->
+    Db = erlfdb_util:get_test_db([empty]),
+    open(Db, <<1,2,3>>, 4),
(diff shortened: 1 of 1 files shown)


src/ebtree/src/ebtree.erl
```diff
@@ -940,9 +941,9 @@ collate(#tree{} = Tree, A, B, Allowed) ->
     lists:member(collate(Tree, A, B), Allowed).
 
 
-umerge_members(#tree{} = Tree, List1, List2) ->
+umerge_members(#tree{} = Tree, Level, List1, List2) ->
     CollateWrapper = fun
-        ({K1, _V1}, {K2, _V2}) ->
+        ({K1, _V1}, {K2, _V2}) when Level == 0 ->
             collate(Tree, K1, K2, [lt, eq]);
         ({_F1, L1, _V1, _R1}, {_F2, L2, _V2, _R2}) ->
```
davisp, line 948: Seems like you might as well add a `when Level > 0` guard here?

davisp: And in the corresponding places below.

PR #3068, opened 2020-08-10
Unlink index pid and swallow EXIT message if present

### LINKED ISSUE #3061: couch_index_server crashes too easily
We have a problem with couch_index_server.

handle_info({'EXIT', Pid, Reason}, Server) ->
    case ets:lookup(?BY_PID, Pid) of
        [{Pid, {DbName, Sig}}] ->
            DDocIds = [DDocId || {_, {DDocId, _}}
                <- ets:match_object(?BY_DB, {DbName, {'$1', Sig}})],
            rem_from_ets(DbName, Sig, DDocIds, Pid);
        [] when Reason /= normal ->
            exit(Reason); <--

whenever an index process crashes abnormally couch_index_server also terminates, taking all other index processes with it.

The cause of the couch_index crash I saw this with was from couch_mrview_index;

close(State) ->
    erlang:demonitor(State#mrst.fd_monitor, [flush]),
    couch_file:close(State#mrst.fd).

where the "fd" was not a live process.

couch_index_server should _not_ terminate if there is an abnormal exit by a single index.

rnewson:
## Overview

This should prevent unexpected exit messages arriving which crash
couch_index_server.

Patch suggested by davisp.

Closes #3061.

## Testing recommendations

Under high query load with ddoc deletes, observe that couch_index_server does not terminate.

N/A

## Checklist

### DIFF
--- a/src/couch_index/src/couch_index_server.erl
+++ b/src/couch_index/src/couch_index_server.erl
@@ -243,9 +243,15 @@ reset_indexes(DbName, Root) ->
     end, dict:new(), ets:lookup(?BY_DB, DbName)),
     Fun = fun({Sig, DDocIds}) ->
         [{_, Pid}] = ets:lookup(?BY_SIG, {DbName, Sig}),
-        MRef = erlang:monitor(process, Pid),
+        unlink(Pid),
         gen_server:cast(Pid, delete),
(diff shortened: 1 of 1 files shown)


wohali:
@rnewson Please be sure to backport this to 3.x.

PR #3075, opened 2020-08-13
Don't crash couch_index_server if the db isn't known yet

rnewson:
## Overview

If a ddoc is added immediately after database creation (_users and
_replicator when couchdb is used in a multi-tenant fashion), we can
crash couch_index_server in handle_db_event, as mem3_shards:local
throws an error.

## Testing recommendations

Create new $prefix/_replicator ddocs while under heavy indexing load elsewhere and observe a crash without this patch, and no crash with it.

#3068

## Checklist


chewbranca:
Good catch!

wohali:
Please, please backport this to 3.x.

PR #3092, opened 2020-08-20
Don't log client disconnects

rnewson:
Don't log client disconnects


davisp:
We should add a clause at [1] that just re-exit's normally.

[1] https://github.com/apache/couchdb/blob/25428e48eba96ca1eb6a3099a1b1cfe4b35ee13d/src/chttpd/src/chttpd.erl#L997

wohali:
Ooh, can we backport this to 3.x? People have been asking for this one for a while.

Issue #2795, opened 2020-04-20
Partitioned _find drops results unexpectedly


willholley:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

When a partioned `_find` performs an index scan, it only scans a subset of the index limited to `partition_query_limit`. This can result in matches being silently dropped.

## Steps to Reproduce

 ```
# Configure CouchDB with `query_server_config.partition_query_limit = 1`:
$ curl -u admin:password 'http://localhost:5984/_node/_local/_config/query_server_config/partition_query_limit' -H 'Content-Type:application/json' -d '"1"' -XPUT

# Create a partitioned database
$ curl -u admin:password 'http://localhost:5984/test?partitioned=true' -X PUT
{"ok":true}

# Add 2 documents with a common "user"  field
$ curl -u admin:password 'http://localhost:5984/test/mypartition%3Aalice1' -X PUT -H 'Content-Type: application/json' -d '{"_id":"mypartition:alice","user":"alice"}'
{"ok":true,"id":"mypartition:alice1","rev":"1-908c4364e2555b9067d1b07be0094d58"}

$ curl -u admin:password 'http://localhost:5984/test/mypartition%3Aalice2' -X PUT -H 'Content-Type: application/json' -d '{"_id":"mypartition:alice2","user":"alice"}'
{"ok":true,"id":"mypartition:alice2","rev":"1-908c4364e2555b9067d1b07be0094d58"}

# Create a partitioned index on field "user"
$ curl -u admin:password 'http://localhost:5984/test/_index' -H 'Content-Type: application/json' -d '{"index":{"fields":["user"]},"partitioned":true}'
{"result":"created","id":"_design/78d1b777945c322cf4c49b3f5f8a5d5c7b8a4321","name":"78d1b777945c322cf4c49b3f5f8a5d5c7b8a4321"}

# Run a query for "user" equals "alice" with limit 10
$ curl -u admin:password 'http://localhost:5984/test/_partition/mypartition/_find' -H 'Content-Type: application/json' -d '{"selector":{"user":"alice"}, "limit": 10}'
{"docs":[
{"_id":"mypartition:alice1","_rev":"1-908c4364e2555b9067d1b07be0094d58","user":"alice"}
],
[...]

PR #3105, opened 2020-08-26
bypass partition query limit for mango

tonysun83:
## Overview
When partition_query_limit is set for couch_mrview, it limits how many
docs can be scanned when executing partitioned queries. But this limits
mango's doc scans internally. This leads to documents not being scanned
to fulfill a query. This fixes:
https://github.com/apache/couchdb/issues/2795

## Testing recommendations
Unit Test added.
## Checklist

### DIFF
--- a/dev/run
+++ b/dev/run
@@ -427,7 +427,10 @@ def boot_haproxy(ctx):
 def hack_default_ini(ctx, node, contents):
 
     contents = re.sub(
-        "^\[httpd\]$", "[httpd]\nenable = true", contents, flags=re.MULTILINE,
+        "^\[httpd\]$",
+        "[httpd]\nenable = true",
+        contents,
+        flags=re.MULTILINE,
     )
 
     if ctx["enable_erlang_views"]:
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -116,7 +116,11 @@ base_args(#cursor{index = Idx, selector = Selector} = Cursor) ->
         start_key = StartKey,
         end_key = EndKey,
         include_docs = true,
-        extra = [{callback, {?MODULE, view_cb}}, {selector, Selector}]
+        extra = [
+            {callback, {?MODULE, view_cb}},
+            {selector, Selector},
+            {ignore_partition_query_limit, true}
+        ]
     }.
 
 
--- a/test/elixir/test/partition_mango_test.exs
+++ b/test/elixir/test/partition_mango_test.exs
@@ -547,6 +547,59 @@ defmodule PartitionMangoTest do
   end
 
   @tag :with_partitioned_db
+  test "partitioned query with query server config set", context do
+    db_name = context[:db_name]
+    create_partition_docs(db_name)
+    create_index(db_name, ["value"])
+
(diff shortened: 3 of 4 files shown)


src/mango/src/mango_cursor_view.erl
```diff
@@ -116,7 +116,8 @@ base_args(#cursor{index = Idx, selector = Selector} = Cursor) ->
         start_key = StartKey,
         end_key = EndKey,
         include_docs = true,
-        extra = [{callback, {?MODULE, view_cb}}, {selector, Selector}]
+        extra = [{callback, {?MODULE, view_cb}}, {selector, Selector},
```
davisp, line 119: Style nit, but make this multi-lined:

```erlang
extra = [
    {callback, {?MODULE, view_cb}},
    {selector, Selector},
    {ignore_partition_query_limit, true}
]
```

test/elixir/test/partition_mango_test.exs
```diff
@@ -502,6 +502,9 @@ defmodule PartitionMangoTest do
     create_partition_docs(db_name)
     create_index(db_name, ["value"])
 
+    # this is to test that we bypass partition_query_limit for mango
+    set_config({"query_server_config", "partition_query_limit", "1"})
+
```
davisp, line 507: Lets create a specific test for this rather than mixing our assertions.

tonysun83:
@wohali CI seems to be failing with `python-black` issues. I'm not sure if it's related to this PR as I have not made any changes to python files. I've rebuilt it a few times already with no luck. See anything I'm missing?

wohali:
Hey @tonysun83 ,

```
[2020-08-27T19:38:24.347Z] would reformat /home/****/workspace/****-cm1_PullRequests_PR-3105/20.3.8.11/build/apache-couchdb-3.0.0-7dbd0ad/dev/run
```

Looking at the commit history, `dev/run` got changed 5 months ago, and my change to ensure black was run across all files dates from April, and commits have passed since then. So this must be related to a new check in python black.

If you run `make python-black-update` you can fix the problem, then commit the change it makes to this PR.

I realize this, strictly speaking, is an unrelated change to fixing mango, but your help is greatly appreciated :) and a single PR is faster than 2.

tonysun83:
thx @wohali , it looks good!

dev/run
```diff
@@ -427,7 +427,10 @@ def boot_haproxy(ctx):
 def hack_default_ini(ctx, node, contents):
 
     contents = re.sub(
-        "^\[httpd\]$", "[httpd]\nenable = true", contents, flags=re.MULTILINE,
+        "^\[httpd\]$",
+        "[httpd]\nenable = true",
+        contents,
+        flags=re.MULTILINE,
```
davisp, line 433: What's this change for? Looks like styling which we shouldn't mix.

wohali: Line's too long, required change to make python black happy. It is unrelated but needs to go in because otherwise tests fail. Don't give him grief on this one.

davisp: Ah! Fair enough.

wohali:
@tonysun83 Would you backport this (mango) fix to 3.x too?

PR #3108, opened 2020-08-28
Make restricted partition search return bad request

bessbd:
This is a backport of `master`'s 6659dbbd7c556b8dc00c075e331d7b106d44088d to `3.x`.

PR #3109, opened 2020-08-28
Allow drilldown for search to always be specified as list of lists

bessbd:
This is a backport of `master`'s 42403914a8c86a26cf58363f0eaf35551400aa30 to `3.x`.

### DIFF
--- a/src/dreyfus/src/dreyfus_httpd.erl
+++ b/src/dreyfus/src/dreyfus_httpd.erl
@@ -239,6 +239,8 @@ validate_index_query(counts, Value, Args) ->
     Args#index_query_args{counts=Value};
 validate_index_query(ranges, Value, Args) ->
     Args#index_query_args{ranges=Value};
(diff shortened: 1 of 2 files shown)


wohali:
Hey @bessbd, thanks for the fixes (this and #3108). I think this may be the first time I've seen you here, but you do seem to already be an Apache committer. 

Obviously you know what you're doing here, but @iilyak or I or someone else will have to merge these for you, I think, as I don't have any record of having put you in as a CouchDB committer.

bessbd:
Thank you for the review, @iilyak !

@wohali : the original commits which I wanted to backport here and in #3108 were actually (at least partly) created by me, but it's always great to be noticed!
> already be an Apache committer
> any record of having put you in as a CouchDB committer

Right, I'm a member of Apache Flume and I'm not a CouchDB committer (yet, though I'm doing my best to become one someday).

Also, thank you for the merge, @wohali .

PR #3116, opened 2020-08-31
fix bookmark passing with text indexes

tonysun83:
## Overview

Previously, we passed in the unpacked version of the bookmark with
the cursor inside the options field. This worked fine for _find because
we didn't need to return it to the user. But for _explain, we return
the value back as unpacked tuple instead of a string and jiffy:encode/1
complains. Now we correctly extract the bookmark out of options, unpack
it, and then pass it separately in it's own field. This way options
retains it's original string form for the user so that invalid_ejson
is not thrown.

## Testing recommendations

Regression tests should pass. If you create a text index, simply use _explain with a bookmark and verify the results. 

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_text.erl
+++ b/src/mango/src/mango_cursor_text.erl
@@ -43,7 +43,7 @@
 }).
 
 
-create(Db, Indexes, Selector, Opts0) ->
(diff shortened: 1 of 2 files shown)


tonysun83:
text index tests also pass with full regression

PR #3120, opened 2020-09-01
Fix stale=false for views

bessbd:
## Overview

The docs mention an option of "false" for "stale" which is currently not supported and is rejected by validation.

This commit fixes the problem described above.

## Testing recommendations

`make eunit apps=couch_mrview suites=couch_mrview_http`

N/A

## Checklist

- [x] Any new configurable parameters are documented in `rel/overlay/etc/default.ini` -- not applicable
- [x] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation -- this PR makes behavior match behavior described in https://docs.couchdb.org/en/stable/api/ddoc/views.html#api-ddoc-view


wohali:
I agree with @eiri here. The `stale` parameter was supposed to be deprecated in 3.x, but we gave it a reprieve until 4.0 (and maybe beyond that).

bessbd:
Thank you for the feedback, @eiri and @wohali . I'll close this PR and open a new one for docs.

PR #3126, opened 2020-09-03
Prototype/fdb layer ebtree improvements

davisp:
## Overview

Various improvements to ebtree.

The `lookup_multi` and `insert_multi` functions exist to minimize the cost of view updates by minimizing the amount of serialization and collation costs.

Nodes are made immutable so that clients can cache nodes to reduce latency due to round trips to foundationdb.

The `umerge_members/4` is optimized to avoid the usage of `lists:umerge/3`. This is because `lists:umerge/3` uses a `=<` operation which increases the number of calls to our collation function. By re-implementing using the `lt, eq, gt` return values we can minimize the number of collation calls.

Finally, we disable `validate_node` in production. This validation is useful during tests but the collation cost is significant during production work loads. A tree can still be validated in production manually.

## Testing recommendations

`make check`

## Checklist

### DIFF
--- a/src/ebtree/src/ebtree.erl
+++ b/src/ebtree/src/ebtree.erl
@@ -49,15 +49,15 @@
     collate_fun,
     reduce_fun,
     encode_fun,
-    persist_fun
+    persist_fun,
+    cache_fun
 }).
 
 -define(META, 0).
 -define(META_ORDER, 0).
--define(META_NEXT_ID, 1).
 
 -define(NODE, 1).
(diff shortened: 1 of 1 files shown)


rnewson (approved):
very nice. agree on disabling validate_node at production though it would be good to have it as an option when we run load testing in preprod environments. Clearly validate_node never throws during the test suite or any other testing so far, though.

Have confirmed tests pass, test coverage at 97%, and no dialyzer errors.

PR #3129, opened 2020-09-04
Add option to delay responses until the end

rnewson:
## Overview

When set, every response is sent once fully generated on the server
side. This increases memory usage on the nodes but simplifies error
handling for the client as it eliminates the possibility that the
response will be deliberately terminated midway through due to a
timeout.

The config value can be changed at runtime without impacting any
in-flight responses.

## Testing recommendations

Set `[chttpd] delay_until_end` to true then confirm that streamed responses are now sent non-stream (i.e, with a Content-Length header and not in chunked transfer mode).

N/A

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -130,6 +130,9 @@ prefer_minimal = Cache-Control, Content-Length, Content-Range, Content-Type, ETa
 ; _dbs_info in a request
 max_db_number_for_dbs_info_req = 100
 
+; set to true to delay the start of a response until the end has been calculated
+;buffer_response = false
+
 ; authentication handlers
 ; authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, default_authentication_handler}
 ; uncomment the next line to enable proxy authentication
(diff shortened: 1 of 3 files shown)


janl:
Neat! Have you considered making this a per-request option instead of a global config?

rnewson:
I'm open to that, sure, would not be hard to make it a request parameter as we have the original Req at the start.

rnewson:
@janl done. I am interested in feedback on the name of this config setting and request param (i.e, I don't much like my choice but can't think of a better name).

janl:
OTOH:

- `buffer_response(s)`
- `disable_streaming_response(s)`

rnewson:
thinking I might reference "transfer encoding" explicitly, with "identity" and "chunked" instead of boolean?

rnewson:
enh, but the point of the change is the delay. not sending the response until we're sure we can send the whole thing, it's not really about the transfer encoding, that's a side-effect (it just seemed silly send a chunked response when I have the whole thing in memory and can calculate the content-length)

rnewson:
buffer_response it is.

janl (approved):

PR #3131, opened 2020-09-07
feat: add same_site = none_secure option

janl:
c.f. https://web.dev/samesite-cookies-explained/

also https://github.com/apache/couchdb/discussions/3012


janl:
I’d be happy to biked this to make it its own config variable `cookie_secure = bool` and folks who want both can set both, this was the least effort way to propose a change tho :)

rnewson:
hm, the secure flag is added automatically if couchdb is serving a TLS response. If something else is doing TLS (like haproxy) then it should also add the secure flag.

rnewson:
I'm -1 on this on principle, but please update the PR if I'm missing something obvious.

janl:
This might still be useful for scenarios where folks are behind proxies where they can’t add the Secure flag to the `Set-Cookie` header dynamically, and if they don’t want to go through the particular fun of setting up TLS with CouchDB itself.

But unless we have folks reporting they are in this situation, I’m happy to hold off on this.

PR #3132, opened 2020-09-07
Add option to delay responses until the end

rnewson:
When set, every response is sent once fully generated on the server
side. This increases memory usage on the nodes but simplifies error
handling for the client as it eliminates the possibility that the
response will be deliberately terminated midway through due to a
timeout.

The config value can be changed at runtime without impacting any
in-flight responses.

backported from master.


bessbd:
+1 (non-binding), content of this change matches the content merged to master for the original PR.

wohali:
+1 post-merge for the record (consent given on IRC)

PR #3145, opened 2020-09-10
Fix buffer_response=true

rnewson:
We need to call StartFun as it might add headers, etc.

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -855,13 +855,18 @@ end_delayed_json_response(#delayed_resp{buffer_response=false}=DelayedResp) ->
 
 end_delayed_json_response(#delayed_resp{buffer_response=true}=DelayedResp) ->
     #delayed_resp{
+        start_fun = StartFun,
         req = Req,
         code = Code,
         headers = Headers,
         chunks = Chunks
     } = DelayedResp,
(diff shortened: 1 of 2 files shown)


src/chttpd/src/chttpd.erl
```diff
@@ -855,13 +855,19 @@ end_delayed_json_response(#delayed_resp{buffer_response=false}=DelayedResp) ->
...
         chunks = Chunks
     } = DelayedResp,
-    {ok, Resp} = start_response_length(Req, Code, Headers, iolist_size(Chunks)),
-    send(Resp, lists:reverse(Chunks)).
+    {ok, Resp} = StartFun(Req, Code, Headers),
+    lists:foreach(
+        fun
+            ("") -> ok;
+            (Chunk) -> send_chunk(Resp, Chunk)
+        end, lists:reverse(Chunks)),
```
davisp, line 869: This formatting is kinda weird. Mebbe something like such?

```erlang
lists:foreach(fun
    ([]) -> ok;
    (Chunk) -> send_chunk(Resp, Chunk)
end, lists:reverse(Chunks)),
```

davisp (approved):
+1. Those multi-clause-anonymous-inline-functions are always kinda weird so whatever you feel like doing on that style suggestion is up to you.

PR #3147, opened 2020-09-10
Fix buffer_response=true (#3145)

rnewson:
We need to call StartFun as it might add headers, etc.


wohali:
+1, I will cut an RC2 for this.

Issue #3160, opened 2020-09-17
Work around statistics(run_queue) returning incorrect data


chewbranca:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

In the `/_system` endpoint, particularly in the `chttpd_node:get_stats/0` function, one of the data points returned is an integer reflecting the total sum of the length of all run queues [1]. With the introduction of dirty schedulers in Erlang, the `statistics(run_queue)` statistic now includes information about the depth of the dirty CPU and IO queues, however, the documentation indicates it should not be including those values. At least that's how it appears to me, I've filed a bug on the matter [2] and I'll keep an eye on that.

In the meantime, we can isolate the scheduler run queues from the dirty scheduler run queues, and report those values separately which will ensure we have an accurate value for `run_queue` and also allow us to introduce a new value for the dirty CPU queue.

For more context, CouchDB uses several NIFs, but none of them are dirty NIFs. However, OTP's `rsa_generate_key` function _is_ a dirty NIF [3], and will flow through the dirty NIF CPU queue. If you're using something like HAProxy for SSL termination, you won't normally encounter this, however, we can still end up making dirty NIF calls for a push replication to a remote instance over https. The end result is that the `run_queue` statistic in the `/_system` endpoint will reflect the amount of crypto work going through the dirty cpu queue and will give false positives on run_queue overload, as elevated run_queues are usually indicative of the Erlang VM having more work to handle that resources available.

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your actual CouchDB! )

## Steps to Reproduce

[NOTE]: # ( Include commands to reproduce, if possible. curl is preferred. )

See the bug report in [2] for more details.

## Expected Behaviour

[NOTE]: # ( Tell us what you expected to happen. )

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )
[TIP]:  # ( You can paste the output of curl http://YOUR-COUCHDB:5984/ here. )

* CouchDB version used:
* Browser name and version:
* Operating system and version:

## Additional Context

[TIP]:  # ( Add any other context about the problem here. )

[1] https://github.com/apache/couchdb/blob/master/src/chttpd/src/chttpd_node.erl#L216
[2] https://bugs.erlang.org/browse/ERL-1355
[...]

PR #3161, opened 2020-09-17
Workaround dirty schedulers in run_queue stats

chewbranca:
## Overview

This is a first pass at fixing https://github.com/apache/couchdb/issues/3160 for reporting proper run_queue statistics.

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_node.erl
+++ b/src/chttpd/src/chttpd_node.erl
@@ -15,7 +15,8 @@
 
 -export([
     handle_node_req/1,
-    get_stats/0
+    get_stats/0,
+    run_queues/0
 ]).
 
 -include_lib("couch/include/couch_db.hrl").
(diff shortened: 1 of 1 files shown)


davisp:
Why not do the version that includes the dirty CPU schedulers given that we're doing dirty IO schedulers?

chewbranca:
> Why not do the version that includes the dirty CPU schedulers given that we're doing dirty IO schedulers?

@davisp I'm assuming you meant the opposite, I'm using the version with dirty CPU schedulers since we're not using IO schedulers.

More to the point, I'm using `statistics(run_queue_lengths)` [1], which returns the list or normal scheduler queues followed by a single entry for the dirty CPU queue. This works for our immediate needs and what the PR does. There's also `statistics(run_queue_lengths_all)` [2] which returns the list of normal scheduler queues followed by an entry for the dirty CPU queue and another entry for the IO dirty queue.

There isn't any dirty IO in Erlang yet, and we're not using any either, so I didn't include that. I think it does make sense to report the dirty IO queues as Erlang will switch to using them at some point in the future, however, I did not do that in this PR as the `statistics(run_queue_lengths_all)` statistic didn't land until OTP-20, and we still have partial support for OTP-19.
The `statistics(run_queue_lengths)` statistic landed in OTP-18.3 so we're free to use that. I'm not opposed to using `statistics(run_queue_lengths_all)`, but if we're going to do that I think we should drop support for OTP-19 in the process.

[1] http://erlang.org/doc/man/erlang.html#statistics_run_queue_lengths
[2] http://erlang.org/doc/man/erlang.html#statistics_run_queue_lengths_all

davisp:
That's fair. I didn't realize those two weren't released in the same version. I'm fine leaving it until we've moved our minimum version to 20.

davisp (approved):
+1

PR #3164, opened 2020-09-18
Feature ebtree views

davisp:
## Overview

Full map/reduce implementation using ebtree's

## Testing recommendations

`make check`

These are a few other attempts at reduce functions with various tradeoffs and limitations.

#2456
#2984 
#3018 

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -337,6 +337,12 @@ iterations = 10 ; iterations for password hashing
 ; The maximum allowed value size emitted from a view for a document (in bytes)
 ;value_size_limit = 64000
 ;
+; The maximum size of B+Tree nodes used by the id btree
+;id_btree_node_size = 100
+;
+; The maximum size of B+Tree nodes used by view btrees
+;view_btree_node_size = 100
+;
 ; Batch size sensing parameters
 ; batch_initial_size = 100 ; Initial batch size in number of documents
 ; batch_search_increment = 500 ; Size change when searching for the threshold
--- a/src/couch_views/src/couch_views_fdb.erl
+++ b/src/couch_views/src/couch_views_fdb.erl
@@ -13,6 +13,8 @@
 -module(couch_views_fdb).
 
 -export([
+    get_view_state/2,
+
     new_interactive_index/3,
     new_creation_vs/3,
     get_creation_vs/2,
@@ -40,52 +42,89 @@
 -include_lib("fabric/include/fabric2.hrl").
 
 
-new_interactive_index(Db, Mrst, VS) ->
-    couch_views_fdb:new_creation_vs(Db, Mrst, VS),
-    couch_views_fdb:set_build_status(Db, Mrst, ?INDEX_BUILDING).
+get_view_state(Db, #mrst{} = Mrst) ->
+    get_view_state(Db, Mrst#mrst.sig);
+
+get_view_state(Db, Sig) when is_binary(Sig) ->
+    #{
+        tx := Tx
+    } = Db,
+
+    VersionF = erlfdb:get(Tx, version_key(Db, Sig)),
+    ViewSeqF = erlfdb:get(Tx, seq_key(Db, Sig)),
+    ViewVSF = erlfdb:get(Tx, creation_vs_key(Db, Sig)),
+    BuildStatusF = erlfdb:get(Tx, build_status_key(Db, Sig)),
+
+    Version = case erlfdb:wait(VersionF) of
+        not_found -> not_found;
+        VsnVal -> element(1, erlfdb_tuple:unpack(VsnVal))
+    end,
+
+    ViewSeq = case erlfdb:wait(ViewSeqF) of
+        not_found -> <<>>;
+        SeqVal -> SeqVal
+    end,
+
+    ViewVS = case erlfdb:wait(ViewVSF) of
+        not_found -> not_found;
+        VSVal -> element(1, erlfdb_tuple:unpack(VSVal))
+    end,
+
+    State = #{
+        version => Version,
+        view_seq => ViewSeq,
+        view_vs => ViewVS,
+        build_status => erlfdb:wait(BuildStatusF)
+    },
+
+    maybe_upgrade_view(Db, Sig, State).
+
+
+new_interactive_index(Db, #mrst{} = Mrst, VS) ->
+    new_interactive_index(Db, Mrst#mrst.sig, VS);
+
+new_interactive_index(Db, Sig, VS) ->
+    set_version(Db, Sig),
+    new_creation_vs(Db, Sig, VS),
+    set_build_status(Db, Sig, ?INDEX_BUILDING).
 
 
 %Interactive View Creation Versionstamp
 %(<db>, ?DB_VIEWS, ?VIEW_INFO, ?VIEW_CREATION_VS, Sig) = VS
 
 new_creation_vs(TxDb, #mrst{} = Mrst, VS) ->
-    #{
-        tx := Tx
-    } = TxDb,
-    Key = creation_vs_key(TxDb, Mrst#mrst.sig),
-    Value = erlfdb_tuple:pack_vs({VS}),
-    ok = erlfdb:set_versionstamped_value(Tx, Key, Value).
+    new_creation_vs(TxDb, Mrst#mrst.sig, VS);
 
-
-get_creation_vs(TxDb, #mrst{} = Mrst) ->
-    get_creation_vs(TxDb, Mrst#mrst.sig);
-
-get_creation_vs(TxDb, Sig) ->
+new_creation_vs(TxDb, Sig, VS) ->
     #{
         tx := Tx
     } = TxDb,
     Key = creation_vs_key(TxDb, Sig),
-    case erlfdb:wait(erlfdb:get(Tx, Key)) of
-        not_found ->
-            not_found;
-        EK ->
-            {VS} = erlfdb_tuple:unpack(EK),
-            VS
-    end.
+    Value = erlfdb_tuple:pack_vs({VS}),
+    ok = erlfdb:set_versionstamped_value(Tx, Key, Value).
+
+
+get_creation_vs(TxDb, MrstOrSig) ->
+    #{
+        view_vs := ViewVS
+    } = get_view_state(TxDb, MrstOrSig),
+    ViewVS.
 
 
 %Interactive View Build Status
 %(<db>, ?DB_VIEWS, ?VIEW_INFO, ?VIEW_BUILD_STATUS, Sig) = INDEX_BUILDING | INDEX_READY
 
-get_build_status(TxDb, #mrst{sig = Sig}) ->
+get_build_status(TxDb, MrstOrSig) ->
     #{
-        tx := Tx
-    } = TxDb,
-    Key = build_status_key(TxDb, Sig),
-    erlfdb:wait(erlfdb:get(Tx, Key)).
+        build_status := BuildStatus
+    } = get_view_state(TxDb, MrstOrSig),
+    BuildStatus.
 
 
-set_build_status(TxDb, #mrst{sig = Sig}, State) ->
+set_build_status(TxDb, #mrst{} = Mrst, State) ->
+    set_build_status(TxDb, Mrst#mrst.sig, State);
+
+set_build_status(TxDb, Sig, State) ->
     #{
         tx := Tx
     } = TxDb,
@@ -98,24 +137,18 @@ set_build_status(TxDb, #mrst{sig = Sig}, State) ->
 % (<db>, ?DB_VIEWS, Sig, ?VIEW_UPDATE_SEQ) = Sequence
 
 
-get_update_seq(TxDb, #mrst{sig = Sig}) ->
+get_update_seq(TxDb, MrstOrSig) ->
     #{
-        tx := Tx,
-        db_prefix := DbPrefix
-    } = TxDb,
-
-    case erlfdb:wait(erlfdb:get(Tx, seq_key(DbPrefix, Sig))) of
-        not_found -> <<>>;
-        UpdateSeq -> UpdateSeq
-    end.
+        view_seq := ViewSeq
+    } = get_view_state(TxDb, MrstOrSig),
+    ViewSeq.
 
 
 set_update_seq(TxDb, Sig, Seq) ->
     #{
-        tx := Tx,
-        db_prefix := DbPrefix
+        tx := Tx
     } = TxDb,
-    ok = erlfdb:set(Tx, seq_key(DbPrefix, Sig), Seq).
+    ok = erlfdb:set(Tx, seq_key(TxDb, Sig), Seq).
 
 
 list_signatures(Db) ->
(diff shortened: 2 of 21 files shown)


src/couch_views/src/couch_views_indexer.erl
```diff
@@ -589,6 +613,51 @@ fail_job(Job, Data, Error, Reason) ->
...
+
+stat_dump() ->
+    case get('$view_stats') of
+        #{} = Stats ->
+            KVs = lists:sort(maps:to_list(Stats)),
+            Strs = lists:foldl(fun({Id, Value}, Acc) ->
+                Str = io_lib:format("~s:~w", [Id, Value]),
+                [Str | Acc]
+            end, [], KVs),
+            Msg = "XKCD VIEW STATS: " ++ string:join(lists:reverse(Strs),  " "),
```
nickva, line 653: Debug log

davisp: Whoops. That whole commit is debugging. Will remove it.

davisp: And its like it never even happened.

src/couch_views/test/couch_views_indexer_test.erl
```diff
@@ -388,8 +384,8 @@ multiple_design_docs(Db) ->
 
     % This is how we check that no index updates took place
     meck:new(couch_views_fdb, [passthrough]),
-    meck:expect(couch_views_fdb, write_doc, fun(TxDb, Sig, ViewIds, Doc) ->
-        meck:passthrough([TxDb, Sig, ViewIds, Doc])
+    meck:expect(couch_views_fdb, update_views, fun(TxDb, Mrst, Docs) ->
```
iilyak, line 387: This is confusing.  We create an `expect` for an `update_views`, but use `meck:num_calls/3` on `write_doc`.

garrensmith (changes requested):
Generally looking good. The big issue I have is the tree related functions in `couch_views_fdb`, I think they should be in a new module. I'm not sure why we need the `id_tree` though, it wasn't obvious to me what that buys us other than everything is a btree?

Finally, It would be nice to see some performance results and how it compares to the old map implementation.

rel/overlay/etc/default.ini
```diff
@@ -337,6 +337,12 @@ iterations = 10 ; iterations for password hashing
 ; The maximum allowed value size emitted from a view for a document (in bytes)
 ;value_size_limit = 64000
 ;
+; The order of B+Tree nodes used by the id btree
```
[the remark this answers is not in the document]
davisp: I've just changed the comment to say "maximum size" rather than using the term from literature.

src/couch_views/test/couch_views_indexer_test.erl
davisp: Good catch. write_doc was renamed and I forgot to update the test. Will update with a fix.

src/couch_views/test/couch_views_red_test.erl
```diff
@@ -0,0 +1,764 @@
...
+-include_lib("couch/include/couch_eunit.hrl").
+-include_lib("couch/include/couch_db.hrl").
+-include("couch_views.hrl").
+
+
+-define(TDEF(A), {atom_to_list(A), fun A/1}).
+-define(TDEFI(A), {atom_to_list(A), fun A/0}).
+
+
+with(Tests) ->
```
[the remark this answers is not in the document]
davisp: Good catch. I didn't realize till after that I could include a `.hrl` from a test directory. Have updated to remove these macros.

rel/overlay/etc/default.ini
```diff
@@ -337,6 +337,12 @@ iterations = 10 ; iterations for password hashing
...
+;id_btree_node_size = 100
+;
+; The order of B+Tree nodes used by view btrees
+;view_btree_node_size = 100
```
nickva, line 344: What are some trade-offs for view be when setting this higher or lower. Say what would setting to 10 do vs say 1000

src/couch_views/src/couch_views_indexer.erl
```diff
@@ -110,6 +112,10 @@ init() ->
         error:database_does_not_exist ->
             fail_job(Job, Data, db_deleted, "Database was deleted");
         Error:Reason  ->
+            Stack = erlang:get_stacktrace(),
+            Fmt = "Error building view for ddoc ~s in ~s: ~p:~p ~p",
```
nickva, line 116: Good idea to add this

src/couch_views/src/couch_views_indexer.erl
```diff
@@ -102,6 +102,8 @@ init() ->
         update_stats => #{}
     },
 
+    process_flag(sensitive, false),
```
nickva, line 105: Why is this needed?

rel/overlay/etc/default.ini
davisp: There's a bunch of different variables that we'll want to balance:

* Size of resultant FDB values
* Height of the tree affects performance (lower limits cause taller trees)
* Number of FDB KVs touched in a transaction
* Speed of reductions
* Speed of iterating over maps
* Lots of other things

100 is basically a hand wave guess that seems to work well in practice. But I didn't want it hard coded so exposed the knobs.

src/couch_views/src/couch_views_indexer.erl
davisp: Ah derps. That was leftover from testing when trying to trace when aegis was accidentally setting it to true. The other spot has been fixed so this can be removed now.

src/couch_views/test/couch_views_indexer_test.erl
davisp: Fixed.

src/couch_views/src/couch_views_fdb.erl
```diff
@@ -244,200 +290,284 @@ clear_index(Db, Signature) ->
...
+    Prefix = id_tree_prefix(DbPrefix, Sig),
+    TreeOpts = [
+        {persist_fun, fun persist_chunks/3},
+        {cache_fun, create_cache_fun(id_tree)}
+    ],
+    ebtree:open(Tx, Prefix, get_order(id_btree), TreeOpts).
 
 
-update_row_count(TxDb, Sig, ViewId, Increment) ->
+open_view_tree(TxDb, Sig, _Lang, View) ->
```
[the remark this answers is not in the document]
davisp: I've gone and created `couch_views_trees.erl` which contains all of the tree reading and writing related functions.

src/couch_views/src/couch_views_fdb.erl
```diff
@@ -244,200 +290,284 @@ clear_index(Db, Signature) ->
...
+    InfoAcc1 = #{
+        ids => BaseIdMap,
+        views => BaseViewMaps,
+        delete_ref => DeleteRef
+    },
 
-row_count_key(DbPrefix, Sig, ViewId) ->
-    Key = {?DB_VIEWS, ?VIEW_INFO, ?VIEW_ROW_COUNT, Sig, ViewId},
-    erlfdb_tuple:pack(Key, DbPrefix).
+    lists:foldl(fun(Doc, InfoAcc2) ->
```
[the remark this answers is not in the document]
davisp: This was changed so the anonymous function is named and includes this function head match.

src/couch_views/src/couch_views_fdb.erl
```diff
@@ -126,92 +128,136 @@ set_update_seq(TxDb, Sig, Seq) ->
...
-        end,
-        update_map_idx(TxDb, Sig, ViewId, DocId, ExistingKeys, NewRows)
-    end, lists:zip3(ViewIds, Results, KVSizes)).
+        ids := IdMap,
+        views := ViewMaps,
+        delete_ref := DeleteRef
+    } = gather_update_info(Tx, Mrst, Docs),
+
+    % Generate a list of Keys to delete and Rows to insert from a map
+    UpdateBTree = fun(BTree, Map) ->
```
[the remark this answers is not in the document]
davisp: Done.

src/couch_views/test/couch_views_red_test.erl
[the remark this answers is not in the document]
davisp: Updated to use fabric2_test.hrl.

src/couch_views/test/couch_views_red_test.erl
```diff
@@ -0,0 +1,764 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
```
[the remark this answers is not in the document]
davisp: Added your tests.

src/couch_views/src/couch_views_indexer.erl
davisp: Removed.

src/couch_views/src/couch_views_fdb.erl
```diff
@@ -126,92 +128,136 @@ set_update_seq(TxDb, Sig, Seq) ->
...
+        {ToRemove, ToInsert} = maps:fold(fun(Key, Value, {Keys, Rows}) ->
+            case Value of
+                DeleteRef -> {[Key | Keys], Rows};
```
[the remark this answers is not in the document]
davisp: I don't like that. The loop splits into two groups whcih are then acted on separately. Acting on some keys while reserving others seems more awkward than a split.

davisp:
@garrensmith I've updated things to use a `couch_views_trees` module for all the tree related functions. On using ebtree for the `id_tree` it simplifies things greatly as we don't have to maintain a separate set of encoding/decoding functions.

I'll run through a 1M doc update again locally and post a link to the scripts and results. Last I checked it was roughly 2-5 times faster than the existing view implementation but I'll double check I've not messed anything up.

nickva (approved):
I built views with some built-in reducers (`_count`, `_approx_count_distinct`) then a few custom ones. Tested view updates, queries, group_levels, descending=true|false, start_key/end_key. Set `id_btree_node_size` and `view_btree_node_size` to smaller values like 6, 10 and 20 and it worked.

Good idea to enable some more elixir tests and bring in @garrensmith's work

There are also a few basic tests that use reduce that were commented out, we can uncomment them now as well:

https://github.com/apache/couchdb/blob/fa3c3c8dd3d5b77371cd74dfbe149623797160ad/test/elixir/test/basics_test.exs#L195

https://github.com/apache/couchdb/blob/fa3c3c8dd3d5b77371cd74dfbe149623797160ad/test/elixir/test/basics_test.exs#L208

One thing that changed from 3.x is how we handle errors:

```
{
    "rows": [
        {
            "error": "reduce_overflow_error",
            "key": null,
            "reason": "Reduce output must shrink more rapidly: input size: 671 output size: 10003"
        }
    ]
}
```

vs

```
{
    "rows": [
        {
            "key": null,
            "value": {
                "reduce_overflow_error": "Reduce output must shrink more rapidly: input size: 779 output size: 10003"
            }
        }
    ]
}
```

However this can be done in a separate PR.

Nice work!

+1

davisp:
For the reduce error test I'm working on a second PR that re-enables most of the Elixir test suite that now passes. This error is flagged there so we can keep track of it that way.

nickva:
@davisp that sounds good!

davisp:
For reference, that PR is #3175 but has this PR branch as its base.

davisp:
These are three replicas building a simple view against 1M randomized documents. Each test is local to my laptop so take that for what its worth.

main: 7m40s, 8m40s, 9m8s
feature-ebtree-views: 8m46s, 9m36s, 8m47s

I'm fairly surprised to see the main branch looking a bit faster with that first run at 7m40s. When we've run similar tests against FoundationDB clusters that version is significantly slower than the ebtree branch. My best guess at this point is that its due to touching a significantly larger number of fdb key/value pairs since each row is separate.

davisp:
Also, the scripts for generating those numbers are here:

https://gist.github.com/davisp/3863ab02fb5def733f1a6991651840f9

garrensmith:
I did a quick group_level benchmarking this morning to compare it to the map + ebtree work I built-in #3018. The benchmark builds a view with keys that are a group_level = 3. Group_levels 1 and 2 will group results. 

The full ebtree builds a lot faster, which is great. And queries for group_level = 1 and group_level = 2 are very similar. But when doing a reduce query for group_level = 3 the full ebtree is noticeably slower.

Example results for 300K docs:
full ebtree (this PR):
```
Insert Time  71435.896557 ms
Querying:
Group 1 Time 12.577319 ms
Result Length 5
Group 2 Time 20.666643 ms
Result Length 20
Group 3 Time 3712.46744 ms
Result Length 300002
```

Map + ebtree reduce (PR #3018):
```
Insert Time  362359.221368 ms
Querying:
Group 1 Time 13.821064 ms
Result Length 5
Group 2 Time 23.708709 ms
Result Length 20
Group 3 Time 2672.549299 ms
Result Length 300002
```

Code for the benchmarking is here https://gist.github.com/garrensmith/9bd090cb00f6b641e58a3c490b2300e7
Same warning as Paul's this is run locally on my machine.

davisp:
@garrensmith That seems reasonably close. I'd suspect that the difference there is likely the number of times that ebtree-only has to call out to the reduce function to process individual groups. Though 80k rows/sec vs 110k rows/sec is honestly a bit surprising given that we're running the reduce function 300k times roughly.

Are you concerned about the difference there? It seems not a deal breaker given that the full ebtree PR supports 100% API compatibility for reduce functions rather than just pre-specified group levels.

garrensmith:
Hi Paul,

That isnt correct. The benchmark is with my branch that uses ebtree for reduce along with the old map index. So it also supports the full reduce api. I found it surprising that it was faster than this PR specially since it uses a much older version of ebtree. I think it would be worth investigating why this work is slower than my PR.

davisp:
Oh derps. I was thinking of the pre-computed group levels. Somewhat intriguing. I'll investigate tomorrow and post back.

davisp:
@garrensmith A bit of profiling showed that I was doing a bit of extra work compared to #3018 so there were a few simple optimizations to be made. Namely:

1. Avoid calculating unused reductions while reading from a reduce view (coincidentally an optimization that exists even in the legacy view engine
2. Avoid some extra work in couch_query_servers that I re-used instead of duplicating the functionality of this PR.

For 1, the #3018 PR achieves this same end by having a separate ebtree instance for each reduce function on the view. For 2, its similar other than the size calculation is done on at write time.

Screen Shot 2020-09-29 at 3 18 02 PM

Updated numbers generated locally. I ported the Node.js script to Python [1] and removed deleting the database part. I then ran it four times. Once to generate the index, and then three times to gather numbers after everything was warmed up in cache to try and get an equal footing for each replica.

There's still about 200ms difference between the two implementations. However, based on profiling both implementations it would appear that a good chunk of that is due to #3018 performing significantly fewer invocations of collation which is a bug in a previous version of ebtree that has since been fixed [2]. I also went fixed the lack of JSON collation in #3018 [3] and that appears to have made it faster.

I'm fairly confident I've covered the differences given the similar algorithmic properties of the implementations. Let me know if you're OK merging this now.

[1] https://gist.github.com/davisp/48303dcd877b31632fb75055d6758129
[2] https://github.com/apache/couchdb/commit/f8fdf9721e2ac932022065bc075301641568d67c
[3] https://github.com/cloudant/couchdb/compare/fdb-btree-reduce...apache:fdb-btree-reduce-davisp

garrensmith:
Nice work. Thanks for checking that. I’m away until next week so I can’t give a final review until then. I can’t remember if anyone else has approved this or not. If they have then go ahead and merge.

rnewson (approved):
Thanks Garren.

I've read all the above, and the code (both current and previous versions as its evolved), I'm happy to move forward.

PR #3174, opened 2020-09-24
Port rewrite and rewrite_js tests into elixir

jjrodrig:
## Overview

This PR ports these tests to Elixir
 - rewrite.js
 - rewrite_js.js

 rewrite.js was partially ported before this PR, now it is completely ported. 

 Rawresp module has been moved and now can be used by any test if a non-parsed http response body is required.

## Testing recommendations

```
 make elixir
```
N/A

## Checklist


test/elixir/lib/couch.ex
```diff
@@ -175,3 +175,109 @@ defmodule Couch do
     end
   end
 end
+
+defmodule Rawresp do
```
dottorblaster, line 179: couldn't we just write a new file for this? Too much pain? (at the moment I can't remember if we could call those functions inside the module this way)

dottorblaster (changes requested):
Astonishing 🐱 

Could you just look into that comment?

test/elixir/lib/couch.ex
jjrodrig: Sure, i'll put it in a new file

Really, a tried to do it with the current Couch module at couch.ex but I don't know how to parametrize this behaviour.

jjrodrig: Done

dottorblaster (approved):
LGTM

jjrodrig:
Thanks!

PR #3188, opened 2020-10-01
vm.args: add note about changing -name

sdahlbac:
Try to gently suggest using option 2)

## Overview

As per discussions in slack a few days ago:

Since changing -name after the fact involves moving all the shards, which is a bit of work and potentially errorprone.

Try to nudge user towards not using the couchdb@127.0.0.1 if there is a possibility that e.g. clustering is needed down the road.

## Testing recommendations

N/A

## Checklist

### DIFF
--- a/rel/overlay/etc/vm.args
+++ b/rel/overlay/etc/vm.args
@@ -33,6 +33,10 @@
 # Do not rely on tricks with /etc/hosts or libresolv to handle anything
 # other than the above 3 approaches correctly. They will not work reliably.
 #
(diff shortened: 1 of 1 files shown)


wohali:
As this is a footnote and only applies to people changing the value after the fact, It would be preferable to put this in the documentation, not in this file.

Would you consider opening a pull request against our documentation repository instead?  https://github.com/apache/couchdb-documentation/ Thanks!

sdahlbac:
Sure, I'll do that.

PR #3190, opened 2020-10-05
Port users_db_security javascript tests to elixir

jjrodrig:
## Overview
This PR ports `users_db_security.js` tests into the elixir test suite. During the porting of this test I needed to add support for raw responses to Couch.Session module.

This is the last active test to be ported from the javascript test suite. The javascript test suite can be removed from the main build process as there are no more test being ran. I'll prepare a separate PR for this.
  

## Testing recommendations
`make javascript` All test should be skipped or ported
`make elixir` All test should pass

N/A

## Checklist

### DIFF
--- a/test/elixir/README.md
+++ b/test/elixir/README.md
@@ -95,7 +95,7 @@ X means done, - means partially
   - [ ] Port stats.js
   - [X] Port update_documents.js
   - [X] Port users_db.js
-  - [ ] Port users_db_security.js
+  - [X] Port users_db_security.js
   - [X] Port utf8.js
   - [X] Port uuids.js
   - [X] Port view_collation.js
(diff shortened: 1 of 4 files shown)


dottorblaster:
Tests are red 🚨 did the modification to `couch.ex` affect some older stuff maybe?

jjrodrig:
oops... The problem with this test is that it is using _users database. If I'm not properly cleaning it we can have a side effect... I'll check it.

couch.ex changes are not the problem here. It's the execution order.

dottorblaster:
Thanks for checking!

jjrodrig:
I've rewritten the test cleaning process.  I remove each user that has been previously created. 
Now, it does not collide with any other test.
Thanks!

dottorblaster (approved):

jjrodrig:
Thanks!! the last one :tada:

wohali:
@jjrodrig is that the last JS test now?

If so, I can start the work to rip the JS test harness out of our build chain, makefiles, etc.

jjrodrig:
@wohali I'm just creating a PR that removes `@$(MAKE) javascript` from the `check` target

There are some tests that have not been ported. These tests are currently skipped in the JS test suite. I think, that the work is done.

PR #3192, opened 2020-10-05
Proposal for removing javascript tests from main build process

jjrodrig:
## Overview
All the active JS tests have been ported to elixr. The current `make javascript` execution is not performing any test.
I propose to remove it from the main build process.

This PR removes `@$(MAKE) javascript` from the `check` target

## Testing recommendations
``` make check``` should not run the javascript test suite

N/A

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,6 @@ fauxton: share/www
 check: all python-black
 	@$(MAKE) emilio
 	@$(MAKE) eunit
-	@$(MAKE) javascript
 	@$(MAKE) mango-test
 	@$(MAKE) elixir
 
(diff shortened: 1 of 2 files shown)


wohali:
Hi @jjrodrig , this isn't sufficient to fully remove JS from our build process, but if you want to commit this and leave the rest to me, +1.

There's more in `configure` that needs to be dealt with, and certain things like our Windows builds will change. We will also no longer depend on `libcurl` which is an advantage in and of itself.

wohali:
Also you probably want to target `main` instead of `master`, and don't forget `3.x`.

jjrodrig:
Yes, more work is required to fully remove the JS support.  But, as an initial step, we could just stop running javascript target during the build process.
I'll move all these commits to main and 3.x
Thanks.

PR #3201, opened 2020-10-07
Fix boolean POST to view-like endpoints

bessbd:
## Overview
When using a POST with request body to pass parameters to a view-like request the boolean parameters are accepting only JSON strings, but not booleans. For example the stable parameter works with "true" and "false" but not true and false.

## Testing recommendations

`mix test --trace test/elixir/test/all_docs_test.exs`
and
`mix test --trace test/elixir/test/view_test.exs`

## Related Issues or Pull Requests

This fixes #3200 .

## Checklist

- [x] Code is written and works correctly
- [x] Changes are covered by tests
- [x] Any new configurable parameters are documented in `rel/overlay/etc/default.ini` -- N/A
- [x] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation -- N/A

PR #3210, opened 2020-10-09
Fix boolean POST to view-like endpoints

bessbd:
This is a backport of #3201


jjrodrig (approved):
+1 since it is already merged into master

Thanks @bessbd

bessbd:
Thank you for the review and merge, @jjrodrig

PR #3216, opened 2020-10-14
Direct people to Discussions/docs from new issue

wohali:
This presents new links on the New Issue page, similar to this:

Sample image


nickva (approved):
Good idea!

PR #3268, opened 2020-11-16
Use `req_body` field if present

iilyak:
## Overview

When we call `couch_httpd:json_body/1` we can have `req_body` already set.
decompress or decode it. This PR brings the approach we use in `chttpd`
into `couch_httpd`.

## Testing recommendations

```
make check
```

We port an approach from https://github.com/apache/couchdb/pull/2991 into `couch_httpd`

## Checklist


nickva (approved):
+1

rnewson:
needs backporting to 3.x.

noting also that this only matters if code modifies the MochiReq and sets req_body outside of mochiweb, which is something Cloudant does but CouchDB does not currently do.

PR #3271, opened 2020-11-16
Couch att erroneous md5 mismatch

rnewson:
## Overview

If an attachment was stored uncompressed but later is replicated
internally to a node that wants to compress it (based on
content-type), couchdb compares the uncompressed md5 with the
compressed md5 and fails. This breaks eventual consistency between
replicas.

This PR removes the unnecessary MD5 check that is, in these specific
circumstances, always called with mismatched arguments.

## Testing recommendations

1. Upload an attachment with a content-type not included in compressible_type config
2. change compressible_type to include this content-type
3. stop a node
4. delete the hosting shard of the attachment
5. start the node.
6. observe md5_mismatch errors in the log.

N/A

## Checklist

### DIFF
--- a/src/couch/src/couch_att.erl
+++ b/src/couch/src/couch_att.erl
@@ -593,13 +593,9 @@ flush_data(Db, {stream, StreamEngine}, Att) ->
             % Already written
             Att;
         false ->
-            NewAtt = couch_db:with_stream(Db, Att, fun(OutputStream) ->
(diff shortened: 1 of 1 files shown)


davisp:
To be clear, the check that's being removed is both incorrect and redundant. There's a compression aware check that happens inside the `couch_db:with_stream` function so even if this check had been correct it would still be redundant.

PR #3276, opened 2020-11-19
Use `req_body` field if present

rnewson:
When we call `couch_httpd:json_body/1` we can have `req_body` already set.
decompress or decode it. This PR brings the approach we use in `chttpd`
into `couch_httpd`.

backported from https://github.com/apache/couchdb/pull/3268


iilyak (approved):
+1

iilyak:
This also depends on https://github.com/apache/couchdb/pull/2991

iilyak:
The PR which backport both is here https://github.com/apache/couchdb/pull/3277

PR #3277, opened 2020-11-19
Req body json 3.x

iilyak:
## Overview

When we call couch_httpd:json_body/1 we can have req_body already set.
decompress or decode it. This PR brings the approach we use in chttpd
into couch_httpd.

backported from #3268 and #2991

## Testing recommendations

`make check`

- #2991
- #3268

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -413,8 +413,7 @@ possibly_hack(#httpd{path_parts=[<<"_replicate">>]}=Req) ->
     {Props0} = chttpd:json_body_obj(Req),
     Props1 = fix_uri(Req, Props0, <<"source">>),
     Props2 = fix_uri(Req, Props1, <<"target">>),
-    put(post_body, {Props2}),
-    Req;
+    Req#httpd{req_body={Props2}};
 possibly_hack(Req) ->
     Req.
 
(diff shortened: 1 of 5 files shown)


rnewson:
In all existing cases, req_body is either undefined or it is the result of MochiReq:recv_body/1 or recv_body/2.

In this PR we are doing something different, req_body will be the json decoded version of that body.

I think that's too confusing, especially as the request body caching code is already confusing enough due to its history.

iilyak:
> I think that's too confusing, especially as the request body caching code is already confusing enough due to its history.

The two PRs backported here move away from implicit value propagation via process dictionary to explicit propagation via a field in the chttpd record. Which is easier to understand and maintain.

rnewson:
Could you address my body about the change of the format of req_body in this PR versus all the current uses of req_body?

iilyak:
> Could you address my comment about the change of the format of req_body in this PR versus all the current uses of req_body?

The purpose of https://github.com/apache/couchdb/pull/2991 is to allow injecting modified body from EPI plugins. Without this change it is quite hard to modify body because the one would need to either:
- find a way how to override HTTP headers in Mochiweb to change "Content-Encoding" to identity (there is no official way).
- convert JSON object into binary and optionally gzip it if `"Content-Encoding"` header contains `gzip` (which would add an overhead for big requests).

It also let us to remove process dictionary use in `chttpd_misc:handle_replicate_req` (leaky abstraction).

The change allows EPI users to easily modify body. 
- `Req#httpd{req_body={Props}}` - for JSON
- `Req#httpd{req_body=<<"non-JSON body">>}` - for non JSON

The `json_body/1` always returns JSON obj:
- when `req_body` is undefined we decompress and decode JSON from Mochiweb
- when `req_body` is set we return it as is (we assume it is set as `#httpd{req_body={Props}}` i.e. contains JSON obj, this is a responsibility of EPI plugin author).

rnewson (approved):

iilyak:
@rnewson I noticed a problem with `_list` endpoint and pushed a fix after your approval. This fixup don't need to be ported to `main` because `_list` endpoint is [disabled](https://github.com/apache/couchdb/blob/main/src/chttpd/src/chttpd_httpd_handlers.erl#L55) there

PR #3286, opened 2020-12-02
Add ability to control which Elixir integration tests to run

iilyak:
## Overview

New `elixir-suite` Makefile target is added. It runs a predefined set of elixir
integration tests.

The feature is controlled by two files:
- test/elixir/test/config/suite.elixir - contains list of all available tests
- test/elixir/test/config/skip.elixir - contains list of tests to skip

In order to update the `test/elixir/test/config/suite.elixir` when new tests
are added. The one would need to run the following command:

```
MIX_ENV=integration mix suite > test/elixir/test/config/suite.elixir
```

The number of enabled tests in this PR is 237, while `make check` has only 80.

## Testing recommendations

1) Make sure that tests discovery works
```
MIX_ENV=integration mix suite
```
2) Make sure the tests run and pass by using `make elixir-suite`

- somewhat related to https://github.com/apache/couchdb/issues/1515 in a sense that it allows to have different skip files for builds with different extensions (geo+search) 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -271,6 +271,17 @@ elixir-cluster-with-quorum: elixir-init elixir-check-formatted elixir-credo devc
 		--degrade-cluster 1 \
 		--no-eval 'mix test --trace --only with_quorum_test $(EXUNIT_OPTS)'
 
+.PHONY: elixir-suite
+elixir-suite: export MIX_ENV=integration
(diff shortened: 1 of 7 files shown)


nickva (approved):
Looks good. Nice work!

Since our `make check` is a manual suite of tests we run (https://github.com/apache/couchdb/blob/main/Makefile#L164) could we use elixir-suite there?

PR #3299, opened 2020-12-15
Support pluggable custodian monitor

jaydoane:
## Overview

Enable build time configurable monitor for custodian and remove custom
sensu events.

## Testing recommendations

```
make eunit apps=custodian
```

## Checklist

### DIFF
--- a/src/custodian/src/custodian_monitor.erl
+++ b/src/custodian/src/custodian_monitor.erl
@@ -10,20 +10,19 @@
 % License for the specific language governing permissions and limitations under
 % the License.
 
-{application, custodian,
- [
-  {description, "in your cluster, looking after your stuff"},
-  {vsn, git},
-  {registered, []},
-  {applications, [
-                  kernel,
-                  stdlib,
-                  couch_log,
-                  config,
-                  couch_event,
-                  couch,
-                  mem3
-                 ]},
-  {mod, { custodian_app, []}},
-  {env, []}
- ]}.
+-module(custodian_monitor).
+
+
+% N.B. that callback return values are ignored
+
+-callback send_missing_db_alert(DbName :: binary()) ->
+    Ignored :: any().
+
+
+-callback clear_missing_dbs_alert() ->
+    Ignored :: any().
+
+
+-callback send_event(
+    Name :: string(), Count :: non_neg_integer(), Description :: string()) ->
+    Ignored :: any().
--- a/src/custodian/src/custodian_noop_monitor.erl
+++ b/src/custodian/src/custodian_noop_monitor.erl
@@ -1,0 +1,35 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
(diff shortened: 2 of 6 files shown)


src/custodian/src/custodian_monitor.erl
```diff
@@ -10,20 +10,16 @@
...
-                  mem3
-                 ]},
-  {mod, { custodian_app, []}},
-  {env, []}
- ]}.
+-module(custodian_monitor).
+
+
+-callback send_missing_db_alert(DbName :: binary()) ->
+    Output :: string().
```
rnewson, line 17: I'd encode the fact that we don't use the result of these functions, we call them only for their implementation. so perhaps insist on `false` atom in all cases or say `any()` and document that the result of the function is ignored.

src/custodian/src/custodian_noop_monitor.erl
```diff
@@ -0,0 +1,35 @@
...
+
+-export([
+    send_missing_db_alert/1,
+    clear_missing_dbs_alert/0,
+    send_event/1
+]).
+
+
+send_missing_db_alert(_DbName) ->
+    "no-op".
```
rnewson, line 27: adjust these to `false`.

src/custodian/src/custodian_server.erl
```diff
@@ -143,52 +143,7 @@ handle_db_event(_DbName, _Event, _St) ->
     {ok, nil}.
 
 check_shards() ->
-    [send_sensu_event(Item) || Item <- custodian:summary()].
```
rnewson, line 146: this is a cut too far imo. we still want to log the descriptive error message to couch_log. the pluggable module allows a sysadmin to send that message (or one of its own invention) to some other system.

rnewson (changes requested):
Almost :)

src/custodian/src/custodian_monitor.erl
jaydoane: Done in fixup

src/custodian/src/custodian_noop_monitor.erl
jaydoane: Done

src/custodian/src/custodian_server.erl
jaydoane: Hah, I didn't even notice that logging cleverly interleaved in the case statement. In any case, I've returned them, and changed the `send_event/3` callback to DRY things up a bit.

src/custodian/src/custodian_server.erl
```diff
@@ -143,7 +143,49 @@ handle_db_event(_DbName, _Event, _St) ->
...
+        0 ->
+            "--ok";
+        1 ->
+            couch_log:critical("~s", [Description]),
+            "--critical";
+        _ ->
+            couch_log:warning("~s", [Description]),
+            "--warning"
+    end,
+    ?CUSTODIAN_MONITOR:send_event(Name, Level, Description).
```
rnewson, line 162: think we should send `Count` itself, the transformation to `Level` is sensu-specific. 😬

rnewson (approved):
nice! please squash before merging!

PR #3303, opened 2020-12-19
treat 408 as a retryable error condition

rnewson:
## Overview

While CouchDB itself never responds with a 408 Request Timeout it is possible for a proxy to do so. The CouchDB Replicator treats a 408 as a permanent error, increments doc_write_failures, and keeps going. This is undesirable. This patch reorders the clauses in couch_replicator_httpc so we can easily treat a 408 the same way we treat any 5xx code (by retrying).

## Testing recommendations

TBD

N/A

## Checklist


nickva (approved):
+1

Nice work re-organizing the clauses to make more sense.

I think this PR should apply cleanly to the 3.x branch as well

wohali:
Would love to see this on 3.x @rnewson

PR #3323, opened 2021-01-12
Allow gzipped requests to _session

### LINKED ISSUE #3300: _session does not support compressed body

## Description

Posting a gzipped body to `_session` with `Content-Encoding: gzip` results in:
`{"error":"bad_request","reason":"invalid UTF-8 JSON"}`

## Steps to Reproduce

1. `echo '{"username":"user123","password":"123password"}' | gzip > session.json.gz`
1. `curl -H 'Content-Encoding: gzip' -H 'Content-Type: application/json' -d @session.json.gz http://localhost:5984/_session`
1. `{"error":"bad_request","reason":"invalid UTF-8 JSON"}`

## Expected Behaviour

As far as I can tell every endpoint in CouchDB supports `gzip` or `deflate` encoded bodies so I would expect this request to be accepted and suceed the same way it does if the body is not compressed.

Alternatively, if `gzip` encoding is really not acceptable here then the server should respond with the [correct status code](https://tools.ietf.org/html/rfc7231#section-6.5.13) `415`. The current `400` and error message is incorrect - the JSON is completely valid (ofc I can see why the server thinks it's bad JSON when it reads a bunch of compressed bytes instead, but it shouldn't do that).

## Your Environment

* CouchDB version used: 3.1.1
* Browser name and version: N/A
* Operating system and version: macOS

## Additional Context

[TIP]:  # ( Add any other context about the problem here. )

It would be preferable to accept the compressed bodies for this endpoint in the same way as all the others do because it is far easier in clients and proxies to enable request body compression for all requests than it is to have to control it per endpoint.

bessbd:
## Overview
All endpoints but `_session` support gzip encoding and there's no practical reason for that.
This PR enables gzip decoding on compressed requests to `_session`.

## Testing recommendations

`make eunit apps=chttpd suites=chttpd_session_tests`

This fixes #3300 .
This is #3322 cherry-picked to `main`.

## Checklist

- [ ] Any new configurable parameters are documented in `rel/overlay/etc/default.ini` - N/A
- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation - N/A

### DIFF
--- a/src/chttpd/test/eunit/chttpd_session_tests.erl
+++ b/src/chttpd/test/eunit/chttpd_session_tests.erl
@@ -44,7 +44,8 @@ session_test_() ->
                 fun cleanup/1,
                 [
                     ?TDEF_FE(session_authentication_db_absent),
-                    ?TDEF_FE(session_authentication_db_present)
+                    ?TDEF_FE(session_authentication_db_present),
+                    ?TDEF_FE(session_authentication_gzip_request)
                 ]
             }
         }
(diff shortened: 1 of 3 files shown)


rnewson (approved):

bessbd:
Thank you for the review, @rnewson ! Can you please merge this also for me?

PR #3332, opened 2021-01-18
Timeout transaction

garrensmith:
## Overview

If a mango query exceeds the 5 seconds for a transaction it sends a nasty error:
```
89e90f4c55 req_err(2721591552) erlfdb_error : 1031
    [<<"erlfdb_nif:erlfdb_future_get/1">>,<<"erlfdb:do_transaction/2 L686">>,<<"fabric2_fdb:do_transaction/2 L171">>,<<"fabric2_fdb:transactional/2 L142">>,<<"mango_cursor_view:execute/3 L142">>,<<"mango_httpd:handle_find_req/2 L197">>,<<"mango_httpd:handle_req/2 L37">>,<<"chttpd:handle_req_after_auth/2 L330">>]
```

This cleans that up and adds a better error. We also remove the `{restart_tx, true}` for a mango query that uses `_all_docs` to make it consistent with a mango query that uses a view.

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -364,6 +364,8 @@ catch_error(HttpReq, error, not_ciphertext) ->
     send_error(HttpReq, not_ciphertext);
 catch_error(HttpReq, error, {erlfdb_error, 2101}) ->
     send_error(HttpReq, transaction_too_large);
+catch_error(HttpReq, error, {erlfdb_error, 1031}) ->
+    send_error(HttpReq, transaction_timeout);
 catch_error(HttpReq, Tag, Error) ->
     Stack = erlang:get_stacktrace(),
     % TODO improve logging and metrics collection for client disconnects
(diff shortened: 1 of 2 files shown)


src/chttpd/src/chttpd.erl
```diff
@@ -1014,6 +1016,9 @@ error_info({request_entity_too_large, DocID}) ->
 error_info(transaction_too_large) ->
     {413, <<"transaction_too_large">>,
         <<"The request transaction is larger than 10MB" >>};
+error_info(transaction_timeout) ->
+    {413, <<"transaction_timeout">>,
```
iilyak, line 1020: 413 is Payload Too Large response. I think we need 408.

iilyak (approved):
+1 if CI pass

rnewson:
@garrensmith the 408 status code is for when the client takes too long to send the request body, that doesn't appear to be the reason we get a 1031 from fdb?

PR #3336, opened 2021-01-18
Add a .devcontainer configuration for 3.x

kocolosk:
## Overview

This PR adds a Dockerfile and associated configuration to enable developers to quickly provision an environment with all dependencies installed to work on CouchDB 3.x.

The container configuration also installs the [Erlang Language Server](https://erlang-ls.github.io) extension. That extension needs a minimal configuration file in the root of the project in order to find the include files, so I've added that as well. We could likely iterate and enhance that configuration file further with linters, dialyzers configurations, etc.

## Testing recommendations

If you have VS Code with the "Remote - Containers" extension installed, opening the CouchDB repo should cause the editor to offer to start the development container. The container will include all the dependences necessary to

* build CouchDB using `./configure; make`
* run the full test suite
* build the documentation
* build Fauxton

## Checklist

(No code changes here, just a quicker Getting Started experience)

- [ ] Code is written and works correctly
- [ ] Changes are covered by tests
- [ ] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation

PR #3343, opened 2021-01-24
Add a development container for VS Code / GitHub Codespaces

kocolosk:
## Overview

This creates a development environment with a FoundationDB server and a CouchDB layer in two containers, sharing a network through Docker Compose.

It uses the FDB image published to Docker Hub for the FDB container, and downloads the FDB client packages from foundationdb.org to provide the development headers and libraries. www.foundationdb.org is actually not trusted in Debian Buster by default, so we have to download the GeoTrust_Global_CA.pem. The following link has more details:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=962596

Once the Docker Compose setup is running, VS Code executes the `create_cluster_file.bash` script to write down a cluster file containing the IP address in the compose network where the FDB service can be found. This cluster file is used both for a user-driven invocation of `./dev/run`, as well as for unit tests that require a running CouchDB.

Additionally, I've got a small fix to the Makefile that corrects the approach used to running eunit tests with multiple apps specified on the command line to match the behavior of `make eunit` without any arguments.

## Testing recommendations

If you have VS Code with the "Remote - Containers" extension installed, opening the CouchDB repo should cause the editor to offer to start the development container. The container will include all the dependences necessary to

- build CouchDB using ./configure; make
- run the full test suite
- build the documentation
- build Fauxton

#3336 added a container configuration for 3.x. This one is a step more involved because of the use of Docker Compose with a separate FDB container.

## Checklist

(No code changes here, just a quicker Getting Started experience)

### DIFF
--- a/rel/files/eunit.config
+++ b/rel/files/eunit.config
@@ -13,5 +13,18 @@
 [
     {kernel, [{error_logger, silent}]},
     {sasl, [{sasl_error_logger, false}]},
-    {fabric, [{eunit_run, true}]}
+    {fabric, [{eunit_run, true}]},
+
+    % When fabric is configured with eunit_run=true it will ask erlfdb for a
+    % test database. The default behavior for erlfdb in this case is start a
+    % new fdbserver process for the test. If you would rather have erlfdb
+    % connect to an existing FoundationDB cluster, you can supply the path
+    % to the cluster file as a binary string here.
+    %
+    % NOTE: the unit tests will erase all the data in the cluster!
+    %
+    % The docker-compose configuration in the .devcontainer activates this
+    % application setting using ERL_ZFLAGS in the container environment, so
+    % any tests will use the fdbserver running in the fdb container.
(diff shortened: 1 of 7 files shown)


davisp (approved):
+1

rel/files/eunit.config
```diff
@@ -13,5 +13,18 @@
 [
     {kernel, [{error_logger, silent}]},
     {sasl, [{sasl_error_logger, false}]},
-    {fabric, [{eunit_run, true}]}
+    {fabric, [{eunit_run, true}]},
```
davisp, line 16: Looks like the trailing comma here is causing your parse error. Maybe move your comment and config value before the eunit_run setting to avoid that?

kocolosk: Oh sheesh that's a dumb mistake on my last fixup of this commit adding the extra explanation for the `test_cluster_file`. Will fix.

PR #3347, opened 2021-01-26
Set a finite default for max_attachment_size

bessbd:
## Overview
The current default for `max_attachment_size` is `infinity`. This PR changes that to 1 gibibyte.

## Testing recommendations
No testing, this is just a default change.

N/A

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -47,7 +47,7 @@ changes_doc_ids_optimization_threshold = 100
 max_document_size = 8000000 ; bytes
 ;
 ; Maximum attachment size.
-; max_attachment_size = infinity
+; max_attachment_size = 1073741824 : 1 gibibyte
 ;
 ; Do not update the least recently used DB cache on reads, only writes
 ;update_lru_on_read = false
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -47,7 +47,7 @@ changes_doc_ids_optimization_threshold = 100
 max_document_size = 8000000 ; bytes
 ;
 ; Maximum attachment size.
-; max_attachment_size = 1073741824 : 1 gibibyte
+; max_attachment_size = 1073741824 ; 1 gibibyte
 ;
 ; Do not update the least recently used DB cache on reads, only writes
 ;update_lru_on_read = false
--- a/src/couch/src/couch_att.erl
+++ b/src/couch/src/couch_att.erl
@@ -960,7 +960,7 @@ test_transform() ->
 
 max_attachment_size_test_() ->
     {"Max attachment size tests", [
-        ?_assertEqual(infinity, max_attachment_size(infinity)),
+        ?_assertEqual(infinity, max_attachment_size("infinity")),
         ?_assertEqual(5, max_attachment_size(5)),
         ?_assertEqual(5, max_attachment_size("5"))
     ]}.
--- a/src/couch/src/couch_att.erl
+++ b/src/couch/src/couch_att.erl
@@ -178,6 +178,9 @@
 
 -type att() :: #att{} | attachment() | disk_att().
 
+% 1024 * 1024 * 1024
+-define(GB, 1073741824).
+
 new() ->
     %% We construct a record by default for compatability. This will be
     %% upgraded on demand. A subtle effect this has on all attachments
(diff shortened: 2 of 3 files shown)


src/couch/src/couch_att.erl
```diff
@@ -734,7 +734,7 @@ upgrade_encoding(Encoding) -> Encoding.
 
 
 max_attachment_size() ->
-    case config:get("couchdb", "max_attachment_size", "infinity") of
+    case config:get("couchdb", "max_attachment_size", 1024 * 1024 * 1024) of
```
iilyak, line 737: Erlang doesn't do compile time optimizations. I think multiplications would be slow. Could you use a define instead.

```
% 1024 * 1024 * 1024
-define(GB, 1073741824).

....
max_attachment_size() ->
    case config:get("couchdb", "max_attachment_size", ?GB) of
```

iilyak:
This PR might require a vote on ML.

rel/overlay/etc/default.ini
```diff
@@ -47,7 +47,7 @@ changes_doc_ids_optimization_threshold = 100
 max_document_size = 8000000 ; bytes
 ;
 ; Maximum attachment size.
-; max_attachment_size = infinity
+; max_attachment_size = 1073741824 : 1 gibibyte
```
iilyak, line 50: Use `;` instead of `:` to introduce comment.


`; max_attachment_size = 1073741824 : 1 gibibyte`
............................................................... ^----- should be `;`

bessbd: Done in fa8288bdd

bessbd:
Thank you for the review, @iilyak . I've just noticed that this PR is not passing some tests. I'll push a commit soon to fix that.

src/couch/src/couch_att.erl
```diff
@@ -732,13 +735,19 @@ upgrade_encoding(true) -> gzip;
...
-    case config:get("couchdb", "max_attachment_size", "infinity") of
+    max_attachment_size(config:get("couchdb", "max_attachment_size", ?GB)).
+
+max_attachment_size(MaxAttSizeConfig) ->
+    case MaxAttSizeConfig of
         "infinity" ->
             infinity;
+        MaxAttSize when is_list(MaxAttSize) ->
+            list_to_integer(MaxAttSize);
+        MaxAttSize when is_integer(MaxAttSize) ->
```
iilyak, line 747: You wouldn't need this clause if you change the define to be a string.
```
-define(GB, "1073741824").
```

src/couch/src/couch_att.erl
```diff
@@ -732,13 +735,19 @@ upgrade_encoding(true) -> gzip;
...
+            MaxAttSize;
         MaxAttSize ->
-            list_to_integer(MaxAttSize)
+            erlang:error({invalid_max_attachment_size, MaxAttSize})
```
iilyak, line 750: If we are throwing new type of error we might need to define it here https://github.com/apache/couchdb/blob/main/src/chttpd/src/chttpd.erl#L930 and maybe here https://github.com/apache/couchdb/blob/main/src/chttpd/src/chttpd.erl#L335

src/couch/src/couch_att.erl
bessbd: Considering https://github.com/apache/couchdb/pull/3347#discussion_r564547120 , I'd hope going for a type check has better performance than a conversion. What do you think?

src/couch/src/couch_att.erl
bessbd: Good catch! What do you think of 921aa502e instead though?

iilyak (approved):
+1 subject to ML vote pass

bessbd:
Thank you for the review, @iilyak !
I've just opened vote thread: https://lists.apache.org/thread.html/r815b25fe34996ab3c54e2fb9759de2026ddccffc8bb59966b1168063%40%3Cdev.couchdb.apache.org%3E

src/couch/src/couch_att.erl
```diff
@@ -178,6 +178,9 @@
 
 -type att() :: #att{} | attachment() | disk_att().
 
+% 1024 * 1024 * 1024
+-define(GB, 1073741824).
```
rnewson, line 182: better to say `-define(GB, (1024*1024*1024)).` I'm sure the compiler will do the right hing.

rel/overlay/etc/default.ini
```diff
@@ -47,7 +47,7 @@ changes_doc_ids_optimization_threshold = 100
 max_document_size = 8000000 ; bytes
 ;
 ; Maximum attachment size.
-; max_attachment_size = infinity
+; max_attachment_size = 1073741824 ; 1 gibibyte
```
rnewson, line 50: I'd go even lower tbh but anything is an improvement over unlimited.

rnewson (approved):
+1 but please squash this into one commit before or during merge.

src/couch/src/couch_att.erl
bessbd: Done in 83be3921b

rel/overlay/etc/default.ini
bessbd: I support decreasing this over time. I wouldn't want to change this number right now, because I believe that would change the scope of the ongoing vote on the ML.

bessbd:
Thank you for the review, @iilyak and @rnewson !
I'm about to merge this.

PR #3372, opened 2021-02-16
Clean up logs 3.x

iilyak:
## Overview

Port https://github.com/apache/couchdb/pull/3031 (Clean up logs) into 3.x branch.

This PR deals with cleaning up logs from user's data (potentially sensitive) in the cases when OTP processes crash. All commits except "Add format_status/2 callback in gen_server implementations" were cherry picked from https://github.com/apache/couchdb/pull/3031. The "Add format_status/2 callback in gen_server implementations" was extended to include applications relevant for 3.x.

## Testing recommendations

I tested it by running test suite and by comparing logs before and after. I did use following query to reduce noise.

```
find src/ -name couch.log | xargs -I '{}' cat '{}' | grep error | grep -v 'No Admin Account configured' | grep -v 'no_db_file' | grep -v 'no such file' | grep -v 'Request to create N=' | grep -v 'ddoc_cache_ev :: ' | grep -v '"error":"not_found"' | grep -v 'httpd 404' | grep -v boom | grep -v bam | grep -v forbidden | grep -v request_body_too | grep -v xxxxxxxxxxxxxxxxxxxxxxxxx | grep -v 'http 405' | grep -v method_not_allowed | grep -v db_not_found | grep -v 'source missing' | grep -v missing_source | grep -v 'list of tags' | grep -v 'Failed to set security object on' | grep -v 'fabric_rpc:group_info/2 exit:{killed,' | grep -v 'mem3_reshard db monitor' | grep -v unauthorized | grep -v too_large | grep -v 'httpd 413' | grep -v 'httpd 405' | grep -v not_acceptable | grep -v 'httpd [0-9]*' | grep -v 'unsupported encoding' | grep -v content_md5_mismatch | grep -v changes_req_failed | grep -v 'ignoring document with empty ID' | grep -v 'some_id <<"fail">>' | grep -v query_parse_error | grep -v 'Db was deleted' | grep -v kapow | grep -v bad_content_type | grep -v 'purge-test-stuff' | grep -v '{file,"src/couch_native_process.erl"},{line,258}' | grep -v 'file already exists' | grep -v 'Expression does not eval' | grep -v 'Transient job' | grep -v 'mem3_reshard_job cleaning' | grep -v 'mem3_reshard_job source' | grep -v 'Purge checkpoin' | grep -v 'missing_target' | grep -v 'target db missing'
```

I did run `make eunit` on vanilla 3.x branch and redirected the output of the above query into a file `3.x.vanilla.log`. Then I did remove all `couch.log` files and switched to `clean-up-logs-3.x` branch and repeated the process. The result is saved in `3.x.updated.log`. I did compare two files manually.

- https://github.com/apache/couchdb/pull/3031

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_node.erl
+++ b/src/chttpd/src/chttpd_node.erl
@@ -71,7 +71,9 @@ handle_node_req(#httpd{method='PUT', path_parts=[_, Node, <<"_config">>, Section
     Value = couch_util:trim(chttpd:json_body(Req)),
     Persist = chttpd:header_value(Req, "X-Couch-Persist") /= "false",
     OldValue = call_node(Node, config, get, [Section, Key, ""]),
-    case call_node(Node, config, set, [Section, Key, ?b2l(Value), Persist]) of
+    IsSensitive = Section == <<"admins">>,
+    Opts = #{persisit => Persist, sensitive => IsSensitive},
+    case call_node(Node, config, set, [Section, Key, ?b2l(Value), Opts]) of
         ok ->
             send_json(Req, 200, list_to_binary(OldValue));
         {error, Reason} ->
--- a/src/couch/src/couch_os_process.erl
+++ b/src/couch/src/couch_os_process.erl
@@ -14,7 +14,7 @@
 -behaviour(gen_server).
 -vsn(1).
 
--export([start_link/1, start_link/2, start_link/3, stop/1]).
+-export([start_link/1, start_link/2, start_link/3, stop/1, format_status/2]).
 -export([set_timeout/2, prompt/2, killer/1]).
 -export([send/2, writeline/2, readline/1, writejson/2, readjson/1]).
 -export([init/1, terminate/2, handle_call/3, handle_cast/2, handle_info/2, code_change/3]).
@@ -255,6 +255,24 @@ code_change(_, {os_proc, Cmd, Port, W, R, Timeout} , _) ->
     {ok, State};
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
+
+format_status(_Opt, [_PDict, #os_proc{} = State]) ->
+    % command field can potentially contain secrets
+    #os_proc{
+        port = Port,
+        writer = Writer,
+        reader = Reader,
+        timeout = TimeOut,
+        idle = Idle
+    } = State,
+    MapState = #{
+        port => couch_term:format_port(Port),
+        writer => couch_term:format_fun(Writer),
+        reader => couch_term:format_fun(Reader),
+        timeout => TimeOut,
+        idle => Idle
+    },
+    [{data, [{"State", MapState}]}].
 
 killer(KillCmd) ->
     receive _ ->
--- a/src/couch/src/couch_term.erl
+++ b/src/couch/src/couch_term.erl
@@ -1,0 +1,95 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(couch_term).
+
+-include("couch_db.hrl").
+-include("couch_db_int.hrl").
+
+-export([
+    format_record/1,
+    format_term/1,
+    format_fun/1,
+    format_pid/1,
+    format_port/1,
+    format_ref/1,
+    record_to_map/1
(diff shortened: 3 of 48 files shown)


src/chttpd/src/chttpd_node.erl
```diff
@@ -71,7 +71,9 @@ handle_node_req(#httpd{method='PUT', path_parts=[_, Node, <<"_config">>, Section
     Value = couch_util:trim(chttpd:json_body(Req)),
     Persist = chttpd:header_value(Req, "X-Couch-Persist") /= "false",
     OldValue = call_node(Node, config, get, [Section, Key, ""]),
-    case call_node(Node, config, set, [Section, Key, ?b2l(Value), Persist]) of
+    IsSensitive = Section == <<"admins">>,
+    Opts = #{persisit => Persist, sensitive => IsSensitive},
```
rnewson, line 75: typo

src/couch/src/couch_term.erl
```diff
@@ -0,0 +1,95 @@
...
+format_pid(Pid) when is_pid(Pid) ->
+    pid_to_list(Pid);
+
+format_pid(Term) ->
+    format_term(Term).
+
+format_port(Port) when is_port(Port) ->
+    erlang:port_to_list(Port);
+
+format_port(Term) ->
```
rnewson, line 82: why does format_port take non-port arguments? ditto for the other functions structured this way.

src/chttpd/src/chttpd_node.erl
iilyak: Good catch!!

rnewson (commented):
I'm not sure this is appropriate for couchdb. I understand the motivation but it does not apply to all users/admins of couchdb and they should be allowed to keep the current behaviour where stack traces and process states are unaltered. This should be configurable.

I left some specific comments but this also seems like a very large diff for a conceptually much smaller change.

src/couch/src/couch_term.erl
iilyak: The main purpose for these functions is to format record fields. In some cases the value of a field can be `nil` or `undefined`. The default clause was added to simplify the caller side. Otherwise I would have to add a case for each filed I try to convert to each `format_status/2` function which uses this facility.

src/couch/src/couch_os_process.erl
```diff
@@ -256,6 +256,24 @@ code_change(_, {os_proc, Cmd, Port, W, R, Timeout} , _) ->
...
+        idle = Idle
+    } = State,
+    MapState = #{
+        port => couch_term:format_port(Port),
+        writer => couch_term:format_fun(Writer),
+        reader => couch_term:format_fun(Reader),
+        timeout => TimeOut,
+        idle => Idle
+    },
+    [{data, [{"State", MapState}]}].
```
rnewson, line 275: it is odd that we convert everything to maps for format_status even though that is not the format it takes really.

iilyak:
>  I understand the motivation but it does not apply to all users/admins of couchdb and they should be allowed to keep the current behaviour where stack traces and process states are unaltered. 

I also tried to remove big terms from state. Do we want it to be configurable as well?

iilyak:
> I'm not sure this is appropriate for couchdb. I understand the motivation but it does not apply to all users/admins of couchdb and they should be allowed to keep the current behaviour where stack traces and process states are unaltered. This should be configurable.

What do you think about the following approach. The format_status/2 would return a tuple `{Status :: term(), Options :: #{senstive => boolean(), records => %{name :: atom() => Fields :: list(atom())}}}`?

```
format_status(_Opt, [_PDict, State]) ->
    {[{data, [{"State", State}}]}], #{sensitive => false, records => #{
        state => record_info(fields, state),
        httpd => record_info(fields, httpd)
    }}}.
```

Then in couch_log application we would call user configured sanitizer module.

rnewson:
I'd like to hear from non-IBM CouchDB committers that might have a different view on our scrubbing of stacktraces and process states and logs first. If it's generally desired then I'd agree it doesn't need to be configurable. My concern, with both my IBM hat and CouchDB hat on, is that we are degrading the quality of our diagnostics (albeit with good intention) and I think we need to proceed with a broader consensus.

I don't think format_status should return an erlang term that is completely different from the state of the process, it should look the same but with some items removed or redacted.

in summary, the goal is sound, we should ask if a solution is desired widely, whether it should be configurable and how closely the redacted output should match the output it replaces.

@janl @wohali @dch

src/couch/src/couch_term.erl
iilyak: Another option I came up with is this (the `format_port/2` can be expressed using `case` as well).

```
format_port(Term) ->
    format_port(Term, nil).
    
format_port(Term, _Default) when is_port(Term) ->
    erlang:port_to_list(Port);
    
format_port(_Term, Default) ->
    Default.    
```

wohali:
If this is configurable and not enabled by default, I have no objection.

iilyak:
Closing in favor of https://github.com/apache/couchdb/pull/3380

PR #3373, opened 2021-02-17
Read and validate JSON payload on POST to _changes

### LINKED ISSUE #3087: POST to _changes timeout when request.body is included
## Description

When POSTing to the _changes endpoint with a request.body, and not using one of the builtin filters (doc_ids for example) the connection always times out.
@rnewson has looked into this, and says "the code starts the response, and may even send it, but the web server itself (mochiweb) won't proceed because it still needs to read the request body first. we should send a 400 Bad Request if you send a body when it's unexpected"

## Steps to Reproduce
 curl 'foo:bar@localhost:15984/db1/_design/foo'
{"_id":"_design/foo","_rev":"1-1cfd4ef84175f09eb3d41bda94062244","filters":{"bar":"function() { return true; }"}}
​
curl 'foo:bar@localhost:15984/db1/_changes?style=all_docs&since=0&limit=100&filter=foo/bar' -XPOST -Hcontent-type:application/json --data-raw '{"doc_ids": ["foobar"]}'

## Expected Behaviour
Return a 400 Bad Request error in preference to waiting for a timeout if a req.body is provided but unexpected (see enhancement request https://github.com/apache/couchdb/issues/3086 which may solve this bug as a side-effect)

## Your Environment
Couchdb 3.1

eiri:
## Overview

During `POST` on `/{db}/_changes` with a custom filter, nodes attempt to [read the request payload](https://github.com/apache/couchdb/blob/3.x/src/couch/src/couch_changes.erl#L241), unless the request body already stored on a request record. When this is happening on a node foreign to the original http request, reader process quits with `exit:normal` and as a result the whole fabric request hangs until a timeout.

This patch switches filter reader to `chttpd_external:json_req_obj/2` for more descriptive exit message, adds reading of the payload on the request handler and validates that payload, when present, is a proper JSON object.

## Testing recommendations

Related tests been added to the elixir tests

`make elixir tests=test/elixir/test/changes_test.exs`

Closes #3087 

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -85,7 +85,13 @@ handle_request(#httpd{path_parts=[DbName|RestParts],method=Method}=Req)->
 
 handle_changes_req(#httpd{method='POST'}=Req, Db) ->
     chttpd:validate_ctype(Req, "application/json"),
-    handle_changes_req1(Req, Db);
+    case chttpd:body_length(Req) of
+        0 ->
+            handle_changes_req1(Req, Db);
+        _ ->
+            {JsonProps} = chttpd:json_body_obj(Req),
+            handle_changes_req1(Req#httpd{req_body = {JsonProps}}, Db)
+    end;
 handle_changes_req(#httpd{method='GET'}=Req, Db) ->
     handle_changes_req1(Req, Db);
 handle_changes_req(#httpd{path_parts=[_,<<"_changes">>]}=Req, _Db) ->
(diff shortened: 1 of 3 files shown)


nickva (approved):
Good find!

+1

I noticed `couch_httpd_external` was now used in only one other place https://github.com/apache/couchdb/blob/3.x/src/couch/src/couch_query_servers.erl#L494 wonder if that can be made to call into `chttpd_external` too and then we'd remove the `couch_httpd_external` altogether?

rnewson (approved):

eiri:
@nickva good idea! Done in #3374

PR #3380, opened 2021-02-23
Clean up logs 3.x.v2

iilyak:
## Overview

This PR started as a port of https://github.com/apache/couchdb/pull/3031 (Clean up logs) into 3.x branch. The main goal of the PR is to remove sensitive data from the logs.
However as it was pointed out by reviewers it might be not the right approach to convert shape of gen_server state for a CouchDB version which is already widely deployed. Another problem with the scope of the PR is lack of definition of sensitive data. There are multiple points of view on the sensitive data:

- user credentials
- user documents
- design documents 

This means that there is no generic solution which would fit all. In order to solve this problem for the purposes of this PR the sensitive data are defined as "user credentials". 

Before going further down with description of the solution presented in the PR. I wanted to mention the requirements as I understood them:

1) `sys:get_status/1` should return state as is (wrapped in `[{data, [{"State", Term}]}]`) see http://erlang.org/doc/man/gen_server.html#Module:format_status-2
2) the `couch_log_formatter` should be able to call a configurable sanitizer module which would deal with changing the shape of log entries and removing user's data from terms
3) "user credentials" should be removed from state for both `sys:get_status/1` use case and termination of a process.

In order to achieve 1) we need to handle `:format_status(normal, ...)` and `:format_status(terminate, ...)` differently.

In order to achieve 2) we need to pass the callback module name so it is available in `couch_log_formatter`. We also pass the process dictionary so sanitizer module could get some meta information about the process (for example tracing span_id). The process dictionary is not logged by default. However sanitizer module can include elements from process dictionary in a term it returns. 

The approach to `format_status/2` is to return term as is when first argument is `normal` in order for `sys:get_status` to return original term. In both cases we remove "sensitive data" before returning the term from `format_status/2`. The `terminate` clause of a `format_status/2` returns `{Term, Ctx :: map()}`. The `Ctx` includes `module` and `dictionary` keys for now. We need this information so sanitizer could distinguish messages coming from different gen_servers.

## Testing recommendations

`make eunit`

- [ ] https://github.com/apache/couchdb/pull/3372 - previous attempt
- [ ] https://github.com/apache/couchdb/pull/3031 - implementation on main

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_node.erl
+++ b/src/chttpd/src/chttpd_node.erl
@@ -71,7 +71,9 @@ handle_node_req(#httpd{method='PUT', path_parts=[_, Node, <<"_config">>, Section
     Value = couch_util:trim(chttpd:json_body(Req)),
     Persist = chttpd:header_value(Req, "X-Couch-Persist") /= "false",
     OldValue = call_node(Node, config, get, [Section, Key, ""]),
-    case call_node(Node, config, set, [Section, Key, ?b2l(Value), Persist]) of
+    IsSensitive = Section == <<"admins">>,
+    Opts = #{persist => Persist, sensitive => IsSensitive},
+    case call_node(Node, config, set, [Section, Key, ?b2l(Value), Opts]) of
         ok ->
             send_json(Req, 200, list_to_binary(OldValue));
         {error, Reason} ->
--- a/src/setup/src/setup_httpd.erl
+++ b/src/setup/src/setup_httpd.erl
@@ -19,7 +19,7 @@ handle_setup_req(#httpd{method='POST'}=Req) ->
     ok = chttpd:verify_is_server_admin(Req),
     couch_httpd:validate_ctype(Req, "application/json"),
     Setup = get_body(Req),
-    couch_log:notice("Setup: ~p~n", [Setup]),
+    couch_log:notice("Setup: ~p~n", [remove_sensitive(Setup)]),
     Action = binary_to_list(couch_util:get_value(<<"action">>, Setup, <<"missing">>)),
     case handle_action(Action, Setup) of
     ok ->
@@ -91,7 +91,7 @@ handle_action("enable_cluster", Setup) ->
 
 
 handle_action("finish_cluster", Setup) ->
-    couch_log:notice("finish_cluster: ~p~n", [Setup]),
+    couch_log:notice("finish_cluster: ~p~n", [remove_sensitive(Setup)]),
 
     Options = get_options([
         {ensure_dbs_exist, <<"ensure_dbs_exist">>}
@@ -105,7 +105,7 @@ handle_action("finish_cluster", Setup) ->
     end;
 
 handle_action("enable_single_node", Setup) ->
-    couch_log:notice("enable_single_node: ~p~n", [Setup]),
+    couch_log:notice("enable_single_node: ~p~n", [remove_sensitive(Setup)]),
 
     Options = get_options([
         {ensure_dbs_exist, <<"ensure_dbs_exist">>},
@@ -125,7 +125,7 @@ handle_action("enable_single_node", Setup) ->
 
 
 handle_action("add_node", Setup) ->
-    couch_log:notice("add_node: ~p~n", [Setup]),
+    couch_log:notice("add_node: ~p~n", [remove_sensitive(Setup)]),
 
     Options = get_options([
         {username, <<"username">>},
@@ -147,10 +147,10 @@ handle_action("add_node", Setup) ->
     end;
 
 handle_action("remove_node", Setup) ->
-    couch_log:notice("remove_node: ~p~n", [Setup]);
+    couch_log:notice("remove_node: ~p~n", [remove_sensitive(Setup)]);
 
 handle_action("receive_cookie", Setup) ->
-    couch_log:notice("receive_cookie: ~p~n", [Setup]),
+    couch_log:notice("receive_cookie: ~p~n", [remove_sensitive(Setup)]),
     Options = get_options([
        {cookie, <<"cookie">>}
     ], Setup),
@@ -173,3 +173,6 @@ get_body(Req) ->
         couch_log:notice("Body Fail: ~p~n", [Else]),
         couch_httpd:send_error(Req, 400, <<"bad_request">>, <<"Missing JSON body'">>)
     end.
+
+remove_sensitive(KVList) ->
+    lists:keyreplace(<<"password">>, 1, KVList, {<<"password">>, <<"****">>}).
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -144,7 +144,7 @@ SubDirs = [
 
 DepDescs = [
 %% Independent Apps
-{config,           "config",           {tag, "2.1.7"}},
+{config,           "config",           {tag, "2.1.8"}},
 {b64url,           "b64url",           {tag, "1.0.2"}},
 {ets_lru,          "ets-lru",          {tag, "1.1.0"}},
 {khash,            "khash",            {tag, "1.1.0"}},
--- a/src/couch_log/src/couch_log_config.erl
+++ b/src/couch_log/src/couch_log_config.erl
@@ -49,7 +49,8 @@ entries() ->
     [
         {level, "level", "info"},
         {level_int, "level", "info"},
-        {max_message_size, "max_message_size", "16000"}
+        {max_message_size, "max_message_size", "16000"},
+        {strip_last_msg, "strip_last_msg", "true"}
      ].
 
 
@@ -97,4 +98,10 @@ transform(max_message_size, SizeStr) ->
         Size -> Size
     catch _:_ ->
         16000
-    end.
+    end;
+
+transform(strip_last_msg, "false") ->
+    false;
+
+transform(strip_last_msg, _) ->
+    true.
--- a/src/couch_log/src/couch_log_config_dyn.erl
+++ b/src/couch_log/src/couch_log_config_dyn.erl
@@ -25,4 +25,5 @@
 
 get(level) -> info;
 get(level_int) -> 2;
-get(max_message_size) -> 16000.
+get(max_message_size) -> 16000;
+get(strip_last_msg) -> true.
--- a/src/couch_log/src/couch_log_formatter.erl
+++ b/src/couch_log/src/couch_log_formatter.erl
@@ -68,7 +68,13 @@ format(Event) ->
 
 do_format({error, _GL, {Pid, "** Generic server " ++ _, Args}}) ->
     %% gen_server terminate
-    [Name, LastMsg, State, Reason | Extra] = Args,
+    [Name, LastMsg0, State, Reason | Extra] = Args,
+    LastMsg = case couch_log_config:get(strip_last_msg) of
+        true ->
+            redacted;
+        false ->
+            LastMsg0
+    end,
     MsgFmt = "gen_server ~w terminated with reason: ~s~n" ++
                 "  last msg: ~p~n     state: ~p~n    extra: ~p",
     MsgArgs = [Name, format_reason(Reason), LastMsg, State, Extra],
@@ -76,7 +82,13 @@ do_format({error, _GL, {Pid, "** Generic server " ++ _, Args}}) ->
 
 do_format({error, _GL, {Pid, "** State machine " ++ _, Args}}) ->
     %% gen_fsm terminate
-    [Name, LastMsg, StateName, State, Reason | Extra] = Args,
+    [Name, LastMsg0, StateName, State, Reason | Extra] = Args,
+    LastMsg = case couch_log_config:get(strip_last_msg) of
+        true ->
+            redacted;
+        false ->
+            LastMsg0
+    end,
     MsgFmt = "gen_fsm ~w in state ~w terminated with reason: ~s~n" ++
                 " last msg: ~p~n     state: ~p~n    extra: ~p",
     MsgArgs = [Name, StateName, format_reason(Reason), LastMsg, State, Extra],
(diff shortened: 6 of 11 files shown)


src/chttpd/src/chttpd_node.erl
```diff
@@ -71,7 +71,9 @@ handle_node_req(#httpd{method='PUT', path_parts=[_, Node, <<"_config">>, Section
     Value = couch_util:trim(chttpd:json_body(Req)),
     Persist = chttpd:header_value(Req, "X-Couch-Persist") /= "false",
     OldValue = call_node(Node, config, get, [Section, Key, ""]),
-    case call_node(Node, config, set, [Section, Key, ?b2l(Value), Persist]) of
+    IsSensitive = Section == <<"admins">>,
+    Opts = #{persisit => Persist, sensitive => IsSensitive},
```
rnewson, line 75: typo

src/setup/src/setup_httpd.erl
```diff
@@ -173,3 +173,8 @@ get_body(Req) ->
         couch_log:notice("Body Fail: ~p~n", [Else]),
         couch_httpd:send_error(Req, 400, <<"bad_request">>, <<"Missing JSON body'">>)
     end.
+
+remove_sensitive(KVList0) ->
+    KVList1 = lists:keyreplace(<<"username">>, 1, KVList0, {<<"username">>, <<"****">>}),
```
rnewson, line 178: username isn't sensitive (or shouldn't be!)

src/setup/src/setup_httpd.erl
```diff
@@ -173,3 +173,8 @@ get_body(Req) ->
...
+    KVList2 = lists:keyreplace(<<"password">>, 1, KVList1, {<<"password">>, <<"****">>}),
+    KVList2.
```
rnewson, line 180: a style point (and I guess you're anticipating other fields in future) but I would list return the lists:keyreplace result

src/couch/src/couch_bt_engine.erl
```diff
@@ -192,6 +193,44 @@ handle_db_updater_info({'DOWN', Ref, _, _, _}, #st{fd_monitor=Ref} = St) ->
     {stop, normal, St#st{fd=undefined, fd_monitor=closed}}.
 
 
+format_status(normal, [_PDict, #st{} = St]) ->
+    [{data, [{"State", St}]}];
```
rnewson, line 197: I do like that we don't interfere with a sys:get_status call from a remsh (etc).

src/couch/src/couch_bt_engine.erl
```diff
@@ -192,6 +193,44 @@ handle_db_updater_info({'DOWN', Ref, _, _, _}, #st{fd_monitor=Ref} = St) ->
...
+
+format_status(terminate, [_PDict, #st{} = St]) ->
+    [{data, [{"State", reduce_record(St)}]}].
+
+
+reduce_record(#st{} = St) ->
```
rnewson, line 203: I think `redact` is a better word as we have a quite different meaning for `reduce` in couchdb.

src/chttpd/src/chttpd_node.erl
bessbd, line 75: Nit
```suggestion
    Opts = #{persist => Persist, sensitive => IsSensitive},
```

src/couch/src/couch_bt_engine.erl
```diff
@@ -192,6 +193,44 @@ handle_db_updater_info({'DOWN', Ref, _, _, _}, #st{fd_monitor=Ref} = St) ->
...
+        local_tree = reduce_record(LocalTree, undefined),
+        purge_tree = reduce_record(PurgeTree, undefined),
+        purge_seq_tree = reduce_record(PurgeSeqTree, undefined)
+    };
+
+reduce_record(Header, _) when element(1, Header) =:= db_header ->
+    couch_bt_engine_header:reduce_record(Header);
+
+reduce_record(Tree, _) when element(1, Tree) =:= btree ->
+    {tree_size, couch_btree:size(Tree)};
```
rnewson, line 228: this redacts every part of the btree's and replaces it with just the tree size (which isn't part of the state). I'm not seeing anything sensitive in the btree attributes to warrant this.

src/couch/src/couch_bt_engine.erl
```diff
@@ -192,6 +193,44 @@ handle_db_updater_info({'DOWN', Ref, _, _, _}, #st{fd_monitor=Ref} = St) ->
...
+    St#st{
+        header = reduce_record(Header, undefined),
+        id_tree = reduce_record(IdTree, undefined),
+        seq_tree = reduce_record(SeqTree, undefined),
+        local_tree = reduce_record(LocalTree, undefined),
+        purge_tree = reduce_record(PurgeTree, undefined),
+        purge_seq_tree = reduce_record(PurgeSeqTree, undefined)
+    };
+
+reduce_record(Header, _) when element(1, Header) =:= db_header ->
```
rnewson, line 224: should be `when is_record(Header, db_header)`.

src/couch/src/couch_bt_engine.erl
```diff
@@ -192,6 +193,44 @@ handle_db_updater_info({'DOWN', Ref, _, _, _}, #st{fd_monitor=Ref} = St) ->
...
+    couch_bt_engine_header:reduce_record(Header);
+
+reduce_record(Tree, _) when element(1, Tree) =:= btree ->
```
rnewson, line 227: should be `when is_record(Tree, btree)`.

src/couch/src/couch_bt_engine_header.erl
```diff
@@ -327,6 +328,27 @@ latest(N) when is_integer(N), N < ?LATEST_DISK_VERSION ->
...
+        seq_tree_state = {tree_size, tree_state_size(SeqTreeState)},
+        local_tree_state = {tree_size, tree_state_size(LocalTreeState)}
+    }.
+
+
+tree_state_size(TreeState) ->
+    %% We just need a correctly shaped btree record,
+    %% so we can pass Fd = nil
+    {ok, Tree} = couch_btree:open(TreeState, nil),
+    couch_btree:size(Tree).
```
rnewson, line 349: same comment as above

src/couch/src/couch_multidb_changes.erl
```diff
@@ -174,6 +175,18 @@ code_change(_OldVsn, State, _Extra) ->
...
 
+format_status(normal, [_PDict, State]) ->
+    [{data, [{"State", State}]}];
+
+format_status(terminate, [_PDict, State]) ->
+    #state{
+        pids=Pids
+    } = State,
+    Scrubbed = State#state{
+        pids={length, length(Pids)}
```
rnewson, line 186: pids aren't sensitive.

src/couch/src/couch_native_process.erl
```diff
@@ -125,6 +125,20 @@ handle_info({'EXIT',_,Reason}, State) ->
...
+format_status(normal, [_PDict, State]) ->
+    [{data, [{"State", State}]}];
+
+format_status(terminate, [_PDict, State]) ->
+    #evstate{
+        ddocs = DDocs,
+        funs = Funs
+    } = State,
+    Scrubbed = State#evstate{
+        ddocs = {dict_size, dict:size(DDocs)},
```
rnewson, line 137: arguably not sensitive either

rnewson (commented):
I closely reviewed the first few cases and skimmed the rest as they seemed the same idea but adapted for the gen_server in question.

I like the structure of the code but I am concerned about the choices made. This is a great deal of redaction and most of it appears not to be related to credentials or sensitive information (reducing the btree record to just the size of the btree, for example). 

our logs are an important source of debugging information, particularly for users who aren't comfortable with digging around in a remsh (or are in an environment where they are prohibited from doing so).

I don't think redaction beyond obvious things like passwords can be the default behaviour of CouchDB. I'm less sure if there should be an _option_ for the kind of deep "cleanup" here. With my couchdb support hat on, would I be happy to diagnose a problem from a CouchDB log this heavily modified?

iilyak:
> I don't think redaction beyond obvious things like passwords can be the default behaviour of CouchDB. I'm less sure if there should be an option for the kind of deep "cleanup" here. With my couchdb support hat on, would I be happy to diagnose a problem from a CouchDB log this heavily modified?

Personally I hit the same problem over and over. I have a bug and go to check logs. The log contains huge term which we truncate and all important information about other fields of a record omitted.

rnewson:
Ah, sure, I get that point, thank you.

src/couch/src/couch_bt_engine.erl
iilyak: The `is_record/2` works only for records in scope of a module (defined locally or included). The `db_header` is a private record defined in `couch_bt_engine_header.erl`. 

```
couchdb/src/couch/src/couch_bt_engine.erl:224: record db_header undefined
Compiling couchdb/src/couch/src/couch_bt_engine.erl failed:
ERROR: compile failed while processing couchdb/src/couch: rebar_abort
```

rnewson: hm, true.

src/couch/src/couch_bt_engine.erl
iilyak: The `format_status/`2 was added to prevent huge btree structures hitting logs. Since the reduction of terms is not longer a goal I am thinking about removing this function.

janl:
Hey hey, just casually following this, but the diff is a bit complex. I think this might have a better chance if we focus on redacting sensitive data first, and tackle garbage cleanup later.

iilyak:
> I think this might have a better chance if we focus on redacting sensitive data first, and tackle garbage cleanup later.

Agree. I'll update the PR to decouple garbage cleanup out of this PR.

rnewson:
I don't understand these;

```
format_status(normal, [_PDict, State]) ->
    [{data, [{"State", State}]}];

format_status(terminate, [PDict, State]) ->
    {State, #{module => ?MODULE, dictionary => PDict}}.
```

Why so many conversions to map?

rnewson:
we also can't use the ableist slur "sanitize" but that's a mechanical change ("scrub" perhaps?). will ponder if the boilerplating can be reduced/removed (my concern is not just the duplication but also the deviation from OTP and the need to remember to do this for every new gen_server).

iilyak:
The alternative to `map()` for the `Ctx` is a KV list. However map is superior because it can be matched in the function clause. The KV list preserve the order of elements so cannot be matched against. Which simplify dispatch in couch_log_formatter and sanitizer module.

```
format_status(terminate, [PDict, State]) ->
    {State, #{module => ?MODULE, dictionary => PDict}}.
    
sanitizer:sanitize_state(State, #{module := couch_server}) ->
    ....;
sanitizer:sanitize_state(State, #{module := smoosh_server}) ->
    ....;
```

compared to
```
format_status(terminate, [PDict, State]) ->
    {State, [{module, ?MODULE}, {dictionary, PDict}]}.
    
sanitizer:sanitize_state(State, Ctx) ->
    Module = case lists:keyfind(module, 1, Ctx) -> of
        {module, M} -> M;
        false -> undefined
     end,
    PDict = case lists:keyfind(dictionary, 1, Ctx) -> of
        {dictionary, D} -> D;
        false -> []
     end,
     sanitize_state(State, Module, PDict).
     
sanitizer:sanitize_state(State, smoosh_server, PDict) ->
    ....;

sanitizer:sanitize_state(State, couch_server, PDict) ->
    ....;
```

rnewson:
now you've explained I'm fine with the "context" item being a map. What's really awkward is having to alter every gen_server in an identical way in order to pass it.

iilyak:
> I am very uncomfortable to see that pattern cut and paste so many times, it strongly implies a bad design. I think I see how you got here though.

The alternative is a common module somewhere.

```
format_status(terminate, [PDict, State]) ->
      Scrubed = ...
      some_module:format_status(Scrubbed, PDict).
```

However the gen_servers we update are in different applications. We don't have an application which is included in all the others. I could create a new app which would hold the module. This app would be only for utility functions and couldn't depend on other apps to avoid circular dependencies.

rnewson:
could this be done with a parse transform? When we used lager it added a parse transform (https://github.com/apache/couchdb-lager/blob/master/src/lager_transform.erl) which captures module name among other things.

iilyak:
In fact my first take on this problem was a parse transform. The problem with this approach is use of `rebar`. Rebar has bugs the parse transform configured in main application is not passed to included application for umbrella apps. We would have to update `rebar.config` for each app we include.

iilyak:
I think I would need to create a new repo for the parse_transform code. Because it need to be used as dependency for other applications which are not part of the monolithic repository.

I am trying to choose between
1. a repository specifically for this transform
2. a repository which we would use to hold all transforms we ever implement
3. a repository with utility functions which would also have parse transforms in there

janl:
Good cleanup @iilyak! This is a lot easer to comprehend.

The one place where this could do with a comment about what happens is https://github.com/apache/couchdb/pull/3380/files#diff-33000329e1e5eb2554d72f4675fb32c2bf3e0599c981f044683467cb25efc02aR378 which seems very meta and it is not clear to me what is happening there, or if we still need it, now that ddocs are out of scope for redacting.

I agree with @rnewson on the duplication and agree with methods to mitigate. A separate repo sounds good. I would go with option 1. until we have more transforms, at which point we can move things around.

iilyak:
I rolled back the controversial changes and included only the changes needed to prevent leaking of passwords. I think we can have separate PR if we would want to prevent leaking of users' data or minimize state terms.

rnewson (approved):
@iilyak this is very tidy, thank you for persevering. +1 on merging though do please squash beforehand.

PR #3384, opened 2021-02-26
use collate in lookup

rnewson:
Replace the `=:=` comparison with a collate `eq` call.

### DIFF
--- a/src/ebtree/src/ebtree.erl
+++ b/src/ebtree/src/ebtree.erl
@@ -132,13 +132,13 @@ max() ->
     {Key :: term(), Value :: term()} | false.
 lookup(Db, #tree{} = Tree, Key) ->
     Fun = fun
-        ({visit, K, V}, _Acc) when K =:= Key ->
-            {stop, {K, V}};
-        ({visit, K, _V}, Acc) ->
-            case collate(Tree, K, Key, [gt]) of
-                true ->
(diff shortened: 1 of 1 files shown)


src/ebtree/src/ebtree.erl
```diff
@@ -132,13 +132,13 @@ max() ->
     {Key :: term(), Value :: term()} | false.
 lookup(Db, #tree{} = Tree, Key) ->
     Fun = fun
-        ({visit, K, V}, _Acc) when K =:= Key ->
-            {stop, {K, V}};
-        ({visit, K, _V}, Acc) ->
-            case collate(Tree, K, Key, [gt]) of
-                true ->
+        ({visit, K, V}, Acc) ->
+            case {collate(Tree, K, Key, [eq]), collate(Tree, K, Key, [gt])} of
```
davisp, line 136: Should we bother delaying the second `[gt]` comparison until it fails the `[eq]` comparison. Its a micro optimization to be sure.

rnewson: I'm following the pattern for traverse, but your point is valid. I do think it's a micro-optimization and would rather fix the bug separately.

davisp: Cool beans.

davisp (approved):
+1

PR #3386, opened 2021-02-27
Optimize lookup/3

rnewson:
A tidier version of https://github.com/apache/couchdb/pull/3384 that
saves an unnecessary call to collate.


nickva (approved):
+1

The `{traverse, ...}` clause below could probably be similarly optimized to do the collate `gt` first then based on the result do `lt, eq`.

PR #3404, opened 2021-03-05
Fix and re-enable ChangesAsyncTest

bessbd:
## Overview
The `ChangesAsyncTest` tests were broken and disabled. This PR fixes and re-enables them.

## Testing recommendations

`mix test --trace test/elixir/test/changes_async_test.exs`

## Checklist

### DIFF
--- a/src/chttpd/src/chttpd_db.erl
+++ b/src/chttpd/src/chttpd_db.erl
@@ -222,9 +222,13 @@ changes_callback(waiting_for_updates, Acc) ->
         mochi = Resp1,
         chunks_sent = ChunksSent + 1
     }};
-changes_callback({timeout, _ResponseType}, Acc) ->
+changes_callback({timeout, ResponseType}, Acc) ->
     #cacc{mochi = Resp, chunks_sent = ChunksSent} = Acc,
-    {ok, Resp1} = chttpd:send_delayed_chunk(Resp, "\n"),
+    Chunk = case ResponseType of
+        "eventsource" -> "event: heartbeat\ndata: \n\n";
+        _ -> "\n"
+    end,
+    {ok, Resp1} = chttpd:send_delayed_chunk(Resp, Chunk),
     {ok, Acc#cacc{mochi = Resp1, chunks_sent = ChunksSent + 1}};
 changes_callback({error, Reason}, #cacc{mochi = #httpd{}} = Acc) ->
     #cacc{mochi = Req} = Acc,
--- a/test/elixir/lib/couch.ex
+++ b/test/elixir/lib/couch.ex
@@ -80,6 +80,10 @@ defmodule Couch do
   @request_timeout 60_000
   @inactivity_timeout 55_000
 
+  def base_url() do
+    System.get_env("EX_COUCH_URL") || "http://127.0.0.1:15984"
+  end
+
   def process_url("http://" <> _ = url) do
     url
   end
@@ -110,7 +114,7 @@ defmodule Couch do
   end
 
   def process_options(options) do
-    base_url = System.get_env("EX_COUCH_URL") || "http://127.0.0.1:15984"
+    base_url = base_url()
     options = Keyword.put_new(options, :base_url, base_url)
 
     options
--- a/test/elixir/test/changes_async_test.exs
+++ b/test/elixir/test/changes_async_test.exs
@@ -34,12 +34,11 @@ defmodule ChangesAsyncTest do
       )
 
     changes = process_response(req_id.id, &parse_chunk/1)
-    {changes_length, last_seq_prefix} = parse_changes_response(changes)
-    assert changes_length == 1, "db should not be empty"
-    assert last_seq_prefix == "1-", "seq must start with 1-"
-
+    assert length(changes["results"]) == 1, "db should not be empty"
     last_seq = changes["last_seq"]
-    {:ok, worker_pid} = HTTPotion.spawn_link_worker_process(Couch.process_url(""))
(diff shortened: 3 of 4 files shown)


src/chttpd/src/chttpd_db.erl
```diff
@@ -222,9 +222,13 @@ changes_callback(waiting_for_updates, Acc) ->
         mochi = Resp1,
         chunks_sent = ChunksSent + 1
     }};
-changes_callback({timeout, _ResponseType}, Acc) ->
+changes_callback({timeout, ResponseType}, Acc) ->
     #cacc{mochi = Resp, chunks_sent = ChunksSent} = Acc,
-    {ok, Resp1} = chttpd:send_delayed_chunk(Resp, "\n"),
+    Chunk = case ResponseType of
+        "eventsource" -> "event: heartbeat\ndata: \n\n";
```
jjrodrig, line 228: Why is this required here? Previously, this seems to be managed at line [171](https://github.com/apache/couchdb/blob/db6a7822b9c5bc501bb20acb6739a1c0cf7f929c/src/chttpd/src/chttpd_db.erl#L171). We are addressing the root issue here or we are duplicating some logic?

test/elixir/lib/couch.ex
```diff
@@ -80,6 +80,10 @@ defmodule Couch do
   @request_timeout 60_000
   @inactivity_timeout 55_000
 
+  def base_url() do
```
jjrodrig, line 83: Nice, I should have done this during the porting.

test/elixir/test/changes_async_test.exs
```diff
@@ -34,12 +34,11 @@ defmodule ChangesAsyncTest do
       )
 
     changes = process_response(req_id.id, &parse_chunk/1)
-    {changes_length, last_seq_prefix} = parse_changes_response(changes)
-    assert changes_length == 1, "db should not be empty"
-    assert last_seq_prefix == "1-", "seq must start with 1-"
```
jjrodrig, line 39: It is the first time I've seen a 4.x change seq. It is completely different from the previous format. I see that we can not check anything about the seq.

test/elixir/test/changes_async_test.exs
```diff
@@ -51,13 +50,13 @@ defmodule ChangesAsyncTest do
...
+    changes =
       req_id.id
       |> process_response(&parse_chunk/1)
-      |> parse_changes_response()
 
-    assert changes_length == 1, "should return one change"
-    assert last_seq_prefix == "2-", "seq must start with 2-"
+    assert length(changes["results"]) == 1, "should return one change"
+    next_last_seq = changes["last_seq"]
+    assert next_last_seq > last_seq
```
jjrodrig, line 59: Now seqs are comparable?

jjrodrig (commented):
Thanks @bessbd. The elixir test looks ok.

The main worry is about not addressing the root issue with the proposed fix. I'm not an expert on this but this seems to be solved previously with another callback. Are we duplicating code here?

src/chttpd/src/chttpd_db.erl
bessbd: The thing is, the server was sending plain `\n`s as heartbeat (which is not what is expected for eventsource). You can try it yourself with and without the change. A `GET` `/<db>/_changes?feed=eventsource&heartbeat=1000` should help reproducing the issue.
We may want to remove the lines you've mentioned above, but I don't know the codebase well enough to tell if that would break anything.
I'm hoping I'm not duplicating logic here, but I'm not 100% the previous related code isn't dead code.
IIRC https://github.com/apache/couchdb/pull/3404/files#diff-0ad4fb5d609221dac3b3f5faa48e061bd84f0745ecb0b6953b820d117b55cf5cR106 was the test case that was broken for this.

test/elixir/lib/couch.ex
bessbd: Thank you!

test/elixir/test/changes_async_test.exs
bessbd: The format of the `seq`s is different now. They no longer have the predictability they had in previous releases.

test/elixir/test/changes_async_test.exs
bessbd: I think they were comparable previously, too. Now the only thing about them is that they are monotonously increasing.
+CC @kocolosk

bessbd:
Thank you for the review, @jjrodrig . Great comments. Please let me know if I've missed anything.

bessbd:
Hi @jjrodrig ,
Can you please do a second / final pass on this PR?

src/chttpd/src/chttpd_db.erl
jjrodrig: Sorry for the delay. I've been checking this and I think we are leaving dead code during the move to fdb. I suppose this should be addressed in some moment. 

I did some checks changing the callback params to match with the old code [here](https://github.com/apache/couchdb/blob/82bd4d450711b0ff62616229851e5d19fd20e42f/src/chttpd/src/chttpd_changes.erl#L392) and it seems to work with some side effects in the auth cache changes monitor that can be easily solved. 

The old timeout callbacks seem to be dead code, but with the test suite partially I'm not sure. 

Your patch solves the issue in the new execution path, so we will go ahead.

jjrodrig (approved):
+1

Thanks!

bessbd:
Thank you for the review, @jjrodrig ! I'm about to merge this.

PR #3415, opened 2021-03-12
Remove error message on mix test

bessbd:
## Overview

All `mix test` runs were starting with
```
Cannot configure base applications: [:kernel]

These applications are already started by the time the configuration
executes and these configurations have no effect.

If you want to configure these applications for a release, you can
specify them in your vm.args file:

    -kernel config_key config_value

Alternatively, if you must configure them dynamically, you can wrap
them in a conditional block in your config files:

    if System.get_env("RELEASE_MODE") do
      config :kernel, ...
    end

and then configure your releases to reboot after configuration:

    releases: [
      my_app: [reboot_system_after_config: true]
    ]

This happened when loading config/config.exs or
one of its imports.
```

I haven't been able to find references to `Mix.Config`'s `error_logger` in `kernel` so I've opened this PR to remove it.
I understand this is probably a [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) so I'm okay with keeping it, but I'd prefer this PR be `-1`d with a reference to why those lines need to stay.
+CC @iilyak (`git blame` tells me it was introduced in d427f353ab4be03998685f6a167d9559c7b50a76 ).

## Checklist

### DIFF
--- a/config/integration.exs
+++ b/config/integration.exs
@@ -5,8 +5,5 @@ config :logger,
   compile_time_purge_level: :debug,
   level: :debug
 
-config :kernel,
-  error_logger: false
-
 config :sasl,
   sasl_error_logger: false


bessbd:
Thank you for the review and merge, @janl !

PR #3455, opened 2021-03-24
Couch server shard simplification

rnewson:
## Overview

In testing I've found that it would be much easier if max_dbs_open defines the maximum number of .couch files on a per couch_server basis (now that we have sharded the workload over several such processes). A few reasons that come into play;

1. In many test setups, including some couchdb tests in this repo, we set max_dbs_open quite low. The runtime division of that value by the number of schedulers then causes tests to fail in a confusing way. Performance tests are particularly impacted.
2. It is surprising to find that a config setting is not honored directly.
3. max_dbs_open itself exists to address two concerns. Firstly, as a cap on the number of open files directly. Secondly, to reflect the prohibitive cost of updating the server state for very large numbers of entries. For the first item, for small systems, the number of cores will also be small, so the multiplying effect of this change should not be a concern. For the second item, max_dbs_open's new definition is a significant improvement, as it can be set to whatever an individual couch_server can handle in terms of updating the server state.

## Testing recommendations

N/A

https://github.com/apache/couchdb/pull/3366

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -10,10 +10,9 @@ view_index_dir = {{view_index_dir}}
 ; plugin_dir =
 os_process_timeout = 5000 ; 5 seconds. for view servers.
 
-; Maximum number of .couch files to open at once.
-; The actual limit may be slightly lower depending on how
-; many schedulers you have as the allowance is divided evenly
-; among them.
+; Maximum number of .couch files to open at once per couch_server
+; process. There will be as many couch_server processes as there are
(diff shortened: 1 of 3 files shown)


rnewson:
@chewbranca in particular for finding some of the operational difficulties of my original approach.
@janl for his thoughts on the impact of the multiplying effect inherent to this change.

nickva (approved):
+1

rnewson:
am having second thoughts after discussing with rob allen.

rnewson:
closing for now, might come back to it.

PR #3481, opened 2021-03-30
Validate ddoc uses couch_eval

garrensmith:
## Overview

Change the validate ddoc check to use couch_eval.
Also add in some extra functions in couch_eval so that the try_compile
will work.

## Testing recommendations
All current design doc tests should pass.

## Checklist

### DIFF
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -895,7 +895,7 @@ validate_doc_update(Db, Doc, GetDiskDocFun) ->
 
 validate_ddoc(Db, DDoc) ->
     try
-        ok = couch_index_server:validate(Db, couch_doc:with_ejson_body(DDoc))
+        ok = couch_mrview:validate(Db, couch_doc:with_ejson_body(DDoc))
     catch
         throw:{invalid_design_doc, Reason} ->
             {bad_request, invalid_design_doc, Reason};
--- a/src/couch_eval/src/couch_eval.erl
+++ b/src/couch_eval/src/couch_eval.erl
@@ -17,7 +17,9 @@
 -export([
     acquire_map_context/6,
     release_map_context/1,
-    map_docs/2
+    map_docs/2,
+    with_context/2,
+    try_compile/4
 ]).
 
 
@@ -35,6 +37,10 @@
 -type result() :: {doc_id(), [[{any(), any()}]]}.
 -type api_mod() :: atom().
 -type context() :: {api_mod(), any()}.
+-type function_type() :: binary().
+-type function_name() :: binary().
+-type function_src() :: binary().
+-type error(_Error) :: no_return().
 
 -type context_opts() :: #{
     db_name := db_name(),
@@ -46,10 +52,17 @@
     api_mod => api_mod()
 }.
 
+-type with_context_opts() :: #{
+    language := language()
+}.
+
 
 -callback acquire_map_context(context_opts()) -> {ok, any()} | {error, any()}.
 -callback release_map_context(context()) -> ok | {error, any()}.
 -callback map_docs(context(), [doc()]) -> {ok, [result()]} | {error, any()}.
+-callback acquire_context() -> {ok, any()} | {error, any()}.
+-callback release_context(context()) -> ok | {error, any()}.
+-callback try_compile(context(), function_type(), function_name(), function_src()) -> ok.
 
 
 -spec acquire_map_context(
@@ -59,7 +72,10 @@
         sig(),
         lib(),
         map_funs()
-    ) -> {ok, context()} | {error, any()}.
+    ) ->
+    {ok, context()}
+    | error({invalid_eval_api_mod, Language :: binary()})
+    | error({unknown_eval_api_language, Language :: binary()}).
 acquire_map_context(DbName, DDocId, Language, Sig, Lib, MapFuns) ->
     ApiMod = get_api_mod(Language),
     CtxOpts = #{
@@ -85,6 +101,37 @@ release_map_context({ApiMod, Ctx}) ->
 -spec map_docs(context(), [doc()]) -> {ok, result()} | {error, any()}.
 map_docs({ApiMod, Ctx}, Docs) ->
     ApiMod:map_docs(Ctx, Docs).
+
+
+-spec with_context(with_context_opts(), function()) -> 
+    any() 
+    | error({invalid_eval_api_mod, Language :: binary()})
+    | error({unknown_eval_api_language, Language :: binary()}).
+with_context(#{language := Language}, Fun) ->
+    {ok, Ctx} = acquire_context(Language),
+    try 
+        Fun(Ctx)
+    after
+        release_context(Ctx)
+    end.
+
+
+-spec try_compile(context(), function_type(), function_name(), function_src()) -> ok.
+try_compile({ApiMod, Ctx}, FuncType, FuncName, FuncSrc) -> 
+    ApiMod:try_compile(Ctx, FuncType, FuncName, FuncSrc).
+
+
+acquire_context(Language) ->
+    ApiMod = get_api_mod(Language),
+    {ok, Ctx} = ApiMod:acquire_context(),
+    {ok, {ApiMod, Ctx}}.
+
+
+release_context(nil) ->
+    ok;
+
+release_context({ApiMod, Ctx}) ->
+    ApiMod:release_context(Ctx).
 
 
 get_api_mod(Language) when is_binary(Language) ->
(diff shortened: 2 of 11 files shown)


bessbd (commented):
Thank you for this PR, @garrensmith .
The change looks good to me on a first read.
Do you think there's an easy way to add a test for this?

bessbd:
> @bessbd this functionality is used throughout all the fabric2 and couch_views test. I think it's pretty well covered there.

That I get. I'd just like to make sure we avoid accidentally (partially) reverting this.

garrensmith:
@bessbd could you give me an example.

bessbd:
> @bessbd could you give me an example?

If I understand correctly, the essence of this PR is https://github.com/apache/couchdb/pull/3481/files#diff-eb8693c8aa6baf90841ed8bc66c04cbc525af0bb34368e3201ab4891b954190eL2157-R2157 . I'd just like to make sure that we don't inadvertently undo that change.
I had no heavy QA in mind, just a single-line eunit `meck`-y check to assert that `couch_mrview:validate` is called at the right time. (We could even assert `couch_index_server:validate` wasn't called, if that's a fundamental part of this change.)

src/couch_eval/src/couch_eval.erl
```diff
@@ -86,6 +94,26 @@ release_map_context({ApiMod, Ctx}) ->
 map_docs({ApiMod, Ctx}, Docs) ->
     ApiMod:map_docs(Ctx, Docs).
 
+-spec acquire_context(language()) -> {ok, context()} | {error, any()}.
```
iilyak, line 97: The way function is written it cannot return `{error, any()}`. I think the correct spec would be:

```
-type error(_Error) :: no_return().

....

spec acquire_context(language()) -> 
   {ok, context()} 
   | error({invalid_eval_api_mod, Language :: binary()})
   | error({unknown_eval_api_language, Language :: binary()}).
```

src/couch_mrview/src/couch_mrview.erl
```diff
@@ -203,22 +203,23 @@ validate(DbName, _IsDbPartitioned,  DDoc) ->
...
+                couch_eval:try_compile(Ctx, RedName, RedSrc)
         end, Reds)
     end,
     {ok, #mrst{
         language = Lang,
         views = Views
     }} = couch_mrview_util:ddoc_to_mrst(DbName, DDoc),
 
-    try Views =/= [] andalso couch_query_servers:get_os_process(Lang) of
+    try Views =/= [] andalso couch_eval:acquire_context(Lang) of
```
iilyak, line 214: `couch_eval:acqure_context/1` can raise `{invalid_eval_api_mod, Language}` or `{unknown_eval_api_language, Language}. Do we need to handle this on line [224](https://github.com/apache/couchdb/pull/3481/files#diff-637f0ee4524b44455ad1fcbfa99fdbdc757c0a0c7a1bd462f4e9b5f51a548dc9R224)?

src/couch_js/src/couch_js.erl
```diff
@@ -49,3 +52,15 @@ map_docs(Proc, Docs) ->
...
+    Ctx = couch_query_servers:get_os_process(?JS),
+    {ok, Ctx}.
+
+
+release_context(Proc) ->
+    couch_js_query_servers:ret_os_process(Proc).
+
+
+try_compile(Proc, FunName, FunSrc) ->
+    couch_query_servers:try_compile(Proc, map, FunName, FunSrc).
```
iilyak, line 66: Is it always would be a `map` in a second argument?

src/couch_eval/src/couch_eval.erl
iilyak: Don't forget to update `-callback acquire_context()....` above as well.

src/couch_js/src/couch_js.erl
iilyak: I can see that we used to call `couch_query_servers:try_compile(Proc, reduce, RedName, RedSrc)` [here](https://github.com/apache/couchdb/pull/3481/files#diff-637f0ee4524b44455ad1fcbfa99fdbdc757c0a0c7a1bd462f4e9b5f51a548dc9L206)

iilyak: Looks like we either need to pass it explicitly as try_compile argument or have it as part of a context.

iilyak:
There are very few places where we need to call `acquire_context/1` and `release_context/1` so I am ok with proposed change. However I wanted to make sure you did consider a different option, which is something like the following:

```
-module(couch_eval).

-map_docs/2
+map_docs/2,
+with_context/2,
+try_compile/3

with_context(#{language := Language}, Fun) ->
    case acquire_context(Language) of
    try acquire_context(Language) of
        {ok, Ctx} -> Fun(Ctx)
    after
        release_context(Ctx)
    end.
    

%% this would be private function
-spec acquire_context(language()) -> {ok, context()} | {error, any()}.
acquire_context(Language) ->
    ApiMod = get_api_mod(Language),
    {ok, Ctx} = ApiMod:acquire_context(),
    {ok, {ApiMod, Ctx}}.
```

Then the caller would look like:

```
-module(couch_mrview).

validate(DbName, _IsDbPartitioned,  DDoc) ->
    ...
    Views =/= [] andalso couch_eval:with_context(#{language => Lang}, fun(Ctx) ->
         lists:foreach(fun(V) -> ValidateView(Ctx, V) end, Views)     
    end),
    ok.
```

src/couch_js/src/couch_js.erl
iilyak: Never mind the idea of passing the `map | reduce` in the context. I can see now that we are using both in `couch_mrview:validate/3`

- map on [line 192](https://github.com/apache/couchdb/pull/3481/files#diff-637f0ee4524b44455ad1fcbfa99fdbdc757c0a0c7a1bd462f4e9b5f51a548dc9L192)
- reduce on [line 206](https://github.com/apache/couchdb/pull/3481/files#diff-637f0ee4524b44455ad1fcbfa99fdbdc757c0a0c7a1bd462f4e9b5f51a548dc9L206)

src/couch_mrview/src/couch_mrview.erl
garrensmith: I don't think so. I'm happy for that error to bubble up.

iilyak:
> I had no heavy QA in mind, just a single-line eunit meck-y check to assert that couch_mrview:validate is called at the right time. (We could even assert couch_index_server:validate wasn't called, if that's a fundamental part of this change.)

Since `couch_index_server:validate/2` is no longer used we could break it and other tests would fail if we partially revert the PR.

```
-module(couch_index_server).
...
validate(_Db, _DDoc) ->
    error({assert, ?MODULE, ?FUNCTION_NAME, ?FUNCTION_ARITY, ?LINE}).
```

bessbd:
> > I had no heavy QA in mind, just a single-line eunit meck-y check to assert that couch_mrview:validate is called at the right time. (We could even assert couch_index_server:validate wasn't called, if that's a fundamental part of this change.)
> 
> Since `couch_index_server:validate/2` is no longer used we could break it and other tests would fail if we partially revert the PR.
> 
> ```
> -module(couch_index_server).
> ...
> validate(_Db, _DDoc) ->
>     error({assert, ?MODULE, ?FUNCTION_NAME, ?FUNCTION_ARITY, ?LINE}).
> ```

I didn't know that `couch_index_server:validate` would become dead code after this change. So we might as well remove the `fun` altogether then, right?

garrensmith:
@bessbd right. I've deleted the function. I think that is the best approach.

src/couch/src/couch_db.erl
```diff
@@ -895,7 +895,7 @@ validate_doc_update(Db, Doc, GetDiskDocFun) ->
 
 validate_ddoc(Db, DDoc) ->
     try
-        ok = couch_index_server:validate(Db, couch_doc:with_ejson_body(DDoc))
+        ok = couch_mrview:validate(Db, couch_doc:with_ejson_body(DDoc))
```
rnewson, line 898: isn't this a layering violating? the index/mrview split was intended to make the m/r view a specific kind of index

garrensmith: This isn't ideal. But it will be temporary. I want to remove couch_index_server but I want to do it in a few smaller PR's rather than one large one. This first step will get the validation_ddoc using couch_eval instead of couch_index_server and the old javascript proc system. 

The next one will be fixes to the replication filtering and the vdu. A final one will be where I can remove couch_index_server and the fix this.

rnewson: ok, I'm onboard with this change and will keep an eye out for the subsequent PR's. There is a lot of cleanup of dead code to do before a 4.0 release can be considered so I hope to see couch_index_server disappear in the near future.

iilyak (changes requested):

iilyak (approved):
+1 from me.

PR #3483, opened 2021-03-30
password requirement

### LINKED ISSUE #2951: Password requirement validation when creating users in CouchDB
## Summary

When creating users in CouchDB with `PUT /_users/org.couchdb.user:<name>` I would like to have the ability to enforce password rules. For example, a password should have a minimum length of 10 characters, special characters or whatever the developer of the app wants.

I tried to accomplish this with the `validate_doc_update` function but realised that the passed document already contains `derived_key` and `salt` but no password.

My current solution involves a serverless function that acts as proxy and checks the password requirements before passing the request to CouchDB.

## Desired Behaviour

When a request is made to `PUT /_users/org.couchdb.user:<name>`, CouchDB should validate the password and return an HTTP `400` error if the validation fails. Ideally the response contains a message like: `Password does not conform to requirements.`

## Possible Solution

I had a quick chat with @janl about this problem and his idea was to add a `passwordRegex` config option to CouchDB.

Example:

```ini
# Check that a password has 10 or more characters
passwordReqex = /.{10,}/ 
```

Terreii:
## Overview

This pull request adds a new setting for password requirements.
Closes #2951.

## Testing recommendations

A user should not be able to change their password, if it does not
match the requirements set by the admin.

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -279,6 +279,10 @@ iterations = 10 ; iterations for password hashing
 ; min_iterations = 1
 ; max_iterations = 1000000000
 ; password_scheme = pbkdf2
+; List of Erlang RegExp or tuples of RegExp and an optional error message.
+; Where a new password must match all RegExp.
+; Example: [{".{10,}", "Password min length is 10 characters."}, "\\d+"]
+; password_regexp = []
 ; proxy_use_secret = false
 ; comma-separated list of public fields, 404 if empty
 ; public_fields =
--- a/src/couch/src/couch_users_db.erl
+++ b/src/couch/src/couch_users_db.erl
@@ -26,6 +26,8 @@
 -define(ITERATIONS, <<"iterations">>).
 -define(SALT, <<"salt">>).
 -define(replace(L, K, V), lists:keystore(K, 1, L, {K, V})).
+-define(REQUIREMENT_ERROR, "Password does not conform to requirements.").
+-define(PASSWORD_SERVER_ERROR, "Server cannot hash passwords at this time.").
 
 % If the request's userCtx identifies an admin
 %   -> save_doc (see below)
@@ -67,6 +69,7 @@ save_doc(#doc{body={Body}} = Doc) ->
     {undefined, _} ->
         Doc;
     {ClearPassword, "simple"} -> % deprecated
+        ok = validate_password(ClearPassword),
         Salt = couch_uuids:random(),
         PasswordSha = couch_passwords:simple(ClearPassword, Salt),
         Body0 = ?replace(Body, ?PASSWORD_SCHEME, ?SIMPLE),
@@ -75,6 +78,7 @@ save_doc(#doc{body={Body}} = Doc) ->
         Body3 = proplists:delete(?PASSWORD, Body2),
         Doc#doc{body={Body3}};
     {ClearPassword, "pbkdf2"} ->
+        ok = validate_password(ClearPassword),
         Iterations = list_to_integer(config:get("couch_httpd_auth", "iterations", "1000")),
         Salt = couch_uuids:random(),
         DerivedKey = couch_passwords:pbkdf2(ClearPassword, Salt, Iterations),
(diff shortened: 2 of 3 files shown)


rel/overlay/etc/default.ini
```diff
@@ -279,6 +279,7 @@ iterations = 10 ; iterations for password hashing
 ; min_iterations = 1
 ; max_iterations = 1000000000
 ; password_scheme = pbkdf2
+; password_reqexp = ""
```
wohali, line 282: This should be `regexp` not `reqexp`.

src/couch/src/couch_users_db.erl
```diff
@@ -86,7 +90,63 @@ save_doc(#doc{body={Body}} = Doc) ->
...
+% Throws if not.
+validate_password(ClearPassword) ->
+    case config:get("couch_httpd_auth", "password_reqexp", "") of
+    "" ->
+        ok;
+    "[]" ->
+        ok;
+    ValidateConfig ->
+        case couch_util:parse_term(ValidateConfig) of
+        {ok, RegExpList} when is_list(RegExpList) ->
```
jaydoane, line 106: This line should be indented 4 spaces

src/couch/src/couch_users_db.erl
```diff
@@ -86,7 +90,63 @@ save_doc(#doc{body={Body}} = Doc) ->
...
+        ok;
+    ValidateConfig ->
+        case couch_util:parse_term(ValidateConfig) of
+        {ok, RegExpList} when is_list(RegExpList) -> 
+            % Check the password on every RegExp.
+            Loop = fun(RegExp) ->
+                check_password_with_regexp(ClearPassword, RegExp)
+            end,
+            lists:foreach(Loop, RegExpList),
+            ok;
```
jaydoane, line 112: Stylistically, I think this is preferred:
```erlang
            lists:foreach(fun(RegExp) ->
                check_password_with_regexp(ClearPassword, RegExp)
            end, RegExpList);
```

src/couch/src/couch_users_db.erl
```diff
@@ -86,7 +90,63 @@ save_doc(#doc{body={Body}} = Doc) ->
...
+        {ok, NonListValue} ->
+            couch_log:error(
+                "[couch_httpd_auth] password_reqexp value of '~p' is invalid.",
```
jaydoane, line 115: `?MODULE` is preferred to hard-coding the module name here. Also, please omit the `[]`

src/couch/src/couch_users_db.erl
```diff
@@ -86,7 +90,63 @@ save_doc(#doc{body={Body}} = Doc) ->
...
+    check_password(Password, RegExp, ?REQUIREMENT_ERROR);
+% Not correct RegExpValue.
+check_password_with_regexp(_, RegExpValue) ->
+    couch_log:error(
+    "[couch_httpd_auth] password_reqexp value of '~p' is invalid.", [RegExpValue]),
+    throw({forbidden, ?PASSWORD_SERVER_ERROR}).
+
+check_password(Password, RegExp, ErrorMsg) ->
+    case re:run(Password, RegExp, [{capture, none}]) of
+    match ->
```
jaydoane, line 146: Indentation

src/couch/src/couch_users_db.erl
```diff
@@ -86,7 +90,63 @@ save_doc(#doc{body={Body}} = Doc) ->
...
+                check_password_with_regexp(ClearPassword, RegExp)
+            end,
+            lists:foreach(Loop, RegExpList),
+            ok;
+        {ok, NonListValue} ->
+            couch_log:error(
+                "[couch_httpd_auth] password_reqexp value of '~p' is invalid.",
+                [NonListValue]
+            ),
+            throw({forbidden, ?PASSWORD_SERVER_ERROR});
```
jaydoane, line 118: Is `forbidden` the correct error here? I would argue 400 is more appropriate, since the password is "malformed" based on not passing the regexp

jaydoane (changes requested):
This looks pretty good, although there are some stylistic issues, and more seriously, I'm not convinced 403 is the right error for failing the password regex check.

src/couch/src/couch_users_db.erl
Terreii: I changed it the the latest commit a little. What do you think?

src/couch/src/couch_users_db.erl
Terreii: This is the configuration name, inspired by the error message of [`save_doc/1`](https://github.com/apache/couchdb/blob/776f920021dd15c51355e9a904bdc5a2250b19d9/src/couch/src/couch_users_db.erl#L88).

src/couch/src/couch_users_db.erl
Terreii: If in [`save_doc/1`](https://github.com/apache/couchdb/blob/776f920021dd15c51355e9a904bdc5a2250b19d9/src/couch/src/couch_users_db.erl#L89) the `[couch_httpd_auth] password_scheme` config has an invalid value, the user will receive an 403 status. I also used 403 for all instances where a part of `[couch_httpd_auth] password_regexp` is wrong.
In the latest commit, I changed the status code for "malformed" password to 400.

src/couch/src/couch_users_db.erl
Terreii: Done

src/couch/src/couch_users_db.erl
Terreii: Is your request to change the error logs to be more like [this](https://github.com/apache/couchdb/blob/776f920021dd15c51355e9a904bdc5a2250b19d9/src/mem3/src/mem3_rep.erl#L496)
Like this so:
`couch_log:error("~p : ~p couldn't be parsed. '~p' is not a list.", [?MODULE, "password_regexp", NonListValue])`

rel/overlay/etc/default.ini
```diff
@@ -279,6 +279,8 @@ iterations = 10 ; iterations for password hashing
 ; min_iterations = 1
 ; max_iterations = 1000000000
 ; password_scheme = pbkdf2
+; List of tuples of an Erlang RegExp and an optional error message. Where a new password must match all RegExp. Example: [{".{10,}", "Password min length is 10 characters."}]
```
janl, line 282: nitpick, please keep lines under 80 chars.

src/couch/src/couch_users_db.erl
janl: @jaydoane if another auth module supports PW validation it can hardcode its own name, so I think this is fine here. — And I prefer the convention of `[section] key` for referring to config value.

src/couch/src/couch_users_db.erl
janl: 403 for “we can’t hash any PW because we are misconfigured” is fine because we used that before, although a 5xx code might be more appropriate, but no need to change all that now.

400 for “your password is bad” is ok too. Sadly, there is no 4xx for “you failed whatever custom validation I have going on here”, only ones specific to other HTTP behaviour like `Accept` or `Expect` headers,

janl:
this is neat, let’s get it into 3.2.0 :)

jaydoane (approved):
Just some minor line length issues remain for me. Tests pass, although they seem to add about 20 seconds to the suite on my machine:
```
UsersDbTest
  * test users password requirements (76.2ms)
  * test users password requirements with non list value (4973.3ms)
  * test users db (5267.9ms)
  * test users password requirements with not correct syntax (4968.3ms)
  * test users password requirements with a string in a list (4969.9ms)
  ```
Please remember to squash and rebase once @wohali and @janl approve.

**Comment by bessbd** [comment]
Once this is merged, I think it needs a docs update. @Terreii : are you interested in opening a PR for docs, too?

File: src/couch/src/couch_users_db.erl [hunk 8]
```diff
@@ -86,7 +90,72 @@ save_doc(#doc{body={Body}} = Doc) ->
...
+                            {ok, RegExp, PasswordErrorMsg} ->
+                                check_password(ClearPassword, RegExp, PasswordErrorMsg);
+                            {error} ->
+                                couch_log:error(
+                                    "[couch_httpd_auth] password_regexp value of '~p' is invalid.",
+                                    [RegExpTuple]
+                                ),
+                                throw({forbidden, ?PASSWORD_SERVER_ERROR})
+                        end
+                    end, RegExpList),
```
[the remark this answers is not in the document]
Terreii: Done

rel/overlay/etc/default.ini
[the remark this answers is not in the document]
janl: Alternatively it could encourage folks to provide proper Reasons, which is likely beneficial down the road :) — I wouldn’t worry about shortcuts here.

test/elixir/test/users_db_test.exs
```diff
@@ -298,4 +298,142 @@ defmodule UsersDbTest do
 
     assert resp.body["userCtx"]["name"] == "foo@example.org"
   end
+
+  @tag :with_db
+  test "users password requirements", context do
```
janl, line 303: `context` seems unused:

```
[2021-04-09T17:17:07.011Z] Excluding tags: [:without_quorum_test, :with_quorum_test, :pending, :skip_on_****]
[2021-04-09T17:17:07.011Z] 
[2021-04-09T17:17:14.578Z] warning: variable "string" is unused (if the variable is not meant to be used, prefix it with an underscore)
[2021-04-09T17:17:14.578Z]   test/elixir/test/utf8_test.exs:32: UTF8Test."test UTF8 support"/1
[2021-04-09T17:17:14.578Z] 
[2021-04-09T17:17:14.578Z] warning: variable "context" is unused (if the variable is not meant to be used, prefix it with an underscore)
[2021-04-09T17:17:14.578Z]   test/elixir/test/users_db_test.exs:303: UsersDbTest."test users password requirements"/1
[2021-04-09T17:17:14.578Z] 
[2021-04-09T17:17:14.578Z] warning: variable "context" is unused (if the variable is not meant to be used, prefix it with an underscore)
[2021-04-09T17:17:14.578Z]   test/elixir/test/users_db_test.exs:391: UsersDbTest."test users password requirements with non list value"/1
[2021-04-09T17:17:14.578Z] 
[2021-04-09T17:17:14.578Z] warning: variable "context" is unused (if the variable is not meant to be used, prefix it with an underscore)
[2021-04-09T17:17:14.578Z]   test/elixir/test/users_db_test.exs:416: UsersDbTest."test users password requirements with not correct syntax"/1
```

janl:
> Just some minor line length issues remain for me. Tests pass, although they seem to add about 20 seconds to the suite on my machine:
> 
> ```
> UsersDbTest
>   * test users password requirements (76.2ms)
>   * test users password requirements with non list value (4973.3ms)
>   * test users db (5267.9ms)
>   * test users password requirements with not correct syntax (4968.3ms)
>   * test users password requirements with a string in a list (4969.9ms)
> ```
> 

looks like wrapping all tests cases in `with_db` starts a new couch server each time, we could probably get away with only starting it once for all these tests.

Terreii:
There wasn't a change without the `with_db` tag. But all password requirement tests without it.
But I could, with the `with a string in a list` test now being removed, combine the remaining error path tests into one test. Reducing the time added to "only" 10s. Or I could combine them all into one password requirements test.

Terreii:
Down to 5s - 6s.

janl (approved):
nicely done!

Terreii:
Thanks  😄

bessbd:
Should we port this to `main`?

Terreii:
> Should we port this to `main`?

I did port it on #3539.

PR #3500, opened 2021-04-07
[3.x] Add support for Spidermonkey 78

### LINKED ISSUE #3476: Add support for SpiderMonkey 78
Necessary for debian bullseye, which is now in hard freeze:

https://packages.debian.org/search?keywords=mozjs&searchon=names&suite=testing&section=all

This will incidentally allow compiling against Ubuntu `groovy (20.10)` and `hirsuite`, though these are not officially supported releases, Fedora 34, and Alpine Linux 3.13. (CentOS 8 "Stream" is still on mozjs 52/60.)

Hopefully this is "most of, but not all" of the changes @janl landed in #3434 plus a couple of minor `rebar.config.script` tweaks.

wohali:
Good to go, got the failing test fixed. Closes #3476 

/cc @janl @dch @davisp

### DIFF
--- a/src/couch/rebar.config.script
+++ b/src/couch/rebar.config.script
@@ -61,6 +61,8 @@ SMVsn = case lists:keyfind(spidermonkey_version, 1, CouchConfig) of
         "60";
     {_, "68"} ->
         "68";
+    {_, "78"} ->
+        "78";
     {_, "86"} ->
         "86";
     undefined ->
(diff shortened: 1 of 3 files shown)


davisp:
I'm pretty sure that same test was failing the last time we updated SpiderMonkey as well. Its not the greatest test and is basically showing that SpiderMonkey's garbage collection/memory efficiency is improving. I'd be on board if we just add another test function that basically blows up with the same error code rather than relying on specific memory behavior of SpiderMonkey.

wohali:
@davisp Yeah, it was: https://github.com/apache/couchdb/commit/c4aee4dc5b19cdcc55bb44e8863bc62403d635e6

wohali:
Got it, this is ready for review.

janl (approved):

PR #3505, opened 2021-04-13
[CI] Move to new containers for full build

wohali:
This is the companion PR to #3504, and moves our "full" builds to the new containers as well.

It drops CentOS 6 because we are no longer able to build these containers, now that their repositories are gone.


wohali:
I intend to merge this proactively if https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FFullPlatformMatrix/detail/jenkins-3.x-new-containers/2/pipeline passes.

PR #3508, opened 2021-04-14
Add docker creds to CI run

wohali:
This should work around rate limiting that @bessbd and I have seen.

The creds are stored in Jenkins on https://ci-couchdb.apache.org/ . If you would like the account password, let me know.


wohali:
@bessbd this should work around our rate limiting.

bessbd:
> @bessbd this should work around our rate limiting.

Woot woot! Thank you for this change, @wohali

PR #3525, opened 2021-04-23
Export parse_int/1, parse_pos_int/1 and parse_boolean/1

iilyak:
## Overview

We used to have public `parse_pos/1`, `parse_pos_int/1` and `parse_boolean/1` and they were used by one of the EPI plugins.
This PR exports these functions again.

## Testing recommendations

Compilation without new warnings is enough to test it.

- This was changed in https://github.com/apache/couchdb/pull/3503

## Checklist

### DIFF
--- a/src/couch_views/src/couch_views_http_util.erl
+++ b/src/couch_views/src/couch_views_http_util.erl
@@ -28,7 +28,10 @@
     row_to_obj/1,
     row_to_obj/2,
     row_to_json/1,
-    row_to_json/2
+    row_to_json/2,
+    parse_boolean/1,
+    parse_int/1,
+    parse_pos_int/1
 ]).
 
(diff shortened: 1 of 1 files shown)


rnewson:
we prefer to expose functions via the "top" module, in this case that would be couch_views.erl, so that we aren't coupling applications so delicately. These functions seem like they belong somewhere else entirely, though.

iilyak:
> These functions seem like they belong somewhere else entirely, though.

Maybe new application where we would be having only utility functions. 

- parsing
- json

The rationale for a new application is to avoid circular dependencies. The new application shouldn't depend on anything else from couchDB.

rnewson:
a new library application sounds right.

iilyak:
Closing in favor of https://github.com/apache/couchdb/pull/3529

PR #3536, opened 2021-04-28
fix default values for prometheus templates

tonysun83:
## Overview
Couch_prometheus's additional http server should be off by default
and the port should have a default value of 17986 when running

`make release`.  The http server was mistakenly left on without a port
and that led to startup issues.

## Testing recommendations

`make release` should install correctly after running `configure`.
I performed the above step and the `default.ini` in `rel/couchdb` shows:

```
[prometheus]
additional_port = false
bind_address = 127.0.0.1
port = 17986
```

## Checklist


iilyak (approved):
+1

PR #3544, opened 2021-05-04
Handle the case when md5 field is undefined

iilyak:
## Overview

Previously the default value for md5 field was `<<>>`. This value changed to
undefined when we switch to using maps instead of erlang records.
The change break the `couch_att:to_json/4` funciton because `base64:encode/1`
cannot handle atom `undefined`.

The issue happens for inline attachments only this is why it wasn't discovered
earlier.

## Testing recommendations

1. run test suite
2. there is a manual way to test it as well
```
curl -u adm:pass -X PUT -H 'Content-Type: application/json' http://127.0.0.1:15984/test
curl -u adm:pass -X PUT -H 'Content-Type: application/json' -d '{"_id": "_design/ddd", "language": "javascript", "views": {"diet": {"map": ""}}, "_attachments": {"foo.txt": {"content_type": "text/plain", "data": "VGhpcyBpcyBhIGJhc2U2NCBlbmNvZGVkIHRleHQ="}}}' http://127.0.0.1:15984/test/_design/ddd
```

N/A

## Checklist

### DIFF
--- a/src/couch/src/couch_att.erl
+++ b/src/couch/src/couch_att.erl
@@ -331,7 +331,8 @@ digest_from_json(Props) ->
     end.
 
 
-to_json(Att, OutputData, DataToFollow, ShowEncoding) ->
+to_json(#{md5 := Md5} = Att, OutputData, DataToFollow, ShowEncoding)
+        when is_binary(Md5) ->
     #{
         name := Name,
         type := Type,
@@ -339,7 +340,6 @@ to_json(Att, OutputData, DataToFollow, ShowEncoding) ->
         disk_len := DiskLen,
         att_len := AttLen,
         revpos := RevPos,
-        md5 := Md5,
         encoding := Encoding
     } = Att,
     Props = [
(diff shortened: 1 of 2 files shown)


src/couch/src/couch_att.erl
```diff
@@ -346,9 +346,14 @@ to_json(Att, OutputData, DataToFollow, ShowEncoding) ->
         {<<"content_type">>, Type},
         {<<"revpos">>, RevPos}
     ],
-    DigestProp = case base64:encode(Md5) of
-        <<>> -> [];
-        Digest -> [{<<"digest">>, <<"md5-", Digest/binary>>}]
+    DigestProp = case Md5 of
+        <<>> ->
```
eiri, line 350: But you are changing logic here, right? Previously match for empty bin was on `base64` encode of `Md5` and here you are doing this on `Md5` itself. With this change an empty string will give you `<<"md5-">>, because `base64:encode(<<"">>) == <<>>`

iilyak: Nice catch!

iilyak: fixed in https://github.com/apache/couchdb/pull/3544/commits/ea6e208d0a30a8b9095ce38116afcc377f71429a

eiri (approved):
legit

PR #3588, opened 2021-05-31
Aegis improvements

jaydoane:
## Overview

- Add aegis eunit coverage
- Decrease likelihood of key expiring from cache while in use
- Fix test which breaks when aegis configured with operational key manager
- Add `aegis:key_manager/0` function

## Testing recommendations

`make eunit apps=aegis` should pass with default `aegis_noop_key_manager` and any operational key manager you might choose.

## Checklist

### DIFF
--- a/src/aegis/src/aegis_server.erl
+++ b/src/aegis/src/aegis_server.erl
@@ -49,6 +49,7 @@
 -define(CACHE_MAX_AGE_SEC, 1800).
 -define(CACHE_EXPIRATION_CHECK_SEC, 10).
 -define(LAST_ACCESSED_INACTIVITY_SEC, 10).
+-define(CACHE_DELETION_GRACE_SEC, 5). % Keep in cache after expiration
 
 
 -record(entry, {uuid, encryption_key, counter, last_accessed, expires_at}).
@@ -168,43 +169,11 @@ handle_call({insert_key, UUID, DbKey}, _From, #{cache := Cache} = St) ->
     NewSt = insert(St, UUID, DbKey),
     {reply, ok, NewSt, ?TIMEOUT};
 
-handle_call({encrypt, #{uuid := UUID} = Db, Key, Value}, From, St) ->
+handle_call({encrypt, Db, Key, Value}, From, St) ->
+    handle_crypto_call(fun do_encrypt/4, Db, Key, Value, From, St);
 
-    {ok, DbKey} = lookup(St, UUID),
-
-    erlang:spawn(fun() ->
-        process_flag(sensitive, true),
-        try
-            do_encrypt(DbKey, Db, Key, Value)
-        of
-            Resp ->
-                gen_server:reply(From, Resp)
-        catch
-            _:Error ->
-                gen_server:reply(From, {error, Error})
-        end
-    end),
-
-    {noreply, St, ?TIMEOUT};
-
-handle_call({decrypt, #{uuid := UUID} = Db, Key, Value}, From, St) ->
-
-    {ok, DbKey} = lookup(St, UUID),
-
-    erlang:spawn(fun() ->
-        process_flag(sensitive, true),
-        try
-            do_decrypt(DbKey, Db, Key, Value)
-        of
-            Resp ->
-                gen_server:reply(From, Resp)
-        catch
-            _:Error ->
-                gen_server:reply(From, {error, Error})
-        end
-    end),
-
-    {noreply, St, ?TIMEOUT};
+handle_call({decrypt, Db, Key, Value}, From, St) ->
+    handle_crypto_call(fun do_decrypt/4, Db, Key, Value, From, St);
 
 handle_call(_Msg, _From, St) ->
     {noreply, St}.
@@ -235,6 +204,29 @@ code_change(_OldVsn, St, _Extra) ->
 
 
 %% private functions
+
+
+handle_crypto_call(DoCryptoOp, Db, Key, Value, From, St) ->
+    #{uuid := UUID} = Db,
+    case lookup(St, UUID) of
+        {error, not_found} ->
+            gen_server:reply(From, {error, db_encryption_key_not_cached});
+        {ok, DbKey} ->
+            erlang:spawn(fun() ->
+                process_flag(sensitive, true),
+                try
+                    DoCryptoOp(DbKey, Db, Key, Value)
+                of
+                    Resp ->
+                        gen_server:reply(From, Resp)
+                catch
+                    _:Error ->
+                        gen_server:reply(From, {error, Error})
+                end
+            end),
+            {noreply, St, ?TIMEOUT}
+    end.
+
 
 do_open_db(#{uuid := UUID} = Db) ->
     case ?AEGIS_KEY_MANAGER:open_db(Db) of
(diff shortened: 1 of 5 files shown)


iilyak (approved):
+1.

```
module 'aegis_noop_key_manager'
module 'aegis_basic_test'
module 'aegis_sup'
module 'aegis'
module 'aegis_server'
module 'aegis_server_test'
  aegis_server_test: fun.test_init_db (init_db returns true when encryption enabled)...[0.006 s] ok
  aegis_server_test: fun.test_open_db (open_db returns true when encryption enabled)...ok
  aegis_server_test: fun.test_get_db_info (get_db_info returns encryption enabled true)...ok
  aegis_server_test: fun.test_init_db_cache (init_db caches key)...[0.002 s] ok
  aegis_server_test: fun.test_open_db_cache (open_db caches key)...[0.002 s] ok
  aegis_server_test: fun.test_encrypt_cache (encrypt fetches and caches key when it's missing)...[0.001 s] ok
  aegis_server_test: fun.test_decrypt_cache (decrypt fetches and caches key when it's missing)...ok
  aegis_server_test: fun.test_disabled_init_db (init_db returns false when encryptions disabled)...ok
  aegis_server_test: fun.test_disabled_open_db (open_db returns false when encryptions disabled)...ok
  aegis_server_test: fun.test_disabled_encrypt (pass through on encrypt when encryption disabled)...ok
  aegis_server_test: fun.test_disabled_decrypt (pass through on decrypt when encryption disabled)...ok
  aegis_server_test: fun.test_advance_counter (counter moves forward on access bump)...[0.003 s] ok
  aegis_server_test: fun.test_evict_old_entries (oldest entries evicted)...[0.005 s] ok
  aegis_server_test: fun.test_bump_accessed (access bump preserves entries)...[0.002 s] ok
  aegis_server_test: fun.test_remove_expired (expired entries removed)...[0.875 s] ok
  [done in 88.672 s]
module 'aegis_key_manager'
module 'aegis_app'
module 'aegis_keywrap'
  aegis_keywrap:89: test_wrap_unwrap...ok
  aegis_keywrap:90: test_wrap_unwrap...ok
  aegis_keywrap:89: test_wrap_unwrap...ok
  aegis_keywrap:90: test_wrap_unwrap...ok
  aegis_keywrap:89: test_wrap_unwrap...ok
  aegis_keywrap:90: test_wrap_unwrap...ok
  aegis_keywrap:89: test_wrap_unwrap...ok
  aegis_keywrap:90: test_wrap_unwrap...ok
  aegis_keywrap:89: test_wrap_unwrap...ok
  aegis_keywrap:90: test_wrap_unwrap...ok
  aegis_keywrap:89: test_wrap_unwrap...ok
  aegis_keywrap:90: test_wrap_unwrap...ok
  aegis_keywrap: fail_test...ok
  [done in 0.039 s]
  All 28 tests passed.
  ```

File: src/aegis/src/aegis_server.erl [hunk 1]
```diff
@@ -286,8 +278,10 @@ is_key_fresh(UUID) ->
     Now = fabric2_util:now(sec),
 
     case ets:lookup(?KEY_CHECK, UUID) of
-        [{UUID, ExpiresAt}] when ExpiresAt >= Now -> true;
-        _ -> false
+        [{UUID, ExpiresAt}] when ExpiresAt - ?KEY_EARLY_EXPIRE_SEC > Now ->
```
eiri, line 281: You lost me here - so this makes keys to expire 5 secs earlier, i.e. this is virtually the same as reducing `[aegis] cache_max_age_sec` config parameter on 5 sec less. Expired keys and missing from cache keys are treated the same way by doing sync call, so what the point of this change?

rnewson: same question here. In epep we had a 'fresh', 'stale' and 'expired' category for this kind of thing (where a stale entry was usable until expired, and we forced a refresh when we first notice its stale). Did I really forget to do the same thing here? :(

eiri: Aegis cache's refresh needs to read/write from fdb, hence it needs fdb context and can't be run on background off the  "parent" process, so 'stale' strategy doesn't really work here.

jaydoane: > this is virtually the same as reducing [aegis] cache_max_age_sec config parameter on 5 sec less

I don't think so, because `ExpiresAt` is being compared to `now()` in both `is_key_fresh` and also `remove_expired_entries`, whereas this change basically says that the key is not "fresh" if it's 5 seconds away from expiring.  I think the `badmatch` crash exposed this race condition, and `KEY_EARLY_EXPIRE_SEC` is an attempt to minimize the impact of that race by not trying to use a cached key that's too close to expiring. I wonder if 10 seconds would be better though since that's the frequency used by the expiration reaper?

> Did I really forget to do the same thing here? :(

I think so, but that is effectively what adding this parameter does: `is_key_fresh` returns false not only when the key is expired, but also when it is stale. Maybe a better name would be `KEY_STALE_SEC`?

eiri: So there is a race between `is_key_fresh` and `lookup` when `remove_expired_entries` called at the same second as `is_key_fresh`? Wouldn't then changing ` MatchConditions = [{'=<', '$1', fabric2_util:now(sec)}],`  in `remove_expired_entries` work better from logical point of view, because right now you are not really refreshing a key, but expiring it ahead of the configured period and a client will just experience a latency bump earlier, while really the intent here is to delay cache eviction.

This will also address a potential surprise in case "cache_max_age_sec"  happened to be set lower than `?KEY_EARLY_EXPIRE_SEC` and cache will stop working.

jaydoane: Yeah, I think something like
```erlang
    MatchConditions = [{'=<', '$1', fabric2_util:now(sec) + ?KEY_EXPIRE_GRACE_SEC}],

```
would also work, and is likely better for the reasons you mention.

> you are not really refreshing a key, but expiring it ahead of the configured period and a client will just experience a latency bump earlier

By returning false from `is_key_fresh`, it effectively forces a refresh, but yes, it sneakily shortens `CACHE_MAX_AGE_SEC` by a few seconds. Seems like just adding a grace period where the cache is still honored is maybe more intuitive.

eiri: As I see it "refresh" is different from "expire" is that it's happening transparently for the client, i.e. async and in background. Here both behave the same, because fdb context limitation, so from user's perspective it'll looks like they set expiration for 10 sec and got key expired every 5 sec instead, doesn't look intuitive.

jaydoane: I pushed a fixup that reverts changes to `is_key_fresh` and instead honors a new configurable grace period before deleting entries in `remove_expired_entries`. It also required a tweak to one of the lru tests.

eiri (approved):
lgtm

bessbd:
@jaydoane : https://github.com/apache/couchdb/pull/3568 is very close to a merge. Do you want me to hold off with it so that this can be merged before?

jaydoane:
> @jaydoane : #3568 is very close to a merge. Do you want me to hold off with it so that this can be merged before?

Yes, please. I'll merge this today after squashing.

PR #3617, opened 2021-06-10
COUCHDB_FAUXTON_DOCROOT fails on {{fauxton_root}}

sklassen:
"${COUCHDB_FAUXTON_DOCROOT:-{{fauxton_root}}}" substitution fails as it gobbles the the third "}" yielding:

`./rel/couchdb/bin/couchdb: 51: Syntax error: Unterminated quoted string`

Adding a surplus white space succeeds in substitution, but then fauxton fails to locate the directory with trailing white space. 

Two alternatives: 

"${COUCHDB_FAUXTON_DOCROOT:-{{fauxton_root}}/}" removes the triple brace, succeeds in substitution, and should be a valid directory.
"${COUCHDB_FAUXTON_DOCROOT:-{{prefix}}/share/www}" works the same, avoids the doubled // and looks similar to the lines above. 

I went with the second.

## Checklist


nickva:
Thanks for your contribution. I wonder if we could just use a separate variable:

```
+# Use a separate var to work around rebar's mustache template bug
+DEFAULT_FAUXTON_ROOT={{fauxton_root}}
+export COUCHDB_FAUXTON_DOCROOT="${COUCHDB_FAUXTON_DOCROOT:-${DEFAULT_FAUXTON_ROOT}}"
```

That way we can keep the fauxton_root template var and don't have to add slashes to the end.

sklassen:
Super. Yes, that is a clever idea.

nickva:
PR with a default var fix https://github.com/apache/couchdb/pull/3708

nickva:
Merged the PR with the default var fix.

Closing this one

PR #3629, opened 2021-06-15
Fix custodian default system dbs

jaydoane:
These system db defaults were left unchanged when this code was
imported from Cloudant. This changes them to CouchDB defaults.

## Checklist

### DIFF
--- a/src/custodian/src/custodian_db_checker.erl
+++ b/src/custodian/src/custodian_db_checker.erl
@@ -132,7 +132,7 @@ get_dbs() ->
 
 
 get_users_db() ->
-    UsersDb = config:get("couch_httpd_auth", "authentication_db", "users"),
+    UsersDb = chttpd_auth_cache:dbname(),
     [list_to_binary(UsersDb)].
 
 
--- a/src/custodian/src/custodian_util.erl
+++ b/src/custodian/src/custodian_util.erl
@@ -43,7 +43,7 @@ report() ->
     fold_dbs([], Fun).
 
 ensure_dbs_exists() ->
-    DbName = config:get("mem3", "shards_db", "dbs"),
+    DbName = mem3_sync:shards_db(),
     {ok, Db} = mem3_util:ensure_exists(DbName),
     ensure_custodian_ddoc_exists(Db),
     {ok, Db}.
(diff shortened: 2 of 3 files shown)


src/custodian/src/custodian_util.erl
```diff
@@ -43,7 +43,7 @@ report() ->
     fold_dbs([], Fun).
 
 ensure_dbs_exists() ->
-    DbName = config:get("mem3", "shards_db", "dbs"),
+    DbName = config:get("mem3", "shards_db", "_dbs"),
```
chewbranca, line 46: Any reason to not use: https://github.com/apache/couchdb/blob/3.x/src/mem3/src/mem3_sync.erl#L315-L316

src/custodian/src/custodian_db_checker.erl
```diff
@@ -132,7 +132,7 @@ get_dbs() ->
 
 
 get_users_db() ->
-    UsersDb = config:get("couch_httpd_auth", "authentication_db", "users"),
+    UsersDb = config:get("couch_httpd_auth", "authentication_db", "_users"),
```
chewbranca, line 135: Any reason to not use: https://github.com/apache/couchdb/blob/3.x/src/chttpd/src/chttpd_auth_cache.erl#L188-L189

src/custodian/src/custodian_util.erl
jaydoane: Other than laziness, the only reason is that I was recently chastised by @rnewson for using a function in a different module that was not in an `_util` module. However, I do prefer keeping things DRY, so I've changed it!

src/custodian/src/custodian_db_checker.erl
jaydoane: That particular function is not exported. If you prefer, I can also export `dbname/1` from chttpd_auth_cache and then use it.

jaydoane: Fixed up with exported `dbname/0`

chewbranca (approved):
Thanks @jaydoane, LGTM.

If folks think we should rearrange where these databases are declared so they're in a `_util` module then that works too, but let's get rid of the duplication of these string values, especially when they're not correct :+1:

rnewson:
I was speaking of the convention of only calling the "top level" module of an application, in order to allow changes within an application with minimal risk of breaking couchdb as a whole. so if we need something from mem3 from other places, add a function in mem3.erl that delegates to the internal mem3_sync module. high cohesion / low coupling.

PR #3634, opened 2021-06-18
Use correct socks5 proxy ibrowse options

jaydoane:
## Overview

With the move from using a forked ibrowse to upstream [1], the
ibrowse options for socks5 proxy settings all changed to a `socks5_`
prefix.

[1] https://github.com/apache/couchdb/pull/3551

## Testing recommendations

```
make eunit apps=couch_replicator suites=couch_replicator_docs tests=parse_proxy_params_test
```

## Checklist

### DIFF
--- a/src/couch_replicator/src/couch_replicator_docs.erl
+++ b/src/couch_replicator/src/couch_replicator_docs.erl
@@ -571,8 +571,7 @@ parse_proxy_params(ProxyUrl) ->
         password = Passwd,
         protocol = Protocol
     } = ibrowse_lib:parse_url(ProxyUrl),
-    [
-        {proxy_protocol, Protocol},
+    Params = [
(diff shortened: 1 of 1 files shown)


nickva (approved):
Good find, @jaydoane!

I totally missed this one when switching to upstream ibrowse

+1

This would apply to `main` too, but not as a cherry pick, I think.  There we parse the endpoint data into a map in https://github.com/apache/couchdb/blob/main/src/couch_replicator/src/couch_replicator_parse.erl#L227 but that also, later gets "reconstituted" into an `#httpdb` record in https://github.com/apache/couchdb/blob/main/src/couch_replicator/src/couch_replicator_api_wrap.erl#L1020 so we would have to make sure the socks5_* params are handled properly in those cases.

Issue #3367, opened 2021-02-11
The authorization scheme for HTTP Basic Authentication should not be case sensitive


eiri:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

Basic authentication in CouchDB expects `auth-scheme` in `Authorization` to be case sensitive and equal to "Basic".

According to [HTTP Authentication RFC2617](https://tools.ietf.org/html/rfc2617) it should be case-insensitive.

From section 1.2, emphasis is mine:

> HTTP provides a simple challenge-response authentication mechanism
   that MAY be used by a server to challenge a client request and by a
   client to provide authentication information. It uses an extensible,
   **case-insensitive token to identify the authentication scheme**...

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your actual CouchDB! )

## Steps to Reproduce

This is from dev instance with temp creds:

[NOTE]: # ( Include commands to reproduce, if possible. curl is preferred. )

```
$ curl http://localhost:15984/test1 -X PUT -H'Authorization: basic YWRtaW46YWRtaW4='
{"error":"unauthorized","reason":"You are not a server admin."}

$ curl http://localhost:15984/test1 -X PUT -H'Authorization: Basic YWRtaW46YWRtaW4='
{"ok":true}

 $ curl http://localhost:15984/test1 -X GET -H'Authorization: BASIC YWRtaW46YWRtaW4='
{"error":"unauthorized","reason":"You are not authorized to access this db."}

 # curl http://localhost:15984/test1 -X GET -H'Authorization: Basic YWRtaW46YWRtaW4='
{"db_name":"test1", ...
```

## Expected Behaviour

[NOTE]: # ( Tell us what you expected to happen. )

Header `Authorization` for basic authentication should accept all  "Basic", "basic" and "BASIC" `auth-scheme` as valid.

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )
[TIP]:  # ( You can paste the output of curl http://YOUR-COUCHDB:5984/ here. )

* CouchDB version used: HEAD of 3.x branch
* Browser name and version: curl 7.64.1 (x86_64-apple-darwin19.0) libcurl/7.64.1 (SecureTransport) LibreSSL/2.8.3 zlib/1.2.11 nghttp2/1.39.2
* Operating system and version: Darwin 19.6.0 Darwin Kernel Version 19.6.0

## Additional Context

[...]

PR #3637, opened 2021-06-22
Remove case sensitivity for basic auth

noahshaw11:
## Overview

Basic authentication in CouchDB currently expects `auth-scheme` in `Authorization` to be equal to "Basic" (case sensitive). However, according to [HTTP Authentication RFC2617](https://datatracker.ietf.org/doc/html/rfc2617#section-1.2), it should be case-insensitive.

## Testing recommendations

Three test cases have been added to use different cases of "Basic" including "basic", "BASIC", and "BAsIc". Furthermore, you can execute the following curl requests and see that you will receive an appropriate response each time.

```
% curl http://localhost:15984/db1 -X PUT -H'Authorization: basic YWRtOnBhc3M='
{"ok":true}
% curl http://localhost:15984/db2 -X PUT -H'Authorization: Basic YWRtOnBhc3M='
{"ok":true}
% curl http://localhost:15984/db -X GET -H'Authorization: BASIC YWRtOnBhc3M='
{"cluster":{"n":0,"q":0,"r":0,"w":0},"compact_running":false,"data_size":0,"db_name":"db","disk_format_version":0,"disk_size":0,"instance_start_time":"0","purge_seq":0,"encryption":{"enabled":false,"key_manager":{}},"update_seq":"0000000a1f123cb400000000","uuid":"00974c2f618fc85a3baede507945551b","doc_del_count":0,"doc_count":102,"sizes":{"external":98233562,"views":0}}
% curl http://localhost:15984/db -X GET -H'Authorization: bAsIc YWRtOnBhc3M='
{"cluster":{"n":0,"q":0,"r":0,"w":0},"compact_running":false,"data_size":0,"db_name":"db","disk_format_version":0,"disk_size":0,"instance_start_time":"0","purge_seq":0,"encryption":{"enabled":false,"key_manager":{}},"update_seq":"0000000a1f123cb400000000","uuid":"00974c2f618fc85a3baede507945551b","doc_del_count":0,"doc_count":102,"sizes":{"external":98233562,"views":0}}
```

Before, you would receive responses like:

```
$ curl http://localhost:15984/test1 -X PUT -H'Authorization: basic YWRtaW46YWRtaW4='
{"error":"unauthorized","reason":"You are not a server admin."}

$ curl http://localhost:15984/test1 -X PUT -H'Authorization: Basic YWRtaW46YWRtaW4='
{"ok":true}

 $ curl http://localhost:15984/test1 -X GET -H'Authorization: BASIC YWRtaW46YWRtaW4='
{"error":"unauthorized","reason":"You are not authorized to access this db."}

 # curl http://localhost:15984/test1 -X GET -H'Authorization: Basic YWRtaW46YWRtaW4='
{"db_name":"test1", ...)
```

The authorization scheme for HTTP Basic Authentication should not be case sensitive: https://github.com/apache/couchdb/issues/3367

## Checklist


test/elixir/test/security_validation_test.exs
```diff
@@ -12,19 +12,19 @@ defmodule SecurityValidationTest do
...
+      authorization: "BAsIc dG9tOmNhdA=="
     ],
     spike_cat: [
       # spike:cat - which is wrong
       authorization: "Basic c3Bpa2U6Y2F0"
     ],
     spike: [
       # spike:dog
-      authorization: "Basic c3Bpa2U6ZG9n"
+      authorization: "BASIC c3Bpa2U6ZG9n"
```
bessbd, line 27: This seems like a [piggyback](https://stackoverflow.com/a/333883) to me. Can you please add a test case that would fail without your production code change, but passes with it? That way when we're changing code parts related to this fix, we have one specific test case that shows us what exactly is about to break and it is clearly listed in the test report.

noahshaw11: Interesting term haha. I will make this change.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -71,22 +71,39 @@ special_test_authentication_handler(Req) ->
...
-            {User, Pass};
         _ ->
-            nil
-        catch
-        error:function_clause ->
-            throw({bad_request, "Authorization header has invalid base64 value"})
-        end;
-    _ ->
-        nil
+            try string:split(AuthorizationHeader, " ") of
```
iilyak, line 78: I think `string:split/2` can fail only if non string is passed into it. Therefore `try` is not needed here.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -71,22 +71,39 @@ special_test_authentication_handler(Req) ->
...
+                [Basic, Base64Value] ->
+                    BasicLower = string:casefold(Basic),
+                    AuthorizationHeader1 = BasicLower ++ " " ++ Base64Value,
+                    case AuthorizationHeader1 of
+                        "basic " ++ Base64Value ->
```
iilyak, line 83: I think comparing `BasicLower`  is simpler. Something like the following would do:
```
case string:casefold(Basic) of
   "basic" ->
         try re:split(base64:decode(Base64Value), ":",
                 [{return, list}, {parts, 2}]) of
                 ...
```

test/elixir/test/security_validation_test.exs
noahshaw11: Added two new tests cases to test lowercase and mixed case "Basic". See https://github.com/apache/couchdb/pull/3637/commits/a6c2028adf8f717b3a9ec97835388a456fb1ce03.

src/couch/src/couch_httpd_auth.erl
noahshaw11: Understandable. Fixed in https://github.com/apache/couchdb/pull/3637/commits/a6c2028adf8f717b3a9ec97835388a456fb1ce03.

src/couch/src/couch_httpd_auth.erl
noahshaw11: Agreed, comparing on `"basic " ++ Base64Value` was not needed. See https://github.com/apache/couchdb/pull/3637/commits/a6c2028adf8f717b3a9ec97835388a456fb1ce03.

test/elixir/test/security_validation_test.exs
bessbd: Perfect, thank you!

src/couch/src/couch_httpd_auth.erl
```diff
@@ -99,27 +99,32 @@ special_test_authentication_handler(Req) ->
...
-                [User, Pass] ->
-                    {User, Pass};
-                _ ->
-                    nil
-            catch
-                error:function_clause ->
-                    throw({bad_request, "Authorization header has invalid base64 value"})
-            end;
+        % undefined is sent during startup process
+        undefined ->
```
iilyak, line 103: This `undefined` appear to me as a magic value. I think we can flip the order of clauses to make it more readable and be closer to functionality of original version (In original version we ignore any unexpected values).

```
basic_name_pw(Req) ->
    AuthorizationHeader = header_value(Req, "Authorization"),
    case AuthorizationHeader of
        Header when is_list(Header) ->
            [Basic, Base64Value] = string:split(AuthorizationHeader, " "),
            ...
        _ ->
            nil
   end.
```

noahshaw11: Your approach makes sense. See https://github.com/apache/couchdb/pull/3637/commits/dd0b666e743f6466910b278000a865bb087da451.

iilyak (approved):
+1

PR #3697, opened 2021-08-07
Improve handling of + in urls 3.x

noahshaw11:
## Overview

Currently CouchDB encodes '+' into ' ' in non-query parts of URLs. This is against CouchDB standards. See the original issue [here](https://issues.apache.org/jira/browse/COUCHDB-883).

Note: This is an opt-in policy for 3.x.

###  Backward Incompatibility
Since changing the way that '+' is encoded is a breaking change, there is a backward incompatibility issue with 3.x. For example, prior to this change if a user had created a document named `my doc`, they would have used an HTTP request similar to `% curl -X PUT http://localhost:15984/mydb/my+doc -H"Authorization: Basic YWRtOnBhc3M="` and the `my+doc` would have been encoded into `my doc`. However, with this change, `my+doc` would be encoded into `my+doc`.

## Testing recommendations

A test case has been added to test creating a document whose id contains a '+' with `encode_plus = true` and another test case has been added to test creating a document whose id contains a '+' with `encode_plus = false`. Furthermore, a test case has been added to test creating a dbname with '+' and `%2B` with `encode_plus = true`.

```
mix test test/elixir/test/basics_test.exs
```

- Replication of attachments fails when the attachment name contains a +: https://github.com/apache/couchdb/issues/3239
- Non-standard treatment of + character in URLs: https://github.com/apache/couchdb/issues/2235

## Checklist

### DIFF
--- a/test/elixir/test/basics_test.exs
+++ b/test/elixir/test/basics_test.exs
@@ -58,6 +58,62 @@ defmodule BasicsTest do
     assert context[:db_name] in Couch.get("/_all_dbs").body, "Db name in _all_dbs"
   end
 
+  test "Database name with '+' should encode to '+'", _context do
+    set_config({"chttpd", "decode_plus_to_space", "false"})
+
+    random_number = :rand.uniform(16_000_000)
+    db_name = "random+test+db+#{random_number}"
+    resp = Couch.put("/#{db_name}")
+
+    assert resp.status_code == 201
+    assert resp.body["ok"] == true
+
+    resp = Couch.get("/#{db_name}")
+
+    assert resp.status_code == 200
+    assert resp.body["db_name"] == db_name
+  end
+
+  test "Database name with '%2B' should encode to '+'", _context do
+    set_config({"chttpd", "decode_plus_to_space", "true"})
+
+    random_number = :rand.uniform(16_000_000)
+    db_name = "random%2Btest%2Bdb2%2B#{random_number}"
+    resp = Couch.put("/#{db_name}")
+
+    assert resp.status_code == 201
+    assert resp.body["ok"] == true
+
+    resp = Couch.get("/#{db_name}")
+
+    assert resp.status_code == 200
+    assert resp.body["db_name"] == "random+test+db2+#{random_number}"
+  end
+
+  @tag :with_db
+  test "'+' in document name should encode to '+'", context do
+    set_config({"chttpd", "decode_plus_to_space", "false"})
+
+    db_name = context[:db_name]
+    doc_id = "test+doc"
+    resp = Couch.put("/#{db_name}/#{doc_id}", body: %{})
+
+    assert resp.status_code == 201
+    assert resp.body["id"] == "test+doc"
+  end
+
+  @tag :with_db
+  test "'+' in document name should encode to space", context do
+    set_config({"chttpd", "decode_plus_to_space", "true"})
+
+    db_name = context[:db_name]
+    doc_id = "test+doc+2"
+    resp = Couch.put("/#{db_name}/#{doc_id}", body: %{})
+
+    assert resp.status_code == 201
+    assert resp.body["id"] == "test doc 2"
+  end
+
   @tag :with_db
   test "Empty database should have zero docs", context do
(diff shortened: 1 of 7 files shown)


test/elixir/test/basics_test.exs
```diff
@@ -58,6 +58,64 @@ defmodule BasicsTest do
     assert context[:db_name] in Couch.get("/_all_dbs").body, "Db name in _all_dbs"
   end
 
+  @tag :with_db
+  test "Database name with '+' should encode to '+'", _context do
+    set_config({"chttpd", "encode_plus", "false"})
+
+    random_number = :rand.uniform(16_000_000)
```
iilyak, line 65: You don't need  to use randomness here. The `with_db` annotation would create a random db  for you (see https://github.com/apache/couchdb/blob/main/test/elixir/lib/couch/db_test.ex#L27:L28)

test/elixir/test/basics_test.exs
```diff
@@ -58,6 +58,64 @@ defmodule BasicsTest do
...
+    db_name = "random+test+db+#{random_number}"
```
iilyak, line 66: the  database is created automatically and the name can be accessed as `db_name = context[:db_name]`.

iilyak: Hmm. Never mind. I see why you are doing it this way.

iilyak: I  think you should use either the following (rely on db  creation during setup) or remove `with_db` tag from your  solution.
```
@tag :with_db_name random_db_name("random+test+db")
test "Database name with '+' should encode to '+'",  context do
    set_config({"chttpd", "encode_plus", "false"})
    db_name = context[:db_name]
    resp = Couch.get("/#{db_name}")
    assert resp.body["db_name"] == db_name
end
```

test/elixir/test/basics_test.exs
```diff
@@ -58,6 +58,64 @@ defmodule BasicsTest do
...
+    assert resp.status_code == 201
+    assert resp.body["ok"] == true
+
+    resp = Couch.get("/#{db_name}")
+
+    assert resp.status_code == 200
+    assert resp.body["db_name"] == db_name
+  end
+
+  @tag :with_db
```
iilyak, line 78: Again with `@tag :with_db`  you are creating a database which you don't use in test.

test/elixir/test/basics_test.exs
noahshaw11: I have removed the `:with_db` tag from two of the tests.

test/elixir/test/basics_test.exs
noahshaw11: I have removed the `:with_db` tag from two of the tests.

test/elixir/test/basics_test.exs
noahshaw11: See above.

test/elixir/test/basics_test.exs
iilyak: The [`random_db_name/1`](https://github.com/apache/couchdb/blob/main/test/elixir/lib/couch/db_test.ex#L79) is injected into the  scope of each test module (the module which has `use CouchTestCase` stanza).

Therefore you can use `db_name = random_db_name("random+test+db")`. However I  don't have strong opinion about it.

test/elixir/test/basics_test.exs
```diff
@@ -58,6 +58,62 @@ defmodule BasicsTest do
     assert context[:db_name] in Couch.get("/_all_dbs").body, "Db name in _all_dbs"
   end
 
+  test "Database name with '+' should encode to '+'", _context do
```
iilyak, line 61: The tests present in the module do not show that encoding of path part and query part is independent. 
It would be better to add an additional check into each test case. It seems you can use `GET /{db}/_changes?filter=_doc_ids&doc_ids=["hello+world"]'`. Where used `doc_ids` would have `+` in there.

iilyak:
Nice job! Only test case coverage question left.

nickva (approved):
+1 well done

PR #3754, opened 2021-09-15
Fix limit0 for views again

rnewson:
## Overview

    Restrict the limit=0 clause to the sorted=false case as originally intended

    The limit=0 clause was introduced in commit 4e0c97bf which added
    sorted=false support. It accidentally matches when the user specifies
    limit=0 and causes us not to apply the logic that ensures we collect a
    {meta, Meta} message from each shard range and then send the
    total_rows and offset fields.

## Testing recommendations

repro steps described in https://github.com/apache/couchdb/issues/3750

https://github.com/apache/couchdb/issues/3750

## Checklist


nickva (approved):
+1

PR #3795, opened 2021-10-21
fix case_clause when HitId and DocId do not match when include_docs=true

tonysun83:
## Overview

Sometimes, a clouseau index may become out of sync with the db or
become corrupt under certain conditions (heap exhaustion, garbage collection).
This leads to certain HitIds that get returned which may not exist
in the current db. Before this fix, we assume the Hit's Id will
automatically match an Id returned from:

`
fabric:all_docs(DbName, fun callback/2, [], [{keys, DocIds}, {include_docs, true}]).
`

When the document does not exist, `{error, false}` is returned for the row.
This fix  accounts for this scenario and possible other scenarios
by returning an error to the user and also logging the error when the
Hit Id does not match what is returned from the _all_docs call.

Another option is to implement a read-repair solution but I don't think that it's possible without rebuilding the index.
Note that this issue surfaced when a user ran similar queries across several databases with similar documents and indexes. One of the queries resulted in a document id being returned in a Hit that the user claims belongs to another database. That caused the `case_clause`. A rebuild of the index resolved the issue temporarily but surfaced again as heap levels reached max capacity again.

## Testing recommendations
Unit test added for possible error scenarios. Current regression tests should pass.

## Checklist

### DIFF
--- a/src/dreyfus/src/dreyfus_httpd.erl
+++ b/src/dreyfus/src/dreyfus_httpd.erl
@@ -532,16 +532,18 @@ hits_to_json(DbName, IncludeDocs, Hits) ->
     if IncludeDocs ->
         chttpd_stats:incr_reads(length(Hits)),
         {ok, JsonDocs} = dreyfus_fabric:get_json_docs(DbName, Ids),
-        lists:zipwith(fun(Hit, {Id, Doc}) ->
+        lists:zipwith(fun
+            (Hit, {Id, Doc}) ->
                 case Hit of
-                    {Id, Order, Fields} ->
-                        {[{id, Id}, {order, Order}, {fields, {Fields}}, Doc]};
-                    {Id, Order, Fields, Highlights} ->
+                    {HitId, Order, Fields} ->
+                        Doc1 = hits_id_validator(HitId, {Id, Doc}),
+                        {[{id, Id}, {order, Order}, {fields, {Fields}}, Doc1]};
+                    {HitId, Order, Fields, Highlights} ->
+                        Doc1 = hits_id_validator(HitId, {Id, Doc}),
                         {[{id, Id}, {order, Order}, {fields, {Fields}},
-                          {highlights, {Highlights}}, Doc]}
+                          {highlights, {Highlights}}, Doc1]}
                 end
-            end, HitData, JsonDocs);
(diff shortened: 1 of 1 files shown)


tonysun83:
@rnewson , I'm unable to reproduce the issue for this particular user. However, they confirmed that the document id belongs to a different database that resides on the same cluster.

The code shows that the segment files are written in a directory path based simply on DbName + Hash:
https://github.com/apache/couchdb/blob/3.x/src/dreyfus/src/dreyfus_index.erl#L250. A

On the clouseau side, as you're fully aware, the PATH is mapped to PIDs which get returned to dreyfus:

https://github.com/cloudant-labs/clouseau/blob/master/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala#L44-L45

For the user behavior to manifest, one possibility is that the PID mapping is modified and the PID  writes to a different path. I don't see that occurring anywhere in the  Clouseau codebase. That doesn't exclude the possibility it's happening, but unfortunately, I cannot reproduce it.

It's a corner case and the user is getting a function_clause. I propose that instead of return results, that we do: 

> logging a 'fatal' error for the administrator and returning a 500 status code indicating the index is broken to the user.

rnewson:
@tonysun83 thanks for the details, this worries me. I can, sadly, imagine there might be bugs in the pid<>index path logic, either in dreyfus or clouseau or in combination.  Clearly, though, they must manifest rarely. Have you confirmed that the document from "another database" is genuinely only present in that database and is not (now, or in the past) been present in the expected database?

The fix in this PR remains the wrong approach, I think we both agree. As for returning a fatal/500 error, it's not entirely clear that's right either. For one thing the 500 would only manifest if a search query matched one of these misplaced documents, and not if they don't. If we have indexed documents into the wrong search index then the index is corrupt and no results can be trusted. So really the index should be deleted at that moment and forced to rebuild or, if we prefer, marked as corrupt and no longer allowed to serve requests until someone (the user? the administrator?) intervenes.

The true fix, I think, is to remove the pid<>path mapping entirely, changing the rpc messages to state the index path each time.

tonysun83:
closing this as the right approach requires a fairly large refactoring. Will use a different PR.

PR #3818, opened 2021-11-05
Re enable most elixir tests redux

kocolosk:
## Overview

This is essentially a rebase of #3175, updated to skip additional incompatible elixir tests for "couchapp" functionality that have been added in the interim.

There is a bug on `main` right now where the `_users` DB cannot be recreated once it has been deleted, and the cluster is rendered completely inoperable until CouchDB restarts. That's the reason for d8eea48 (and, I believe, for Paul's exclusion of other tests in that module). I've called that out separately so we can revert that commit once the fix is found.

If anyone is interested in consolidating the `elixir`, `elixir-only`, and `elixir-suite` into a single target I'm all ears. The first two are identical save for the `MIX_ENV` and `COUCHDB_TEST_ADMIN_PARTY_OVERRIDE` environment settings, so I'm guessing we only need one target. The last one uses a skip file instead of the `@pending` tag; I could certainly imagine going that route but stuck with @davisp 's original approach for this PR at this time.

## Testing recommendations

`make elixir`

For me this yields

> Finished in 86.2 seconds
440 tests, 0 failures, 136 excluded, 1 skipped

#3175

## Checklist


nickva (approved):
+1

Thank you for rebasing and cleaning up the PR

PR #3842, opened 2021-11-22
feat(couchjs): add support for SpiderMonkey 91esr

lostnet:
## Overview

This PR introduces spidermonkey 91esr. As with 78 and 86 there are no relevant API changes but it became necessary in practice as well as in theory to destroy the context and shutdown JS before any proper exit to prevent assertions in destructors. (For more information, the issue tracking the esr91 migration guide is [here](https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/issues/41.)

## Testing recommendations

Since many distributions don't ship with moz-js 91 yet, I've included a github workflow to test couchdb in ubuntu with the latest spidermonkey, foundationdb 6.x, and icu. It is set to be manually run on a branch from the actions menu.

## Checklist

- [X]  Any new configurable parameters are documented in `rel/overlay/etc/default.ini` [^1]
- [X] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation [^2]
[^1]: no configurable parameters for runtime
[^2]: spidermonkey versions are only referenced in whatsnew for release branches.

### DIFF
--- a/src/couch/priv/couch_js/86/util.cpp
+++ b/src/couch/priv/couch_js/86/util.cpp
@@ -330,7 +330,7 @@ void
 couch_oom(JSContext* cx, void* data)
 {
     fprintf(stderr, "out of memory\n");
-    exit(1);
+    _Exit(1);
 }
 
 
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -1,0 +1,92 @@
+name: Ubuntu build CI
+
+on:
+  workflow_dispatch:
+
+jobs:
+
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: update
+      run: sudo apt-get update
+    - name: install
+      run: sudo apt-get --no-install-recommends -y install autoconf2.13 build-essential pkg-config libcurl4-openssl-dev erlang-dev rebar elixir
+
+    - name: couch checkout
+      uses: actions/checkout@v2
+      with:
+        path: couch
+
+    - name: ICU checkout
+      uses: actions/checkout@v2
+      with:
+        repository: unicode-org/icu
+        path: icubuild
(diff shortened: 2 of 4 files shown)


lostnet:
I don't see how the erlfdb test failure relates to the PR. If this is from [fdb errors](https://apple.github.io/foundationdb/api-error-codes.html) it is `key_outside_legal_range` which apparently could make sense with a random key since some keys are special but it is highly unlikely to happen with multiple runs if they are getting different random seeds?

src/couch/priv/couch_js/86/util.cpp
```diff
@@ -330,7 +330,7 @@ void
 couch_oom(JSContext* cx, void* data)
 {
     fprintf(stderr, "out of memory\n");
-    exit(1);
+    _Exit(1);
```
kocolosk, line 333: Is this change something that ought to be backported to our wrappers for older versions of SpiderMonkey, or is it motivated by something specific in ESR91?

kocolosk (approved):
Thanks for this contribution! It seems like we might need to modify some project configuration for the Action to show up, but I don't want that to block the merge of this PR.

src/couch/priv/couch_js/86/util.cpp
lostnet: It is a little of both something that changed and a bug that existed.. Technically it was never correct to do a proper exit() without cleaning up the js context which we can't do in the middle of being called from the js context (and having a shortage of memory its risky to assume we can get a chance to exit later), but it was only in recent versions that mozilla added destructors that notice they are being called with a still active context.

Since it was a bug I saw no problem with keeping it combined and fixing 86 and 78 in the process, but none of the earlier versions should ever get patches again so it probably doesn't need to be backported to the earlier wrappers.

PR #3853, opened 2021-11-28
Remove the old ddoc_cache_opener process

kocolosk:
## Overview

As it turns out we've been bouncing eviction messages through it for years as part of a half-finished cluster upgrade strategy. I've split this PR into two commits in case anyone wanted to be fastidious about the removal and stop the flow of eviction messages to the process in a separate rollout before eliminating it from the supervision tree.

## Testing recommendations

Run the test suite.

## Checklist


nickva (approved):
A very nice cleanup!

+1

PR #3864, opened 2021-12-07
improve erlang_ls.config

rnewson:
This fixes some of the include issues.

### DIFF
--- a/erlang_ls.config
+++ b/erlang_ls.config
@@ -1,2 +1,5 @@
+apps_dirs:
+    - "src/*"
 include_dirs:
-    - "src/"
+    - "src"
+    - "src/*/include"


nickva (approved):

kocolosk:
What does the `apps_dirs` setting do?

nickva:
@kocolosk I think it's for "umbrella" apps,  like what we have with our top level applications https://erlang-ls.github.io/configuration/#rebar3-umbrella-project

But I am not entirely sure what the difference is between deps and apps is, though. They seems to be doing largely the same thing. 

With this setting go to definition, find references,  "times used" and such seems to work within emacs across apps 
Screen Shot 2021-12-07 at 6 17 48 PM

kocolosk:
Yes, understood. I was being too terse, sorry. I was wanting to understand if that setting was fixing some "include issues". For some reason when I added this config file to the `main` branch in #3343 I did use the correct list of `include_dirs`, but I hadn't found a reason to set `apps_dirs`.

+1 to disabling `suggest-spec`. I was having problems with `typer` crashing, and I've seen some other reports of that happening as well. I also found it distracting.

I might also suggest disabling the `function-references` lens. The "Used N times" link is broken right now, and per erlang-ls/erlang_ls#991 they are thinking of removing it anyway. At least in VS Code the "Show Call Hierarchy" function provides that level of detail and more in a nicer presentation.

On the other hand, we many not want to be _too_ prescriptive about putting this config file in source control. Users can and will want to customize it themselves. Worth further discussion, but it might make sense to maintain a good sample config file in git but leave it to each individual developer to decide whether to copy it into place / modify it further.

But for now, +1 from me.

kocolosk (approved):

rnewson:
without this change it wasn't possibly for erlang_ls to see/index all the exported functions in other apps, etc. There might be other combinations of settings that work too but this worked for me so I stopped trying.

PR #3881, opened 2021-12-20
Enforce row limit in mango

rnewson:
## Overview

We specify two config variables under query_server_config called partition_query_limit and query_limit. These are not enforced in the code since https://github.com/apache/couchdb/commit/b2f9cfaceb121bf391f25e5edc51ad75b71967b4 was merged in response to issue https://github.com/apache/couchdb/issues/2795.

The right fix is to move the enforcement to later in the mango execution sequence.

 https://github.com/apache/couchdb/issues/2795

## Checklist

### DIFF
--- a/src/couch_mrview/src/couch_mrview_util.erl
+++ b/src/couch_mrview/src/couch_mrview_util.erl
@@ -439,28 +439,33 @@ validate_args(#mrst{} = State, Args0) ->
             mrverror(Msg2)
     end.
 
-apply_limit(ViewPartitioned, Args) ->
-    Options = Args#mrargs.extra,
-    IgnorePQLimit = lists:keyfind(ignore_partition_query_limit, 1, Options),
+apply_limit(ViewPartitioned, #mrargs{} = Args) ->
     LimitType =
-        case {ViewPartitioned, IgnorePQLimit} of
-            {true, false} -> "partition_query_limit";
-            {true, _} -> "query_limit";
-            {false, _} -> "query_limit"
+        case ViewPartitioned of
+            true -> "partition_query_limit";
+            false -> "query_limit"
         end,
 
-    MaxLimit = config:get_integer(
-        "query_server_config",
-        LimitType,
-        ?MAX_VIEW_LIMIT
-    ),
+    IsMangoQuery = lists:keyfind(is_mango_query, 1, Args#mrargs.extra),
+    MaxLimit =
+        case IsMangoQuery of
+            {is_mango_query, true} ->
+                ?MAX_VIEW_LIMIT;
+            false ->
+                config:get_integer(
+                    "query_server_config",
+                    LimitType,
+                    ?MAX_VIEW_LIMIT
+                )
+        end,
 
     % Set the highest limit possible if a user has not
-    % specified a limit
+    % specified a limit or if this is a mango query
     Args1 =
-        case Args#mrargs.limit == ?MAX_VIEW_LIMIT of
-            true -> Args#mrargs{limit = MaxLimit};
-            false -> Args
+        case {Args#mrargs.limit == ?MAX_VIEW_LIMIT, IsMangoQuery} of
+            {_, {is_mango_query, true}} -> Args#mrargs{limit = ?MAX_VIEW_LIMIT};
+            {true, false} -> Args#mrargs{limit = MaxLimit};
+            {false, false} -> Args
         end,
 
     if
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -120,7 +120,7 @@ base_args(#cursor{index = Idx, selector = Selector} = Cursor) ->
         extra = [
             {callback, {?MODULE, view_cb}},
             {selector, Selector},
-            {ignore_partition_query_limit, true}
+            {is_mango_query, true}
         ]
     }.
 
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -40,7 +40,8 @@ create(Db, Indexes, Selector, Opts) ->
     Composited = composite_indexes(Indexes, FieldRanges),
     {Index, IndexRanges} = choose_best_index(Db, Composited),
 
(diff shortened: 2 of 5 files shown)


src/mango/src/mango_cursor_view.erl
```diff
@@ -40,7 +40,8 @@ create(Db, Indexes, Selector, Opts) ->
     Composited = composite_indexes(Indexes, FieldRanges),
     {Index, IndexRanges} = choose_best_index(Db, Composited),
 
-    Limit = couch_util:get_value(limit, Opts, mango_opts:default_limit()),
+    MRLimit = mango_opts:mr_limit(couch_db:is_partitioned(Db)),
+    Limit = erlang:min(MRLimit, couch_util:get_value(limit, Opts, mango_opts:default_limit())),
```
tonysun83, line 44: There was some discussion around what takes precedence:
1) `partition_query_limit`
2) `user defined limit/default limit`
The decision was that 2) would take precedence as @willholley mentioned in https://github.com/apache/couchdb/issues/2795

```
I would expect the partitioned query limit to only be enforced for user-specified limits rather than Mango internals (i.e. partitioned Mango index scans should **effectively be limited by the maximum partition size instead**).
```

If the new code is applied, I think we revert back to the previous behavior:

Given the following config:
````
[query_server_config]
partition_query_limit = 5
```
If the user ran a query that provided a `limit=10` and assuming 100 docs are matched, then he would get 5 documents back. I think what Will is saying is that this seems unintuitive. Rather we should get 10. 

Of course, this leads to the case where if the default:

```
default_limit() ->
    config:get_integer("mango", "default_limit", 10000000).
```
is really large, then we could scan and return the entire partition, but that's the behavior we expect.

rnewson: Hm, ok. There's a mixup here. The default limit if one is not specified is one thing, the _maximum_ limit that anyone is allowed to set is another. Which meaning does query_limit and partition_query_limit in [query_server_config] have? We might need more attributes.

The point of a maximum, that a client request cannot exceed, is a mechanism to control resources on the server side and to encourage sensible client patterns (like pagination).

rnewson: Also, Will's point is that the code prior to your change passed that upper limit to couch_mrview, which limits how many rows/docs can be examined, even if they don't match the selector, and so unless every doc matches your selector you would get less than you asked for even if your limit (implicit or explicit) was not exceeded.

My change corrects that. I preserve your 'ignore' logic at the couch_mrview level _if_ the request is from mango, though I rename it for clarity. Then the limit is capped on the mango side, so you can only get up to partition_query_limit / query_limit docs back for your _find response (even if many more docs were examined than that limit).

Queries directly to _view have the limit enforced on the couch_mrview side since there is no possibility of a non-matching row by definition there.

tonysun83: > Also, Will's point is that the code prior to your change passed that upper limit to couch_mrview, which limits how many rows/docs can be examined, even if they don't match the selector, and so unless every doc matches your selector you would get less than you asked for even if your limit (implicit or explicit) was not exceeded.
My change corrects that. I preserve your 'ignore' logic at the couch_mrview level if the request is from mango, though I rename it for clarity. Then the limit is capped on the mango side, so you can only get up to partition_query_limit / query_limit docs back for your _find response (even if many more docs were examined than that limit).

Ya, I forgot to mention this but agree that your new logic preserves the behavior on the `couch_mrview` side. Perhaps it's up to the DBA to notify their users that they are enforcing a particular maximum for the query side? The default is MAX_VIEW_LIMIT so out of the box, users won't run into this problem. It's only when the DBA enforces a tiny limit that users will perhaps see an issue.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -439,28 +439,30 @@ validate_args(#mrst{} = State, Args0) ->
...
     % Set the highest limit possible if a user has not
-    % specified a limit
+    % specified a limit or if this is a mango query
     Args1 =
-        case Args#mrargs.limit == ?MAX_VIEW_LIMIT of
-            true -> Args#mrargs{limit = MaxLimit};
-            false -> Args
+        case {Args#mrargs.limit == ?MAX_VIEW_LIMIT, IsMangoQuery} of
+            {_, {is_mango_query, true}} -> Args#mrargs{limit = ?MAX_VIEW_LIMIT};
+            {true, false} -> Args#mrargs{limit = MaxLimit};
```
tonysun83, line 464: line 463 and 464 seems redundant give lines 449-457. Maybe combine it somehow? But for clarity/readability, I think it's fine to leave it.

src/mango/src/mango_cursor_view.erl
tonysun83: If we're going to go with your change, you need to modify this test: https://github.com/apache/couchdb/blob/3.x/test/elixir/test/partition_mango_test.exs#L555-L594

src/couch_mrview/src/couch_mrview_util.erl
rnewson: ah, sorry, that is really for the `{false, {is_mango_query, true}}` case where we _don't_ want to pass the user-specified limit down to couch_mrview (and get the incorrect responses from before your change). it is (obviously) redundant to change limit to MAX_VIEW_LIMIT for the `{true, {is_mango_query, true}}` case since it already is, but it avoids writing a separate clause. Unless I've misunderstood this section of code? I have verified the behaviour with dev/run and it seemed to do the right thing.

src/mango/src/mango_cursor_view.erl
rnewson: ```My only concern is that a previous user who had limit=50 in their queries suddenly starts getting 25 rows back when the DBA suddenly sets partition_query_limit=25```

That would be exactly what the DBA wanted to happen by making that change, isn't it?

tonysun83:
jenkins requires `erlfmt` check to pass before it'll build. I think we added this enforcement recently:

https://github.com/apache/couchdb/commit/b78ccf18cb4ed6e183ed0bf4e5cbe40d7a7dc493

tonysun83:
+1

PR #3883, opened 2021-12-23
Feat/support sm91 3.x from lostnet

lostnet:
## Overview

  Supporting changes necessary to build/test spidermonkey 91esr on windows and linux/arm64 for the 3.x.

## Testing recommendations

  - Linux/AMD_64: normal CI Tests
  - Windows I am using the following:
https://github.com/lostnet/couchdb-glazier/blob/main/.github/workflows/ci.yml
  - Linux/Arm64: I previously [tested 68/78](https://github.com/lostnet/couchdb-nix-debug) and yocto seems to be using 60, hopefully with arm specific patches.

## Related Issues or Pull Requests

 - #3873
 - #3149
 - Passing the fopen'ed JS file to mozjs was fragile due to [MS Boundaries](https://docs.microsoft.com/en-us/cpp/c-runtime-library/potential-errors-passing-crt-objects-across-dll-boundaries?view=msvc-170). I have moved JS::CompileUtf8File to JS::CompileUtf8Path which implements the logic within mozjs, but will result in lumping failures to fopen with other compile errors that return error 1 instead of error 3.

## Checklist

- [X] Code is written and works correctly
- [X] Changes are covered by tests
- [ ] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation

Issue #3149, opened 2020-09-13
spidermonkey_68 identified as spidermonkey_60 and erroneously(?) blocked by configure on aarch64


lostnet:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

I created a couchdb3/spidermonkey_68 package on nixos, on x86_64 it works fine, but on aarch64 it errors out with the warning:
ERROR: SpiderMonkey 60 is known broken on ARM 64 (aarch64). Use 1.8.5 instead.
which comes from [couchdb/configure](https://github.com/apache/couchdb/blob/master/configure#L181)

## Steps to Reproduce

run configure --spidermonkey-version=68 on aarch64.

[NOTE]: # ( Include commands to reproduce, if possible. curl is preferred. )

## Expected Behaviour
If there is a problem with _68 this should be given in the configure and referenced in an issue, but it seems to block all versions except 1.8.5 and refer to them as _60 so it is impossible to tell if problems exist or the check is out of date.

I was under the impression that 68 was deemed fine on linux/aarch64 and I can only find references to problems building spidermonkey itself on windows/aarch64 in mozilla documentation, which is not relevant. Couchdb bugs seem to indicate that a  SEGV problem with _60 was resolved or no longer reproducable, so possibly the check should be removed entirely.

[NOTE]: # ( Tell us what you expected to happen. )

## Your Environment

* CouchDB version used: 3.1.0
* Browser name and version: N/A
* Operating system and version: nixos-unstable

## Additional Context

[TIP]:  # ( Add any other context about the problem here. )

wohali:
We've never formally tested libmozjs-68 on aarch64, and have a conservative stance on whitelisting dependencies like this under the "principle of least surprise."

To get this changed it would be best to run our test suite multiple times in a loop (`make check`) looking for any segmentation violation crashes of couchjs. If you want you can try overriding that line and running the test a few times, and let us know how it goes.

lostnet:
Thanks @wohali, so far I've pulled in libcurl and run the make javascript part of the tests a few times. The tests still running there instead of in elixir didn't have any problems, so I'll look at bringing in any other dependencies and getting a full make check running in the next few days.

lostnet:
[...]

PR #3884, opened 2022-01-05
Feat/support sm91 4.x from lostnet

lostnet:
## Overview

Changes to couchdb4 with spidermonkey 91 to function on windows and hopefully linux/arm64 (I do not yet have a build of a recent enough fdb in my distribution to test on arm).

## Testing recommendations

For Windows it is necessary to install foundationdb and then edit the \bin\couchdb.cmd to add foundationdb's bin folder, then fauxton verify should pass.
[Reference binaries](https://github.com/lostnet/couchdb-glazier/releases/tag/msdeps7)

#3883 #3149

## Checklist

### DIFF
--- a/src/couch/rebar.config.script
+++ b/src/couch/rebar.config.script
@@ -155,8 +155,13 @@ end.
         };
     {unix, _} when SMVsn == "91" ->
         {
-            "-DXP_UNIX -I/usr/include/mozjs-91 -I/usr/local/include/mozjs-91 -I/opt/homebrew/include/mozjs-91/ -std=c++17 -Wno-invalid-offsetof",
-            "-L/usr/local/lib -L /opt/homebrew/lib/ -std=c++17 -lmozjs-91 -lm"
+            "$CFLAGS -DXP_UNIX -I/usr/include/mozjs-91 -I/usr/local/include/mozjs-91 -I/opt/homebrew/include/mozjs-91/ -std=c++17 -Wno-invalid-offsetof",
+            "$LDFLAGS -L/usr/local/lib -L /opt/homebrew/lib/ -std=c++17 -lmozjs-91 -lm"
+        };
+    {win32, _} when SMVsn == "91" ->
+        {
+            "/std:c++17 /DXP_WIN",
+            "$LDFLAGS mozjs-91.lib"
         }
 end.
 
@@ -191,7 +196,7 @@ IcuDarwinEnv = [{"CFLAGS", "-DXP_UNIX -I/usr/local/opt/icu4c/include -I/opt/home
 IcuBsdEnv = [{"CFLAGS", "-DXP_UNIX -I/usr/local/include"},
              {"LDFLAGS", "-L/usr/local/lib"}].
 IcuWinEnv = [{"CFLAGS", "$DRV_CFLAGS /DXP_WIN"},
-             {"LDFLAGS", "icuin.lib icudt.lib icuuc.lib"}].
+             {"LDFLAGS", "$LDFLAGS icuin.lib icudt.lib icuuc.lib"}].
 
 ComparePath = "priv/couch_ejson_compare.so".
 CompareSrc = ["priv/couch_ejson_compare/*.c"].
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -23,7 +23,7 @@ jobs:
       uses: actions/cache@v2
       id: wgetcache
       with:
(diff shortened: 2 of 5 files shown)


src/couch/rebar.config.script
```diff
@@ -155,8 +155,13 @@ end.
         };
     {unix, _} when SMVsn == "91" ->
         {
-            "-DXP_UNIX -I/usr/include/mozjs-91 -I/usr/local/include/mozjs-91 -I/opt/homebrew/include/mozjs-91/ -std=c++17 -Wno-invalid-offsetof",
-            "-L/usr/local/lib -L /opt/homebrew/lib/ -std=c++17 -lmozjs-91 -lm"
+            "$CFLAGS -DXP_UNIX -I/usr/include/mozjs-91 -I/usr/local/include/mozjs-91 -I/opt/homebrew/include/mozjs-91/ -std=c++17 -Wno-invalid-offsetof",
```
kocolosk, line 158: Two questions:

1. Inheriting existing flags makes sense. Is there a good reason to only start doing this with SM91? Alternatively we could make this change for all versions by doing the inheritance in the construction of the `CouchJSEnv` term below.
2. This copies $CFLAGS into $CXXFLAGS, would it be better to inherit $CXXFLAGS?

src/couch/rebar.config.script
```diff
@@ -191,7 +196,7 @@ IcuDarwinEnv = [{"CFLAGS", "-DXP_UNIX -I/usr/local/opt/icu4c/include -I/opt/home
 IcuBsdEnv = [{"CFLAGS", "-DXP_UNIX -I/usr/local/include"},
              {"LDFLAGS", "-L/usr/local/lib"}].
 IcuWinEnv = [{"CFLAGS", "$DRV_CFLAGS /DXP_WIN"},
-             {"LDFLAGS", "icuin.lib icudt.lib icuuc.lib"}].
+             {"LDFLAGS", "$LDFLAGS icuin.lib icudt.lib icuuc.lib"}].
```
kocolosk, line 199: Should this one be inheriting `$DRV_LDFLAGS` for consistency?

kocolosk (commented):
Looks good, just a few questions about the env inheritance stuff

src/couch/rebar.config.script
lostnet: (1) My thinking was that the standard build environment variables inheritance could be in parity between the OSes on 91 since linux meta distributions and builders often support setting up includes/libraries/hardening flags/etc in them. But with windows I think the %INCLUDE% variable is bypassing eports setup entirely in place of needing $CFLAGS inheritance. The separation of 91 vs previous versions was that anyone who setup an earlier build may have done so while ignoring bad CFLAGS since they had no consequences while any completely new setup is hopefully (ideally) able to use the only mozilla maintained version.

(2) My main concern for inheriting build system variables is that the I think any alternatives to "-std=c++17" are possibly harmful and I think (hope) most systems are supposed to be adding up those settings with CFLAGS to make CXXFLAGS.

src/couch/rebar.config.script
lostnet: Here I think the $DRV_CFLAGS is a now harmless redundancy (AFAICT the template for a C in a DRV context be combining $CFLAGS $DRV_CFLAGS again so it is equivalent to $DRV_CFLAGS $CFLAGS $DRV_CFLAGS) rather than achieving the same thing as $CFLAGS would, my guesses for its existence are:
  - could fix something in terms of order
  - was maybe important in some version of rebar2? 

I think this comes back to comment 1-(1) and whether $CFLAGS support should be added in windows for equivalence even though %INCLUDE% covers everything currently being set.

lostnet: I should probably note that the 3 times an EXE_FLAGS is used in compiles it is generating warnings consistent with using this default [EXE_FLAGS](https://github.com/blt/port_compiler/blob/master/src/pc_port_env.erl#L240) while [win32 has DRV_CFLAGS overrides by default](https://github.com/blt/port_compiler/blob/master/src/pc_port_env.erl#L305) which makes a clean compile, that's how I surmised that I don't want to override rebar's DRV_CFLAGS.

src/couch/rebar.config.script
kocolosk: OK got it, thanks for the explanation.

src/couch/rebar.config.script
kocolosk: Thanks Will.

PR #3891, opened 2022-01-12
Expose `decode/4` to skip decoding steps

jaydoane:
## Overview

Currently, `decode/3` performs various checks on a JWT, and then
base64 decodes and finally JSON decodes the token. However, in some
cases, it's desirable to skip the decoding steps, and just return the
token payload in binary form.

This exposes `decode/4` where the 4th argument is a decoder fun that
defaults to `decode_b64url_json/1` for `decode/3` to retain existing
behavior, but also exposes `decode_passthrough/1` in case a client
wants to avoid any decoding steps.

## Testing recommendations
```
make check
```

## Checklist

### DIFF
--- a/src/jwtf/src/jwtf.erl
+++ b/src/jwtf/src/jwtf.erl
@@ -20,6 +20,9 @@
 -export([
     encode/3,
     decode/3,
+    decode/4,
+    decode_b64url_json/1,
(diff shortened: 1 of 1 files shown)


nickva (approved):
+1

PR #3895, opened 2022-01-13
Use Debian Stable, add Erlang 24 to CI

kocolosk:
## Overview

This PR switches away from using a single container image with multiple Erlang versions installed via kerl and instead uses specific images for each version of Erlang against which we want to test a PR.

It also relies on tags like '20', '21' ... '24' instead of specific Erlang versions, so we can push updated images with new patch releases without needing to also file a PR to modify the Jenkinsfile configuration. The "Update Docker Containers" Jenkins job may need a small tweak to ensure that we can pick up the updated images behind those tags.

The matrix we are testing PRs against after this change looks like

| Erlang | OS | Elixir | JS |
|-------|-------|-----| ------ |
| 20.3.8.26 | Debian 10 (Buster) | v1.9.4 | 60 |
| 21.3.8.24 | Debian 11 (Bullseye) | v1.11.4 | 78 |
| 22.3.4.24 | Debian 11 (Bullseye) | v1.12.3 | 78 | 
| 23.3.4.10 | Debian 11 (Bullseye) | v1.12.3 | 78 |
| 24.2 | Debian 11 (Bullseye) | v1.12.3 | 78 |

I _could_ hack in better support for Erlang 20 on Debian 11 but didn't bother. I encountered linker errors out of the box on Erlang 20 with gcc 10, and figured this was just one more reason to move towards dropping support for it rather than hacking in conditional usage of gcc 9 and/or putting in special linker flags (`-fcommon` probably would have done it).

## Testing recommendations

Jenkins

### DIFF
--- a/build-aux/Jenkinsfile.pr
+++ b/build-aux/Jenkinsfile.pr
@@ -20,7 +20,6 @@ mkdir build
 cd build
 tar -xf ${WORKSPACE}/apache-couchdb-*.tar.gz
 cd apache-couchdb-*
-. /usr/local/kerl/${ERLANG_VERSION}/activate
 ./configure
 make check || (make build-report && false)
 '''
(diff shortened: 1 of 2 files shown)


kocolosk:
I like the direction that this is headed. I was probably too aggressive on upgrading to Elixir v1.13.2 (released today!) in the Erlang 22/23/24 images -- the `credo` static analysis tool is not yet compatible with the 1.13 series at all. Will try 1.12 locally and see how that goes.

kocolosk:
OK, dropping back to Elixir 1.12 helped out. I will file a PR for `couchdb-ci` with the changes I made over there. Technically credo says full Elixir 1.12 compatibility landed in v1.5.6 so I guess I'll bump that one as well, but I think this is ready for a review if anyone wants to take a look.

PR #3900, opened 2022-01-15
Bump Erlang and Debian versions in Full Platform CI

kocolosk:
## Overview

* Parameterize Erlang version selection, bump to 24.2
* Use Debian 11 (Stable) instead of Debian 10 (Old Stable) as base image

## Testing recommendations

I'm hoping the use of a `jenkins-*` branch will execute this pipeline automatically. If memory serves that's how it used to work.

### DIFF
--- a/build-aux/Jenkinsfile.full
+++ b/build-aux/Jenkinsfile.full
@@ -288,51 +288,6 @@ pipeline {
           } // post
         } // stage
 
-        stage('Ubuntu Xenial') {
-          agent {
-            docker {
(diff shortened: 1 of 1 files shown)


kocolosk:
Will close this PR for now while I work on the tests for the Full Platform Build changes in the `jenkins-*` branch. Using an open PR just doubles the work that Jenkins needs to do since it runs the PR suite as well, which is wasteful since I'm not actually making any changes on that front.

PR #3905, opened 2022-01-19
Improve CI pipeline for PRs

kocolosk:
## Overview

This PR has a small improvement to the Elixir test suite, and adds the `credo` style check earlier in the pipeline. It also reconfigures the per-stage timeouts so that they ignore time spent in the build queue waiting for a docker-based agent to free up.

Finally, it refactors the pipeline to use a larger number of explicit steps in each stage. This improves the developer experience navigating the Jenkins UI, as we can see timings for each phase of the build, and if a step fails the logs that are displayed will be scoped just to that step.

## Testing recommendations

Jenkins

## Related PRs

#3895 was the first PR to convert the Jenkinsfile to use specific images for each Erlang version.

### DIFF
--- a/build-aux/Jenkinsfile.pr
+++ b/build-aux/Jenkinsfile.pr
@@ -12,17 +12,6 @@
 // WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 // License for the specific language governing permissions and limitations under
 // the License.
-build_and_test = '''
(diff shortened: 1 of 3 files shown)


nickva (approved):
+1 once CI passes

kocolosk:
Went ahead and cherry-picked from #3903 to get this moving.

PR #3910, opened 2022-01-24
Fix corner case for GET `_dbs_info`

jiahuili430:
## Overview
If a design document exists in shards db, GET `_dbs_info` will throw a `time out` error. 

## Testing recommendations
`make eunit apps=chttpd suites=chttpd_dbs_info_test`

https://github.com/apache/couchdb/pull/3832

## Checklist


nickva (approved):
Good find! Indeed we had removed the auto-inserted _design document not too long and replaced with a BDU, but that doesn't mean users (operators most likely) cannot insert their own _design docs in the shards db.

PR #3914, opened 2022-01-26
Random bits of cleanup

kocolosk:
## Overview

I had forgotten to update `mix.lock` on a recent change to include support for Elixir v1.12, so that's included here. I also removed a few scripts that are leftovers from old build systems.


kocolosk:
Hah, looks like the build system expects a `bin` directory to exist, but I just removed all the version-controlled files from that directory. Will see if it makes sense to `mkdir -p` there.

nickva (approved):
+1 

`mkdir -p` before copying `rebar` or other files or a `.keep` file in to handle `bin` would work, whichever looks better overall.

kocolosk:
Good thought on the .keep file, I added that for now. I feel like there's still some duplication and overlap between the `bin/`, `build-aux/` and `support/`, but not going to try to resolve that here so taking the most conservative approach.

nickva (approved):

PR #3932, opened 2022-02-17
Autodetect spidermonkey version in ./configure

frapa:
Hi, this is my first contribution here!

I have struggled with this point when building couchDB, because it's written nowhere in the documentation that you need to configure the version properly.

It seems that other people had the problem as well: https://couchdb.slack.com/archives/C49LEE7NW/p1632249155263400

The updated configure script automatically finds out the installed library version. It also introduces a dependency on `sed`, `ldconfig` and `head` which I hope are fine (will they work on MacOS?), and remove the default to version 1.8.5 which is old and not in most distributions anymore.

The argument was also removed from the CI scripts because it does not exist anymore.


kocolosk:
Thanks! I've been wanting to do something like this ever since I refactored those CI config files and found myself littering them with `--spidermonkey-version` flags everywhere. How do you feel about still allowing the use of `--spidermonkey-version`, and just using the autodetected version as a default if the user doesn't specify one?

It seems we have an unrelated problem with Jenkins today, we'll investigate that separately.

Also, Trento? That takes me back. I spent a few weeks at ECT when I was a graduate student years ago. Lovely place :)

build-aux/Jenkinsfile.full
```diff
@@ -278,7 +267,7 @@ pipeline {
       steps {
         timeout(time: 15, unit: "MINUTES") {
           sh (script: 'rm -rf apache-couchdb-*', label: 'Clean workspace of any previous release artifacts' )
-          sh "./configure --spidermonkey-version ${spidermonkey}"
+          sh "./configure"
```
kocolosk, line 270: I think you can also remove the `environment` block above this; it was only there to supply `${spidermonkey}` for this command.

configure
```diff
@@ -29,9 +29,14 @@ ERLANG_MD5="false"
...
 COUCHDB_USER="$(whoami 2>/dev/null || echo couchdb)"
-SM_VSN=${SM_VSN:-"1.8.5"}
 ARCH="$(uname -m)"
 
+# Try to detect spidermonkey (libmozjs) version automatically
+SM_VSN=${SM_VSN:-$(/sbin/ldconfig -p | sed -n 's/.libmozjs-*\([0-9]*\).*/\1/p' | head -n 1)}
+if [ "$SM_VSN" = "185" ]; then
+  SM_VSN="1.8.5"
+fi
+
```
kocolosk, line 39: We need this to fail gracefully if it can't find SpiderMonkey. Currently I'm finding that `./configure` will fail with a line like

> Unsupported SpiderMonkey version: 

which comes from `src/couch/rebar.config.script`. I don't have a specific suggestion of where to put the improvement but I think it would be good for the system to tell the user that it could not find a SpiderMonkey library  and bail a little more clearly

kocolosk:
> In that case, should the CI script still be updated?

Yes, I like what you did there.

build-aux/Jenkinsfile.full
frapa: Great catch, removed.

configure
frapa: If the version cannot be detected, the variable `SM_VSN` will be empty. I added an `elif` case exiting the configure script and warning the user. I hope that's good enough.

frapa:
@kocolosk I reintroduced the `--spidermonkey-version`, added a check for empty `SM_VSN` and removed the environment block. Let me know if can do anything else.

frapa:
@kocolosk Any news on this PR?

configure
kocolosk: Yep, looks good except for the typos ("SpiderMonkey", "explicitly")

kocolosk:
@frapa I think this is looking good modulo a small typo or two in the error message. I pushed it to a specially-named `jenkins-`branch so we can test your changes to `Jenkinsfile.full` in CI:

https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FFullPlatformMatrix/detail/jenkins-frapa-autodetect_spidermonkey_version/1/pipeline

Assuming all goes well there we can fix the typo and merge.

kocolosk:
Looks like this fails on macOS and FreeBSD. FreeBSD doesn't have the `-p` option to `ldconfig`, while macOS doesn't use it at all.

I'd be OK to make the `ldconfig` dance conditional on discovering Linux as the underlying platform if you want to go that route, otherwise maybe you can figure out a way to support this discovery for the other Unix-like systems?

frapa:
Thanks for running the CI. I made it Linux-only because I have no access to freeBSD or macOS to test alternatives. Could you update the branch so we see if the CI passes everywhere?

kocolosk:
OK that's fine. Can you add the `spidermonkey_vsn` back into the Jenkinsfile for the macOS and FreeBSD builds and add the flag back in `generateNativeStage`? Otherwise macOS is going to fail because the 1.8.5 default is incorrect.

configure
```diff
@@ -29,11 +29,22 @@ ERLANG_MD5="false"
...
+if [ "$SM_VSN" = "185" ]; then
+  SM_VSN="1.8.5"
+fi
+
 . ${rootdir}/version.mk
-COUCHDB_VERSION=${vsn_major}.${vsn_minor}.${vsn_patch}
+COUCHDB_VERSION=${vsn_major}.${vsn_minor}.${vsn_patch}src/couch/rebar.config.script
```
kocolosk, line 47: Just noticed this, it's not supposed to be there is it?

frapa:
I think this MR should now be finished.

big-r81:
Hi Francesco,

would you like to update and rebase your PR against the current main branch to easier review it?

PR #3987, opened 2022-04-06
Fix busy functions to work with pids

noahshaw11:
## Overview

Fix `busy` functions to work with pids.

## Checklist

### DIFF
--- a/src/couch/src/couch_debug.erl
+++ b/src/couch/src/couch_debug.erl
@@ -25,6 +25,7 @@
 
 -export([
     process_name/1,
+    get_pid/1,
     link_tree/1,
     link_tree/2,
     mapfold_tree/3,
@@ -49,6 +50,7 @@
 -type throw(_Reason) :: no_return().
 
 -type process_name() :: atom().
+-type process() :: process_name() | pid().
 -type function_name() :: atom().
 -type busy_properties() ::
     heap_size
@@ -66,6 +68,7 @@ help() ->
         opened_files_by_regexp,
         opened_files_contains,
         process_name,
+        get_pid,
         link_tree,
         mapfold,
         map,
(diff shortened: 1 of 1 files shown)


src/couch/src/couch_debug.erl
```diff
@@ -281,18 +285,20 @@ help(Unknown) ->
     io:format("    ---~n", []),
     ok.
 
--spec busy(ProcessList :: [process_name()], Threshold :: pos_integer()) ->
+-spec busy([process_name()] | pid(), Threshold :: pos_integer()) ->
```
iilyak, line 288: This is not the intend. The intend is to make the following spec possible.

```
-spec busy([process_name() | pid()], Threshold :: pos_integer()) ->
    [Process :: process_name() | pid()].
```

And to make it clearer and allow re-usability of a type we can define it as.

```
-type process_name() :: atom().
-type process() :: process_name() | pid().
....

-spec busy([process()], Threshold :: pos_integer()) ->
    [Process :: process()].
```

src/couch/src/couch_debug.erl
```diff
@@ -134,6 +137,15 @@ help(process_name) ->
     - '$initial_call' key in process dictionary
     - process_info(Pid, initial_call)
 
+    ---
+    ", []);
+help(get_pid) ->
+    io:format("
+    get_pid(Process)
```
iilyak, line 144: ```get_pid(PidOrName)```

src/couch/src/couch_debug.erl
```diff
@@ -134,6 +137,15 @@ help(process_name) ->
...
+    -----------------
+
+    Get the pid for a process name.
```
iilyak, line 147: I think we need to elaborate a bit more about help message.

Something  like.

```
Get the pid for a process name given either name or `pid`. When `pid` is given it returns it as is. In a sense it is just a safer alternative to `whereis/1` which wouldn't crash when `pid` is given.
```

iilyak (approved):
+1

PR #4037, opened 2022-05-24
Fix jwtf support for ES* algorithms

rnewson:
ES256 support is apparently broken. Here's a test to prove it. I don't know the fix, though. I tried the same example (from jwt.io) with another library (https://github.com/G-Corp/jwerl) and it also fails. Is... Is erlang the problem?

### DIFF
--- a/src/jwtf/src/jwtf.erl
+++ b/src/jwtf/src/jwtf.erl
@@ -27,6 +27,8 @@
     verification_algorithm/1
 ]).
 
+-include_lib("public_key/include/public_key.hrl").
+
 -define(ALGS, [
     % RSA PKCS#1 signature with SHA-256
     {<<"RS256">>, {public_key, sha256}},
@@ -70,7 +72,13 @@ encode(Header = {HeaderProps}, Claims, Key) ->
         SignatureOrMac =
             case verification_algorithm(Alg) of
                 {public_key, Algorithm} ->
-                    public_key:sign(Message, Algorithm, Key);
+                    Signature = public_key:sign(Message, Algorithm, Key),
+                    case Alg of
+                        <<"ES", _/binary>> ->
+                            der_to_jose(Alg, Signature);
+                        _ ->
+                            Signature
+                    end;
                 {hmac, Algorithm} ->
                     hmac(Algorithm, Key, Message)
             end,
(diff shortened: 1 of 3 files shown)


rnewson:
nope. turns out both jwerl and jwtf are wrong, the `ES*` signatures have a different format and need converting to and from. 

Except jwtf is now fixed.

src/jwtf/src/jwtf.erl
```diff
@@ -70,7 +72,13 @@ encode(Header = {HeaderProps}, Claims, Key) ->
...
             case verification_algorithm(Alg) of
                 {public_key, Algorithm} ->
-                    public_key:sign(Message, Algorithm, Key);
+                    Signature = public_key:sign(Message, Algorithm, Key),
+                    case Alg of
+                        <<"ES", _/binary>> ->
+                            der_to_jose(Alg, Signature);
+                        _ ->
+                            Signature
+                    end;
```
jaydoane, line 81: Can we factor this into a separate function and use it on lines 279-285 as well?

src/jwtf/src/jwtf.erl
```diff
@@ -292,6 +307,24 @@ hmac_verify(Algorithm, Message, HMAC, SecretKey) ->
...
+    #'ECDSA-Sig-Value'{r = R, s = S} = public_key:der_decode('ECDSA-Sig-Value', Signature),
+    Len = rs_len(Alg),
+    <<R:Len, S:Len>>.
+
+rs_len(<<"ES256">>) ->
+    256;
+rs_len(<<"ES384">>) ->
+    384;
+rs_len(<<"ES512">>) ->
+    512.
```
jaydoane, line 326: Is there a reason to prefer this implementation over e.g.
```erlang
rs_len(<<"ES", SizeBin/binary>>) ->
    binary_to_integer(SizeBin).

```

jaydoane (approved):

src/jwtf/src/jwtf.erl
rnewson: I like your version better.

src/jwtf/src/jwtf.erl
rnewson: I'm not following. lines 279-285 do the inverse operation. what are you suggesting?

src/jwtf/src/jwtf.erl
```diff
@@ -268,12 +276,19 @@ key(Props, Checks, KS) ->
 verify(Alg, Header, Payload, SignatureOrMac0, Key) ->
     Message = <<Header/binary, $., Payload/binary>>,
     SignatureOrMac1 = b64url:decode(SignatureOrMac0),
+    SignatureOrMac2 =
+        case Alg of
+            <<"ES", _/binary>> ->
+                jose_to_der(SignatureOrMac1);
+            _ ->
+                SignatureOrMac1
+        end,
```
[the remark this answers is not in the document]
rnewson: ?

src/jwtf/src/jwtf.erl
jaydoane: My bad, I failed to notice this one uses `der_to_jose` while the other `jose_to_der`. All I can say is "derp derp derp"

Issue #3758, opened 2021-09-19
JWT auth doesn't pick up roles


loylick:
## Description
Hi! I've encountered an issue when using JWT authentication in couchdb. I had to compile Couchdb from sources to make JWT authentication work. I use the branch 3.x. First I tried to pass roles to couchdb through the custom claim _couchdb.roles. When I failed I looked at recent issues resolved and found that there was a new configuration field "roles_claim_name" in [jwt_auth] section. So, I added a line  roles_claim_name = couchdbroles into [jwt_auth] section of local.ini file. The authentication succeeds, the username is passed correctly from the "sub" claim, but the roles are not passed from JWT token. 

## Steps to Reproduce
I used such request to get authenticated:
curl -L -X GET 'http://127.0.0.1:5984/_session' \
-H 'Content-Type: application/json; charset=utf-8' \
[...]

PR #3166, opened 2020-09-21
feat(auth): Allow a custom JWT claim for roles

rnewson:

PR #4041, opened 2022-05-27
Allow and evaluate nested json claim roles

big-r81:
## Overview

In a JWT token it is possible to add an attribute for role claims. If the roles are presented as top-level attribute like
```json
{
  "couchdb-roles": [
    "my_role_1",
    "my_role_2"
  ]
}
```
and setting the parameter `roles_claim_name` in the config file to 
```
[jwt_auth]
roles_claim_name = couchdb-roles
```
CouchDB was able to read that attributed and take over that roles.

This doesn't work, if the claim roles are nested, eg:
```
{
  "my" :{
    "nested": {
       "couchdb-roles": [
         "my_role_1",
         "my_role_2"
       ]
    }
  }
}
```
To allow this and for backwards compatibility, a new config parameter `roles_claim_path` is introduced to allow nested role claims. To allow the example from above, yo can use the following syntax:
```
[jwt_auth]
roles_claim_path = my.nested.couchdb-roles
```
If `roles_claim_path` is defined, then `roles_claim_name` is ignored.

## Testing recommendations

Example JWT token:
```
eyJ0eXAiOiJKV1QiLCJraWQiOiJteWp3dHRlc3RrZXkiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlhdCI6MTY1NTI5NTgxMCwiZXhwIjoxNzU1Mjk5NDEwLCJteSI6eyJuZXN0ZWQiOnsiX2NvdWNoZGIucm9sZXMiOlsibXlfbmVzdGVkX2NvdWNoZGIucm9sZXNfMSIsIm15X25lc3RlZF9jb3VjaGRiLnJvbGVzXzEiXX19LCJfY291Y2hkYi5yb2xlcyI6WyJfY291Y2hkYi5yb2xlc18xIiwiX2NvdWNoZGIucm9sZXNfMiJdLCJteS5fY291Y2hkYi5yb2xlcyI6WyJteS5fY291Y2hkYi5yb2xlc18xIiwibXkuX2NvdWNoZGIucm9sZXNfMiJdLCJmb28iOnsiYmFyLnpvbmsiOnsiYmF6LmJ1dSI6eyJiYWEiOnsiYmFhLmJlZSI6eyJyb2xlcyI6WyJteV9uZXN0ZWRfcm9sZV8xIiwibXlfbmVzdGVkX3JvbGVfMiJdfX19fX19.F6kQK-FK0z1kP01bTyw-moXfy2klWfubgF7x7Xitd-0
```

Example config file:
```
[chttpd]
authentication_handlers = {chttpd_auth, jwt_authentication_handler}, {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, default_authentication_handler}

[jwt_auth]
roles_claim_path = foo.bar\.zonk.baz\.buu.baa.baa\.bee.roles

[jwt_keys]
hmac:myjwttestkey = NTNv7j0TuYARvmNMmWXo6fKvM4o6nv/aUi9ryX38ZH+L1bkrnD1ObOQ8JAUmHCBq7Iy7otZcyAagBLHVKvvYaIpmMuxmARQ97jUVG16Jkpkp1wXOPsrF9zwew6TpczyHkHgX5EuLg2MeBuiT/qJACs1J0apruOOJCg/gOtkjB4c=
```

#3758 
#3166 

## Checklist

- [x] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation, see https://github.com/apache/couchdb-documentation/pull/737

### DIFF
--- a/src/couch/src/couch_httpd_auth.erl
+++ b/src/couch/src/couch_httpd_auth.erl
@@ -227,6 +227,7 @@ jwt_authentication_handler(Req) ->
             RequiredClaims = get_configured_claims(),
             case jwtf:decode(?l2b(Jwt), [alg | RequiredClaims], fun jwtf_keystore:get/2) of
                 {ok, {Claims}} ->
+                    Roles = get_roles_claim(Claims),
                     case lists:keyfind(<<"sub">>, 1, Claims) of
                         false ->
                             throw({unauthorized, <<"Token missing sub claim.">>});
@@ -234,15 +235,7 @@ jwt_authentication_handler(Req) ->
                             Req#httpd{
                                 user_ctx = #user_ctx{
                                     name = User,
-                                    roles = couch_util:get_value(
-                                        ?l2b(
-                                            config:get(
-                                                "jwt_auth", "roles_claim_name", "_couchdb.roles"
-                                            )
-                                        ),
-                                        Claims,
-                                        []
-                                    )
+                                    roles = Roles
                                 }
                             }
                     end;
@@ -251,6 +244,53 @@ jwt_authentication_handler(Req) ->
             end;
         _ ->
             Req
+    end.
+
+tokenize_json_path(Path, SliceStart, [], Result) ->
+    Result1 = Result ++ [?l2b(string:slice(Path, SliceStart))],
+    [?l2b(string:replace(X, "\\.", ".", all)) || X <- Result1];
+tokenize_json_path(Path, SliceStart, [[{Pos, _}] | T], Result) ->
+    Slice = string:slice(Path, SliceStart, Pos - SliceStart),
+    NewResult = Result ++ [?l2b(Slice)],
+    tokenize_json_path(Path, Pos + 1, T, NewResult).
+
+tokenize_json_path(Path, SplitPositions) ->
+    tokenize_json_path(Path, 0, SplitPositions, []).
+
+get_roles_claim(Claims) ->
+    RolesClaimPath = config:get(
+        "jwt_auth", "roles_claim_path"
+    ),
+    Result =
+        case RolesClaimPath of
+            undefined ->
+                couch_util:get_value(
+                    ?l2b(
+                        config:get(
+                            "jwt_auth", "roles_claim_name", "_couchdb.roles"
+                        )
+                    ),
+                    Claims,
+                    []
+                );
+            Defined when is_list(Defined) ->
+                % find all "." but no "\."
+                PathRegex = "(?<!\\\\)\\.",
+                MatchPositions =
+                    case re:run(RolesClaimPath, PathRegex, [global]) of
+                        nomatch -> [];
+                        {match, Pos} -> Pos
+                    end,
+                TokenizedJsonPath = tokenize_json_path(RolesClaimPath, MatchPositions),
+                couch_util:get_nested_json_value({Claims}, TokenizedJsonPath)
+        end,
+    case lists:all(fun erlang:is_binary/1, Result) of
+        true ->
+            Result;
+        false ->
+            throw(
+                {bad_request, <<"Malformed token">>}
+            )
     end.
 
 get_configured_claims() ->
--- a/test/elixir/test/jwt_roles_claim_test.exs
+++ b/test/elixir/test/jwt_roles_claim_test.exs
@@ -1,0 +1,167 @@
+defmodule JwtRolesClaimTest do
+  use CouchTestCase
+
+  @global_server_config [
+    %{
+      :section => "chttpd",
+      :key => "authentication_handlers",
+      :value => [
+                  "{chttpd_auth, jwt_authentication_handler}, ",
+                  "{chttpd_auth, cookie_authentication_handler}, ",
+                  "{chttpd_auth, default_authentication_handler})"
+                ] |> Enum.join
+    },
+    %{
+      :section => "jwt_keys",
+      :key => "hmac:myjwttestkey",
+      :value => ~w(
+        NTNv7j0TuYARvmNMmWXo6fKvM4o6nv/aUi9ryX38ZH+L1bkrnD1ObOQ8JAUmHCBq7
+        Iy7otZcyAagBLHVKvvYaIpmMuxmARQ97jUVG16Jkpkp1wXOPsrF9zwew6TpczyH
+        kHgX5EuLg2MeBuiT/qJACs1J0apruOOJCg/gOtkjB4c=) |> Enum.join()
+    }
+  ]
+
+  test "case: roles_claim_name (undefined) / roles_claim_path (undefined)" do
+    server_config = @global_server_config
+
+    run_on_modified_server(server_config, fn ->
+      test_roles(["_couchdb.roles_1", "_couchdb.roles_2"])
+    end)
+  end
+
+  test "case: roles_claim_name (defined) / roles_claim_path (undefined)" do
+    server_config =
+      [
+        %{
+          :section => "jwt_auth",
+          :key => "roles_claim_name",
+          :value => "my._couchdb.roles"
+        }
+      ] ++ @global_server_config
+
+    run_on_modified_server(server_config, fn ->
+      test_roles(["my._couchdb.roles_1", "my._couchdb.roles_2"])
+    end)
+  end
+
+  test "case: roles_claim_name (undefined) / roles_claim_path (defined)" do
+    server_config =
+      [
+        %{
+          :section => "jwt_auth",
+          :key => "roles_claim_path",
+          :value => "foo.bar\\.zonk.baz\\.buu.baa.baa\\.bee.roles"
+        }
+      ] ++ @global_server_config
+
+    run_on_modified_server(server_config, fn ->
+      test_roles(["my_nested_role_1", "my_nested_role_2"])
+    end)
+  end
+
+  test "case: roles_claim_name (defined) / roles_claim_path (defined)" do
+    server_config =
+      [
+        %{
+          :section => "jwt_auth",
+          :key => "roles_claim_name",
+          :value => "my._couchdb.roles"
+        },
+        %{
+          :section => "jwt_auth",
+          :key => "roles_claim_path",
+          :value => "foo.bar\\.zonk.baz\\.buu.baa.baa\\.bee.roles"
+        }
+      ] ++ @global_server_config
+
+    run_on_modified_server(server_config, fn ->
+      test_roles(["my_nested_role_1", "my_nested_role_2"])
+    end)
+  end
+
+  test "case: roles_claim_path with bad input" do
+    server_config =
+      [
+        %{
+          :section => "jwt_auth",
+          :key => "roles_claim_path",
+          :value => "<<foo.bar\\.zonk.baz\\.buu.baa.baa\\.bee.roles"
+        }
+      ] ++ @global_server_config
+
+    run_on_modified_server(server_config, fn ->
+      test_roles_with_bad_input()
+    end)
+
+    server_config =
+      [
+        %{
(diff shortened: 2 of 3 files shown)


src/couch/src/couch_httpd_auth.erl
```diff
@@ -227,22 +227,27 @@ jwt_authentication_handler(Req) ->
             RequiredClaims = get_configured_claims(),
             case jwtf:decode(?l2b(Jwt), [alg | RequiredClaims], fun jwtf_keystore:get/2) of
                 {ok, {Claims}} ->
+                    couch_log:info("Claims: ~p", [Claims]),
```
janl, line 230: debugging leftover?

big-r81: for the moment yes, i left them in for the draft, but they are not in the final pr...

src/couch/src/couch_httpd_auth.erl
```diff
@@ -227,22 +227,27 @@ jwt_authentication_handler(Req) ->
...
+                    Roles_Claim_Name = config:get(
+                        "jwt_auth", "roles_claim_name", "_couchdb.roles"
+                    ),
+                    Roles_Claim_Path = [list_to_binary(Item) || Item <- string:split(config:get(
+                        "jwt_auth", "roles_claim_path", ""
```
rnewson, line 235: I think it would be cleaner if roles_claim_path was independent of roles_claim_name. so you'd set it to "foo.bar.baz.roles". Ignore roles_claim_name if roles_claim_path is defined.

And obviously remove the couch_log:info lines.

rnewson: and use the shorter` ?l2b` and `?b2l` macros

big-r81: Should we mark `roles_claim_name` as deprecated (log a message, documentation) to only have one setting for this in future versions? I would suggest current behavior:

1. If `roles_claim_name` is found, use this - old behavior and doesn't break anything and log a info message to migrate to `roles_claim_path`
2. If `roles_claim_path` is found
  2.1 Use this
  2.2 If `roles_claim_name`is defined too, log an info message, that this value is ignoried, because of `roles_claim_path`

big-r81:
I updated the PR with the following logic:

We have now different code paths, how a list of couchdb roles can get into the system (inclusive backwards compatibility). 
I want to explain it in detail:

| roles_claim_name | roles_claim_path | _couchdb.roles in JWT token | used in CouchDB | log message |
| ---  | --- | --- | --- | --- |
| `undef`  | `undef` | `undef` | `---` | `---` |
| `undef`  | `undef` | `def` | `_couchdb.roles` |  implicit values found and used, migrate to path |
| `undef`  | `def` | `undef` | `roles_claim_path` | --- |
| `undef`  | `def` | `def` | `_couchdb.roles` |  implicit values found and used, migrate to path |
| `def`  | `undef` | `undef` | `roles_claim_name` | Use of 'roles_claim_name' is deprecated. Please migrate to 'roles_claim_path'! |
| `def`  | `undef` | `def` | `roles_claim_name` | Use of 'roles_claim_name' is deprecated. Please migrate to 'roles_claim_path'! |
| `def`  | `def` | `undef` | `roles_claim_name` | Both, 'roles_claim_name' and 'roles_claim_path' are set. For backwards compatibility, only `roles_claim_name`is used! |
| `def`  | `def` | `def` | `roles_claim_name` | Both, 'roles_claim_name' and 'roles_claim_path' are set. For backwards compatibility, only `roles_claim_name`is used! |

src/couch/src/couch_httpd_auth.erl
```diff
@@ -253,6 +246,68 @@ jwt_authentication_handler(Req) ->
...
+        "jwt_auth", "roles_claim_name"
+    ),
+    Roles_Claim_Path_Param = config:get(
+        "jwt_auth", "roles_claim_path"
+    ),
+    Roles =
+        case Roles_Claim_Name_Param of
+            undefined ->
+                % Check for implicitly role_claim "_couchdb.roles" in JWT token
+                case lists:keyfind(<<"_couchdb.roles">>, 1, Claims) of
```
rnewson, line 260: I don't understand this. We definitely do not let the token decide where the role param is. The couchdb server's config is the only authority on that.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -253,6 +246,68 @@ jwt_authentication_handler(Req) ->
             Req
     end.
 
+get_roles_claim(Claims) ->
+    Roles_Claim_Name_Param = config:get(
```
rnewson, line 250: these names do not follow any convention in the rest of the codebase.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -253,6 +246,68 @@ jwt_authentication_handler(Req) ->
...
+                                " `roles_claim_path`!",
+                            []
+                        ),
+                        Token_CouchDBRoles
+                end;
+            _ ->
+                % other value for `roles_claim_name` found.
+                if
+                    Roles_Claim_Path_Param =/= undefined ->
+                        couch_log:info(
```
rnewson, line 288: logging this on every occurrence is not great. You've also decided that roles_claim_name is deprecated but there's been no suggestion of that so far and it doesn't seem reasonable to me.

src/couch/src/couch_httpd_auth.erl
big-r81: Is `RolesClaimNameParam` a better alternative, or what are the conventions?

src/couch/src/couch_httpd_auth.erl
big-r81: It was only a [suggestion](https://github.com/apache/couchdb/pull/4041#discussion_r884104705) from me, because I think it's better to only have one param instead of two. We only have to find a find a common denominator...

src/couch/src/couch_httpd_auth.erl
rnewson: ```config:get("jwt_auth", "roles_claim_name", "_couchdb.roles")``` returns the roles_claim_name value from config or, if there isn't one, the default of _couchdb.roles. I see you've done the equivalent logic but in a novel and longer-form way. It would be better to stick with convention and call config:get/3 with our preferred default value.

big-r81:
I've updated the request with things discussed today. It is now possible to specify nested (& unnested) JSON paths in role_claim_path, like `roles_claim_path = foo.bar\.zonk.baz\.buu.baa.baa\.bee.roles` which is the equivalent of
```json
"foo": {
  "bar.zonk": {
    "baz.buu": {
      "baa": {
        "baa.bee": {
          "roles": [
            "my_nested_role_1",
            "my_nested_role_2"
          ]
        }
      }
    }
  }
}
```

test/elixir/test/jwt_roles_claim_test.exs
```diff
@@ -0,0 +1,126 @@
...
+      :section => "jwt_keys",
+      :key => "hmac:myjwttestkey",
+      :value => ~w(
+        NTNv7j0TuYARvmNMmWXo6fKvM4o6nv/aUi9ryX38ZH+L1bkrnD1ObOQ8JAUmHCBq7
+        Iy7otZcyAagBLHVKvvYaIpmMuxmARQ97jUVG16Jkpkp1wXOPsrF9zwew6TpczyH
+        kHgX5EuLg2MeBuiT/qJACs1J0apruOOJCg/gOtkjB4c=) |> Enum.join()
+    }
+  ]
+
+  setup do
```
jjrodrig, line 26: Hi @big-r81 
Just a question, do you need to create a testdb? It seems that the _session endpoint is the only involved here.

big-r81: Hi, oh yes, we dont' need it! Good catch! First I wanted to evaluate `$db/_security` and found the way to the `_session` endpoint. So I can remove setup/0 and tear_down/0.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -253,6 +246,59 @@ jwt_authentication_handler(Req) ->
...
+            [
+                throw(
+                    {bad_request, <<"Malformed JWT roles claim. Must be a JSON list of strings">>}
+                )
+             || R <- Result, not is_binary(R)
+            ],
+            Result;
+        false ->
+            throw(
+                {bad_request, <<"Malformed JWT roles claim. Needs to be a JSON list of strings.">>}
```
rnewson, line 298: `{bad_request, <<"Malformed token">>}` to match the other errors, we try not to reveal useful information to failed authentication requests.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -253,6 +246,59 @@ jwt_authentication_handler(Req) ->
...
+                    case re:run(RolesClaimPath, PathRegex, [global]) of
+                        nomatch -> [];
+                        {match, Pos} -> Pos
+                    end,
+                TokenizedJsonPath = tokenize_json_path(RolesClaimPath, MatchPositions),
+                couch_util:get_nested_json_value({Claims}, TokenizedJsonPath)
+        end,
+    case is_list(Result) of
+        true ->
+            [
```
rnewson, line 289: a list comprehension for its side-effects is a bad practice, the resulting list is constructed by discard. consider https://www.erlang.org/doc/man/lists.html#all-2 like `case lists:all(fun erlang:is_binary/1, Result) of true -> X; false -> Y`

src/couch/src/couch_httpd_auth.erl
```diff
@@ -253,6 +246,59 @@ jwt_authentication_handler(Req) ->
...
+                couch_util:get_value(
+                    ?l2b(
+                        config:get(
+                            "jwt_auth", "roles_claim_name", "_couchdb.roles"
+                        )
+                    ),
+                    Claims,
+                    []
+                );
+            _ ->
```
rnewson, line 276: `_` can only be a list, so test for that `when is_list(X) ->`

src/couch/src/couch_httpd_auth.erl
big-r81: ok, I "stole" this from here: https://github.com/apache/couchdb/blob/116c4e95636483143830d2c3ca9c3e2f68ff2b23/src/couch/src/couch_db.erl#L814

src/couch/src/couch_httpd_auth.erl
```diff
@@ -253,6 +246,59 @@ jwt_authentication_handler(Req) ->
...
+                        {match, Pos} -> Pos
+                    end,
+                TokenizedJsonPath = tokenize_json_path(RolesClaimPath, MatchPositions),
+                couch_util:get_nested_json_value({Claims}, TokenizedJsonPath)
+        end,
+    case is_list(Result) of
+        true ->
+            [
+                throw(
+                    {bad_request, <<"Malformed JWT roles claim. Must be a JSON list of strings">>}
```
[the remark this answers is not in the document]
big-r81: ok, will shorten that message, same as above, looked at I "stole" this from here https://github.com/apache/couchdb/blob/116c4e95636483143830d2c3ca9c3e2f68ff2b23/src/couch/src/couch_db.erl#L814

src/couch/src/couch_httpd_auth.erl
big-r81: No, RolesClaimPath isn't a list, it's something like `my.nested.couchdb\.roles'

rnewson (approved):
nicely done.

PR #4093, opened 2022-07-07
Improve error handling in smoosh_utils:write_to_file/3

noahshaw11:
## Overview

Currently, no error is written to logs when smoosh fails to write its state to the filesystem. This PR writes the error to the logs.
Note: We do not want smoosh crash with an error since failing to persist the queue should not interfere with smoosh operations. Furthermore, nested case statements were used because we do not wish to continue on to the next step in writing the state to the filesystem if we encounter an error.

## Checklist


rnewson (approved):
Thanks, I think this is better. Until we can use https://www.erlang.org/doc/reference_manual/expressions.html#maybe this is about as idiomatic as we can get, and the clear and distinct log statements will be helpful.

PR #4128, opened 2022-07-22
Revert "Replace SHA-1 with SHA-256 for cookie authentication (#4094)"

rnewson:
This reverts commit daff65d8c8772c637eb60365d6e672543163a740.

as discussed in https://github.com/apache/couchdb/issues/4119


big-r81 (commented):
I would not revert the proxyauth test, I would only change back the crypto function from sha256 to sha1…

rnewson:
I reverted the commit you presented, it's not my fault it has two unrelated changes in it. :)

big-r81 (approved):
Ok, +1, will make a new proxyauth fix commit 😉

PR #4129, opened 2022-07-22
Fix proxyauth_test and remove it from skipping tests

big-r81:
After reverting #4094, bringing this back as a seperate fix.


nickva (approved):
+1

PR #4130, opened 2022-07-25
Fix elixir :logger warnings

big-r81:
Migrate deprecated config option variable to new config option

### DIFF
--- a/config/integration.exs
+++ b/config/integration.exs
@@ -2,7 +2,9 @@ import Config
 
 config :logger,
   backends: [:console],
-  compile_time_purge_level: :debug,
+  compile_time_purge_matching: [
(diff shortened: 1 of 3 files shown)


jaydoane (approved):
LGTM, although it might be nice to show what the warnings being eliminated looked like so they are searchable in the commit history. Also, I'd be curious to read the docs where this changed, and am not having much luck googling for `compile_time_purge_matching`.

big-r81:
Oh yes, good idea!

Fixed this messages:
```
warning: :compile_time_purge_level option for the :logger application is deprecated, use :compile_time_purge_matching instead
```

I didn't find a hint in the docs too about the deprecation. Maybe it's with version 1.13.x ...

PR #4140, opened 2022-08-09
Upgrade hash algorithm for cookie auth

big-r81:
Introduce a new config setting `hash_algorithms`.

The values of the new config parameter is a list of comma-separated values of Erlang [hash algorithms](https://www.erlang.org/doc/apps/crypto/algorithm_details.html#hmac).

An example:
```ini
hash_algorithms = sha256, sha, md5
```

This line will use and generate new cookies with the `sha256` hash algorithm and accept/verify cookies with the given hash algorithms `sha` and `md5`.

Closing #4119.

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -267,6 +267,14 @@ bind_address = 127.0.0.1
 ; Set the SameSite cookie property for the auth cookie. If empty, the SameSite property is not set.
 ;same_site =
 
+; Set the HMAC algorithm used by cookie authentication
+; Possible values: sha,sha224,sha256,sha384,sha512,sha3_224,sha3_256,sha3_384,sha3_512,
+;                  blake2b,blake2s,md4,md5,ripemd160
+; New cookie sessions are generated with the first hash algorithm. All values after the
+; first hash algorithm can be used to decode the session.
+; Default: sha256, sha
+hash_algorithms = sha256, sha
+
 ; [chttpd_auth_cache]
 ; max_lifetime = 600000
 ; max_objects = 
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -270,8 +270,8 @@ bind_address = 127.0.0.1
 ; Set the HMAC algorithm used by cookie authentication
 ; Possible values: sha,sha224,sha256,sha384,sha512,sha3_224,sha3_256,sha3_384,sha3_512,
 ;                  blake2b,blake2s,md4,md5,ripemd160
-; New cookie sessions are generated with the first hash algorithm. All values after the
-; first hash algorithm can be used to decode the session.
+; New cookie sessions are generated with the first hash algorithm.
+; All values can be used to decode the session.
 ; Default: sha256, sha
 hash_algorithms = sha256, sha
 
--- a/src/chttpd/test/eunit/chttpd_auth_hash_algorithms_tests.erl
+++ b/src/chttpd/test/eunit/chttpd_auth_hash_algorithms_tests.erl
@@ -27,7 +27,7 @@ hash_algorithms_test_() ->
         {
             foreach,
             fun setup/0,
-            fun cleanup/1,
+            fun teardown/1,
             [
                 ?TDEF_FE(test_hash_algorithms_should_work),
                 ?TDEF_FE(test_hash_algorithms_should_fail)
@@ -50,7 +50,7 @@ setup() ->
     DisallowedHashes = re:split(?DISALLOWED_HASHES, "\\s*,\\s*", [trim, {return, binary}]),
     {Ctx, {AllowedHashes, DisallowedHashes}}.
 
-cleanup({Ctx, _}) ->
+teardown({Ctx, _}) ->
     config:delete("chttpd_auth", "hash_algorithms", false),
     config:delete("chttpd", "require_valid_user", false),
     config:delete("chttpd_auth", "secret", false),
--- a/src/chttpd/test/eunit/chttpd_auth_hash_algorithms_tests.erl
+++ b/src/chttpd/test/eunit/chttpd_auth_hash_algorithms_tests.erl
@@ -18,7 +18,6 @@
 
 -define(ADM_USER, "adm_user").
 -define(ADM_PASS, "adm_pass").
--define(ADM, {?ADM_USER, ?ADM_PASS}).
 -define(ALLOWED_HASHES, "sha256, sha512, sha, blake2s").
 -define(DISALLOWED_HASHES, "md4, md5, ripemd160").
 
@@ -26,53 +25,37 @@ hash_algorithms_test_() ->
     {
         "Testing hash algorithms for cookie auth",
         {
-            setup,
-            fun setup_all/0,
-            fun teardown_all/1,
-            {
-                foreach,
-                fun setup/0,
-                fun cleanup/1,
-                [
-                    ?TDEF_FE(test_hash_algorithms_should_authorized),
-                    ?TDEF_FE(test_hash_algorithms_should_unauthorized)
-                ]
-            }
+            foreach,
+            fun setup/0,
+            fun cleanup/1,
+            [
+                ?TDEF_FE(test_hash_algorithms_should_work),
+                ?TDEF_FE(test_hash_algorithms_should_fail)
+            ]
         }
     }.
 
-% Utility functions
-
-% Server utility functions
-setup_all() ->
-    Ctx = test_util:start_couch([chttpd]),
-    Hashed = couch_passwords:hash_admin_password(?ADM_PASS),
-    config:set("log", "level", "debug", false),
-    config:set("admins", ?ADM_USER, ?b2l(Hashed), false),
-    NewSecret = ?b2l(couch_uuids:random()),
-    config:set("chttpd_auth", "secret", NewSecret, false),
-    config:set("chttpd", "require_valid_user", "true", _Persist = false),
-    config:set("chttpd_auth", "hash_algorithms", ?ALLOWED_HASHES, false),
-    Ctx.
-
-teardown_all(Ctx) ->
-    config:delete("chttpd_auth", "hash_algorithms", false),
-    config:delete("chttpd", "require_valid_user", false),
-    config:delete("chttpd_auth", "secret", false),
-    config:delete("admins", ?ADM_USER, false),
-    config:delete("log", "level", false),
-    test_util:stop_couch(Ctx).
-
 % Test utility functions
 setup() ->
+    Ctx = test_util:start_couch([chttpd]),
+    Hashed = couch_passwords:hash_admin_password(?ADM_PASS),
+    NewSecret = ?b2l(couch_uuids:random()),
+    config:set("admins", ?ADM_USER, ?b2l(Hashed), false),
+    config:set("chttpd_auth", "secret", NewSecret, false),
+    config:set("chttpd", "require_valid_user", "true", false),
+    config:set("chttpd_auth", "hash_algorithms", ?ALLOWED_HASHES, false),
     AllowedHashes = re:split(config:get("chttpd_auth", "hash_algorithms"), "\\s*,\\s*", [
         trim, {return, binary}
     ]),
     DisallowedHashes = re:split(?DISALLOWED_HASHES, "\\s*,\\s*", [trim, {return, binary}]),
-    {AllowedHashes, DisallowedHashes}.
+    {Ctx, {AllowedHashes, DisallowedHashes}}.
 
-cleanup(_) ->
-    ok.
+cleanup({Ctx, _}) ->
+    config:delete("chttpd_auth", "hash_algorithms", false),
+    config:delete("chttpd", "require_valid_user", false),
+    config:delete("chttpd_auth", "secret", false),
+    config:delete("admins", ?ADM_USER, false),
+    test_util:stop_couch(Ctx).
 
 % Helper functions
 base_url() ->
@@ -109,8 +92,8 @@ test_hash_algorithm([DefaultHashAlgorithm | DecodingHashAlgorithmsList] = _, Sta
     ?assertEqual(Status, ReqStatus),
     test_hash_algorithm(DecodingHashAlgorithmsList, Status).
 
-test_hash_algorithms_should_authorized({AllowedHashes, _} = _) ->
+test_hash_algorithms_should_work({_, {AllowedHashes, _}} = _) ->
     test_hash_algorithm(AllowedHashes, 200).
 
-test_hash_algorithms_should_unauthorized({_, DisallowedHashes} = _) ->
+test_hash_algorithms_should_fail({_, {_, DisallowedHashes}} = _) ->
     test_hash_algorithm(DisallowedHashes, 401).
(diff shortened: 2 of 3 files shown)


src/couch/src/couch_httpd_auth.erl
```diff
@@ -296,9 +296,9 @@ cookie_authentication_handler(#httpd{mochi_req = MochiReq} = Req, AuthModule) ->
                 end,
             % Verify expiry and hash
             CurrentTime = make_cookie_time(),
+            HashAlgorithms = get_config_hash_algorithms(),
             case chttpd_util:get_chttpd_auth_config("secret") of
                 undefined ->
-                    couch_log:debug("cookie auth secret is not set", []),
```
rnewson, line 301: restore this line please.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -308,15 +308,26 @@ cookie_authentication_handler(#httpd{mochi_req = MochiReq} = Req, AuthModule) ->
                         {ok, UserProps, _AuthCtx} ->
                             UserSalt = couch_util:get_value(<<"salt">>, UserProps, <<"">>),
                             FullSecret = <<Secret/binary, UserSalt/binary>>,
-                            ExpectedHash = couch_util:hmac(sha, FullSecret, User ++ ":" ++ TimeStr),
+                            CalculatedHashes = lists:map(
```
rnewson, line 311: this will cause us to hash each cookie with every algorithm each time, which is wasteful. instead we want to stop at the first successful match.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -695,3 +707,12 @@ authentication_warning(#httpd{mochi_req = Req}, User) ->
...
         [?MODULE, User, Peer]
     ).
+
+-spec get_config_hash_algorithms() -> list(atom()).
+get_config_hash_algorithms() ->
+    HashAlgorithmStr = string:to_lower(
+        chttpd_util:get_chttpd_auth_config("hash_algorithms", "sha256, sha")
+    ),
+    lists:map(
+        fun binary_to_atom/1, re:split(HashAlgorithmStr, "\\s*,\\s*", [trim, {return, binary}])
```
rnewson, line 717: use binary_to_existing_atom so we don't accidentally bloat the atom table with garbage.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -695,3 +707,12 @@ authentication_warning(#httpd{mochi_req = Req}, User) ->
         "~p: Authentication failed for user ~s from ~s",
         [?MODULE, User, Peer]
     ).
+
+-spec get_config_hash_algorithms() -> list(atom()).
+get_config_hash_algorithms() ->
+    HashAlgorithmStr = string:to_lower(
```
rnewson, line 713: it's not necessary to force to lower case. we only care that the items match supported hash algorithm names.

rnewson (changes requested):
as commented

src/couch/src/couch_httpd_auth.erl
rnewson: suggest testing if the item is a member of `crypto:supports(hashs)` instead, and should handle the case where none is valid (suggest in that case we log a warning and default to `sha256`)

src/couch/src/couch_httpd_auth.erl
big-r81: oh yeah I was too quick there while removing my debug output.

src/couch/src/couch_httpd_auth.erl
big-r81: Added a predicate function to test the hash in combination with lists:any/2

src/couch/src/couch_httpd_auth.erl
big-r81: removed

src/couch/src/couch_httpd_auth.erl
big-r81: changed to binary_to_existing_atom in combination with checking the crypto hashes

src/couch/src/couch_httpd_auth.erl
```diff
@@ -308,15 +309,18 @@ cookie_authentication_handler(#httpd{mochi_req = MochiReq} = Req, AuthModule) ->
...
+                                couch_passwords:verify(Hmac, Hash)
+                            end,
                             Timeout = chttpd_util:get_chttpd_auth_config_integer(
                                 "timeout", 600
                             ),
                             couch_log:debug("timeout ~p", [Timeout]),
                             case (catch erlang:list_to_integer(TimeStr, 16)) of
                                 TimeStamp when CurrentTime < TimeStamp + Timeout ->
-                                    case couch_passwords:verify(ExpectedHash, Hash) of
+                                    case lists:any(VerifyHash, HashAlgorithms) of
```
rnewson, line 323: nice.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -695,3 +700,33 @@ authentication_warning(#httpd{mochi_req = Req}, User) ->
...
+    end.
+
+-spec get_config_hash_algorithms() -> list(atom()).
+get_config_hash_algorithms() ->
+    SupportedHashAlgorithms = crypto:supports(hashs),
+    HashAlgorithmsStr = chttpd_util:get_chttpd_auth_config("hash_algorithms", "sha256, sha"),
+    HashAlgorithms = re:split(HashAlgorithmsStr, "\\s*,\\s*", [trim, {return, binary}]),
+    VerifiedHashNames = verify_hash_names(HashAlgorithms, SupportedHashAlgorithms),
+    if
+        length(VerifiedHashNames) == 0 -> [sha256];
```
rnewson, line 730: it's more common to compare against the empty list `[]`.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -695,3 +700,33 @@ authentication_warning(#httpd{mochi_req = Req}, User) ->
...
+verify_hash_names(HashAlgorithms, SupportedHashFun) ->
+    verify_hash_names(HashAlgorithms, SupportedHashFun, []).
+verify_hash_names([], _, HashNames) ->
+    HashNames;
+verify_hash_names([H | T], SupportedHashFun, HashNames) ->
+    try
+        HashAtom = binary_to_existing_atom(H),
+        Result =
+            case lists:member(HashAtom, SupportedHashFun) of
+                true -> HashNames ++ [HashAtom];
```
rnewson, line 713: while it's unlikely a problem here, with very short lists, we try to avoid `++`. since order doesn't matter I think `[HashAtom | HashNames]` is preferable.

src/couch/src/couch_httpd_auth.erl
```diff
@@ -695,3 +700,33 @@ authentication_warning(#httpd{mochi_req = Req}, User) ->
...
+                false -> HashNames
+            end,
+        verify_hash_names(T, SupportedHashFun, Result)
+    catch
+        _:_ ->
```
rnewson, line 718: catch only the error you expect `error:badarg`

src/couch/src/couch_httpd_auth.erl
```diff
@@ -695,3 +700,33 @@ authentication_warning(#httpd{mochi_req = Req}, User) ->
...
+            couch_log:warning("~p: Hash algorithm ~s is not available.", [?MODULE, H]),
```
rnewson, line 719: suggest 'valid' instead of 'available'.

src/couch/src/couch_httpd_auth.erl
big-r81: Okay, but the order of the hash algos is important. Assume we have
```ini
hash_algorithms = sha256, sha
```
then we get as result
```erlang
[sha, sha256]
```
and the cookie is generated with the first algorithm.
Maybe your solution 
```erlang
true -> [HashAtom | HashNames]
```
in combination with
```
verify_hash_names([], _, HashNames) ->
    lists:reverse(HashNames);
```

Is this better?

rnewson: maybe combine these things to avoid the problem entirely. iterate through the `re:split` list of hash algorithm names in order, moving to the next if its not a valid algorithm name or if the hash doesn't match for valid names, and then return a failure if the list is empty.

big-r81: I added both combinations and I think it's fine...

src/couch/src/couch_httpd_auth.erl
```diff
@@ -695,3 +700,33 @@ authentication_warning(#httpd{mochi_req = Req}, User) ->
...
+            couch_log:warning("~p: Hash algorithm ~s is not valid.", [?MODULE, H]),
+            verify_hash_names(T, SupportedHashFun, HashNames)
+    end.
+
+-spec get_config_hash_algorithms() -> list(atom()).
+get_config_hash_algorithms() ->
+    SupportedHashAlgorithms = crypto:supports(hashs),
+    HashAlgorithmsStr = chttpd_util:get_chttpd_auth_config("hash_algorithms", "sha256, sha"),
+    HashAlgorithms = re:split(HashAlgorithmsStr, "\\s*,\\s*", [trim, {return, binary}]),
+    VerifiedHashNames = verify_hash_names(HashAlgorithms, SupportedHashAlgorithms),
```
rnewson, line 728: suggest;

```
case verify_hash_names(HashAlgorithms, SupportedHashAlgorithms) of
    [] ->
        [sha256];
    List ->
        List
end.
```

big-r81: Oh yes, that is much easier to read and looks more "Erlangish". ;-)

jaydoane (commented):
This looks really great!

I wonder if it might be useful to implements tests that:
- Demonstrate successfully decoding a cookie with a non-primary hash
- Demonstrate logging an unsupported hash algorithm

?

rel/overlay/etc/default.ini
```diff
@@ -268,6 +268,14 @@ bind_address = 127.0.0.1
 ; Set the SameSite cookie property for the auth cookie. If empty, the SameSite property is not set.
 ;same_site =
 
+; Set the HMAC algorithm used by cookie authentication
+; Possible values: sha,sha224,sha256,sha384,sha512,sha3_224,sha3_256,sha3_384,sha3_512,
+;                  blake2b,blake2s,md4,md5,ripemd160
+; New cookie sessions are generated with the first hash algorithm. All values after the
+; first hash algorithm can be used to decode the session.
```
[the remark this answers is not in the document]
big-r81: oh yes, not only does it sound better, nor can it be misunderstood.

big-r81:
> This looks really great!
> 
> I wonder if it might be useful to implements tests that:
> 
>     * Demonstrate successfully decoding a cookie with a non-primary hash
> 
>     * Demonstrate logging an unsupported hash algorithm
> 
> 
> ?
Added some test cases.

jaydoane (approved):
Thanks for writing these tests!
```erlang
module 'chttpd_auth_hash_algorithms_tests'
  Testing hash algorithms for cookie auth
    chttpd_auth_hash_algorithms_tests:31: -with/1-fun-0- (test_hash_algorithms_should_work)...[0.258 s] ok
    chttpd_auth_hash_algorithms_tests:31: -with/1-fun-0- (test_hash_algorithms_should_fail)...[0.002 s] ok
    [done in 0.266 s]
  [done in 4.361 s]
  2 tests passed.
```
LGTM! 👍

src/chttpd/test/eunit/chttpd_auth_hash_algorithms_tests.erl
```diff
@@ -0,0 +1,99 @@
...
+    Ctx = test_util:start_couch([chttpd]),
+    Hashed = couch_passwords:hash_admin_password(?ADM_PASS),
+    NewSecret = ?b2l(couch_uuids:random()),
+    config:set("admins", ?ADM_USER, ?b2l(Hashed), false),
+    config:set("chttpd_auth", "secret", NewSecret, false),
+    config:set("chttpd", "require_valid_user", "true", false),
+    config:set("chttpd_auth", "hash_algorithms", ?ALLOWED_HASHES, false),
+    AllowedHashes = re:split(config:get("chttpd_auth", "hash_algorithms"), "\\s*,\\s*", [
+        trim, {return, binary}
+    ]),
```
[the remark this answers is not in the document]
big-r81: We could extract this into a function, but I would suggest this as a separate PR, because we have this pattern in many modules and with many helper functions `split_list(...)`. If there is no veto, I would merge this as is.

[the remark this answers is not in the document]
jaydoane: Yeah, I think it could wait for a separate PR as well. 

> If there is no veto, I would merge this as is.

I would squash and rebase first 😉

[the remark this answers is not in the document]
big-r81: > I would squash and rebase first 😉

Oh yes, definitely ;-)

nickva:
Noticed a recent failure in the full CI run referencing a "**error:{notsup,{"mac.c",264},"Unsupported digest algorithm"}"

https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FFullPlatformMatrix/detail/main/327/pipeline/203

```
[2022-08-24T16:57:41.113Z] module 'chttpd_auth_hash_algorithms_tests'

[2022-08-24T16:57:41.113Z]   Testing hash algorithms for cookie auth

[2022-08-24T16:57:41.467Z]     chttpd_auth_hash_algorithms_tests:31: with (test_hash_algorithms_should_work)...*failed*

[2022-08-24T16:57:41.467Z] in function crypto:mac_nif/4

[2022-08-24T16:57:41.467Z]   called as mac_nif(hmac,blake2s,<<"305ead508cce07ab432b7cd9d6455349bc1dd8ebfa0524f039f502d3"...>>,"adm_user:63065885")

[2022-08-24T16:57:41.467Z] in call from chttpd_auth_hash_algorithms_tests:make_auth_session_string/4 (test/eunit/chttpd_auth_hash_algorithms_tests.erl, line 68)

[2022-08-24T16:57:41.467Z] in call from chttpd_auth_hash_algorithms_tests:test_hash_algorithm/2 (test/eunit/chttpd_auth_hash_algorithms_tests.erl, line 85)

[2022-08-24T16:57:41.467Z] in call from eunit_test:run_testfun/1 (eunit_test.erl, line 71)

[2022-08-24T16:57:41.467Z] in call from eunit_proc:run_test/1 (eunit_proc.erl, line 522)

[2022-08-24T16:57:41.467Z] in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 347)

[2022-08-24T16:57:41.467Z] in call from eunit_proc:handle_test/2 (eunit_proc.erl, line 505)

[2022-08-24T16:57:41.467Z] in call from eunit_proc:tests_inorder/3 (eunit_proc.erl, line 447)

[2022-08-24T16:57:41.467Z] **error:{notsup,{"mac.c",264},"Unsupported digest algorithm"}

[2022-08-24T16:57:41.467Z]   output:<<"">>
```

It may be related to this change, I didn't get a chance to dig in to find out

big-r81:
I will look at this again.

1. why didn’t it fail in the PR-CI?
2. What Erlang version do we use in the Full CI?
3. It should skip a algorithm which doesn’t exists…
4. It seems blake2s as a test algorithm is the trigger…

big-r81:
It fails on:

1. CentOS 7 with `blake2s` and
2. Ubuntu 22.04 with `md4`

I think they are not supported by the openssl version in erlang. Need to investigate why they are executed, if they are not supported...

big-r81:
Found it, the test itself doesn't check for supported hash algorithms.
Refactor the test to use the checks too. See #4156.

PR #4149, opened 2022-08-16
Update couchdb-mochiweb to v3.1.0

big-r81:

PR #4150, opened 2022-08-16
Fix parse_max_age_test_() test after upgrading to mochiweb v3.1.0

big-r81:
After upgrading to mochiweb v3.1.0, whitespaces (incl. newline) are
trimmed from the header field values. This commit adds more tests
and changing the expected behavior of parsing the header fields.

The tests of this PR should pass if #4149 has been merged.


nickva (commented):
Good find and fix @big-r81. I wonder if we should just combine the two PRs then into one so we get a passing CI run - update mochiweb and the test at the same time.

big-r81:
I thought about it too, but then decided to separat the two things. Running both PR combined, the test will succeed. If we merge #4149, then I need to update this PR and then it should work.

big-r81:
Integrated into #4149, so I'm closing this pull request.

Issue #2496, opened 2020-01-28
Query all document (_all_docs) from specific partition with filtering returns wrong total_rows value and document information


vmatyusGitHub:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your actual CouchDB! )
During the usage of `POST /{db}/_partition/{partition_key}/_all_docs` in the response there is a field `total_rows` and its value should return the number of documents in the shard range hosting that partition.
Furthermore it should return only partition specific documents.

If there is `keys` filter included in the query, currently the request returns the number of documents in the database and all partition related document that matches with the filter. There is no partition filtering.

## Steps to Reproduce
[NOTE]: # ( Include commands to reproduce, if possible. curl is preferred. )
1. Create a partitioned database
2. Create sample partitioned documents
3. Query the database with endpoint: `POST /{db}/_partition/{partition_key}/_all_docs --data-raw {}` It shall return a `total_rows` value. Let's call it `n`. 
Returns all partition specific document.
4. Query the database with an additional filter. eg: `POST /{db}/_partition/part2/_all_docs --data-raw {"keys":["part1:doc1", "part2:doc1"]}`
In this case the `total_rows` of the response contains the number of all document of the database . 

In the request the following can be seen:
```
"rows": [
        {
            "id": "part1:doc1",
            "key": "part1:doc1",
            "value": {
                "rev": "5-aa9d2b82de6db1e61fba8e4057299d21"
            }
        },
        {
            "id": "part2:doc1",
            "key": "part2:doc1",
            "value": {
                "rev": "3-a707bb62c86bd430389e4d596aeb6084"
            }
        }
]
```

## Expected Behaviour
The `total_rows` field should have value  `n`.
Only partition specific document information should be returned.
This should be the expected response:
```
"rows": [
        {
            "key": "part1:doc1",
            "error": "not_found"
        },
        "id": "part2:doc1",
            "key": "part2:doc1",
            "value": {
                "rev": "3-a707bb62c86bd430389e4d596aeb6084"
            }
    ]
```
[NOTE]: # ( Tell us what you expected to happen. )

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )
[TIP]:  # ( You can paste the output of curl http://YOUR-COUCHDB:5984/ here. )

[...]

PR #4152, opened 2022-08-19
Fix Partition All Docs Query With Keys

zachlankton:
## Overview

This PR addresses a bug that is described in issue #2496 where posting an all docs query to a partition endpoint with keys that are not a part of the partition also gets returned.

## Testing recommendations

Consider the following documents, each in their own partition on a database named `test`.

```json
"rows": [
        {
            "id": "part1:doc1",
            "key": "part1:doc1",
            "value": {
                "rev": "5-aa9d2b82de6db1e61fba8e4057299d21"
            }
        },
        {
            "id": "part2:doc1",
            "key": "part2:doc1",
            "value": {
                "rev": "3-a707bb62c86bd430389e4d596aeb6084"
            }
        }
]
```

When we run the following query against the `part2` partition:

```bash
curl -X POST http://127.0.0.1:15984/test/_partition/part2/_all_docs \
    -d '{"keys":["part1:doc1", "part2:doc1"]}'
```

We get **_both documents from both partitions_**, but what we expect should look like this

```json
"rows": [
        {
            "key": "part1:doc1",
            "error": "not_found"
        },
        {
        "id": "part2:doc1",
            "key": "part2:doc1",
            "value": {
                "rev": "3-a707bb62c86bd430389e4d596aeb6084"
            }
       }
    ]
```

The code in this PR fixes this issue and returns the results we expect.

Issue #2496 

## Checklist

- [n/a] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [n/a] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation


src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -286,6 +286,20 @@ all_docs_concurrency() ->
...
     end.
 
+validate_if_partition(Row, Acc) ->
+    QueryStr = Acc#vacc.req#httpd.qs,
+    case QueryStr of
+        undefined -> Row;
+        [{"partition", Id}] -> is_doc_in_partition(Row, Id)
+    end.
+is_doc_in_partition(Row, Id) ->
+    BinRowId = element(2, Row),
```
janl, line 296: the more idiomatic way of doing this in Erlang would be:

```
 {_, BinRowId, _, _, _, _} = Row,
```

but I’m not 100% that that’s actually better. I think maybe on revising the code it is nicer to know the explicit row structure rather than having to know the id is in location 2.

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -286,6 +286,20 @@ all_docs_concurrency() ->
             10
     end.
 
+validate_if_partition(Row, Acc) ->
+    QueryStr = Acc#vacc.req#httpd.qs,
+    case QueryStr of
+        undefined -> Row;
+        [{"partition", Id}] -> is_doc_in_partition(Row, Id)
```
janl, line 293: this only matches if the `qs` is either `undefined` or exactly the list `[{"partition", Id}]` — is this always the case? Can there be other options in the qs? If yes, we’d probably need to get the partition tuple out with a `proplists:get()` or `couch_util:get_value()`

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -286,6 +286,20 @@ all_docs_concurrency() ->
...
+    end.
+is_doc_in_partition(Row, Id) ->
+    BinRowId = element(2, Row),
+    IsDocInPartition = binary:match(BinRowId, list_to_binary(Id)) /= nomatch,
+    case IsDocInPartition of
+        true -> Row;
+        false -> {view_row, BinRowId, undefined, undefined, undefined, undefined}
```
janl, line 300: this looks like you create a dummy row for entries that should not be in the result that presumably gets filtered out later. You also pass this into `Callback()` in line 318 below. Wouldn’t it make more sense to just not call `Callback()` with that row?

janl:
Thanks for this, this looks decent overall, especially for a first contribution :)

 I left a few minor notes on the code. I just have one question about the approach taken here.

The way I read it: passing keys to match into the query will return result rows we do not want. So this PR starts preventing those result rows from being returned.

Would another option to solve this be making it so that we don’t read those rows in the first place? That would make this a performance improvement too.

We might even be able to do this at the request validation stage: if `keys` includes a partition that is not the partition of the URL being queried, we return a 400 error with a message that all keys must match the partition provided in the URL.

Just spitballing here tho. Nice work!

src/fabric/src/fabric_view_all_docs.erl
rnewson: and also it should be created as `#view_row{id = BinRowId}`

src/fabric/src/fabric_view_all_docs.erl
rnewson: Neither of those are good approaches. `Row#view_row.id` extracts the required field by name.

janl: thanks, that’s what I was looking for

zachlankton: This is so much better, thank you!

zachlankton: I realize now that I wasn't totally clear with my variable name as I am trying to get the key, not the id.  So I will update as follows:

```erlang
BinRowKey = Row#view_row.key,
```

src/fabric/src/fabric_view_all_docs.erl
zachlankton: 👍

src/fabric/src/fabric_view_all_docs.erl
zachlankton: You are correct, I am newish to the pattern matching in erlang and here it really shows.  I can see how this will fail if there are any other query parameters.

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -286,6 +286,24 @@ all_docs_concurrency() ->
             10
     end.
 
+get_query_list(undefined) ->
+    [];
+get_query_list(List) when is_list(List) ->
+    List.
+validate_if_partition(Row, Acc) ->
+    QueryList = get_query_list(Acc#vacc.req#httpd.qs),
+    case couch_util:get_value("partition", QueryList) of
```
zachlankton, line 295: I just discovered `chttpd:qs_value/2` ... will be refactoring this code.

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -286,6 +286,24 @@ all_docs_concurrency() ->
...
+        undefined -> Row;
+        Id -> is_doc_in_partition(Row, Id)
+    end.
+is_doc_in_partition(Row, Id) ->
+    BinRowKey = Row#view_row.key,
+    IsDocInPartition = binary:match(BinRowKey, list_to_binary(Id)) /= nomatch,
```
zachlankton, line 301: How do we feel about the way I am matching the `BinRowKey` with the requested partition `Id` ?

This works, but wondering if it could be problematic in the future as a partition key of `test` could also match the following row key: `someOtherPartition:with-test-in-the-doc-id`

What would be the recommended way to extract the partition from the `BinRowKey` and do an explicit match?

zachlankton: just discovered `couch_partition:is_member(BinRowKey, Partition)`  will be refactoring this as well

PR #4153, opened 2022-08-22
config section for require_valid_user is only [chttpd]

big-r81:
If this gets merged, please also merge https://github.com/apache/couchdb-documentation/pull/753

### DIFF
--- a/dev/run
+++ b/dev/run
@@ -472,15 +472,10 @@ def hack_default_ini(ctx, node, contents):
 
 
 def hack_local_ini(ctx, contents):
-    # make sure all three nodes have the same secret
-    secret_line = "secret = %s\n" % COMMON_SALT
-    previous_line = "; require_valid_user = false\n"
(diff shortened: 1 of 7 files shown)


nickva (changes requested):
Looks good but there is one ugly corner case we should handle in dev/run.

Notice https://github.com/apache/couchdb/blob/d4ac4457a849ea9b3244da27c4502c3d32c3a880/dev/run#L476-L478 

There we use the commented out require_valid_user line as a replacement marker to know where to write in the secret value. We'd want to make sure the dev/run setup still works with removing both the section name and `; require_valid_user = false`

EDIT: Since the section is now completely removed from local.ini there we could just add the section when we update the secret value so could be something like `file_contents + '\n[chttpd_auth]\nsecret=foobar...\n` ?

big-r81:
Good catch @nickva and thx for the good discussion!

nickva (approved):

PR #4155, opened 2022-08-23
fix missing "=" for admin party in #4153

big-r81:
In #4153 we missed a "=" in the secret option for the admin party


nickva (approved):
+1

PR #4157, opened 2022-08-25
Refactor hash algorithms test for full CI run

big-r81:
The test doesn't check if the hash algorithm is supported by the
erlang vm. The test for supported hash algorithms was only missing
in the test itself and not in CouchDB.
Refactor test and verify hash names during test runs.

See https://github.com/apache/couchdb/pull/4140.


nickva (approved):
+1 if the full CI passes

Thanks for the quick fix @big-r81 !

PR #4158, opened 2022-08-28
Fix variable already bound compiler warnings

jaydoane:
## Overview

OTP 25 generates warnings like the following:

```
src/chttpd/test/eunit/chttpd_util_test.erl:33:48: Warning: variable '_Persist' is already bound. If you mean to ignore this value, use '_' or a different underscore-prefixed name
```

Create an explicit `Persist` variable set `false` to suppress those warnings.

## Testing recommendations

```
make eunit
```
should no longer emit those warnings under OTP 25.

## Checklist


big-r81 (approved):
+1

nickva (approved):

PR #4328, opened 2022-12-21
Fix {eaddrnotavail} in elixir tests under Windows

big-r81:
## Overview

Adding full url to database names when replicating.


big-r81:
A big thx to  @jjrodrig for debugging and applying a nice and generic solution!

janl (approved):

PR #4340, opened 2023-01-03
Use the bootstrap script to build local rebar

big-r81:
This can be merged after https://github.com/apache/couchdb-rebar/pull/4 is merged...

### DIFF
--- a/configure
+++ b/configure
@@ -296,9 +296,10 @@ install_local_rebar() {
         if [ ! -d "${rootdir}/src/rebar" ]; then
             git clone --depth 1 https://github.com/apache/couchdb-rebar.git ${rootdir}/src/rebar
         fi
-        make -C ${rootdir}/src/rebar
+        cd src/rebar
+        ./bootstrap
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -296,9 +296,10 @@ install_local_rebar() {
         if [ ! -d "${rootdir}/src/rebar" ]; then
             git clone --depth 1 https://github.com/apache/couchdb-rebar.git ${rootdir}/src/rebar
         fi
-        make -C ${rootdir}/src/rebar
+        cd src/rebar
+        ./bootstrap
```
jaydoane, line 300: Can you help me understand why it's preferable to run `bootstrap` explicitly when it appears that `make` [should be using bootstrap already?](https://github.com/apache/couchdb-rebar/blob/main/Makefile#L8-L9)

big-r81: I used the rebar3 example as a guide at the beginning.  I will test it again without this change.

big-r81: @jaydoane You are right, there are no advantages and it works without calling bootstrap directly. Closing this PR.

PR #4342, opened 2023-01-04
Set Spidermonkey 91 as default

big-r81:
Set Spidermonkey 91 as the new default version.

### DIFF
--- a/configure.ps1
+++ b/configure.ps1
@@ -9,7 +9,7 @@
   -DisableDocs               request build process skip building documentation (default false)
   -SkipDeps                  do not update Erlang dependencies (default false)
   -CouchDBUser USER          set the username to run as (defaults to current user)
-  -SpiderMonkeyVersion VSN   select the version of SpiderMonkey to use (defaults to 1.8.5)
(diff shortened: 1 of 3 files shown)


janl:
let’s not remove support just yet. +1 for moving the default foward tho

configure.ps1
```diff
@@ -138,6 +138,7 @@ $CouchDBConfig = @"
 {node_name, "-name couchdb@localhost"}.
 {cluster_port, 5984}.
 {backend_port, 5986}.
+{prometheus_port, 17986}.
```
janl, line 141: stray commit?

big-r81: No, I only adapted this to the *nix configure-[version](https://github.com/apache/couchdb/blob/33ee2164e6e73429193bc353a2c7dfb9d27a7790/configure#L261)

janl: ok, but this also has nothing to do with SM91? It’d be good to keep topical things in separate commits/PRs

big-r81: ok, fixed. Should the doc update be separate too?

janl (approved):

PR #4343, opened 2023-01-04
fix undef when parsing replication doc body

rnewson:
Closes https://github.com/apache/couchdb/issues/4344

### DIFF
--- a/src/couch_replicator/src/couch_replicator_parse.erl
+++ b/src/couch_replicator/src/couch_replicator_parse.erl
@@ -490,7 +490,7 @@ get_json_value(Key, Obj) ->
     couch_replicator_utils:get_json_value(Key, Obj).
 
 get_json_value(Key, Obj, Default) ->
-    couch_replicator_util:get_json_value(Key, Obj, Default).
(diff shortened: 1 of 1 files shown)


janl (approved):

src/couch_replicator/src/couch_replicator_parse.erl
```diff
@@ -490,7 +490,7 @@ get_json_value(Key, Obj) ->
     couch_replicator_utils:get_json_value(Key, Obj).
 
 get_json_value(Key, Obj, Default) ->
-    couch_replicator_util:get_json_value(Key, Obj, Default).
+    couch_replicator_utils:get_json_value(Key, Obj, Default).
```
jaydoane, line 493: Can we add a test for this as well?

rnewson: go right ahead.

jaydoane: https://github.com/apache/couchdb/pull/4345

PR #4363, opened 2023-01-10
Fix replication _scheduler/docs total_rows

jaydoane:
## Overview

The total_rows property was decremented by one to account for the VDU that was automatically added to the total. Now that a BDU has replaced the VDU [1] total_rows is one less than it should be.

This removes the decrement so that total_rows equals the actual doc count.

[1] https://github.com/apache/couchdb/pull/4274

## Testing recommendations

```
make eunit apps=couch_replicator suites=couch_replicator_scheduler_docs_tests
```

## Checklist

### DIFF
--- a/src/couch_replicator/src/couch_replicator_httpd_util.erl
+++ b/src/couch_replicator/src/couch_replicator_httpd_util.erl
@@ -158,7 +158,7 @@ docs_cb({meta, Meta}, #vacc{meta_sent = false, row_sent = false} = Acc) ->
     Parts =
         case couch_util:get_value(total, Meta) of
             undefined -> [];
-            Total -> [io_lib:format("\"total_rows\":~p", [adjust_total(Total)])]
+            Total -> [io_lib:format("\"total_rows\":~p", [Total])]
         end ++
             case couch_util:get_value(offset, Meta) of
                 undefined -> [];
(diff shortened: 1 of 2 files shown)


src/couch_replicator/src/couch_replicator_httpd_util.erl
```diff
@@ -158,7 +158,7 @@ docs_cb({meta, Meta}, #vacc{meta_sent = false, row_sent = false} = Acc) ->
     Parts =
         case couch_util:get_value(total, Meta) of
             undefined -> [];
-            Total -> [io_lib:format("\"total_rows\":~p", [adjust_total(Total)])]
+            Total -> [io_lib:format("\"total_rows\":~p", [Total])]
```
big-r81, line 161: Hi Jay,

is there a reason why we use io_lib:format instead couch_log: here?

jaydoane: Yes, because it's not actually doing I/O here, but just [formatting the string](https://www.erlang.org/doc/man/io_lib.html#format-2). Are you thinking of `io:format/2` maybe?

big-r81: Ah, you are completly right! (_shakes fist at my brain... ;-)_)

big-r81 (approved):
+1

PR #4379, opened 2023-01-17
docs(mango): match description of `$mod` with reality

pgj:
The remainder argument for the `$mod` operator can be zero, while its documentation suggests otherwise.  It actually covers a very realistic use case where divisibility is expressed.

[Neither related restrictions could be identified in the sources](https://github.com/apache/couchdb/blob/adf17140e81d0b74f2b2ecdea48fc4f702832eaf/src/mango/src/mango_selector.erl#L512:L513) [nor MongoDB forbids this](https://www.mongodb.com/docs/manual/reference/operator/query/mod/).  Tests also seem to exercise [this specific case](https://github.com/apache/couchdb/blob/0059b8f90e58e10b199a4b768a06a762d12a30d3/src/mango/test/03-operator-test.py#L58).  Thanks @iilyak for checking on these.

Tasks:
- [ ] Backport to all affected branches

### DIFF
--- a/src/docs/src/api/database/find.rst
+++ b/src/docs/src/api/database/find.rst
@@ -705,8 +705,8 @@ operators require the argument to be in a specific JSON format.
 |               |             |            | document. Non-array fields cannot |
 |               |             |            | match this condition.             |
 +---------------+-------------+------------+-----------------------------------+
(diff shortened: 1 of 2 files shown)


willholley (approved):
+1

PR #4402, opened 2023-01-23
Docs: Fix example code in Partitioned Databases

jiahuili430:
## Overview
https://docs.couchdb.org/en/stable/partitioned-dbs/index.html#partitions-by-example

The first example is a partitioned database using a partitioned query, so here we should use `ddoc.json`.

## Checklist


nickva (approved):
+1

Thanks, @jiahuili430

PR #4409, opened 2023-01-27
docs(INSTALL.Unix): update macOS instructions

pgj:
On recent macOS systems, the `xcode-select --install` command will install a version of Python 3, along with `pip3`, hence there is no need to do that explicitly.  At the same time, system-wide installation of the required Python packages should be discouraged.

Fix markdown of the `brew` command, while here.

### DIFF
--- a/INSTALL.Unix.md
+++ b/INSTALL.Unix.md
@@ -117,10 +117,9 @@ You can install Node.JS via the
 
 You can install the documentation dependencies by running:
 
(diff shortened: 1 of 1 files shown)


big-r81:
maybe a duplicate of #4386, as Nick mentioned, we only need Python3 installed, because the necessary packages are installed in a venv for the docs... See https://github.com/apache/couchdb/pull/4386/files#r1073714355

PR #4410, opened 2023-01-27
Mango covering JSON indexes RFC

mikerhodes:
## Overview

RFC for covering indexes in Mango. Propose covering indexes within Mango, for JSON indexes, and outline the implementation steps.

#4394 is a pre-req for this work, because without it we always need to read the full doc to return to the coordinator.


src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,254 @@
...
+choosing an index. This is because we require all keys in the index to be fields
+within the selector -- with predicates implying `$exists=true` -- due to the
+fact that only documents that include _all_ fields in the index are added to the
+index. Therefore, if the selector doesn't imply all fields in the index's keys
+exist, then using that index risks returning an incomplete result set.
+
+Within `view_cb/2`, we'll need to know whether an index is covering. Without
+that, `view_cb/2` will interpret the lack of included documents as an indicator
+that it should do nothing, while in fact we want it to fully process the result
+as it does when `include_docs` is used -- apart from when the user passes `r>=2` in the Mango query because then the coordinator reads and processes
```
janl, line 108:
```suggestion
as it does when `include_docs` is used -- apart from when the user passes `r>=2`
in the Mango query because then the coordinator reads and processes
```

missin `\n`

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,254 @@
...
+to include documents. If the index is covering, this will not be required and
+so the `include_docs` argument to the view fabric call will be `false`. We'll
+need to add a helper method to return whether the index is covering.
+
+When selecting an index, we'll need to be careful of some subtleties. We will
+need to ensure that only fields in the `selector` and not `fields` are used when
+choosing an index. This is because we require all keys in the index to be fields
+within the selector -- with predicates implying `$exists=true` -- due to the
+fact that only documents that include _all_ fields in the index are added to the
+index. Therefore, if the selector doesn't imply all fields in the index's keys
```
garrensmith, line 102: Just checking my understanding. The added subtlety here is because previously we would do an in-memory filter of the document to check the filter completely matches the document. Now if we can use the index alone we have to make sure all fields in the selector are also in the index keys. So if a selector has filters on `name`, `age` and `country` and the `fields` section in the query is `name` and `age`. Mango would have to choose an index with `name`, `age` and `country` even though it is only returning two fields. Is that correct?



What happens if no index satisfies this?

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,254 @@
...
+exist, then using that index risks returning an incomplete result set.
+
+Within `view_cb/2`, we'll need to know whether an index is covering. Without
+that, `view_cb/2` will interpret the lack of included documents as an indicator
+that it should do nothing, while in fact we want it to fully process the result
+as it does when `include_docs` is used -- apart from when the user passes `r>=2` in the Mango query because then the coordinator reads and processes
+documents. (Aside: it'd be good to remove this `r` option to simplify things).
```
garrensmith, line 109: +1 to removing the `r` option. It has been something I wanted to remove for a long time.

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,254 @@
...
+```
+
+Behaviour requirements:
+
+- Unlike `fields`, the fields in `include` _do not have to exist_ in the source
+    document in order that the document be included in the index. This is to
+    allow the index to cover more queries.
+- Including a deeply nested field would follow the same pattern as for other
+    field references in mango, `person.address.zip`.
+- There is no notation to include the whole document, that is, no equivalent of
```
garrensmith, line 136: Is this something we should consider or rather if a user wants the whole document, they would need to list all the fields of the index?

garrensmith (approved):
Nice RFC, it explains the concept clearly and +1 to covering indexes this will be a great feature.

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: Added in 240b91280.

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: It's worth considering. In [the past](https://dx13.co.uk/articles/2017/02/04/why-you-should-generally-avoid-using-include_docs-in-cloudant-and-couchdb-view-queries/) I've used a simple benchmark to show that having documents included in the index does produce faster query response than `include_docs=true`, at least in relatively extreme scenarios, which does argue for using it. In truth, I'd want to look at how the btree is laid out on disk and look at some benchmarking to understand how much storing tons of extra data in the view affects query speed. It feels like there's a potential to be pulling tons of unneeded data from disk.

So, to start with, I wanted to scope this down to what I saw as the minimum useful thing. It's harder to remove things later.

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: Noted for some later pondering 😄

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: Added some clarifying notes in 56ae1854f. I created a new section better describing the relationship required between an index's keys and the `selector` if an index is to be used during query processing. I tried to then cover the specifics for this work within the "Part 1" and "Part 2" sections.

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,360 @@
...
+Now we index two documents. The first document is included in the index while the second is not (because it doesn't include `name`):
+
+
+```json
+{"_id": "foo", "age": 39, "name": "mike"}
+
+{"_id": "bar", "age": 39, "pet": "cat"}
+```
+
+The `selector` `{"age": {"$gt": 30}}` should return both documents. However, if
```
garrensmith, line 128: Nice example thanks.

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,360 @@
...
+A Mango JSON index is implemented as a view with a complex key. The first field
+in the index is the first entry in the complex key, the second field is the
+second key and so on. Even indexes with one field use a complex key with length
+`1`.
+
+When choosing a JSON index to use for a query, there are a couple of things that
+are important to covering indexes.
+
+Firstly, note there are certain predicate operators that can be used with an
+index, currently: `$lt`, $lte`, `$eq`, $gte` and `$gt`. These can easily be
```
nickva, line 101: Missing '`` for `$lte`

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,360 @@
...
+
+- Unlike `fields`, the fields in `include` _do not have to exist_ in the source
+    document in order that the document be included in the index. This is to
+    allow the index to cover more queries.
+- Including a deeply nested field would follow the same pattern as for other
+    field references in mango, `person.address.zip`.
+- There is no notation to include the whole document, that is, no equivalent of
+    `emit(doc.name, doc)`.
+- It will be an error to include a field in both `fields` and `include`. This
+    should be rejected by the `_index` call.
```
nickva, line 195: Would the equivalent of `"include": []` be the same as not specifying an `"include"` field at all?

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Phase 2: add support for included fields in indexes
+
+I propose we add an `include` field into a Mango JSON index definition:
+
+```json
+{
+    "index": {
+        "fields": [ "age", "name" ],
+        "include": [ "occupation", "manager_id" ]
```
nickva, line 178: The order of the fields in `include` is not meaningful in any way? Should we add a note highlighting it in the API, just for completeness.

As an implementation detail, perhaps we'd just want to normalize it by sorting when creating the design doc and the view signature. That would mean that two indexes with the same details an only the include in a different order would be equivalent and "point to" the same view signature.

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,360 @@
...
+but shouldn't alter the correctness of the result. So we shouldn't need a
+special code path to support this case. Which is nice.
+
+### Outdated coordinator, updated shard
+
+In this case, the coordinator won't be checking for covering indexes, meaning
+that `include_docs=true` will be set when `r<2` as today.
+
+I suspect we'll set an option in `viewcbargs` that contains the index field
+names and whether its a covering index. This means that an updated shard will be
```
nickva, line 260: `its` -> `it's` ?

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,360 @@
...
+document are to be interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.txt).
+
+## Terminology
+
+[TIP]:  # ( Provide a list of any unique terms or acronyms, and their definitions here.)
+
+- Mango: CouchDB's Mongo inspired querying system.
+- View / JSON index: Mango index that uses the same index as Cloudant views.
+- Coordinator: the erlang process that handles doing a distributed query across
```
nickva, line 71: Tiny nit: Capitalize `Erlang`. Not sure which one is more correct but just to keep consistent with the other RFCs' style.

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,360 @@
...
+
+I suspect we'll set an option in `viewcbargs` that contains the index field
+names and whether its a covering index. This means that an updated shard will be
+checking for those fields. When it can't find them, it'll fallback to the
+current behaviour in `view_cb/2`, meaning that it reads the document found via
+`include_docs=true`, execute `match_and_extract_doc/3` and return the row if it
+matches the query.
+
+The coordinator will received the final result document as today and assume it's
+correct, and forward it to the client.More work than needed will be carried out
```
nickva, line 267: `client.More` - Add a space between `client.` and `More`

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,360 @@
...
+    "type": "json"
+}
+```
+
+Behaviour requirements:
+
+- Unlike `fields`, the fields in `include` _do not have to exist_ in the source
+    document in order that the document be included in the index. This is to
+    allow the index to cover more queries.
+- Including a deeply nested field would follow the same pattern as for other
```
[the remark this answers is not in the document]
mikerhodes: Goodness -- great point :tada:

I think that we could validate the length of the list of fields when the ddoc is updated, rather than failing during indexing. We could also limit the depth by counting the `.` characters.

One thing we can only validate at index time are things like the length of included strings. I think here that we might want to place a limit on the total size of the values, say 32kb. Even that's quite a few disk pages.

Given it's easier to start with limits and increase them later, perhaps we should think about this more deeply. In a view we allow ~anything I believe, but here potentially we could be more conservative.

As an example, [postgres](https://www.postgresql.org/docs/current/limits.html) limits indexes to 32 columns. Its max field size is 1GB; I think we'd like something a little smaller 😬 

Are there other limits here?

My thought is that we do limit, and make it configurable, and perhaps start relatively low for the defaults:

- `mango_json_index_include_fields_max=16` (why 16? Powers of two always sound nice)
- `mango_json_index_include_depth_max=8`
- `mango_json_index_include_size_bytes_max=32768` (32kb)

We can enforce `mango_json_index_include_fields_max` and `mango_json_index_include_depth_max` in `_index`. (We may have to belt-and-braces this as the user can go behind Mango's back to upload views that are the "right shape").

`mango_json_index_include_size_bytes_max` would need to be checked per document at index time. I worry what the behaviour should be here -- I see options of marking the whole index bad; having rows with "missing" values fields, meaning complexity during query; skipping indexing the document entirely. I lean towards skipping the doc as the least likely to cause unpredictable behaviour, but what's the current behaviour for views if indexing a doc fails?

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: 91f99be2d

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: 91f99be2d

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: 91f99be2d

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: 91f99be2d

src/docs/rfcs/018-mango-covering-json-index.md
mikerhodes: 91f99be2d

src/docs/rfcs/018-mango-covering-json-index.md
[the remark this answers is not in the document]
mikerhodes: I've added a section on limits for `include` in 91f99be2d. I think that the topic is really important and it's a bit facepalm that I skipped it. Too excited about writing Erlang I guess 😬 What do you think?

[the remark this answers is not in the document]
nickva: Looks great. Thank you!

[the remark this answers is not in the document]
nickva: I am not sure what the behavior for `mango_json_index_include_size_bytes_max` should be either. It's pretty tricky. From the top of my head I don't know what currently happens if we fail to index a document. With mango we don't usually fail, if I had to guess it would be we end up in a crash loop on that document. A user then may index for a day and hit a large field value their index crashes, they'd remove the field, try again (get a new view signature) , index for 2 days and crash, etc.

The alternative is to skip the doc but then there is danger of it looking like data loss - user has some very important document (medical record, say allergies to medicine), index skips it, nobody notices until the patient is prescribed the wrong medication (sorry being, a bit dramatic here with a silly user story).  Maybe that's something we solve outside of the RFC and just that if the value made it past the max_document_size limit then it will be indexed or crash horribly...?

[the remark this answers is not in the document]
nickva: What if instead of storing plain values in rows we'd allow also storing a marker which would indicate that a value was too large, and just fallback to reading the doc? We can still have a hard-limit (strict) mode perhaps which deals with failures, but this would allows us not to deal with indexing failures so to speak and punt it for later.

There may be an optimal cut-off value where storing it in the index is more wasteful than reading the doc? Or there may be not, as technically I think we can write arbitrarily large values in the index, it will just spread over a lot of b-tree blocks, but we'd still duplicate the data.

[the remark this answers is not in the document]
mikerhodes: > What if instead of storing plain values in rows we'd allow also storing a marker which would indicate that a value was too large, and just fallback to reading the doc?

I'd like to avoid doing this if we can -- while nice in some ways, this approach makes it more difficult for a customer to understand the performance of their query. I'd like the performance of queries to be predictable. My feeling is that large fields should require a trip to primary data, and that enforcing smaller value sizes keeps it more likely that indexes can live in RAM, where, I think, they should be.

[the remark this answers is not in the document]
janl: We discussed this a bit in [the developer meeting yesterday](https://lists.apache.org/thread/5x22cl09jv5f9o82v3cw3opvpz9ywlqg).

First, we compared this to how this would manifest in M/R views, and the example we used was what if a doc causes a JS exception to be thrown. What we do there is skip the doc and not include any result rows. We do log this in couch.log but it is not surfaced to the HTTP API user.

We went through the motions of whether to add fields inline with the result and also recording a metric for this and concluded from experience that most people do not really care about these things.

However some folks take this seriously, and we want to accommodate those. Imagine this:

1. we increment a metrics each time a mango covering index can’t include a doc because of some limit.
2. a user sees this number growing and finds out that this isn’t good.
3. next they want to know which document caused this and we can point them to couch.log where this should appear (TODO: decide log level)
4. finally, it’d be great if there was a variant to `_explain`, say: `POST /db/_explain?include_result=true` which returns the result like normal, but rows that are missing have an error object in them (or maybe we just show the error rows)

So some handwaving to be defined away, but: let’s do this right this time :)

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,397 @@
...
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.txt).
+
+## Terminology
+
+[TIP]:  # ( Provide a list of any unique terms or acronyms, and their definitions here.)
+
+- Mango: CouchDB's Mongo inspired querying system.
+- View / JSON index: Mango index that uses the same index as Cloudant views.
+- Coordinator: the Erlang process that handles doing a distributed query across
+    a CouchDB cluster.
```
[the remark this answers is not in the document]
mikerhodes: I tend towards four spaces as, at least a few years ago, that used to be safest across Markdown renderers. For consistency, I've found the one item that _wasn't_ indented in that style and corrected it. Given CommonMark states that you'd use two spaces in this context, in future I'll probably drop this habit. But for now I've kept it.

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,397 @@
...
+
+A Mango JSON index is implemented as a view with a complex key. The first field
+in the index is the first entry in the complex key, the second field is the
+second key and so on. Even indexes with one field use a complex key with length
+`1`.
+
+When choosing a JSON index to use for a query, there are a couple of things that
+are important to covering indexes.
+
+Firstly, note there are certain predicate operators that can be used with an
```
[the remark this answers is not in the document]
mikerhodes: I think they were just the easiest to convert into startkey/endkey in JSON indexes. Text indexes, ie Lucene-backed, came later. I'm not sure what the rules are in the text cursor, but these are the rules for JSON. Likely the set can be increased (`$in` comes up commonly).

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,397 @@
...
+that `include_docs=true` will be set when `r<2` as today.
+
+I suspect we'll set an option in `viewcbargs` that contains the index field
+names and whether it's a covering index. This means that an updated shard will
+be checking for those fields. When it can't find them, it'll fallback to the
+current behaviour in `view_cb/2`, meaning that it reads the document found via
+`include_docs=true`, execute `match_and_extract_doc/3` and return the row if it
+matches the query.
+
+The coordinator will received the final result document as today and assume it's
```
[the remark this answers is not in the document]
mikerhodes: 6f5dd5f5d

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,397 @@
...
+Behaviour requirements:
+
+- Unlike `fields`, the fields in `include` _do not have to exist_ in the source
+    document in order that the document be included in the index. This is to
+    allow the index to cover more queries.
+- Including a deeply nested field would follow the same pattern as for other
+    field references in mango, `person.address.zip`.
+- There is no notation to include the whole document, that is, no equivalent of
+    `emit(doc.name, doc)`.
+- `"include": []` is equivalent to omitting the `include` field.
```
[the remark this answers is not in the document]
mikerhodes: 6f5dd5f5d

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,397 @@
...
+
+The `selector` `{"age": {"$gt": 30}}` should return both documents. However, if
+we use the index above, we'd miss out `bar` because it's not in the index.
+Therefore we can't use the index.
+
+On the other hand, the `selector` `{"age": {"$gt": 30}, "name":
+{"$exists"=true}}` requires that the `name` field exist so the index can be used
+because the query predicates can only match documents containing both `age` and
+`name`, just like the index. In both cases, note the predicate `"age": {"$gt":
+30}` implies `"age": {"$exists"=true}`.
```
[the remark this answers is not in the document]
mikerhodes: 6f5dd5f5d

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,397 @@
...
+
+{"_id": "bar", "age": 39, "pet": "cat"}
+```
+
+The `selector` `{"age": {"$gt": 30}}` should return both documents. However, if
+we use the index above, we'd miss out `bar` because it's not in the index.
+Therefore we can't use the index.
+
+On the other hand, the `selector` `{"age": {"$gt": 30}, "name":
+{"$exists"=true}}` requires that the `name` field exist so the index can be used
```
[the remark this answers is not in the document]
mikerhodes: 6f5dd5f5d

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,397 @@
...
+A Mango JSON index is implemented as a view with a complex key. The first field
+in the index is the first entry in the complex key, the second field is the
+second key and so on. Even indexes with one field use a complex key with length
+`1`.
+
+When choosing a JSON index to use for a query, there are a couple of things that
+are important to covering indexes.
+
+Firstly, note there are certain predicate operators that can be used with an
+index, currently: `$lt`, `$lte`, `$eq`, $gte` and `$gt`. These can easily be
```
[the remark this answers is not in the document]
mikerhodes: 6f5dd5f5d

src/docs/rfcs/018-mango-covering-json-index.md
```diff
@@ -0,0 +1,397 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.txt).
+
+## Terminology
+
+[TIP]:  # ( Provide a list of any unique terms or acronyms, and their definitions here.)
+
+- Mango: CouchDB's Mongo inspired querying system.
```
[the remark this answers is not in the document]
mikerhodes: 6f5dd5f5d

PR #4411, opened 2023-02-02
docs(INSTALL.Unix): update instructions

pgj:
This change aims to correct the Unix installation instructions in different ways:

- Note that a too new version of the Erlang OTP can cause build problems.
- On recent macOS systems, the `xcode-select --install` command will install a version of Python 3, along with `pip3`, hence there is no need to do that explicitly.
- Sphinx and the respective theme are installed locally via `venv` for building the documentation, so users should only provide a working Python 3 environment (thanks @nickva for mentioning this).
- Fix markdown of file and command names.

## Checklist


nickva (approved):
+1 thank you for your contribution @pgj 

Sorry you had to encounter the python-black issue, it does happen once in a while. Thanks for fixing up the python code, too!

PR #4419, opened 2023-02-09
docs: update developer instructions

pgj:
The JavaScript-based integration test suites are still referenced by the developer documentation which can create confusions.  Remove those references and mention that other ones that have been added to the `Makefile` in the meantime.


nickva (approved):
+1 thank you for your contribution, @pgj

PR #4420, opened 2023-02-11
Reorder top-level smoosh-settings to the correct section

big-r81:
Moving some top-level settings into the top-level section
to prevent an accidential setup (if enabled) in the wrong 
section (smoosh.slack_views).


nickva (approved):
+1

Good find, @big-r81

PR #4426, opened 2023-02-15
docs/api: use {params} consistently in titles

alxndrsn:
Currently the doc section titles look like this:

Screenshot_2023-02-15_11-22-26

The documentation suggests that `/db/` is the correct form:

Screenshot_2023-02-15_11-31-54

This PR updates inconsistent titles and the suggested format to use `/{db}`.


big-r81:
Hi,

thank you for digging into this topic. I think it should be `{db}` for all url paths, because it should show the user, thats a dnamic value for a database name. If it's `db` then it looks like a static url path, which it isn't. An equivalent is `{node_name}` which should show the user, that the path is dynamic - dependent of the given node name.

alxndrsn:
@big-r81 I agree that would be clearer.  If that's the preferred approach, I can change the PR to use the `/{db}/` format, and remove the guidance shown in the second screenshot.

nickva:
Thank you, @alxndrsn for a very nice improvement @big-r81, for the suggestion to make it consistent.

Wonder if it would be possible to do the same to the rest of the API index?

Screen Shot 2023-02-15 at 4 26 52 PM

I found at least these additional files:

```
api/ddoc/common.rst
api/ddoc/render.rst
api/ddoc/rewrites.rst
api/ddoc/search.rst
api/ddoc/views.rst
api/document/common.rst
api/document/attachments.rst
api/partitioned-dbs.rst
api/local.rst
```

alxndrsn:
@nickva I've updated for all path variables:

Screenshot_2023-02-17_10-23-18
Screenshot_2023-02-17_10-23-02

nickva (approved):
+1

That looks great, Alex.

Thank you for your contribution!

PR #4429, opened 2023-02-16
fix(mango): fix specification of `choose_best_index/1`

pgj:
Comparators are not represented by binary strings in the selection ranges, captured by the `range/0` type.  Although that is how they are coming from the corresponding parsed JSON object, they are being translated to specific atoms on the fly.  Thanks for @nickva noticing this!


nickva (approved):
+1

PR #4430, opened 2023-02-16
docs(`README-DEV`): chase rename of `dreyfus-test` to `elixir-search`

pgj:
This is a documentation follow-up for #4425.


jaydoane (approved):
Good catch!

PR #4445, opened 2023-02-25
Introduce ad-hoc Erlang cookie configuration for the `dev` scripts

pgj:
Extend the `dev/run` and `dev/remsh` scripts in ways that would let the users to set the cookie for inter-node communication directly from the command line.  This may come handy in certain cases, e.g. when there is a specific Clouseau instance to hook up.


rel/overlay/etc/vm.args
```diff
@@ -38,8 +38,7 @@
 {{node_name}}
 
 # All nodes must share the same magic cookie for distributed Erlang to work.
-# Uncomment the following line and append a securely generated random value.
-# -setcookie
+{{erlang_cookie}}
```
big-r81, line 41: Hi,

can this change break "couchdb-pkg" somewhere, if we are maybe grepping for "-setcookie"? @rnewson

pgj: I have not worked with `couchdb-pkg` before but I can verify this concern there.

janl: https://github.com/apache/couchdb-pkg/blob/d284420d445573c726b4fcf1d8940ce22793e364/rpm/SPECS/couchdb.spec.in#L155

https://github.com/apache/couchdb-pkg/blob/d284420d445573c726b4fcf1d8940ce22793e364/debian/couchdb.postinst#L165

https://github.com/apache/couchdb-pkg/blob/d284420d445573c726b4fcf1d8940ce22793e364/debian/couchdb.config#L85

pgj: Thanks @jan for the references.

dev/remsh
```diff
@@ -23,6 +23,12 @@ if [ -z $HOST ]; then
     HOST="127.0.0.1"
 fi
 
+if [ -n "$COOKIE" ]; then
+    _cookie_flag="-setcookie $COOKIE"
+else
+    _cookie_flag=""
```
rnewson, line 29: any reason for the else clause when empty is the default?

rel/overlay/etc/vm.args
rnewson: The package scripts should be (_need to be_)_ robust against this kind of thing and I believe they are.

If the commented-out setcookie line is no longer present after this change the packaging scripts will append a setcookie line to the end of the file. That's less tidy than it being where we planned for it to be, but it will still work.

I would, however, prefer for this change not to have unintended side-effects at all.

dev/remsh
pgj: No, there is not any :-)  Thanks for catching this.

PR #4447, opened 2023-02-27
Update last_check in file logger record

KangTheTerrible:
## Overview

While investigating issues in my deployment around logrotate I discovered the file logger last_check doesn't appear to be updated after init is called, this could result in maybe_reopen being called with every log write 30 seconds after init is called. The fix works as the documentation specifies, but I wasn't able to measure a meaningful performance harm from the previous functionality. Throwing it up in case it's desirable.

## Testing recommendations

With the fix observe it can take up to 30 seconds from the last check before the log is recreated after deletion. Without the fix the log gets recreated much more quickly.

## Checklist


nickva (approved):
+1 

Excellent find, @KangTheTerrible. Thanks for your contribution!

PR #4458, opened 2023-03-07
mango: correct text index selection for queries with `$regex`

pgj:
Upon adding a text index, Mango queries may not return documents that should match a selector that contains the `$regex` operator.  Steps to reproduce (thanks to @willholley), that are also captured in the integration tests:

1. Create an empty database. 
2. Add a document:

```json
{ 
    "_id": "foo",
    "currency": "HUF",
    "location": "EUROPE"
}
```

3. Run a Mango query without indexes -- this should return the document above:

```json
{
    "selector": {
        "currency": {
            "$regex": "HUF"
        }
    }
}
```

4. Add a text index for `location`:

```json
{
    "index": {
        "fields": [
            {
                "name": "location",
                "type": "string"
            }
       ]
   },
   "name": "location-text",
   "type": "text"
}
```

5. Run the query once more -- this time, no documents are returned.

This is because the text index is chosen to answer the query even if it does not contain the field on which the `$regex` match should be run.  This is due to the permissive logic of the index selection where the fields with `$regex` in the selector are not cross-checked against the fields in the index.

The PR implements the necessary changes to ensure that text indexes are chosen more carefully and covers them with tests.  Note that this also means an increase in the unit test coverage of the related functions to be less dependent on a working Clouseau instance.  As an extra, it aims to improve the readability by introducing extra commentary (thanks to @mikerhodes) and type specifications.

Random stats on coverage, before:

```console
$ make eunit apps=mango
==> mango (compile)
==> rel (compile)
==> couchdb (compile)
WARN:  Missing plugins: [pc]
==> couchdb (setup_eunit)
==> mango (eunit)
[..]
  All 55 tests passed.
[..]
Code Coverage:
Code Coverage:
mango_app             :   0%
mango_crud            :   0%
mango_cursor          :   0%
mango_cursor_special  :   0%
mango_cursor_text     :   0%
mango_cursor_view     :  33%
mango_doc             :   3%
mango_epi             : 100%
mango_error           :   0%
mango_execution_stats :   0%
mango_fields          :  50%
mango_httpd           :   0%
mango_httpd_handlers  :   0%
mango_idx             :  18%
mango_idx_special     :   0%
mango_idx_text        :  29%
mango_idx_view        :   1%
mango_json            :  10%
mango_json_bookmark   :   0%
mango_native_proc     :   5%
mango_opts            :  25%
mango_selector        :  52%
mango_selector_text   :   0%
mango_sort            :   0%
mango_sup             :   0%
mango_util            :  23%

Total                 : 17%
==> rel (eunit)
==> couchdb (eunit)
```

after:

```console
$ make eunit apps=mango
==> mango (compile)
==> rel (compile)
==> couchdb (compile)
WARN:  Missing plugins: [pc]
==> couchdb (setup_eunit)
==> mango (eunit)
[..]
module 'mango_idx_text'
[..]
  mango_idx_text: indexable_fields_test...[0.106 s] ok
  mango_idx_text: is_usable_test...ok
  [done in 5.703 s]
[..]
module 'mango_selector_text'
  mango_selector_text: convert_fields_test...ok
  mango_selector_text: convert_default_test...ok
  mango_selector_text: convert_lt_test...ok
  mango_selector_text: convert_lte_test...ok
  mango_selector_text: convert_eq_test...ok
  mango_selector_text: convert_ne_test...ok
  mango_selector_text: convert_gte_test...ok
  mango_selector_text: convert_gt_test...ok
  mango_selector_text: convert_all_test...ok
  mango_selector_text: convert_elemMatch_test...ok
  mango_selector_text: convert_allMatch_test...ok
  mango_selector_text: convert_keyMapMatch_test...ok
  mango_selector_text: convert_in_test...ok
  mango_selector_text: convert_nin_test...ok
  mango_selector_text: convert_exists_test...ok
  mango_selector_text: convert_type_test...ok
  mango_selector_text: convert_mod_test...ok
  mango_selector_text: convert_regex_test...ok
  mango_selector_text: convert_size_test...ok
  mango_selector_text: convert_not_test...ok
  mango_selector_text: convert_and_test...ok
  mango_selector_text: convert_or_test...ok
  mango_selector_text: convert_nor_test...ok
  mango_selector_text: to_query_test...ok
  [done in 0.074 s]
[..]
  All 81 tests passed.
[..]
Code Coverage:
mango_app             :   0%
mango_crud            :   0%
mango_cursor          :   0%
mango_cursor_special  :   0%
mango_cursor_text     :   0%
mango_cursor_view     :  33%
mango_doc             :   3%
mango_epi             : 100%
mango_error           :   0%
mango_execution_stats :   0%
mango_fields          :  50%
mango_httpd           :   0%
mango_httpd_handlers  :   0%
mango_idx             :  18%
mango_idx_special     :   0%
mango_idx_text        :  61%
mango_idx_view        :   1%
mango_json            :  10%
mango_json_bookmark   :   0%
mango_native_proc     :   5%
mango_opts            :  25%
mango_selector        :  62%
mango_selector_text   :  86%
mango_sort            :   0%
mango_sup             :   0%
mango_util            :  38%

Total                 : 31%
==> rel (eunit)
==> couchdb (eunit)
```

### DIFF
--- a/src/couch/src/test_util.erl
+++ b/src/couch/src/test_util.erl
@@ -37,6 +37,8 @@
 
 -export([shuffle/1]).
 
+-export([as_selector/1]).
+
 -include_lib("couch/include/couch_eunit.hrl").
 -include_lib("couch/include/couch_db.hrl").
 -include("couch_db_int.hrl").
@@ -488,3 +490,7 @@ shuffle(List) ->
     Paired = [{couch_rand:uniform(), I} || I <- List],
     Sorted = lists:sort(Paired),
     [I || {_, I} <- Sorted].
+
+%% Create a valid Mango selector from an Erlang map.
+as_selector(Map) ->
+    mango_selector:normalize(jiffy:decode(jiffy:encode(Map))).
--- a/src/mango/src/mango.hrl
+++ b/src/mango/src/mango.hrl
@@ -11,3 +11,12 @@
 % the License.
 
 -define(MANGO_ERROR(R), throw({mango_error, ?MODULE, R})).
+
+-type abstract_text_selector() :: {'op_and', [abstract_text_selector()]}
+				| {'op_or', [abstract_text_selector()]}
+				| {'op_not', {abstract_text_selector(), abstract_text_selector()}}
+				| {'op_not', {_, 'false'}}
+				| {'op_field', {iolist() | binary(), _}}
+				| {'op_fieldname', {_, _}}
+				| {'op_null', {_, _}}
+				| {'op_default', _}.
--- a/src/mango/src/mango.hrl
+++ b/src/mango/src/mango.hrl
@@ -19,4 +19,5 @@
 				| {'op_field', {iolist() | binary(), _}}
 				| {'op_fieldname', {_, _}}
 				| {'op_null', {_, _}}
-				| {'op_default', _}.
+				| {'op_default', _}
+				| {'op_regex', binary()}.
--- a/src/mango/src/mango_idx_text.erl
+++ b/src/mango/src/mango_idx_text.erl
@@ -29,6 +29,10 @@
 -include("mango.hrl").
 -include("mango_idx.hrl").
 
+-ifdef(TEST).
+-import(test_util, [as_selector/1]).
+-endif.
+
 validate_new(#idx{} = Idx, Db) ->
     {ok, Def} = do_validate(Idx#idx.def),
     maybe_reject_index_all_req(Def, Db),
@@ -127,6 +131,14 @@ columns(Idx) ->
                 )
     end.
 
+% Mind that `is_usable/3` is not about "what fields can be answered by
+% the index" but instead more along the lines of "this index will
+% ensure that all the documents that should be returned for the query
+% will be, because we checked that all the bits of the query that
+% imply `$exists` for a field are used when we check that the indexing
+% process will have included all the relevant documents in the index".
+-spec is_usable(#idx{}, SelectorObject, _) -> boolean() when
+    SelectorObject :: any().
 is_usable(_, Selector, _) when Selector =:= {[]} ->
     false;
 is_usable(Idx, Selector, _) ->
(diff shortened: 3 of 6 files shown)


janl (approved):

jaydoane:
I was concerned that during a rolling upgrade containing this change, it might be possible for a selector containing the new `op_regex` to be processed by a node that has not yet been upgraded, so I attempted to test such a configuration.

I configured a dev cluster with `node1` running on  https://github.com/apache/couchdb/commit/f40147b262ca007d07bb5f48c345b31b46b80629, and `node2` & `node3` on https://github.com/apache/couchdb/commit/8ec7b578848903b9a152934c7ef0f57e69d8946d. This was more difficult than I expected, and required:
- Turning off the code reloader I normally use.
- Making the mango ebin directory sticky so it wouldn't automatically reload updated code.
- Making sure the old code was loaded on nodes 2 & 3, so it would stick there.
- Checking the md5 hashes of e.g. `mango_idx_text` on different nodes to ensure they were as expected.

But once I had created the desired configuration, I was unable to induce any errors using the document, selector, and index in the above example. 

I also ran the python mango tests (modified to create dbs with q=2,n=3) against that mixed cluster, and likewise didn't see any errors stemming from `op_regex`.

At this point, I'm fairly convinced that this change shouldn't cause problems during rolling upgrades, but I'm still open to a good argument one way or another.

pgj:
Thanks @jaydoane for testing this change more extensively in a production environment and expecting us to come up with a good reasoning.

I am still new to this area but I believe introduction of `op_regex` itself will not cause any problems with upgrades.  The reason is that it is part of an internal data structure that is not shared between nodes during the evaluation (not passed around as a message), it is more like a helper to implement the selector-to-Lucene-query translation which happens in a single place.

The only potential problem I see in mixed-version environments is that one may node handle the given query differently due to the induced change in the index selection process.  But that is not because of `op_regex`, it is only a vehicle of the fix.

jaydoane:
> I am still new to this area but I believe introduction of op_regex itself will not cause any problems with upgrades. The reason is that it is part of an internal data structure that is not shared between nodes during the evaluation (not passed around as a message), it is more like a helper to implement the selector-to-Lucene-query translation which happens in a single place.

This is exactly what I was unsure about, especially with Mike's recent optimization to push selectors to the individual shards. I could have added some debug logging to prove to myself it was safe, but I'm happy to hear your assurances!

> The only potential problem I see in mixed-version environments is that one may node handle the given query differently due to the induced change in the index selection process

Yes, I noticed that on node1 with your fix, it returned the document as expected, but on the other nodes, it still returned
```json
{
  "docs": [],
  "bookmark": "g2o"
}
```
which seems expected as well :)

PR #4460, opened 2023-03-09
docs(`_find`): tweak formatting and style

pgj:
While reading the documentation, noticed a couple of problems with the HTML version, of which this change covers:

- Parameter types for JSON objects are not uniform.  The type `json` could either ambigiously refer to objects or arrays and used randomly.
- The `Content-Type` header for response descriptions is not always put in a list, though probably that is the right choice because there can be many.
- Indentation of paragraphs and list items is off sometimes which causes rendering mistakes.
- Formatting of JSON examples is not uniform by indentation.
- Add a couple of missed markdown for names, such as operators, to make them stand out from the text for better readability.
- Unify style: use colons for the sentences that introduce examples, end sentences with full stops, lose contractions, regularize elaboration of object details for index descriptions and definitions.
- Fix the reference to views in design documents when talking about indexes.
- Add link for `_find` at the section about the index selection.

### DIFF
--- a/src/docs/src/api/database/find.rst
+++ b/src/docs/src/api/database/find.rst
@@ -29,14 +29,14 @@
 
     :<header Content-Type: - :mimetype:`application/json`
 
-    :<json json selector: JSON object describing criteria used to select
+    :<json object selector: JSON object describing criteria used to select
         documents. More information provided in the section on :ref:`selector
         syntax <find/selectors>`. *Required*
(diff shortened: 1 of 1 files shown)


nickva (approved):
A very nice cleanup, thanks @pgj!

+1

big-r81 (approved):
+1

Thank you for your contribution.

PR #4479, opened 2023-03-16
Migrate configure settings to Windows

big-r81:
Migrate the settings `with_proper` and `erlang_md5` to `configure.ps1` to add it to `config.erl`.


nickva (approved):
+1

`--eralng-md5` configure flag is headed to be deprecated since it can be [enabled at runtime](https://github.com/apache/couchdb/pull/4443), but since it's minimal effort to keep both in sync that's fine to update it to work similar on windows. But if it was a larger piece of work, we could have skipped it.

PR #4483, opened 2023-03-21
Add report logging

chewbranca:
<!-- Thank you for your contribution!

     Please file this form by replacing the Markdown comments
     with your text. If a section needs no action - remove it.

     Also remember, that CouchDB uses the Review-Then-Commit (RTC) model
     of code collaboration. Positive feedback is represented +1 from committers
     and negative is a -1. The -1 also means veto, and needs to be addressed
     to proceed. Once there are no objections, the PR can be merged by a
     CouchDB committer.

     See: http://couchdb.apache.org/bylaws.html#decisions for more info. -->

## Overview

<!-- Please give a short brief for the pull request,
     what problem it solves or how it makes things better. -->

This PR introduces a mechanism for logging metadata about requests and system activity outside of the normal http log entry. In particular, this work is meant to provide tooling for logging things like the quantity of work performed to fulfill a particular request.

For example, given we make a `_find` query like so:

```
(chewbranca)-(jobs:0)-(~)                                                                                                                                                                                                                                                                                                   
(! 2793)-> curl  -u adm:pass -X POST localhost:15984/foo/_find?include_docs=true -d '{"selector":{"blah":"blah blah"},"use_index":"_design/67ca9ba1e20371ee9d7e703c1f04eb1d03521b57","execution_stats":true}' -H content-type:application/json                                                                              
{"docs":[                                                                                                                                                                                                                                                                                                                   
{"_id":"493609ddc70fa41dec7ec28fc3000bf0","_rev":"1-6440c84fecde0669be847bf9639110b5","blah":"blah blah"},                                                                                                                                                                                                                  
...*SNIP*...                                                                                                                                                                                                                                                                                                                
{"_id":"7ea025f22312db0d494f7a3d0f2614d4","_rev":"1-6440c84fecde0669be847bf9639110b5","blah":"blah blah"},                                                                                                                                                                                                                  
{"_id":"7ea025f22312db0d494f7a3d0f2622ee","_rev":"1-6440c84fecde0669be847bf9639110b5","blah":"blah blah"}                                                                                                                                                                                                                   
],                                                                                                                                                                                                                                                                                                                          
"bookmark": "g1AAAABfeJzLYWBgYMpgSmHgKy5JLCrJTq2MT8lPzkzJBYormKcmGhiZphkZGRsapSQZpJhYmqSZJxqnGKQZmRkZpaaC9HHA9OUAdTCCtHEm5SRmKICIrCwAuMMZsw",                                                                                                                                                                               
"execution_stats": {"total_keys_examined":0,"total_docs_examined":1243,"total_quorum_docs_examined":0,"results_returned":1243,"execution_time_ms":190.988}}                                                                                                                                                                 
```

This PR will generate a log entry report of the form:

> [report] 2023-03-20T23:39:03.763470Z node1@127.0.0.1 <0.7071.0> ecded843da mango-stats [type="mango-stats" total_quorum_docs_examined="0" total_keys_examined="0" total_docs_examined="1243" results_returned="1243" nonce="ecded843da" execution_time_ms="190.988"]

The report metadata is logged in format compliant with [RFC5424](https://www.rfc-editor.org/rfc/rfc5424.html#section-6.3) such that syslog can pickup the metadata entries. This allows us to extend what metadata is introduced over time in such a way that parsing at the log/syslog level will not need further modifications for new data. This means we can easily introduce additional data logging around volume or IO operations or other statistics at a lower granularity than the doc level.

It could be argued that this data should be directly coupled with the underlying http log, and I offer two counterpoints:

  1) currently `mango_crud:find/5` returns an accumulated response, rather than a request handle, so we lose the metadata generated inside `mango` when it passes back into the `chttpd` layer, hence `chttpd` no longer has the relevant stats to include in the log when we finally log the post http request entry in `chttpd`. We _could_ modify the relevant code paths to introduce a data structure containing both the response and metadata values allowing us to funnel data to be logged up from lower levels in the system than chttpd; I think this is potentially worthwhile, but I also think item 2) below makes this less of an immediate concern.

  2) I believe strongly we need the ability to collect data about the overhead performed by a particular task and then log that data such that we can more clearly articulate what CouchDB is doing as a function of time. This is certainly incredibly important for http requests that generate load to understand expensive requests, but also just as important for background operations as we have some blind spots in the amount of `couch_btree:foldl` operations induced per operation, be it a view/_all_docs or a background job like compaction of view building. Therefore we have areas to generate these report logs where there's no existing http request context to hook into. Furthermore, I think it's worthwhile to be able to record stats along these lines on the RPC side rather than funneling all the data back to the coordinator to do the logging there.

I do think there's merit to 1) above, however, I think 2) necessitates report logging outside of an http context so I'm not convinced we should do 1) as part of this PR. That said, feedback is welcome.

<hr />

The current structure of this PR is functional and allows for a hybrid approach for the report logging such that the RFC5424 metadata can be introduced _along with_ an additional arbitrary string message afterwards. While working through the logic here, in practice, supporting the hybrid approach complicates the underlying logic in the codebase and also seems to complicate the ability for RFC5424 parsers to extract the metadata and parse the log entry. For example, here's a [RFC5424 parser Splunk plugin](https://splunkbase.splunk.com/app/978) I've been looking at and one of the caveats is:

> The MSG section of the event, if it exists, is not parsed by this app.

and a sample message from RFC5424 demonstrating structured metadata followed by a string `MSG`:

```
           <165>1 2003-10-11T22:14:15.003Z mymachine.example.com
           evntslog - ID47 [exampleSDID@32473 iut="3" eventSource=
           "Application" eventID="1011"] BOMAn application
           event log entry...
```

So it's a valid log entry to introduce structured metadata _and also_ follow that with an arbitrary log `MSG` string, but like I said this seems to complicate a lot of things both from a parsing standpoint and the logistics of short circuiting empty messages and format strings through `couch_log_formatter`. I'm inclined to remove this functionality, especially since it feels that in practice, the most useful thing here is logging the metadata block, rather than an additional `MSG` payload.

I wanted to get broader feedback before going down that route of removing the additional MSG, so the PR is in the current form outlined above, but I've got an initial diff below to drop that functionality. This would also cleanup all the additional logic introduced by `couch_log:log/3 --> couch_log:log/5`, and so my preferred direction for this PR to go is to apply this diff and then drop out all support for metadata logging outside of a very specifically targeted `couch_log:report`. Here's the wip diff:

```diff
diff --git a/src/couch_log/src/couch_log.erl b/src/couch_log/src/couch_log.erl
index 379326b52..2fb5808bb 100644
--- a/src/couch_log/src/couch_log.erl
+++ b/src/couch_log/src/couch_log.erl
@@ -22,7 +22,6 @@
     alert/2,
     emergency/2,
     report/2,
-    report/4,
     set_level/1
 ]).
 
@@ -51,11 +50,7 @@ alert(Fmt, Args) -> log(alert, Fmt, Args).
 emergency(Fmt, Args) -> log(emergency, Fmt, Args).
 
 -spec report(string(), map()) -> ok.
-report(ReportId, Meta) when is_map(Meta) ->
-    report(ReportId, "", [], Meta).
-
--spec report(string(), string(), list(), map()) -> ok.
-report(ReportId, Fmt, Args, Meta0) when is_map(Meta0) ->
+report(ReportId, Meta0) when is_map(Meta0) ->
     Meta = case maps:is_key(type, Meta0) of
         true ->
             Meta0;
@@ -63,7 +58,7 @@ report(ReportId, Fmt, Args, Meta0) when is_map(Meta0) ->
             Meta0#{type => ReportId}
     end,
     couch_stats:increment_counter([couch_log, level, report]),
-    Entry = couch_log_formatter:format(report, self(), ReportId, Fmt, Args, Meta),
+    Entry = couch_log_formatter:format_report(self(), ReportId, Meta),
     ok = couch_log_server:log(Entry).
 
 -spec set_level(atom() | string() | integer()) -> true.
diff --git a/src/couch_log/src/couch_log_formatter.erl b/src/couch_log/src/couch_log_formatter.erl
index bf7571975..5b23f7ed5 100644
--- a/src/couch_log/src/couch_log_formatter.erl
+++ b/src/couch_log/src/couch_log_formatter.erl
@@ -22,6 +22,8 @@
     format/3,
     format/1,
 
+    format_report/3,
+
     format_reason/1,
     format_mfa/1,
     format_trace/1,
@@ -32,6 +34,28 @@
 
 -define(DEFAULT_TRUNCATION, 1024).
 
+-define(REPORT_LEVEL, report).
+
+format_report(Pid, Type, Meta) ->
+    MaxMsgSize = couch_log_config:get(max_message_size),
+    Msg = case format_meta(Meta) of
+        "" ->
+            "";
+        MetaStr when length(MetaStr) > MaxMsgSize ->
+            %% TODO: what to do when meta formatted data is too large?
+            error(what_to_do_here);
+        MetaStr ->
+            ["[", MetaStr, "]"]
+    end,
+    #log_entry{
+        level = couch_log_util:level_to_atom(?REPORT_LEVEL),
+        pid = Pid,
+        msg = Msg,
+        msg_id = couch_log_util:get_msg_id(),
+        time_stamp = couch_log_util:iso8601_timestamp(),
+        type = Type
+    }.
+
 format(Level, Pid, Fmt, Args) ->
     format(Level, Pid, undefined, Fmt, Args, #{}).
```

## Testing recommendations

<!-- Describe how we can test your changes.
     Does it provide any behaviour that the end users
     could notice? -->

<!-- If your changes affect multiple components in different
     repositories please put links to those issues or pull requests here.  -->

## Checklist

### DIFF
--- a/src/couch_log/src/couch_log_formatter.erl
+++ b/src/couch_log/src/couch_log_formatter.erl
@@ -17,6 +17,7 @@
 -module(couch_log_formatter).
 
 -export([
+    format/6,
     format/4,
     format/3,
     format/1,
@@ -32,12 +33,26 @@
 -define(DEFAULT_TRUNCATION, 1024).
 
 format(Level, Pid, Fmt, Args) ->
+    format(Level, Pid, undefined, Fmt, Args, #{}).
+
+format(report = Level, Pid, Type, Fmt, Args, Meta) ->
     #log_entry{
         level = couch_log_util:level_to_atom(Level),
         pid = Pid,
-        msg = maybe_truncate(Fmt, Args),
+        msg = maybe_truncate(Fmt, Args, Meta, false),
         msg_id = couch_log_util:get_msg_id(),
-        time_stamp = couch_log_util:iso8601_timestamp()
+        time_stamp = couch_log_util:iso8601_timestamp(),
+        type = Type
+    };
+
+format(Level, Pid, Type, Fmt, Args, Meta) ->
+    #log_entry{
+        level = couch_log_util:level_to_atom(Level),
+        pid = Pid,
+        msg = maybe_truncate(Fmt, Args, Meta, true),
+        msg_id = couch_log_util:get_msg_id(),
+        time_stamp = couch_log_util:iso8601_timestamp(),
+        type = Type
     }.
 
 format(Level, Pid, Msg) ->
@@ -365,9 +380,16 @@ format_args([H | T], FmtAcc, ArgsAcc) ->
     {Str, _} = couch_log_trunc_io:print(H, 100),
     format_args(T, ["~s" | FmtAcc], [Str | ArgsAcc]).
 
-maybe_truncate(Fmt, Args) ->
+maybe_truncate(Fmt, Args, Meta, TruncateMeta) ->
     MaxMsgSize = couch_log_config:get(max_message_size),
-    couch_log_trunc_io:format(Fmt, Args, MaxMsgSize).
+    case format_meta(Meta) of
+        "" ->
+            couch_log_trunc_io:format(Fmt, Args, MaxMsgSize);
+        MetaStr when TruncateMeta ->
+            couch_log_trunc_io:format(["[", MetaStr, "] " | Fmt], Args, MaxMsgSize);
+        MetaStr ->
+            ["[", MetaStr, "] " | couch_log_trunc_io:format(Fmt, Args, MaxMsgSize)]
+    end.
 
 maybe_truncate(Msg) ->
     MaxMsgSize = couch_log_config:get(max_message_size),
@@ -436,3 +458,29 @@ get_value(Key, List, Default) ->
 
 supervisor_name({local, Name}) -> Name;
 supervisor_name(Name) -> Name.
+
+format_meta(Meta) ->
+    %% https://www.rfc-editor.org/rfc/rfc5424.html#section-6.3
+    %% iut="3" eventSource="Application" eventID="1011"
+    string:join(maps:fold(fun(K, V, Acc) ->
+        [to_str(K, V) | Acc]
+    end, [], Meta), " ").
+
+%% passing complex terms as meta value is a mistake so we are going
+%% to eat it, because we cannot bubble up errors from logger
+%% Therefore following are not supported
+%% - lists
+%% - tuples
+%% - maps
+%% However we are not going to try to distinguish lists from string
+%% Atoms would be printed as strings
+to_str(K, _) when not (is_list(K) or is_atom(K)) ->
+    "";
+to_str(K, Term) when is_list(Term) ->
+    io_lib:format("~s=\"~s\"", [K, Term]);
+to_str(_K, Term) when is_tuple(Term) ->
+    "";
+to_str(_K, Term) when is_map(Term) ->
+    "";
+to_str(K, Term) ->
+    io_lib:format("~s=\"~p\"", [K, Term]).
--- a/src/couch_log/src/couch_log_formatter.erl
+++ b/src/couch_log/src/couch_log_formatter.erl
@@ -380,6 +380,17 @@ format_args([H | T], FmtAcc, ArgsAcc) ->
     {Str, _} = couch_log_trunc_io:print(H, 100),
     format_args(T, ["~s" | FmtAcc], [Str | ArgsAcc]).
 
+maybe_truncate("", [], Meta, TruncateMeta) ->
+    MaxMsgSize = couch_log_config:get(max_message_size),
+    case format_meta(Meta) of
+        "" ->
+            "";
+        MetaStr when TruncateMeta and length(MetaStr) > MaxMsgSize ->
+            %% TODO: what to do when meta formatted data is too large?
+            error(what_to_do_here);
+        MetaStr ->
+            ["[", MetaStr, "]"]
+    end;
 maybe_truncate(Fmt, Args, Meta, TruncateMeta) ->
     MaxMsgSize = couch_log_config:get(max_message_size),
     case format_meta(Meta) of
@@ -388,7 +399,9 @@ maybe_truncate(Fmt, Args, Meta, TruncateMeta) ->
         MetaStr when TruncateMeta ->
             couch_log_trunc_io:format(["[", MetaStr, "] " | Fmt], Args, MaxMsgSize);
         MetaStr ->
-            ["[", MetaStr, "] " | couch_log_trunc_io:format(Fmt, Args, MaxMsgSize)]
+            %% Subtract 3 for open/close bracket and space added below
+            MsgLength = length(MetaStr) + 3,
+            ["[", MetaStr, "] " | couch_log_trunc_io:format(Fmt, Args, MaxMsgSize - MsgLength)]
     end.
 
 maybe_truncate(Msg) ->
@@ -482,5 +495,7 @@ to_str(_K, Term) when is_tuple(Term) ->
     "";
 to_str(_K, Term) when is_map(Term) ->
     "";
+to_str(K, Term) when is_binary(Term) ->
+    io_lib:format("~s=\"~s\"", [K, Term]);
 to_str(K, Term) ->
     io_lib:format("~s=\"~p\"", [K, Term]).
--- a/src/couch_log/src/couch_log_formatter.erl
+++ b/src/couch_log/src/couch_log_formatter.erl
@@ -17,10 +17,11 @@
 -module(couch_log_formatter).
 
 -export([
-    format/6,
     format/4,
     format/3,
     format/1,
+
+    format_report/3,
 
     format_reason/1,
     format_mfa/1,
@@ -32,27 +33,38 @@
 
 -define(DEFAULT_TRUNCATION, 1024).
 
+-define(REPORT_LEVEL, report).
+
+format_report(Pid, Type, Meta) ->
+    MaxMsgSize = couch_log_config:get(max_message_size),
+    Msg = case format_meta(Meta) of
+        "" ->
+            "";
+        MetaStr ->
+            ["[", MetaStr, "]"]
+    end,
+    case length(Msg) > MaxMsgSize of
+        true ->
+            {error, emsgtoolong};
+        false ->
+            Entry = #log_entry{
+                level = couch_log_util:level_to_atom(?REPORT_LEVEL),
+                pid = Pid,
+                msg = Msg,
+                msg_id = couch_log_util:get_msg_id(),
+                time_stamp = couch_log_util:iso8601_timestamp(),
+                type = Type
+            },
+            {ok, Entry}
+    end.
+
 format(Level, Pid, Fmt, Args) ->
-    format(Level, Pid, undefined, Fmt, Args, #{}).
-
-format(report = Level, Pid, Type, Fmt, Args, Meta) ->
     #log_entry{
         level = couch_log_util:level_to_atom(Level),
         pid = Pid,
-        msg = maybe_truncate(Fmt, Args, Meta, false),
+        msg = maybe_truncate(Fmt, Args),
         msg_id = couch_log_util:get_msg_id(),
-        time_stamp = couch_log_util:iso8601_timestamp(),
-        type = Type
-    };
-
-format(Level, Pid, Type, Fmt, Args, Meta) ->
-    #log_entry{
-        level = couch_log_util:level_to_atom(Level),
-        pid = Pid,
-        msg = maybe_truncate(Fmt, Args, Meta, true),
-        msg_id = couch_log_util:get_msg_id(),
-        time_stamp = couch_log_util:iso8601_timestamp(),
-        type = Type
+        time_stamp = couch_log_util:iso8601_timestamp()
     }.
 
 format(Level, Pid, Msg) ->
@@ -380,29 +392,9 @@ format_args([H | T], FmtAcc, ArgsAcc) ->
     {Str, _} = couch_log_trunc_io:print(H, 100),
     format_args(T, ["~s" | FmtAcc], [Str | ArgsAcc]).
 
-maybe_truncate("", [], Meta, TruncateMeta) ->
+maybe_truncate(Fmt, Args) ->
     MaxMsgSize = couch_log_config:get(max_message_size),
-    case format_meta(Meta) of
-        "" ->
-            "";
-        MetaStr when TruncateMeta and length(MetaStr) > MaxMsgSize ->
-            %% TODO: what to do when meta formatted data is too large?
-            error(what_to_do_here);
-        MetaStr ->
-            ["[", MetaStr, "]"]
-    end;
-maybe_truncate(Fmt, Args, Meta, TruncateMeta) ->
-    MaxMsgSize = couch_log_config:get(max_message_size),
-    case format_meta(Meta) of
-        "" ->
-            couch_log_trunc_io:format(Fmt, Args, MaxMsgSize);
-        MetaStr when TruncateMeta ->
-            couch_log_trunc_io:format(["[", MetaStr, "] " | Fmt], Args, MaxMsgSize);
-        MetaStr ->
-            %% Subtract 3 for open/close bracket and space added below
-            MsgLength = length(MetaStr) + 3,
-            ["[", MetaStr, "] " | couch_log_trunc_io:format(Fmt, Args, MaxMsgSize - MsgLength)]
-    end.
+    couch_log_trunc_io:format(Fmt, Args, MaxMsgSize).
 
 maybe_truncate(Msg) ->
     MaxMsgSize = couch_log_config:get(max_message_size),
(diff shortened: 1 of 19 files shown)


src/couch_log/src/couch_log_formatter.erl
```diff
@@ -365,9 +380,29 @@ format_args([H | T], FmtAcc, ArgsAcc) ->
...
     format_args(T, ["~s" | FmtAcc], [Str | ArgsAcc]).
 
-maybe_truncate(Fmt, Args) ->
+maybe_truncate("", [], Meta, TruncateMeta) ->
+    MaxMsgSize = couch_log_config:get(max_message_size),
+    case format_meta(Meta) of
+        "" ->
+            "";
+        MetaStr when TruncateMeta and length(MetaStr) > MaxMsgSize ->
+            %% TODO: what to do when meta formatted data is too large?
```
iilyak, line 389: If I understand the intend of the PR correctly the addition of `report` feature to logger serves a specific purpose. Which is to "report" a handful of integers along with the names of the keys.

Given the fact that we 1) do not support complex terms 2) report counters only; I think the best strategy here would be to drop the report and log an error instead. Because if it  happen it is either:

1. administrator set a `max_message_size` which is too small
2. the developer passing too much data to report call.

If we go with this route the general recommendation for developers would be to avoid using long names for counters.

src/mango/src/mango_execution_stats.erl
```diff
@@ -35,6 +37,15 @@ to_json(Stats) ->
         {execution_time_ms, Stats#execution_stats.executionTimeMs}
     ]}.
 
+to_map(Stats) ->
```
iilyak, line 40: The names of the fields are quite long. On the other hand they are crystal clear and it would be easy to read the logs and analyze the logs.

I don't have any opinion. Therefore, I would just state the fact. 

The current choice of names takes up at least 248 bytes in the log message (including date).

src/couch_log/src/couch_log_formatter.erl
```diff
@@ -436,3 +471,31 @@ get_value(Key, List, Default) ->
...
+    "";
+to_str(K, Term) when is_list(Term) ->
+    io_lib:format("~s=\"~s\"", [K, Term]);
+to_str(_K, Term) when is_tuple(Term) ->
+    "";
+to_str(_K, Term) when is_map(Term) ->
+    "";
+to_str(K, Term) when is_binary(Term) ->
+    io_lib:format("~s=\"~s\"", [K, Term]);
+to_str(K, Term) ->
```
iilyak, line 500: One question I had is whether we should limit precision of floats. I did a quick experiment to see whether "~p" limits the number of decimal places.

```erlang
6> io_lib:format("~p", [1.227773345398009300000000000000000000000000000000000000000000000000001]).
["1.2277733453980093"]
```

src/couch_log/src/couch_log_formatter.erl
chewbranca: In 8f3aabe2c I've switched over to `couch_log:report` returning true or false on successful report when the message is within bounds, and I've added a metric counter `couch_log.level.report_error` to track the error on too long of messages. My rationale is that we're not going to log data about the entry that was too long to be logged, so introducing a log message won't provide additional insight over a counter.

src/mango/src/mango_execution_stats.erl
chewbranca: Yeah... I thought about making shorter versions of the names, but those are the names as they are in the codebase, and gzip'ed log rotation reduces the impact of field length. I think more importantly though, we've already introduced those names into the public API as documented in the `Execution Statistics` portion of the `_find` documentation [1] and so I think it makes sense to stay consistent with those, or at the very least, we should update those as well if we're going to make shorter names here.


[1] https://docs.couchdb.org/en/stable/api/database/find.html#find-statistics

src/couch_log/src/couch_log_formatter.erl
chewbranca: Good call @iilyak. That seems fine to me, are you wanting to see more precise decimal formatting here?

iilyak: My note is for posterity so other reviewers do not spend time verifying this aspect. You can leave the implementation as is.

src/couch_log/src/couch_log_formatter.erl
```diff
@@ -31,6 +33,31 @@
...
+
+format_report(Pid, Type, Meta) ->
+    MaxMsgSize = couch_log_config:get(max_message_size),
+    Msg = case format_meta(Meta) of
+        "" ->
+            "";
+        MetaStr ->
+            ["[", MetaStr, "]"]
+    end,
+    case length(Msg) > MaxMsgSize of
```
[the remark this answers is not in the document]
pgj: Per line #44 `Msg` can also be `""`, where the size is `0`, cannot it be?

src/couch_log/src/couch_log_formatter.erl
```diff
@@ -31,6 +33,30 @@
...
+format_report(Pid, Type, Meta) ->
+    MaxMsgSize = couch_log_config:get(max_message_size),
+    Msg =
+        case format_meta(Meta) of
+            "" -> "";
+            Msg0 -> ["[", Msg0, "]"]
+        end,
+    case iolist_size(Msg) > MaxMsgSize of
+        true ->
+            {error, emsgtoolong};
```
iilyak, line 47: Would it be helpful to include the `ReportId` in the error message? Something like 

`{error, {emsgtoolong, ReportId}}`.

src/couch_log/src/couch_log_writer_syslog.erl
```diff
@@ -55,28 +57,44 @@ init() ->
...
+        report_level = Level,
+        enterprise_number = config:get("log", "syslog_enterprise_number")
     }}.
 
 terminate(_Reason, St) ->
     gen_udp:close(St#st.socket).
 
-write(Entry, St) ->
+write(#log_entry{level = report} = Entry, #st{enterprise_number = undefined} = St) ->
+    do_write(Entry#log_entry{level = error}, St);
```
iilyak, line 77: Is there a reason for changing a log level to `error`? I think more often then not the syslog error is used to signal about errors. I don't think the case when `enterprise_number` is not configured is an error. Omitting configuration of `enterprise_number` is a way to say I don't want structured logging, please give me the report as formatted key/value pairs. This is useful for cases when syslog server doesn't fully support SDATA. 

We could use `level = St#st.report_level`. In this case it would be up to the administrator to decide what level they want in case of a syslog server without SDATA support.

src/couch_log/test/eunit/couch_log_formatter_test.erl
```diff
@@ -25,6 +25,22 @@ truncate_test() ->
     Entry = couch_log_formatter:format(info, self(), Msg),
     ?assert(length(Entry#log_entry.msg) =< 16000).
 
+format_report_etoolong_test() ->
+    Payload = lists:flatten(["a" || _ <- lists:seq(1, 1048576)]),
```
iilyak, line 29:
```
lists:duplicate(1048576, $X).
```

src/couch_log/test/eunit/couch_log_formatter_test.erl
```diff
@@ -25,6 +25,22 @@ truncate_test() ->
...
+    }),
+    ?assertEqual({error, emsgtoolong}, Resp).
+
+format_report_test() ->
+    {ok, Entry} = couch_log_formatter:format_report(self(), report123, #{
+        foo => 123,
+        bar => "barStr",
+        baz => baz
+    }),
+    Formatted = "[foo=\"123\" baz=\"baz\" bar=\"barStr\"]",
```
iilyak, line 41: The expectation of a stable order of keys might not hold in future versions of OTP. The documentation for `maps:fold` used in `format_meta` states:

> Calls F(Key, Value, AccIn) for every Key to value Value association in MapOrIter in any order. 

I think we can accept the risk for a test case. However you can decide whether you want to handle it.

src/couch_log/src/couch_log_formatter.erl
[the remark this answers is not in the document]
chewbranca: Good catch. I switched to `iolist_size` which will handle empty strings, empty lists of strings, and other combinations:

```
Eshell V11.2.2.13  (abort with ^G)
1> iolist_size("").
0
2> iolist_size([""]).
0
3> iolist_size(["","",""]).
0
4> iolist_size(["","foo","", "bar"]).
6
```

src/couch_log/src/couch_log_formatter.erl
chewbranca: I left out an extra identifier like the `ReportId` as that value is provided by the caller into this function, so the caller that would receive `{error, emsgtoolong}` just provided the relevant `ReportId` and we're only incrementing a counter for any report error https://github.com/apache/couchdb/pull/4483/files#diff-d29be5c448a221047cbe9116da9f07ac471fd11716b561ab1f3979ac5683ce86R62-R65

I'm not necessarily opposed to the idea, I had similar thoughts, but I didn't see the value of adding it.

src/couch_log/src/couch_log_writer_syslog.erl
chewbranca: Well we need to change the level value to a syslog appropriate entry, but good catch, we should definitely see this to the configured `report_level` 👍

chewbranca: Fixed in f87be554d.

src/couch_log/test/eunit/couch_log_formatter_test.erl
chewbranca: Good call. Fixed in 9f1664c64.

src/couch_log/test/eunit/couch_log_formatter_test.erl
chewbranca: @iilyak hmmm... so I specifically wanted to avoid needing to sort the keys in the output string and to rely upon a fixed formatting, but you're correct in pointing out that our test expects a consistent output.

Again, I don't want to sort the keys in the report for the normal use case just to get a specific test result. The "fix" that jumps out to me would be to dissect the resulting output string and assert the expected keys. For instance, we could first regex pattern match on `^[` and `]$` to assert the formatted output is in brackets. Then we could split the inner string on spaces, sort the output, and then assert each of the key/value pairs matches up. I'm somewhat inclined to leave this and see if we encounter it as a problem down the road, but I think the approach I just outlined is viable if folks want to avoid that eventuality. Alternative solutions welcome too 👍

iilyak: > The "fix" that jumps out to me would be to dissect the resulting output string and assert the expected keys. For instance, we could first regex pattern match on ^[ and ]$ to assert the formatted output is in brackets. Then we could split the inner string on spaces, sort the output, and then assert each of the key/value pairs matches up.

Something like

```erlang

parse_data(String) ->
   maps:from_list(
       lists:map(fun(X) -> 
           [K, V] = string:split(X, "="),
           {list_to_atom(K), V} 
       end, string:split(
           string:trim(String, both, [$[, $]]), 
           " ", 
           all))).
```

Another approach is based on the fact that we don't need to fully parse the data. For us it is enough to get to the level of

```
[
"bar=\"barStr\"",
"baz=\"baz\"",
"foo=\"123\""
]
``` 

It could be done as following

```erlang
Expected = split_and_order("[foo=\"123\" baz=\"baz\" bar=\"barStr\"]"),
?assertEqual(Expected, lists:flatten(split_and_order(Entry#log_entry.msg))),
?assertEqual($[, lists:first(Entry#log_entry.msg)).
?assertEqual($], lists:last(Entry#log_entry.msg)).


def split_and_order(String) -> 
    lists:sort(
         string:split(
             string:trim(String, both, [$[, $]]), 
             " ", 
             all
         )
    ).
```

chewbranca: Thanks for the feedback. I agree the generic test version is quite a bit more complex and I'm not a huge fan. For now I've added a comment at the test location.

src/mango/src/mango_cursor_text.erl
```diff
@@ -127,9 +127,11 @@ execute(Cursor, UserFun, UserAcc) ->
             JsonBM = dreyfus_bookmark:pack(FinalBM),
             Arg = {add_key, bookmark, JsonBM},
             {_Go, FinalUserAcc} = UserFun(Arg, LastUserAcc),
-            FinalUserAcc0 = mango_execution_stats:maybe_add_stats(
+            {FinalUserAcc0, Stats1} = mango_execution_stats:maybe_add_stats(
                 Opts, UserFun, Stats0, FinalUserAcc
             ),
+            %% This needs Stats1 as log_end is called in maybe_add_stats
+            mango_execution_stats:log_stats(Stats1),
```
iilyak, line 134: We pass Stats0 into the `mango_cursor:maybe_add_warning` here while we pass Stats1 in `mango_cursor_view`. Why there is a difference?

chewbranca: Good catch! Fixed in b4e35020e.

iilyak (approved):
Nice work!!! +1 if tests pass

PR #4485, opened 2023-03-21
catch and log any error from mem3:local_shards

rnewson:
Enhance the fix in 06ab38d0c7607abeea7d63aeb47a5229fe0129da to cover all errors thrown by mem3:local_shards. The consequences of a crash here takes down a couch_index_server, which messily kills any in-flight view or find requests for that server.


davisp (approved):
Now that log is spelt propellarly, +1

nickva (approved):

PR #4490, opened 2023-03-22
fix(doc): reverse definition of `all_nodes` and `cluster_nodes` to ma…

janl:
https://github.com/apache/couchdb/blob/main/src/mem3/src/mem3_httpd.erl#L31-L43


rnewson (approved):
facepalm

PR #4493, opened 2023-03-22
Add log directory to eunit setup template

big-r81:
To log messages of test runs to file we need to add a absoulute path to the file logger.


nickva (approved):
+1

That's a nice ergonomic cleanup. I see logs in tmp/couch.log on linux/macos. It's a bit easier than having to hunt for them in `src/app/.eunit subdir`

Wonder if we could redirect coverage report tmp as well...

PR #4494, opened 2023-03-23
Treat single-element keys as key

jiahuili430:
## Overview
 - Request `_view` with reduce function, treat single-element `keys` as `key`
 - Add treat_single_keys_as_key/2 function for POST `_view`.
    If we query `_all_docs` for deleted or nonexistent docs using single-element
    `keys`, we can get {deleted:true} and {error:not_found}.
- Add documentation on using key, keys, start_key, and end_key

## Testing recommendations
make eunit apps=chttpd,couch_mrview

Fixed https://github.com/apache/couchdb/issues/2243

## Checklist

### DIFF
--- a/src/couch_mrview/src/couch_mrview_http.erl
+++ b/src/couch_mrview/src/couch_mrview_http.erl
@@ -486,7 +486,12 @@ parse_params(Props, Keys, #mrargs{} = Args0, Options) ->
                 Args0;
             _ ->
                 % group_level set to undefined to detect if explicitly set by user
-                Args0#mrargs{keys = Keys, group = undefined, group_level = undefined}
+                case Keys of
+                    [_] ->
+                        Args0#mrargs{group = undefined, group_level = undefined};
+                    _ ->
+                        Args0#mrargs{keys = Keys, group = undefined, group_level = undefined}
+                end
         end,
     lists:foldl(
         fun({K, V}, Acc) ->
@@ -513,8 +518,24 @@ parse_body_and_query(Req, Keys) ->
 
 parse_body_and_query(Req, {Props}, Keys) ->
     Args = #mrargs{keys = Keys, group = undefined, group_level = undefined},
-    BodyArgs = parse_params(Props, Keys, Args, [decoded]),
-    parse_params(chttpd:qs(Req), Keys, BodyArgs, [keep_group_level]).
+    BodyArgs0 = parse_params(Props, Keys, Args, [decoded]),
+    BodyArgs1 =
+        case is_view(Req) of
+            true -> treat_single_keys_as_key(BodyArgs0);
+            false -> BodyArgs0
+        end,
+    parse_params(chttpd:qs(Req), Keys, BodyArgs1, [keep_group_level]).
+
+is_view(#httpd{path_parts = [_, _, _, <<"_view">> | _]}) ->
+    true;
+is_view(#httpd{}) ->
+    false.
+
+treat_single_keys_as_key(#mrargs{keys = Keys} = BodyArgs) ->
+    case Keys of
+        [_] -> BodyArgs#mrargs{keys = undefined};
+        _ -> BodyArgs
+    end.
 
 parse_param(Key, Val, Args, IsDecoded) when is_binary(Key) ->
     parse_param(binary_to_list(Key), Val, Args, IsDecoded);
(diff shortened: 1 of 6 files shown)


src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -513,8 +518,20 @@ parse_body_and_query(Req, Keys) ->
...
 parse_body_and_query(Req, {Props}, Keys) ->
     Args = #mrargs{keys = Keys, group = undefined, group_level = undefined},
-    BodyArgs = parse_params(Props, Keys, Args, [decoded]),
-    parse_params(chttpd:qs(Req), Keys, BodyArgs, [keep_group_level]).
+    BodyArgs0 = parse_params(Props, Keys, Args, [decoded]),
+    BodyArgs1 = treat_single_keys_as_key(Req, BodyArgs0),
+    parse_params(chttpd:qs(Req), Keys, BodyArgs1, [keep_group_level]).
+
+treat_single_keys_as_key(#httpd{path_parts = Path}, #mrargs{keys = Keys} = BodyArgs) ->
+    case lists:member(<<"_view">>, Path) of
```
rnewson, line 526: `_view` only has a special meaning in certain path parts, so this isn't quite right. I think we'll need a caluse that matches `_view` in the correct place and then another clause that doesn't look at path_parts and just returns `BodyArgs` (for the _all_docs case, right?).

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -530,9 +547,15 @@ parse_param(Key, Val, Args, IsDecoded) ->
             JsonKey = ?JSON_DECODE(Val),
             Args#mrargs{start_key = JsonKey, end_key = JsonKey};
         "keys" when IsDecoded ->
-            Args#mrargs{keys = Val};
+            case Val of
+                [SingleKeys] -> Args#mrargs{start_key = SingleKeys, end_key = SingleKeys};
```
rnewson, line 551: `SingleKey` or simply `Key` reads better; it is (by definition) not plural.

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -530,9 +547,15 @@ parse_param(Key, Val, Args, IsDecoded) ->
...
+                _ -> Args#mrargs{keys = Val}
+            end;
         "keys" ->
-            Args#mrargs{keys = ?JSON_DECODE(Val)};
+            case ?JSON_DECODE(Val) of
+                [SingleKeys] -> Args#mrargs{start_key = SingleKeys, end_key = SingleKeys};
```
rnewson, line 556: `Key` here also for the same reason.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -581,13 +581,12 @@ validate_args(Args) ->
             ok;
         {[], _, _} ->
             ok;
+        {[Key], StartKey, EndKey} when Key =:= StartKey andalso Key =:= EndKey ->
```
rnewson, line 584: `{[Key], Key, Key}` is a less verbose equivalent. Not essential.

rnewson (changes requested):
The comment on lists:member("_view") is a blocker for merge but the rest were style comments and can be taken at your discretion.

src/couch_mrview/src/couch_mrview_http.erl
jiahuili430: This way is much safer than just checking if `_view` exists in the path parts, thanks for the suggestion.

src/couch_mrview/src/couch_mrview_util.erl
jiahuili430: Good idea!
Changed, thanks for the review!

src/couch_mrview/src/couch_mrview_http.erl
```diff
@@ -513,8 +518,22 @@ parse_body_and_query(Req, Keys) ->
...
-    parse_params(chttpd:qs(Req), Keys, BodyArgs, [keep_group_level]).
+    BodyArgs0 = parse_params(Props, Keys, Args, [decoded]),
+    BodyArgs1 =
+        case is_view(Req) of
+            true -> treat_single_keys_as_key(BodyArgs0);
+            false -> BodyArgs0
+        end,
+    parse_params(chttpd:qs(Req), Keys, BodyArgs1, [keep_group_level]).
+
+is_view(#httpd{path_parts = Path}) ->
```
rnewson, line 529: If you look at chttpd.erl / chttp_view.erl etc you'll see many examples where we match on `path_parts` in the function head. It's a very widely used pattern and I think we should continue it here. I also note that the lists:member function call is not needed in your version, as the lists:nth is sufficient.

```is_view(#httpd{path_paths = [_, _, _, <<"_view">> | _]}) ->
    true;
is_view(#httpd{}) ->
    false.
```

jiahuili430: `length(path_parts)` is not always greater than 4 if called by other endpoints. E.g.:
- _all_docs: `path_parts`= [<<"db_name">>, <<"_all_docs">>]
- _view: `path_parts`= [<<"db_name">>, <<"_design">>, <<"ddoc_name">>, <<"_view">>, <<"view_name">>]

rnewson (approved):

src/couch_mrview/src/couch_mrview_http.erl
rnewson: right, in those cases we'll hit the second clause. we could be more proscriptive with `[_, _, _, <<"_view">>, _]` but I think it's fine to leave the right side variable.

rnewson:
very nice piece of work, thanks!

PR #4498, opened 2023-03-27
docs(`_find`): catch up with the implementation and further fixes

pgj:
- Unify the style of synopsis lines.
- Mention the `partitioned` parameter where applicable.
- Fix formatting of `warning` in one of the example responses.
- Trade the possibly retired `range` attribute for `mrargs` and expand the attributes within `opts` in the response of `_explain`.


nickva (approved):
+1

PR #4515, opened 2023-04-06
docs(_find): Remove redundancy from sample `_explain` response

pgj:
Acciddentally in `467e14ef`, the `use_index` field got doubled in one of the sample responses.


big-r81 (approved):
Good find! +1

PR #4543, opened 2023-04-24
enable nouveau in CI

rnewson:
enable nouveau in CI

    make index creation idempotent if index definition matches


big-r81 (approved):
+1

nickva (approved):
+1

PR #4555, opened 2023-04-27
fix race condition when creating indexes

rnewson:
two concurrent attempts to create the same index can fail (in our case this is the fabric rpc worker racing against the ken worker). take a lock to prevent this, but use a striped pattern to allow some concurrency between independent index creations.


nickva (approved):
+1 

Should we re-enable ken building the index in the test?

PR #4591, opened 2023-05-10
couchdb 3.3.2 is latest

rnewson:


iilyak:
There is also a version in https://github.com/apache/couchdb/blob/main/version.mk

nickva (changes requested):
Ideally this would be a forward port of the commit which bumped the version in the latest 3.x release.

PR #4600, opened 2023-05-12
add tests for QueryDeserializer

rnewson:
this class is currently unreachable but might be a new way to specify the query in SearchRequest rather than expanding the string-based query syntax.


nickva (approved):
+1

PR #4602, opened 2023-05-16
fix(`mango`): missing parts of the `_index` API

pgj:
Many of the requests outside the scope of the `_index` endpoint are not handled gracefully but trigger an internal server error.  Enhance the `_index` handler logic to return proper answers for invalid queries and supply it with more exhaustive integration tests.

Provide documentation for the existing `_index/_bulk_delete` endpoint as it was missing, and mention that the `_design` prefix is not needed when deleting indexes.

This is a follow-up to #4584 and a revisited version of #4514.

## Checklist


nickva (approved):
+1

Thank you for adding the docs and for writting tests!

PR #4609, opened 2023-05-21
fix: missing stats description for Mango `keys_examined`

pgj:
The stats description was probably not included for the `keys_examined` counter for Mango in #4569, which causes `unknown metric: [mango, keys_examined]` messages in the logs.  This change fixes the problem by providing the necessary item.


nickva (approved):
+1

PR #4610, opened 2023-05-21
fix(mango): integration tests

pgj:
- Fix the faulty `index_array_length` tests:

    - They do not check the count of the returned documents but blindly
      assume that there will be results.  Add an explicit assertion about
      the expected number of documents in the response so it could be
      detected if something breaks.

    - The assertion on the document count revealed that the test database
      is actually empty because it is immediately wiped out after creation
      by a misplaced `recreate()` call.

    - When the indexing on the array lengths is disabled, there shall be no
      documents returned, and not the sizes of the `favorites` arrays in
      the results shall be zero.  That is because there is no information
      available on the sizes of the arrays.

- The result of the `mango.has_text_service()` checks is implicitly forecasted by the `@unittest.skipUnless` annotations in the head of the class declaration.  Hence they are not needed anymore.

### DIFF
--- a/src/mango/test/04-key-tests.py
+++ b/src/mango/test/04-key-tests.py
@@ -48,8 +48,7 @@ class KeyTests(mango.DbPerClass):
         super(KeyTests, klass).setUpClass()
         klass.db.save_docs(TEST_DOCS, w=3)
         klass.db.create_index(["type"], ddoc="view")
-        if mango.has_text_service():
-            klass.db.create_text_index(ddoc="text")
+        klass.db.create_text_index(ddoc="text")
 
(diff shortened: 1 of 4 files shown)


jaydoane (approved):
Nice fixes!
```
running clouseau c59127d5657e3130b55d6e2b9575820a08e438e1

make mango-test

[ * ] Exec command COUCH_USER=adm COUCH_PASS=pass .venv/bin/python3 -m nose2  ... ........s.ss.s.s.............s.ss.s.s............s.ss.s.s.............................................................................................................................................................s...................................................................................................................................................
Ran 362 tests in 25.601s

OK (skipped=16)
```

PR #4632, opened 2023-06-03
chore(`mango`): revisit test database recreation logic

pgj:
Databases used for integration testing are being recreated by unbounded recursion which can blow up the stack in case of unrecoverable errors.  This change revisits the respective parts of the implementation in the following ways:

- Replace the tail recursion for a regular loop.
- Set an upper limit for the number of tries.
- Signal the problem with a more descriptive message.
- Introduce an increasing delay between the subsequent tries for tolerance against transient errors.


nickva (approved):
+1

PR #4636, opened 2023-06-11
clarify default quorum calculation

rnewson:
It was pointed out on the user@ mailing list (https://lists.apache.org/thread/s2t47t9c84bqg50gz2805gnh73qxr7fc) that the docs express a different quorum algorithm from the code (that happens to give the same answers for odd numbers of `N`).

This PR updates the docs to match the code.


rnewson:
An alternative expression is `r=w=⌊n/2⌋+1` but perhaps is too subtle.

nickva (approved):
+1

PR #4645, opened 2023-06-18
docs: add `adm:pass` to URL examples

jiahuili430:
fix https://github.com/apache/couchdb/issues/4642


nickva (approved):
+1

Thanks, @jiahuili430. We switched from admin party but never updated the docs.

PR #4665, opened 2023-07-04
chore(`mango`): simplify implementation of empty selector tests

pgj:
Remove unused imports and make the definition of test suites less handcrafted via multiple inheritance.  This representation makes it possible for developers to add tests as class-specific methods when required, much like it happened for `03-operator-test`.


nickva (approved):
+1  a very nice cleanup, @pgj

PR #4666, opened 2023-07-06
fix(config): updating values

pgj:
Configuration values could be updated by sending JSON strings as the body.  But there is no safeguard in place to stop users from submitting values of arbitrary JSON type, including objects. Violating this assumption silently triggers a HTTP 500 response. Fill in this hole and provide users with a cause of rejection.

## Checklist


pgj:
The [documentation](https://docs.couchdb.org/en/latest/api/server/configuration.html) suggests other types of JSON values might be allowed but I have only seen strings being used so far.  The original code also implies that strings are preferred.

big-r81 (approved):
+1

nickva (approved):
+1

pgj:
Thank you, guys!

PR #4669, opened 2023-07-10
split all

rnewson:
It is not currently possible to override `[replicator] valid_socket_options` with more than two items as we call `string:split/2`, which splits the string at the first occurrence of the separator. We need to call `string:split/3` with third argument of `all` to split out all the fields.


nickva (approved):
+1

Issue #4668, opened 2023-07-08
CouchDB sends A dns request when configured -proto_dist inet6_tcp


sergey-safarov:
## Description
We have an IPv6-only installation. No IPv4 is configured on the server (except loopback). In the `vm.agrs` configured
```
-proto_dist inet6_tcp
```
When I start replication to then CouchDB sends only `A` dns request and does not try to send `AAAA` dns requsts.
[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your actual CouchDB! )

## Steps to Reproduce
Create `request.json` file like
```json
{
  "user_ctx": {
    "name": "admin",
    "roles": [
      "_admin",
      "_reader",
      "_writer"
    ]
  },
  "source": {
    "url": "http://ippbx-1a.la.example.com:5984/mydb",
    "headers": {
      "Authorization": "Basic XXXXXXXXXXXX"
    }
  },
  "target": {
    "url": "http://127.0.0.1:5984/mydb",
    "headers": {
      "Authorization": "Basic XXXXXXXXX"
    }
  },
  "create_target": true,
  "continuous": false
}
```
Then send curl command like
```bash
curl \
 -H 'Content-Type: application/json' \
 -H 'Accept: application/json' \
 -d @request.json \
 'http://admin:change_me@127.0.0.1:5984/_replicator'
```
in the logs i can see
> Jul 07 09:21:51 ippbx-1b.la.example.com docker[3677369]: [error] 2023-07-07T09:21:51.599467Z [couchdb@ippbx-1b.la.example.com](mailto:couchdb@ippbx-1b.la.example.com) <0.6095.0> -------- couch_replicator_httpc: auth plugin initialization failed "http://ippbx-1a.la.example.com:5984/mydb/" {session_request_failed,"[http://ippbx-1a.la.example.com:5984/_session","admin",{conn_failed,{error,nxdomain}}}](http://ippbx-1a.la.example.com:5984/_session%22,%22admin%22,%7Bconn_failed,%7Berror,nxdomain%7D%7D%7D)
[...]

PR #4670, opened 2023-07-10
allow setting of some ibrowse options

rnewson:
in particular this allows;

```
[replicator]
ibrowse_options = [{prefer_ipv6, true}]
```

which allows the replicator to replicate with an ipv6-only source or target.

Closes: https://github.com/apache/couchdb/issues/4668

### DIFF
--- a/src/couch_replicator/src/couch_replicator_parse.erl
+++ b/src/couch_replicator/src/couch_replicator_parse.erl
@@ -34,6 +34,10 @@
     recbuf,
     sndbuf
 ]).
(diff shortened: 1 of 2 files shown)


nickva (approved):
+1

sergey-safarov:
@rnewson could you help me undrstand when fix will be backported to `3.3.x` branch? Or older branch.

rnewson:
@sergey-safarov sorry, I do not know when there will be a release that includes this fix. I expect it to be backported to 3.3.x but not further.

PR #4673, opened 2023-07-12
[WIP, but please review] per doc access 2023

janl:


src/chttpd/src/chttpd_view.erl
```diff
@@ -69,6 +69,21 @@ fabric_query_view(Db, Req, DDoc, ViewName, Args) ->
...
+%    %       However, this *might* be here because we need to handle
+%    %       errors here now, because access might tell us to.
+%    case fabric:query_view(Db, Options, DDoc, ViewName,
+%            fun view_cb/2, VAcc, Args) of
+%        {ok, Resp} ->
+%            {ok, Resp#vacc.resp};
+%        {error, Error} ->
+%            throw(Error)
+%    end.
+
```
jaydoane: Can this be removed, or is more work needed here?

jaydoane (commented):
I pointed out some compiler warnings that I saw when poking around.

I also noticed a few TODO and other comments that I'm guessing need to still be addressed or removed?

Currently this PR has no description, and I wonder if you could add some high level documentation of how this works, or maybe link to an architectural document?

janl:
> Currently this PR has no description, and I wonder if you could add some high level documentation of how this works, or maybe link to an architectural document?

sorry for not making this explicit, but if you follow the superseded PRs backwards, you should find more commentary that I didn’t think was necessary to repeat here. The main design is outlined here: https://github.com/apache/couchdb-documentation/pull/424

janl:
latest push fixes compiler warnings noted by Jay

src/couch/src/couch_db.erl
```diff
@@ -1458,23 +1567,29 @@ collect_results(Pid, MRef, ResultsAcc) ->
...
     MRef = erlang:monitor(process, Pid),
+    UserCtx =
+        case has_access_enabled(Db) of
+            true -> UserCtx0;
+            false -> []
+        end,
+
     try
-        Pid ! {update_docs, self(), DocBuckets, LocalDocs, ReplicatedChanges},
+        Pid ! {update_docs, self(), DocBuckets, LocalDocs, ReplicatedChanges, UserCtx},
```
rnewson, line 1585: note to self to check update_docs receiving end for mixed clustered rpc issues

src/couch/src/couch_db_updater.erl
```diff
@@ -294,12 +304,12 @@ collect_updates(GroupedDocsAcc, ClientsAcc, ReplicatedChanges) ->
         % local docs. It's easier to just avoid multiple _local doc
         % updaters than deal with their possible conflicts, and local docs
         % writes are relatively rare. Can be optmized later if really needed.
-        {update_docs, Client, GroupedDocs, [], ReplicatedChanges} ->
```
rnewson, line 297: do we need a clause with the original message format for internode compatibility during upgrade?

src/couch/src/couch_db_updater.erl
```diff
@@ -711,7 +731,17 @@ update_docs_int(Db, DocsList, LocalDocs, ReplicatedChanges) ->
...
+    % Loop over DocsList, validate_access for each OldDocInfo on Db,
+    %.  if no OldDocInfo, then send to DocsListValidated, keep OldDocsInfo
+    %   if valid, then send to DocsListValidated, OldDocsInfo
+    %.  if invalid, then send_result tagged `access`(c.f. `conflict)
+    %.    and don’t add to DLV, nor ODI
+
+    {DocsListValidated, OldDocInfosValidated} = validate_docs_access(
+        Db, DocsList, OldDocInfos
+    ),
+    % couch_log:notice("~n~n u_d_i: DocsList: ~p~n, OldDocInfos: ~p~n, DocsListValidated: ~p~n, OldDocInfosValidated: ~p~n~n~n", [DocsList, OldDocInfos, DocsListValidated, OldDocInfosValidated]),
```
rnewson, line 743: junk

src/couch/src/couch_db_updater.erl
```diff
@@ -721,7 +751,9 @@ update_docs_int(Db, DocsList, LocalDocs, ReplicatedChanges) ->
     % the trees, the attachments are already written to disk)
     {ok, IndexFDIs} = flush_trees(Db, NewFullDocInfos, []),
     Pairs = pair_write_info(OldDocLookups, IndexFDIs),
-    LocalDocs2 = update_local_doc_revs(LocalDocs),
+    % TODO: local docs access needs validating
```
rnewson, line 754: is this TODO todone already?

src/couch/src/couch_db_updater.erl
```diff
@@ -736,17 +768,91 @@ update_docs_int(Db, DocsList, LocalDocs, ReplicatedChanges) ->
...
+    {ok, commit_data(Db1), UpdatedDDocIds}.
+
+% at this point, we already validated this Db is access enabled, so do the checks right away.
+check_access(Db, UserCtx, Access) ->
+    couch_db:check_access(Db#db{user_ctx = UserCtx}, Access).
+
+validate_docs_access(Db, DocsList, OldDocInfos) ->
+    case couch_db:has_access_enabled(Db) of
+        true -> validate_docs_access_int(Db, DocsList, OldDocInfos);
+        _Else -> {DocsList, OldDocInfos}
```
rnewson, line 786: `s/_Else/false`

src/couch/src/couch_db_updater.erl
```diff
@@ -736,17 +768,91 @@ update_docs_int(Db, DocsList, LocalDocs, ReplicatedChanges) ->
...
+    end.
+
+validate_docs_access_int(Db, DocsList, OldDocInfos) ->
+    validate_docs_access(Db, DocsList, OldDocInfos, [], []).
+
+validate_docs_access(_Db, [], [], DocsListValidated, OldDocInfosValidated) ->
+    % TODO: check if need to reverse this? maybe this is the cause of the test reverse issue?
```
rnewson, line 793: junk

src/couch/src/couch_httpd_auth.erl
```diff
@@ -98,6 +98,13 @@ basic_name_pw(Req) ->
             nil
     end.
 
+extract_roles(UserProps) ->
+    Roles = couch_util:get_value(<<"roles">>, UserProps, []),
+    case lists:member(<<"_admin">>, Roles) of
+        true -> Roles;
+        _ -> Roles ++ [<<"_users">>]
```
rnewson, line 105: `[<<"_users">> | Roles]`

src/couch/src/couch_httpd_auth.erl
```diff
@@ -199,7 +206,7 @@ proxy_auth_user(Req) ->
             Roles =
                 case header_value(Req, XHeaderRoles) of
                     undefined -> [];
-                    Else -> re:split(Else, "\\s*,\\s*", [trim, {return, binary}])
+                    Else -> re:split(Else, "\\s*,\\s*", [trim, {return, binary}]) ++ [<<"_users">>]
```
rnewson, line 209: `[<<"_users">> | re:split ...]`

rnewson (changes requested):

src/couch/src/couch_httpd_auth.erl
janl: we’ve been over. this before, I want to explicitly append the new role to the array so any tests and code that expect their existing roles in the beginning just keep working and you OK’d that reasoning on an older PR :)

src/couch/src/couch_httpd_auth.erl
```diff
@@ -248,7 +255,7 @@ jwt_authentication_handler(Req) ->
                             Req#httpd{
                                 user_ctx = #user_ctx{
                                     name = User,
-                                    roles = Roles
+                                    roles = Roles ++ [<<"_users">>]
```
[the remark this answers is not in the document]
janl: c.f. https://github.com/apache/couchdb/pull/4673#discussion_r1285933876

src/chttpd/src/chttpd_db.erl
```diff
@@ -958,16 +958,18 @@ view_cb(Msg, Acc) ->
     couch_mrview_http:view_cb(Msg, Acc).
 
 db_doc_req(#httpd{method = 'DELETE'} = Req, Db, DocId) ->
-    % check for the existence of the doc to handle the 404 case.
-    couch_doc_open(Db, DocId, nil, []),
-    case chttpd:qs_value(Req, "rev") of
+    % fetch the old doc revision, so we can compare access control
+    % in send_update_doc() later.
+    Doc0 = couch_doc_open(Db, DocId, nil, [{user_ctx, Req#httpd.user_ctx}]),
+    Revs = chttpd:qs_value(Req, "rev"),
```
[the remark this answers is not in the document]
janl: it is used to set the `revs` property in the #doc{} record a few lines further down, but happy to adjust

src/chttpd/src/chttpd_db.erl
```diff
@@ -1967,6 +1972,21 @@ parse_shards_opt("placement", Req, Default) ->
                     throw({bad_request, Err})
             end
     end;
+parse_shards_opt("access", Req, Value) when is_list(Value) ->
```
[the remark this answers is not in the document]
janl: the main purpose of this clause is the `list_to_existing_atom()` call, but I can move that into the next clause

src/chttpd/src/chttpd_db.erl
```diff
@@ -1967,6 +1972,21 @@ parse_shards_opt("placement", Req, Default) ->
...
+    parse_shards_opt("access", Req, list_to_existing_atom(Value));
+parse_shards_opt("access", _Req, Value) when Value =:= true ->
+    case config:get_boolean("per_doc_access", "enabled", false) of
+        true ->
+            true;
+        false ->
+            Err = ?l2b(["The `access` option is not available on this CouchDB installation."]),
```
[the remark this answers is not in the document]
janl: I couldn’t find a nice way to do the binary construction with a parameter further down and I opted to make all `Err` constructions looks the same. I’ll remove the runtime conversion where easily possible

src/couch/src/couch_access_native_proc.erl
```diff
@@ -0,0 +1,139 @@
...
+}).
+
+start_link() ->
+    gen_server:start_link(?MODULE, [], []).
+
+set_timeout(Pid, TimeOut) when is_integer(TimeOut), TimeOut > 0 ->
+    gen_server:call(Pid, {set_timeout, TimeOut}).
+
+prompt(Pid, Data) ->
+    gen_server:call(Pid, {prompt, Data}).
```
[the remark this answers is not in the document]
janl: it probably needs to be configurable like other query server timeouts

src/couch/src/couch_db.erl
```diff
@@ -302,17 +312,23 @@ open_doc(Db, Id, Options) ->
         {ok, #doc{deleted = true} = Doc} ->
             case lists:member(deleted, Options) of
                 true ->
-                    apply_open_options({ok, Doc}, Options);
```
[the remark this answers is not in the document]
janl: `open_doc_int` already runs `apply_open_options`

src/couch/src/couch_db.erl
```diff
@@ -302,17 +312,23 @@ open_doc(Db, Id, Options) ->
...
+                    {ok, Doc};
                 false ->
                     {not_found, deleted}
             end;
         Else ->
-            apply_open_options(Else, Options)
```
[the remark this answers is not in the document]
janl: `open_doc_int` already runs `apply_open_options`

src/couch/src/couch_db.erl
```diff
@@ -769,6 +787,64 @@ security_error_type(#user_ctx{name = null}) ->
...
+        [] ->
+            % if doc has no _access, userCtX must be admin
+            is_admin(Db);
+        Access ->
+            % if doc has _access, userCtx must be admin OR matching user or role
+            case is_admin(Db) of
+                true ->
+                    true;
+                _ ->
+                    case {check_name(UserName, Access), check_roles(UserRoles, Access)} of
```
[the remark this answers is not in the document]
janl: nice cleanup, we can even get rid of the `case` entirely

src/couch/src/couch_db.erl
```diff
@@ -1330,13 +1411,41 @@ update_docs(Db, Docs0, Options, ?REPLICATED_CHANGES) ->
...
+        false ->
+            % we’re done here
+            {ok, DocErrors};
+        _ ->
+            AccessViolations = lists:filter(fun({_Ref, Tag}) -> Tag =:= access end, Results),
+            case length(AccessViolations) of
+                0 ->
+                    % we’re done here
+                    {ok, DocErrors};
+                _ ->
```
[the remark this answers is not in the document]
janl: that would make `N` bound and we do not need it

src/couch/src/couch_db_updater.erl
janl: https://github.com/apache/couchdb/pull/4673#discussion_r1296850568

src/couch/src/couch_db.erl
```diff
@@ -1864,23 +1976,35 @@ open_doc_revs_int(Db, IdRevs, Options) ->
 open_doc_int(Db, <<?LOCAL_DOC_PREFIX, _/binary>> = Id, Options) ->
     case couch_db_engine:open_local_docs(Db, [Id]) of
         [#doc{} = Doc] ->
-            apply_open_options({ok, Doc}, Options);
+            case Doc#doc.body of
+                {Body} ->
+                    Access = couch_util:get_value(<<"_access">>, Body),
+                    apply_open_options(Db, {ok, Doc#doc{access = Access}}, Options);
+                _Else ->
```
rnewson, line 1983: will do a full review tomorrow but I am uncomfortable with a giant `_Else` here. what else would we get except a body? `undefined` or `nil`? let's be specific in couch_db.erl.

arturog:
I am coming back to couchdb updating an old application that *has already* implemented the above (per-document access, with linux-like permissions like owner/read/write access, even per-field protection on the doc).  I personally don't like this feature and wouldn't like to see it merged:

- This kind of security requirement is almost always specific to the application.  In my case, for example, this solution is not fit for purpose.
- Most of the time a front-end application server is better suited to handle security requirements, which might be complex.
- I believe CouchDB should stick to being a database.  We should refine the current codebase instead.
- If we need to implement this kind of functionality (like in my case), we are better off providing a plugin/hook system where can enter the validate and read lifecycle of the document.  Devs are free to implement what they want and we could provide a working example. (This is what *I* did).  Providing a `_custom_<field>` on documents for devs to freely use would be great.
- Are we going to filter the views?  Are we going to filter the document counts on views?  What about leaking information on views?  Some views need some info, some others not?  Will I sometimes use per-user view, sometimes the global views?  And what about searching with Lucene?  I feel we should not touch our indices.
- Replicating documents *into* your database where the _access field changes, are we doing that?  Will devs want a "master" where the `_access` field is allowed to change, and slaves where you won't?  Is the ultimate purpose of this PR to segregate documents within your own DB?  Then why not create per-user DBs?

I would be in favour of document lifecycle plugins/hooks, where we can implement this feature and many others.

rnewson:
@arturog

I think secondary indexes would need to be updated to support this for it to have the desired effect (restricting a user's visibility of a database to a subset). I'm sure we've not done that work for dreyfus/clouseau (and I definitely haven't for nouveau). 

I share your concern that this implementation might not meet everyone's needs, but that's not necessarily a blocker. It just needs to meet a significant number of people's needs.

I don't find "I believe CouchDB should stick to being a database." compelling. Other databases have richer permissions and access controls that we currently do, they're not off-topic. 

The PR changes a number of internal records which prevents a smooth upgrade of existing CouchDB clusters and is blocked from merging until that is fixed imo, so there's time to think.

We currently have `validate_doc_update` where you can implement any kind of write control logic, but no equivalent for reads (or view queries besides the deprecated _list option). If Javascript were evaluated more efficiently (c.f, the `quickjs` embedding idea) we might add that.

arturog:
> We currently have validate_doc_update where you can implement any kind of write control logic, but no equivalent for reads (or view queries besides the deprecated _list option). If Javascript were evaluated more efficiently (c.f, the quickjs embedding idea) we might add that.

We implemented a fork of CouchDB with exactly that (a `validate_doc_read`), with code written in Erlang.  We also added our own `validate_doc_write` in Erlang before the JS one kicks in.  If extended, I believe it can provide not only this feature, but much more.  Also, the issue of leaking out data in indices (specially to the lucene index) lead me to limit indexing to only public fields -- so the field-access was also born.  Indices in CouchDB were kept untouched, and though all views had the full set of `_id`s, on document access (or requesting the full document via views), you would get `<unauthorized>`.  Our implementation was not meant to present a "subset" of documents, but instead control access to the document.

Perhaps the requirement to present a *subset* of the database needs to be revisited?  If this requirement were to be dropped, things might become easier.

janl:
@arturog thank you for your comments, practical experience with variants of the same idea here are invaluable.

> This kind of security requirement is almost always specific to the application. In my case, for example, this solution is not fit for purpose.
> Most of the time a front-end application server is better suited to handle security requirements, which might be complex.

I’m 100% on board with you here. That’s why this is designed as an opt-in feature. If you have a system on top of CouchDB that works for you, that will just continue to work.

To avoid confusion, let me restate the design goal of this PR: **“make the db-per-user pattern obsolete”**. It is specifically not “support arbitrary ACLs inside a database”.

With that out of the way, the _scope_ of the potential work for a complete solution here is _very big_. In accordance with common wisdom: to built a complex system that works, one first has to build a simple system that works. The current PR reflects the _simplest_ system that I could think of to get the minimal functionality working that satisfies the design goal.

To this end, per-user design docs (and all associated indexes of any form, JS, Mango, Search/Nouveau) are _not_ supported. A future version of this PR however can use the new `by-access-seq` index to build per-user indexes, including reduces that are guaranteed to not leak any data. Global views are already supported but are limited to database admins.

> Perhaps the requirement to present a subset of the database needs to be revisited? If this requirement were to be dropped, things might become easier.

Maybe you want a different feature? You are stating the explicit design goal of this PR that should be dropped.

danieltaylor-nz:
hi @janl I think this per-document access control could be extremely helpful for my project. Is there any possibility to get this simple solution into main or is there a super-ceding solution?

janl:
closing in favour of https://github.com/apache/couchdb/pull/5842

PR #4674, opened 2023-07-13
Conditionally exclude nouveau from elixir make target

rnewson:
Only run nouveau tests if configured with `--enable-nouveau`

sidebar: is it appropriate to move the `-include install.mk` directive? Its original placement was just before the install targets, but it seems necessary to know `with_nouveau = X` elsewhere.

A similar switch is needed for dreyfus. `make elixir` still fails if you are not running a clouseau jvm. this predates the nouveau work.


janl:
> sidebar: is it appropriate to move the -include install.mk directive? Its original placement was just before the install targets, but it seems necessary to know with_nouveau = X elsewhere.

looks fine to me

pgj:
For `search`, there [a separate `elixir-search` target](https://github.com/apache/couchdb/commit/395c083026f9735a2a34cda3012151ef94466136) has been added back in mid-February.  But I like the idea of merging it with `elixir` so that the test could be run depending on the configuration.

pgj:
Hrm, I have not noticed that `elixir` runs the search tests too.  But why does not it choke the CI all the time?

PR #4681, opened 2023-07-19
Introduce optional countermeasures as we run out of disk space

rnewson:
## Overview

Add some opt-in countermeasures if disk space gets critically low, with the aim of avoiding hitting end of disk.

## Testing recommendations

Enable the disk monitor, run a CouchDB node on a small disk, and add lots of data. You should find that background indexing stops (at 80% full) and then stale view queries and database updates are blocked. Reduce disk usage (allow compaction to run to completion, or delete databases) and see both interventions automatically unlock. 

N/A

## Checklist

### DIFF
--- a/rel/overlay/etc/vm.args
+++ b/rel/overlay/etc/vm.args
@@ -122,3 +122,15 @@
 # in the features list.
 #
 #-crypto fips_mode true
+
+# OS Mon Settings
+
+# only start disksup
+-os_mon start_cpu_sup false
+-os_mon start_memsup false
+
+# Check disk space every 5 minutes
+-os_mon disk_space_check_interval 5
+
+# don't let disksup send alerts
+-os_mon disk_almost_full_threshold 1.0
--- a/src/couch/src/couch_db.erl
+++ b/src/couch/src/couch_db.erl
@@ -1338,6 +1338,15 @@ update_docs(Db, Docs0, Options, ?REPLICATED_CHANGES) ->
     ),
     {ok, DocErrors};
 update_docs(Db, Docs0, Options, ?INTERACTIVE_EDIT) ->
+    BlockInteractiveDatabaseWrites = couch_disk_monitor:block_interactive_database_writes(),
+    if
+        BlockInteractiveDatabaseWrites ->
+            {ok, [{insufficient_storage, <<"database_dir is too full">>} || _ <- Docs0]};
+        true ->
+            update_docs_interactive(Db, Docs0, Options)
+    end.
+
+update_docs_interactive(Db, Docs0, Options) ->
     Docs = tag_docs(Docs0),
 
     AllOrNothing = lists:member(all_or_nothing, Options),
--- a/src/couch/src/couch_disk_monitor.erl
+++ b/src/couch/src/couch_disk_monitor.erl
@@ -1,0 +1,241 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+%   http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+% WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+% License for the specific language governing permissions and limitations under
+% the License.
+
+-module(couch_disk_monitor).
+-behaviour(gen_server).
+
+% public api
+-export([
+    block_background_view_indexing/0,
+    block_interactive_view_indexing/0,
+    block_interactive_database_writes/0
+]).
+
+% testing
+-export([
+    set_database_dir_percent_used/1,
+    set_view_index_dir_percent_used/1
+]).
+
+% gen_server callbacks
+-export([
+    start_link/0,
+    init/1,
+    handle_call/3,
+    handle_cast/2,
+    handle_info/2
+]).
+
+-include_lib("kernel/include/file.hrl").
+
+-define(SECTION, "disk_monitor").
+
+-record(st, {
+    timer
+}).
+
+block_background_view_indexing() ->
+    Enabled = enabled(),
+    if
+        Enabled ->
+            view_index_dir_percent_used() > background_view_indexing_threshold();
+        true ->
+            false
+    end.
+
+block_interactive_view_indexing() ->
+    Enabled = enabled(),
+    if
+        Enabled -> view_index_dir_percent_used() > interactive_view_indexing_threshold();
+        true -> false
+    end.
+
+block_interactive_database_writes() ->
+    Enabled = enabled(),
+    if
+        Enabled -> database_dir_percent_used() > interactive_database_writes_threshold();
+        true -> false
+    end.
+
+set_database_dir_percent_used(PercentUsed) when PercentUsed >= 0, PercentUsed =< 100 ->
+    gen_server:call(?MODULE, {set_database_dir_percent_used, PercentUsed}).
+
+set_view_index_dir_percent_used(PercentUsed) when PercentUsed >= 0, PercentUsed =< 100 ->
+    gen_server:call(?MODULE, {set_view_index_dir_percent_used, PercentUsed}).
+
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+init(_Args) ->
+    ets:new(?MODULE, [named_table, {read_concurrency, true}]),
+    ets:insert(?MODULE, {database_dir, 0}),
+    ets:insert(?MODULE, {view_index_dir, 0}),
+    update_disk_data(),
+    TRef = erlang:send_after(timer_interval(), self(), update_disk_data),
+    {ok, #st{timer = TRef}}.
+
+handle_call({set_database_dir_percent_used, PercentUsed}, _From, St) ->
+    ets:insert(?MODULE, {database_dir, PercentUsed}),
+    {reply, ok, St};
+handle_call({set_view_index_dir_percent_used, PercentUsed}, _From, St) ->
+    ets:insert(?MODULE, {view_index_dir, PercentUsed}),
+    {reply, ok, St};
+handle_call(_Msg, _From, St) ->
+    {reply, {error, unknown_msg}, St}.
+
+handle_cast(_Msg, St) ->
+    {noreply, St}.
+
+handle_info(update_disk_data, St) ->
+    erlang:cancel_timer(St#st.timer),
+    update_disk_data(),
+    TRef = erlang:send_after(timer_interval(), self(), update_disk_data),
+    {noreply, St#st{timer = TRef}};
+handle_info(_Msg, St) ->
+    {noreply, St}.
+
+update_disk_data() ->
+    DiskData = disksup:get_disk_data(),
+    update_disk_data(DiskData).
+
+update_disk_data([]) ->
+    ok;
+update_disk_data([{Id, _, PercentUsed} | Rest]) ->
+    IsDatabaseDir = is_database_dir(Id),
+    IsViewIndexDir = is_view_index_dir(Id),
+    if
+        IsDatabaseDir ->
+            ets:insert(?MODULE, {database_dir, PercentUsed});
+        true ->
+            ok
+    end,
+    if
+        IsViewIndexDir ->
+            ets:insert(?MODULE, {view_index_dir, PercentUsed});
+        true ->
+            ok
+    end,
+    update_disk_data(Rest).
+
+is_database_dir(MntOn) ->
+    same_device(config:get("couchdb", "database_dir"), MntOn).
+
+is_view_index_dir(MntOn) ->
+    same_device(config:get("couchdb", "view_index_dir"), MntOn).
+
+same_device(DirA, DirB) ->
+    case {device_id(DirA), device_id(DirB)} of
+        {{ok, DeviceId}, {ok, DeviceId}} ->
+            true;
+        _Else ->
+            false
+    end.
+
+device_id(Dir) ->
+    case file:read_file_info(Dir) of
+        {ok, FileInfo} ->
+            {ok, {FileInfo#file_info.minor_device, FileInfo#file_info.major_device}};
+        {error, Reason} ->
+            {error, Reason}
+    end.
+
+database_dir_percent_used() ->
+    [{database_dir, PercentUsed}] = ets:lookup(?MODULE, database_dir),
+    PercentUsed.
+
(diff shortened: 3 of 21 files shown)


janl (approved):
nicely done!

nickva (changes requested):
It's a nice usability improvement and a good use of disk space monitor.

There are a few comments and notes in-line. 

The one tricky aspect I think is how this interacts with the internal replicator. If one node is becoming full, we'd return an ` insufficient_storage` error from workers there. The request overall might even skip over that error and return a 201/202 if other shard copies reply with a success. That's a good behavior on one hand, on the other hand, there a good chance updates we blocked interactively would still make it to the affected nodes via the internal replicator and would continue filling up the disk.

Not sure what a good solution might be. These are a few random thoughts I had:

1) We could crash the whole update_doc request if any shard copy is almost full? But even then, it might be too late as the successful requests on other nodes would still propagate via the internal replicator to the full node.

2) Is there a graceful way to stop internal replicator jobs and avoid them going into a crash loop and making a mess to avoid replicating data to nodes which are getting full?

3) I am wondering if at least for doc writes it would make sense to keep track of cluster wide disk usage (via a gossip between `couch_disk_monitor` gen_server or something like that) and return an immediate error at the fabric level without attempting any worker doc writes. That would increase the complexity of disk monitor, but would avoid the internal replicator loophole issue?

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -190,12 +191,24 @@ get_view_index_state(Db, DDoc, ViewName, Args0, RetryCount) ->
                     couch_index:get_state(Pid, 0);
                 false ->
                     couch_index:get_state(Pid, 0);
+                _ when BlockInteractiveIndexing ->
+                    %% if the view is already fresh enough, proceed.
+                    %% otherwise it's an error until there's room to update.
+                    case couch_index:get_state(Pid, 0) of
```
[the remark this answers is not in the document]
rnewson: oh, true. I think it's pretty niche so maybe no? but I'll look into it.

src/couch/src/couch_disk_monitor.erl
```diff
@@ -0,0 +1,258 @@
...
+-include_lib("couch/include/couch_eunit.hrl").
+
+not_enabled_by_default_test() ->
+    ?assertEqual(false, enabled()),
+    ?assertEqual(false, block_background_view_indexing()),
+    ?assertEqual(false, block_interactive_view_indexing()),
+    ?assertEqual(false, block_database_writes()).
+
+setup_all() ->
+    application:ensure_all_started(os_mon),
```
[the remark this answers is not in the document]
rnewson: yeah, will do. the tests pass in isolation but they fail in the full run because the cleanup isn't right. it's a known blocker

src/couch/src/couch_disk_monitor.erl
```diff
@@ -0,0 +1,258 @@
...
+            {error, Reason}
+    end.
+
+database_dir_percent_used() ->
+    [{database_dir, PercentUsed}] = ets:lookup(?MODULE, database_dir),
+    PercentUsed.
+
+view_index_dir_percent_used() ->
+    [{view_index_dir, PercentUsed}] = ets:lookup(?MODULE, view_index_dir),
+    PercentUsed.
```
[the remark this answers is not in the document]
nickva: Agree an ets table with `{read_concurrency, true}` would work, it's what we use for `config:get(...)` and that's fairly fast, 3-6 microseconds in a silly timer:tc/3 benchmark I ran.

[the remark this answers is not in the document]
rnewson: added that.

src/couch/src/couch_disk_monitor.erl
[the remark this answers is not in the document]
rnewson: that helped.

src/couch/src/couch_disk_monitor.erl
```diff
@@ -0,0 +1,258 @@
...
+            ("disk_monitor", "background_view_indexing_threshold", _) -> 50;
+            ("disk_monitor", "interactive_view_indexing_threshold", _) -> 60;
+            ("disk_monitor", "database_writes_threshold", _) -> 70
+        end
+    ),
+    {ok, Pid} = start_link(),
+    Pid.
+
+teardown_all(Pid) ->
+    true = exit(Pid, normal),
```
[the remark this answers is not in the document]
rnewson: eliminated this line

rnewson:
> It's a nice usability improvement and a good use of disk space monitor.
> 
> There are a few comments and notes in-line.
> 
> The one tricky aspect I think is how this interacts with the internal replicator. If one node is becoming full, we'd return an ` insufficient_storage` error from workers there. The request overall might even skip over that error and return a 201/202 if other shard copies reply with a success. That's a good behavior on one hand, on the other hand, there a good chance updates we blocked interactively would still make it to the affected nodes via the internal replicator and would continue filling up the disk.
> 
> Not sure what a good solution might be. These are a few random thoughts I had:
> 
>     1. We could crash the whole update_doc request if any shard copy is almost full? But even then, it might be too late as the successful requests on other nodes would still propagate via the internal replicator to the full node.
> 
>     2. Is there a graceful way to stop internal replicator jobs and avoid them going into a crash loop and making a mess to avoid replicating data to nodes which are getting full?
> 
>     3. I am wondering if at least for doc writes it would make sense to keep track of cluster wide disk usage (via a gossip between `couch_disk_monitor` gen_server or something like that) and return an immediate error at the fabric level without attempting any worker doc writes. That would increase the complexity of disk monitor, but would avoid the internal replicator loophole issue?

I think I will revise the block_database_writes() to be only _interactive_ writes. replicated writes will be permitted, as anti-entropy is critical.

src/couch/src/couch_db.erl
```diff
@@ -1338,6 +1338,15 @@ update_docs(Db, Docs0, Options, ?REPLICATED_CHANGES) ->
...
 update_docs(Db, Docs0, Options, ?INTERACTIVE_EDIT) ->
+    BlockInteractiveDatabaseWrites = couch_disk_monitor:block_interactive_database_writes(),
+    if
+        BlockInteractiveDatabaseWrites ->
+            {ok, [{insufficient_storage, <<"database_dir is too full">>}|| _ <- Docs0]};
+        true ->
+            update_docs_interactive(Db, Docs0, Options)
+    end.
+
+update_docs_interactive(Db, Docs0, Options) ->
```
nickva, line 1349: We should have some tests to assert what happens with single doc puts and bulk_docs. Could probably some existing fabric doc update tests or add some new ones (could be in chttpd too, if they are at the HTTP api level) then either mock `couch_disk_monitor:block_interactive_database_writes()` or maybe configure a very low limit and check that we get an expected result.

src/couch_mrview/src/couch_mrview_util.erl
```diff
@@ -176,6 +176,7 @@ get_view_index_state(_, DDoc, _, _, RetryCount) when RetryCount < 0 ->
     couch_log:warning("DDoc '~s' recreated too frequently", [DDoc#doc.id]),
     throw({get_view_state, exceeded_retry_count});
 get_view_index_state(Db, DDoc, ViewName, Args0, RetryCount) ->
+    BlockInteractiveIndexing = couch_disk_monitor:block_interactive_view_indexing(),
     try
         {ok, Pid, Args} = get_view_index_pid(Db, DDoc, ViewName, Args0),
         UpdateSeq = couch_util:with_db(Db, fun(WDb) ->
```
nickva, line 182: Wonder if could modify some view tests to run with `block_interactive_view_indexing() = true` and assert it does what we expect?

nickva (commented):
Handling the error at the doc level looks cleaner as it avoids dealing with the extra error in the complicated fabric_doc_update code.

~One interesting side-effect of allowing internal replicator to keep working and have errors at the doc level, is it might also allow the regular replicator to jump over documents and keep checkpointing. We should add a test for that too ("how replicator behaves when the target is full").  I wonder if that's what we'd want? Would we rather crash the replication job to prevent changes from streaming, or maybe it's like a 413 condition and skipping over docs makes sense.~

EDIT: Ignore that, I was silly and confused the previous behavior with the new one. With the new behavior replicated changes both from the internal replicator and the regular one won't throw any new errors.

Another interesting side-effect of allowing replicated changes is if we have uneven storage sizes on nodes, is for each shard range we'd keep filling the data until we reach the maximum of the _largest_ disk. That is if we have 3 nodes A, B, C where some 3 shard copies live, and A, B are 90% full, but C are only 50% full. We'd still try to funnel all the data into A and B until both C is not over 90% as well. This would also make the doc PUT responses random, if we hit A, B and get errors we'd return an error for that doc. If we hit A, C we'd get a success.

src/couch_mrview/src/couch_mrview_util.erl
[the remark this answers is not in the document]
rnewson: so yes, the change I made in maybe_update_others will inhibit lazy updates, as `ShardsNeedUpdated` will be the empty list and therefore we won't cast any `fabric_rpc` `update_mrview` messages.

test/elixir/test/disk_monitor.exs
```diff
@@ -0,0 +1,58 @@
+defmodule DiskMonitorTest do
+  use CouchTestCase
+
+  @moduletag :disk_monitor
+
+  setup_all do
+    set_config({"disk_monitor", "enable", "true"})
+    #on_exit(fn ->
+    #  set_config({"disk_monitor", "enable", "false"})
+    #end
```
nickva, line 10: This automatically gets reset by the test runner? Should we remove the comments then?

src/couch/src/couch_disk_monitor.erl
```diff
@@ -0,0 +1,241 @@
...
+handle_call(_Msg, _From, St) ->
+    {reply, {error, unknown_msg}, St}.
+
+handle_cast(_Msg, St) ->
+    {noreply, St}.
+
+handle_info(update_disk_data, St) ->
+    timer:cancel(St#st.timer),
+    update_disk_data(),
+    {ok, TRef} = timer:send_after(timer_interval(), update_disk_data),
```
nickva, line 101: suggest using `erlang:send_after/3` instead based on https://www.erlang.org/doc/efficiency_guide/commoncaveats.html#timer-module

src/couch/src/couch_disk_monitor.erl
```diff
@@ -0,0 +1,241 @@
...
+
+interactive_database_writes_threshold() ->
+    config:get_integer(?SECTION, "interactive_database_writes_threshold", 90).
+
+enabled() ->
+    config:get_boolean(?SECTION, "enable", false).
+
+%% Align our refresh with the os_mon refresh
+timer_interval() ->
+    application:get_env(os_mon, disk_space_check_interval, 30) * 60_000.
```
nickva, line 173: I think we can use [disksup:get_check_interval()](https://www.erlang.org/doc/man/disksup.html#get_check_interval-0) as that already returns the value in milliseconds and we don't have to hard-code the default here.

test/elixir/test/disk_monitor.exs
rnewson: oof sorry. I forget this. I can't get on_exit to work anywhere at all, I just get an error saying it can only be called from the test function. It doesn't work when I call it there or in setup/setup_all or anywhere at all. I think it's broken somehow.

rnewson: looks like a bug, or at least a very poor error message, in ExUnit.

rnewson: removed this section. as you say set_config adds its own on_exit. doing an on_exit from inside an on_exit is why I was 'not in the test handler function' despite appearances.

nickva (commented):
Looks almost ready to merge. Thanks for adding tests.

Only a few minor notes about using `erlang:send_after/3` and possibly simplify the time interval check.

I also noticed that by default disksup has a "disk almost full" threshold set to 80%

```
disksup:get_almost_full_threshold().
80
```

Do we intend to coordinate or adjust that value as well? As an experiment I forced it to trigger and noticed it set a few alarms:


```
> alarm_handler:get_alarms().
[]
> disksup:set_almost_full_threshold(0.08).
> disksup:set_check_interval(1).
... wait for a few minutes...
alarm_handler:get_alarms().
[{{disk_almost_full,"/System/Volumes/Update/mnt1"},[]},
 {{disk_almost_full,"/System/Volumes/Data"},[]},
 {{disk_almost_full,"/"},[]}]
```

It showed in the logs well:
```
> [notice] 2023-07-31T20:14:14.528679Z node1@127.0.0.1 <0.381.0> -------- rexi_buffer : cluster stable
[info] 2023-07-31T20:23:59.178803Z node1@127.0.0.1 <0.100.0> -------- alarm_handler: {set,{{disk_almost_full,"/"},[]}}
[info] 2023-07-31T20:23:59.178825Z node1@127.0.0.1 <0.100.0> -------- alarm_handler: {set,{{disk_almost_full,"/"},[]}}
[info] 2023-07-31T20:23:59.178845Z node1@127.0.0.1 <0.100.0> -------- alarm_handler: {set,{{disk_almost_full,"/System/Volumes/Data"},[]}}
[info] 2023-07-31T20:23:59.178861Z node1@127.0.0.1 <0.100.0> -------- alarm_handler: {set,{{disk_almost_full,"/System/Volumes/Data"},[]}}
[info] 2023-07-31T20:23:59.178875Z node1@127.0.0.1 <0.100.0> -------- alarm_handler: {set,{{disk_almost_full,"/System/Volumes/Update/mnt1"},[]}}
[info] 2023-07-31T20:23:59.178888Z node1@127.0.0.1 <0.100.0> -------- alarm_handler: {set,{{disk_almost_full,"/System/Volumes/Update/mnt1"},[]}}
```

It's not a big deal either way, but it's just something people might notice a new behavior. It would trigger for `root` or any other mounts, even swap it seems for Linux. We could try to set it as the max value of any of our thresholds (if have an 80 and a 90, we'd set it 90?).

rel/overlay/etc/vm.args
```diff
@@ -122,3 +122,15 @@
...
+
+# only start disksup
+-os_mon start_cpu_sup false
+-os_mon start_memsup false
+
+# Check disk space every 5 minutes
+-os_mon disk_space_check_interval 5
+
+# don't let disksup send alerts
+-os_mon disk_almost_full_threshold 100.0
```
nickva, line 136: I think this needs to be `1.0` 
  * https://www.erlang.org/doc/man/disksup.html#set_almost_full_threshold-1
  * https://www.erlang.org/doc/man/disksup.html#configuration is a bit confusing as it mentions floats and percentages though indicates 0.8 is the default

nickva (approved):
+1

PR #4687, opened 2023-07-21
send cookie on successful basic auth

rnewson:
DRAFT


big-r81:
Don't we need to do more than juste create the FullSecret, like checking the [cookie](https://github.com/apache/couchdb/blob/48338247bdd5fd1410bc4aa8c70579d6eee5af45/src/couch/src/couch_httpd_auth.erl#L349-L396)?

rnewson:
@big-r81 The PR sends a cookie, there isn't one in the request to check yet. If there was the cookie auth handler would have done so.

janl (approved):

PR #4688, opened 2023-07-21
Spawn nouveau process without additional shell process

big-r81:
Spawn nouveau without a parent shell process.


big-r81:
Superseeded by #4689.

PR #4702, opened 2023-07-26
send a session cookie after successful basic auth

rnewson:
## Overview

As a way to migrate users to a strong password hashing scheme without getting a performance hit each time, assuming the client will send back the cookie we send them. we check the cookie first and avoid the hit, but we still fall back to basic auth if necessary.

The AuthSession cookie is extended with a new field that indicates whether it was issued by the default authentication handler or the cookie authentication handler. If the cookie was issued by the default authentication handler the cookie authentication handler will check the username in the basic auth header, if present, is a match for the name in the cookie. If it is not, the cookie is ignored. This ensures that the basic auth header takes precedence for any client that got an unasked-for cookie and ensures their operations are performed as the user they intended.

h/t @glynnbird for the idea

## Testing recommendations

covered by tests in cookie_auth_tests.exs

N/A

## Checklist

### DIFF
--- a/src/couch/src/couch_httpd_auth.erl
+++ b/src/couch/src/couch_httpd_auth.erl
@@ -30,7 +30,6 @@
 
 -export([authenticate/2, verify_totp/2]).
 -export([ensure_cookie_auth_secret/0, make_cookie_time/0]).
--export([cookie_auth_cookie/4, cookie_scheme/1]).
 -export([maybe_value/3]).
 
 -export([jwt_authentication_handler/1]).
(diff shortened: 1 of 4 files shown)


rnewson:
note that I also took out a fair few unnecessary ?b2l and ?l2b's here as all the functions take iodata now. I confirmed locally that it makes no difference to the hmac, etc

src/couch/src/couch_httpd_auth.erl
```diff
@@ -30,7 +30,6 @@
 
 -export([authenticate/2, verify_totp/2]).
 -export([ensure_cookie_auth_secret/0, make_cookie_time/0]).
--export([cookie_auth_cookie/4, cookie_scheme/1]).
```
rnewson, line 33: not used from anywhere afaict, removing them as part of a tidyup

src/couch/src/couch_httpd_auth.erl
```diff
@@ -331,11 +341,15 @@ cookie_authentication_handler(#httpd{mochi_req = MochiReq} = Req, AuthModule) ->
...
+            % TimestampStr is expanded to be a list of options, separated
+            % by commas. The new second option is 'MustMatchBasic', a 0 or
+            % 1 to indicate if the basic auth username must match the cookie
+            % if present.
+            [User, OptionsStr, HashStr] =
                 try
                     AuthSession = couch_util:decodeBase64Url(Cookie),
                     [_A, _B, _Cs] = re:split(
-                        ?b2l(AuthSession),
+                        AuthSession,
```
rnewson, line 352: removed unnecessary b2l

janl (approved):

PR #4707, opened 2023-07-28
fix(`dreyfus`): amend the fix for faceting on zero values

pgj:
Commit 526e203d30acc3a29e4418a7317badddc37f5529 included a version of the fix that did not pass the Elixir tests.  Revert back to the original approach as that one appears to be the best (most flexible, shortest, and correct) solution.


rnewson (approved):
`==` vs `=:=` is the reason the shorter version isn't equivalent. I guess this is as tidy as it can be then.

PR #4732, opened 2023-08-15
Use GNU make vars for paths

big-r81:


nickva (approved):
+1  a very nice cleanup, Ronny

Tested configure, make, make eunit on windows and configure, make, make eunit, make dist, make release on macos.

It looks like all the `make` invocations we do in CI are with GNU make (Windows make.exe, gmake on BSD, my macos has gnu make as well).

PR #4735, opened 2023-08-22
feat(`mango`): rolling execution statistics (exploration)

### LINKED ISSUE #4560: Inconsistent execution statistics for Mango queries
When requested, the Mango execution statistics do not always reflect their actual values but something that is less than or simply zero, most of the times.  This happens on the latest version of CouchDB (as the time of writing, which is `82aa1625`) and the issue can be reproduced as follows.  The commands below create a simple database, called `test` with 25 documents of a single field, for which an index is defined.

```shell
curl -sS -X PUT "$COUCHDB_URL"/test
for i in $(jot - 1 25); do \
  curl -sS -X POST "$COUCHDB_URL"/test -H "Content-Type: application/json" -d '{"a": '"$i"'}'; done
curl -sS -X POST -H "Content-Type: application/json" "$COUCHDB_URL"/test/_index \
  -d '{"index": {"fields": ["a"]}, "name": "a", "type": "json"}'
```

Then, using a selector like `{"a": {"$lt": 20}}` and `limit` of 1, and with execution statistics enabled, the `total_docs_examined` field becomes zero.  The other `_examined` fields are either not implemented or not affected by queries like that.  It is visible that `results_returned` is properly accounted at the same time.

```console
$ curl -sS -X POST -H "Content-Type: application/json" "$COUCHDB_URL"/test/_find \
  -d '{"execution_stats": true, "limit": 1, "selector": {"a": {"$lt": 20}}}' \
  | jq '.execution_stats'
{
  "total_keys_examined": 0,
  "total_docs_examined": 0,
  "total_quorum_docs_examined": 0,
  "results_returned": 1,
  "execution_time_ms": 1.067
}
```

Note that if either `limit` is increased or the search criteria on the `a` field is changed to find less documents, more realistic data is returned.

```console
$ curl -sS -X POST -H "Content-Type: application/json" "$COUCHDB_URL"/test/_find \
  -d '{"execution_stats": true, "limit": 20, "selector": {"a": {"$lt": 20}}}' \
  | jq '.execution_stats'
{
  "total_keys_examined": 0,
  "total_docs_examined": 20,
  "total_quorum_docs_examined": 0,
  "results_returned": 19,
  "execution_time_ms": 2.378
}
```

or

```console
$ curl -sS -X POST -H "Content-Type: application/json" "$COUCHDB_URL"/test/_find \
  -d '{"execution_stats": true, "limit": 1, "selector": {"a": {"$lt": 2}}}' \
  | jq '.execution_stats'
{
  "total_keys_examined": 0,
  "total_docs_examined": 2,
  "total_quorum_docs_examined": 0,
  "results_returned": 1,
  "execution_time_ms": 0.845
}
```

After some debugging, the source of the issue has been identified as the emission of `stop` in `mango_cursor_view:handle_doc/2` when the limit reaches zero (last clause). 

https://github.com/apache/couchdb/blob/82aa1625ca8b652f7db3dadad741e86239641921/src/mango/src/mango_cursor_view.erl#L483-L497

The `stop` action immediately stops the processing of messages from the shards, including the shard-level statistics that are submitted in response to the `complete` message.

https://github.com/apache/couchdb/blob/82aa1625ca8b652f7db3dadad741e86239641921/src/mango/src/mango_cursor_view.erl#L385-L390

That is why when the limit is too low, there is not enough time to receive and process the related message.  Therefore the execution statistics are not, or only partially (from a set of shards) handled.

I could not yet come up with a satisfying solution but I create this ticket to raise awareness about this bug.  Changing `stop` to `ok` in `mango_cursor_view:handle_doc/2` helps with the consistency but then too many results are sent by the shards which will be known to be discarded already.  This potentially has an impact on the performance, although it has not been measured how much.

pgj:
## Overview

In case of map-reduce views, the arrival of the `complete` message is not guaranteed for the view callback (at the shard) when a `stop` is issued during the aggregation (at the coordinator).  Due to that, internally collected shard-level statistics may not be fed back to the coordinator which can cause data loss hence inaccuracy in the overall execution statistics.

Address this issue by switching to a "rolling" model where row-level statistics are immediately streamed back to the coordinator.  Support mixed-version cluster upgrades by activating this model only if requested through the map-reduce arguments and the given shard supports that.

This is only a proposal, a way explore the approach, comments and feedback are welcome.  Remarks:
- It clearly uses more bandwidth as it will send more `execution_stats` messages.  Is this acceptable?
- It does not require changes in fabric itself, the whole logic is kept on the Mango side and it is compatible with the previous solution.
- Are there other ways to stop the aggregation as soon as the limit is reached?
- Are there other kind of messages coming from the shards that could be reliably used to signal the end of statistics collection (alternatives to capturing `complete`)? 

## Testing recommendations

Running the respective Mango unit and integration test suites might suffice (which is done by the CI):

```shell
make eunit apps=mango
make mango-test MANGO_TEST_OPTS="15-execution-stats-test"
```

But there is a detailed description in related the ticket (see below) on how to trigger the problem.  Feel free to kick the tires.

Fixes #4560

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -55,6 +55,15 @@
         covering_index => 'maybe'(#idx{})
     }.
 
+-type mrargs_extra_item() ::
+    {callback, {atom(), atom()}}
+    | {selector, any()}
+    | {callback_args, viewcbargs()}
+    | {ignore_partition_query_limit, boolean()}
+    | {execution_stats_map, boolean()}
+    | {execution_stats_rolling, boolean()}.
+-type mrargs_extra() :: [mrargs_extra_item()].
+
 -spec viewcbargs_new(Selector, Fields, CoveringIndex) -> ViewCBArgs when
     Selector :: selector(),
     Fields :: fields(),
@@ -199,10 +208,13 @@ base_args(#cursor{index = Idx, selector = Selector, fields = Fields} = Cursor) -
 
             {ignore_partition_query_limit, true},
 
-            % Request execution statistics in a map.  The purpose of this option is
-            % to maintain interoperability on version upgrades.
-            % TODO remove this option in a later version.
-            {execution_stats_map, true}
+            % The purpose of the following options is to maintain
+            % interoperability on version upgrades:
+            % TODO: Remove them in a later version
+            % - Return execution statistics in a map
+            {execution_stats_map, true},
+            % - Stream execution statistics
+            {execution_stats_rolling, true}
         ]
     }.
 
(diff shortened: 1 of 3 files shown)


src/mango/src/mango_cursor_view.erl
```diff
@@ -337,19 +349,48 @@ choose_best_index(IndexRanges) ->
...
 -spec view_cb
     (Message, #mrargs{}) -> Response when
         Message :: {meta, any()} | {row, row_properties()} | complete,
         Response :: {ok, #mrargs{}};
     (ok, ddoc_updated) -> any().
-view_cb({meta, Meta}, Acc) ->
+view_cb({meta, Meta}, #mrargs{extra = Options} = Acc) ->
     % Map function starting
-    mango_execution_stats:shard_init(),
+    case couch_util:get_value(execution_stats_rolling, Options, false) of
```
rnewson, line 386: I'd pull this logic into its own function to keep the callbacks short.

rnewson (commented):
seems neat and tidy to me. I'm surprised we're losing this information today or that 'completion' is ambiguous but perhaps I don't understand your concerns in the ticket info.

rnewson:
That's very helpful background. You are right that it was t anticipated we'd need information back about a worker we know we no longer need to calculate the response. 

It would be better to address that directly if we can.

rnewson:
To expand on that last comment, perhaps we alter how workers end? Today we do `exit(Pid, kill)` which is obviously untrappable. As an option to `fabric_util:cleanup/1` perhaps we could instead do `exit(Pid, finish)` (or some other word). I think the workers would get a `{rexi_EXIT, finish}` message. On receipt of that they'd "complete" whatever made sense for them (in our case, simply messaging out the execution stats up to that point, but not performing any new work). The fabric coordinator would have to wait for either that reply or notification that the worker processes terminated for some other reason, before finishing up.

This way ensures we send no more messages than today, at the small cost of making the coordinator wait for one message from each worker it would normally have unilaterally and asynchronously killed.

pgj:
Because I did not want to lose the original description of this PR along with the discussion here, and I wanted to put the change in a different perspective in the light of #4812, I forked it as #4958.  After talking with @chewbranca about the problem and its solution, and adding the fact that he has been working on a model that would follow a similar approach, the current code seems feasible.

I have studied @rnewson's suggestions but neither of them brought a clear solution to the problem.  I am inclined to believe (perhaps I am wrong here) that fixing the issue from the side of `fabric` would be a more complicated change and as such it hides some risks.

pgj:
Closing this in favor of #4958.

PR #4743, opened 2023-08-28
Ban the last monster

big-r81:
Set an example cookie like in the cluster setup docs.


nickva (approved):

kxepal:
Awesome title! Truly there are no monsters under my couch :joy:

PR #4749, opened 2023-09-05
Nouveau ibrowse improvements

rnewson:
## Overview

This is a performance only pull request, but a complicated one.

I benchmarked nouveau against clouseau for the same database and found it was slower (2-4x slower depending on how large the documents were, the smaller the docs the bigger the difference).

This is fundamentally because clouseau's overhead between erlang and the jvm is much lower, as they are exchanging messages over an established, dedicated connection.

To improve nouveau performance I have adopted a similar approach, while ensuring index consistency.

1) nouveau on the jvm side requires each index update request (whether a document update, delete or purge, or a request to advance update_seq or purge_seq independently from an index change) to pass the current update_seq or purge_seq of the index. This is critical in detecting missed updates in the events of failure.
2) nouveau_index_updater spawns its own ibrowse worker (and, thus, its own dedicated http connection) when it starts to update an index. This ensures that all requests issued from the erlang side are received in the same order on the jvm side. Using the ibrowse pool would have allowed for a reordering.
3) finally, the performance improvement itself. nouveau_index_updater sets the `[{stream_to, self()}]` option, so does not wait for the response, successful or otherwise, before making the next one. This is why item 1 is critical. If there is a failure for one update, all subsequent updates will also fail, as the update/purge sequence on the jvm side will not have advanced. At the start of each loop of the callback fun we do a selective receive to collect any and all async responses we might have received. If any are errors, we terminate the update process with that error. We also check after the callback fun is finished but this time we block to receive all the outstanding responses, to ensure the index is in the state we require before issuing the request.

I've separated out these changes across multiple commits for readability.

## Testing recommendations

covered by the test suite

N/A

## Checklist

### DIFF
--- a/nouveau/gradle/wrapper/gradle-wrapper.properties
+++ b/nouveau/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.1.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.3-bin.zip
 networkTimeout=10000
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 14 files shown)


rnewson:
Noting that these changes bring nouveau index build times in line with clouseau index build times; they come out the same.

PR #4755, opened 2023-09-11
Add tests for X-Forwarded-For

iilyak:
## Overview

This PR adds test cases to verify that we log the real IP address of a user when X-Forwarded-For header is present.  

## Testing recommendations

```
make eunit apps=chttpd, suites=chttpd_server_tests
```

- https://github.com/apache/couchdb/pull/4753

## Checklist

### DIFF
--- a/src/chttpd/test/eunit/chttpd_server_tests.erl
+++ b/src/chttpd/test/eunit/chttpd_server_tests.erl
@@ -1,0 +1,239 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
+% http://www.apache.org/licenses/LICENSE-2.0
+%
+% Unless required by applicable law or agreed to in writing, software
+% distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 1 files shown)


rnewson:
I like the coverage but I don't think it tests _couchdb_. It asserts the behaviour of a mochiweb function, but not whether we call it and use the result to populate `peer`

There are a few problems in general;

1) we don't document that `peer` honors x-forwarded-for in the docs afaict, and obviously there's no test for it
2) third party applications can't safely use the internal `httpd` record or assume what `peer` means in the absence of 1)

I think this test needs to be on the elixir integration suite instead and assert that the client ip is logged, but also there should be a public function in couchdb that other apps can call that explicitly says how it calculates peer. 

couchdb's current usage is fine, it asks mochiweb for the peer, using a public function, and logs it. Apps making assumptions about that, particularly about unpublished internals, are skating on thin ice. doing so without an integration test just invites disappointment.

rnewson:
I should say, even if it does test couchdb (that is touch real chttpd and real mochiweb, not mocked versions), the problem we're addressing was not what couchdb logged anyway. it was other apps co-opting the `peer` value to log it in an entirely separate way. A PR for that specifically is in order.

iilyak:
Closing as this is not addressing the real problem.

PR #4759, opened 2023-09-13
Use HTTP rules for hostname verification

rnewson:
## Overview

By default, erlang does not recognise wildcards in Server Name Indication fields (which is bizarre but they seem committed to this error).

Fortunately they also ship a function that does validate them, you just have to hook it up.

So here we do that hooking up.

## Testing recommendations

We are following the published advice at https://www.erlang.org/doc/man/ssl.html#type-customize_hostname_check;

> Customizes the hostname verification of the peer certificate, as different protocols that use TLS such as HTTP or LDAP may want to do it differently. For example the get standard HTTPS handling provide the already implememnted fun from the public_key application for HTTPS. `{customize_hostname_check, [{match_fun, public_key:pkix_verify_hostname_match_fun(https)}]}`

N/A

## Checklist


nickva (approved):
+1

PR #4763, opened 2023-09-15
fix(`mango`): plug hole in unit test coverage of view cursor functions

pgj:
Add a missing `eunit` case for testing error handling of `mango_cursor_view:handle_message/2`.  This was noticed while working on #4735, but I feel it is better to factor this piece out and submit in a separate (and dedicated) commit.

## Testing recommendations

```
make eunit apps=mango
```

The coverage of `mango_cursor_view` should jump from 99% to 100%.


nickva (approved):
+1

pgj:
Thanks @nickva!

PR #4764, opened 2023-09-15
fix(`chttpd`): comment typo

pgj:
A follow-up commit to deddbb2 where a typo was noticed in the contents after it has been merged.


pgj:
Thanks @iilyak!

PR #4772, opened 2023-09-22
Add at least one approval before merging a PR

big-r81:
~~Add additional branch checks before merging a PR.~~

~~1. Require a linear history for better redability.~~
~~2. Disable "merge commits".~~
~~3. Require at least one approval.~~

Edited:

As a first smaller change, add at least one approval before merging a PR.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -27,6 +27,12 @@ github:
         strict: true
         contexts:
           - continuous-integration/jenkins/pr-merge
+      required_pull_request_reviews:
+        # Set to true if you want to automatically dismiss approving reviews when someone pushes a new commit.
+        dismiss_stale_reviews: false
+        # Specifies the number of reviewers required to approve pull requests.
+        # Use a number between 1 and 6 or 0 to not require reviewers.
+        required_approving_review_count: 1
     fdbmain:
       required_status_checks:
         strict: true


rnewson:
I dislike the inability to do a merge commit. For larger pull requests I deliberately split them into multiple commits that are easy to review (at the time but, importantly, retrospectively). CouchDB may not be usable at each step, only before or after the set. Flattening these comments to form a fake linear history will lose that context, will yield broken builds at some of those comments, and the history does not actually reflect the history that the project took.

big-r81:
Okay, than I would like to know, how others review commits. 

For my side, I use the GH UI and check all changes (from all commits) as a whole and if it's a larger change, then I used the checkmark that I viewed the file. If no changes are made again (in the meantime) I would give my approval/request/comment ... 

I think it's totally okay to give seperate commits during development, but if it should land in main, then it should be squashed to exactly one "atomic" commit. One feature/fix one commit. If something is wrong with that, it can be easily reverted or fixed.

This way would solve two of your hints. CouchDB is always usable after each commit and there are no "sub-feature/bug-commits" on the history. If a "sub-feature/bug" is important for history, then it should become an own commit.

big-r81:
Hey, I updated the PR and removed the disabling of merge requests and the force of a linear history.
The only additional check is, that we need at least one approval to be allowed to merge the PR.

Nick @nickva, I had your approval before, but would like to get a new one (from others too). Maybe @rnewson has additional comments?

pgj:
Thanks @big-r81 for working on this!

PR #4774, opened 2023-09-24
fix(`docs`): amend description of `use_index` on `/{db}/_find`

pgj:
The semantics of Mango's `use_index` query parameter has changed over time but this has not been reflected in the documentation, which causes a lot of confusion.

The `use_index` parameter was introduced in 1b0426aa to force the index selection to a specific index.  Unfortunately, this did not work out well in practice, so 743bd882 added a fallback mechanism to make it less brittle.  With that, `use_index` became only a "hint" not an "instruction".

### DIFF
--- a/src/docs/src/api/database/find.rst
+++ b/src/docs/src/api/database/find.rst
@@ -42,9 +42,14 @@
         should be returned. If it is omitted, the entire object is returned.
         More information provided in the section on :ref:`filtering fields
         <find/filter>`. *Optional*
-    :<json string|array use_index: Instruct a query to use a specific index.
-        Specified either as ``"<design_document>"`` or
-        ``["<design_document>", "<index_name>"]``. *Optional*
+    :<json string|array use_index: Request a query to use a specific
+        index. Specified either as ``"<design_document>"`` or
+        ``["<design_document>", "<index_name>"]``. It is not
+        guaranteed that the index will be actually used because if the
+        index is not valid for the selector, fallback to a valid index
+        is attempted. Therefore that is more like a hint. When
+        fallback occurs, the details are given in the ``warning``
+        field of the response. *Optional*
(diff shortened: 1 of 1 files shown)


rnewson:
The discussion in https://github.com/apache/couchdb/pull/4710 is ongoing with advocates for fixing `use_index` to work the way the documentation says.

I think we'll need a vote on it to decide.

pgj:
@rnewson please note that I have included references to the commit history to support that the way `use_index` currently works is not a bug but a feature, a mitigation to the valid problem.

The bug here is that the documentation has not been updated.  Also, it happened a while (7 years) ago and independently of this was a good design decision or not, it is now part of CouchDB and there may be integrations on that work under these assumptions.  Especially that after the expected change CouchDB would immediately start emitting HTTP 400s instead of HTTP 200s in those scenarios.  How this work out in a multi-version cluster?

In #4710 I was proposing ways to change the behavior of `use_index` in backward-compatible ways (i.e. rename `use_index` to `index_hint` and introduce `force_index` to restore the old behavior, also the toggles) but it was rejected.  While I hear you and understand all your concerns and assume good intentions behind your comments, I do not think restoring `use_index` to the old, assumed good behavior right away is the good solution.  Or I might be misunderstanding something?

pgj:
Yes, I am happy to discuss that on `dev@`.  Please give me a day or so to prepare the topic starter email.

pgj:
@rnewson given that commenters are about to agree regarding #4710, do you still think that launching a discussion and potentially a vote on `dev@` would be necessary?

rnewson:
I think no longer needed.

pgj:
Based on the recent conclusion of #4710, @rnewson are you fine with having this PR merged?

rnewson (approved):

pgj:
Thanks @rnewson !

Issue #4775, opened 2023-09-25
Inefficient `$regex` Mango `view` queries on indexed fields


pgj:
On the Mango query interface, when the `$regex` operator is used in a selector with a field that could be backed up by a `view` (`json`) index, it is not leveraged.  The use of `$regex` in its current form rather implies a brute-force scan through all the documents of the underlying view, which makes the whole approach inefficient and resource-intensive.

The source of the problem is that no index ranges could be computed for `$regex` (to narrow down the scan) because regular expressions do not naturally generate an ordering on the set of strings, like for example comparisons do on integers.  However, there is a prior art in for example, Lucene where it is possible to achieve that for specialized cases, such as "starts with"-type of expressions, i.e. `^foo.*`.

Some more links about how Lucene does it, just for inspiration:
- https://lucene.apache.org/core/4_2_0/core/org/apache/lucene/search/RegexpQuery.html
- https://lucene.apache.org/core/4_2_0/core/org/apache/lucene/search/AutomatonQuery.html
- https://blog.mikemccandless.com/2011/03/lucenes-fuzzyquery-is-100-times-faster.html
- https://www.slideshare.net/otisg/finite-state-queries-in-lucene

Another factor that can slow down the evaluation is that documents are included unconditionally, and `$regex` is not matched against the key itself first and only documents for the matched keys are fetched and returned.  Something along the lines of [covering indexes](https://github.com/apache/couchdb/issues/4413).

rnewson:
it would be nice (and I think fairly easy) to at least narrow the query with startkey/endkey if the regex does not start with a wildcard. Parsing the regex to determine that sounds like the tricky part.

PR #4776, opened 2023-09-26
use text prefix in regex to speed up query

rnewson:
## Overview

for selector;

{"selector":{"_id":{"$regex":"doc.+"}}}

before;

{
  "include_docs": true,
  "view_type": "map",
  "reduce": false,
  "partition": null,
  "start_key": [],
  "end_key": [
    "<MAX>"
  ],
  "direction": "fwd",
  "stable": false,
  "update": true,
  "conflicts": "undefined"
}

after;

{
  "include_docs": true,
  "view_type": "map",
  "reduce": false,
  "partition": null,
  "start_key": [
    "doc"
  ],
  "end_key": [
    "doc�",
    "<MAX>"
  ],
  "direction": "fwd",
  "stable": false,
  "update": true,
  "conflicts": "undefined"
}

## Testing recommendations

TBD

https://github.com/apache/couchdb/issues/4775

## Checklist


rnewson:
I've not done the `text` side yet, or any tests. just sounding out the idea.

for `text` I'd prefer to pass the regex through to Lucene (clouseau or nouveau) and document the variation in regex flavour (there's huge overlap). Omitting the optimizations Lucene makes for the sake of purity was a mistake in the original implementation imo.

willholley:
I wonder whether a `$startsWith` operator would be cleaner, as we could then optimize it for text specifically? The `$regex` operator originally did differ for text indexes iirc but we had users experience weirdness when they would add an index and suddenly get different results.

A general principal in Mango over the last ~5 years is that adding an index shouldn't change the result of a query implicitly, so I'd be wary of reintroducing that behaviour.

PR #4800, opened 2023-10-05
revert(`mango`): fix for duplicated text results

pgj:
This PR is to remove the commits from #4782 about the fix itself.  During the review a regression was introduced which went unnoticed until recently.  It causes the Mango integration test suite hang and possibly breaks the integration of `/{db}/_find` with Clouseau.


pgj:
It has been superseded by #4801, so I am now closing this.

PR #4801, opened 2023-10-05
fix(`mango`): correct the fix for the duplicated `text` results

pgj:
This is a follow-up to #4782.  Commits d6d4ad97 and ce2607a5 missed out on the necessity of keeping the bookmark updated all the time.  Otherwise the same hits might be returned which can lead to an infinite loop at the processing and make the integration test suite hang, for example.

This change is covered by the existing unit and integration tests.


rnewson (approved):

nickva (approved):
+1

PR #4806, opened 2023-10-16
mango: do not skip json tests when clouseau installed

willholley:
## Overview

Previously, Mango tests which only apply to JSON indexes were being skipped if the clouseau (text index) service was enabled for the CouchDB instance being tested against, resulting in tests being skipped in some CI cases.

This refactors the tests so that we execute these tests on JSON / all_docs indexes regardless of whether clouseau is present or not.

## Testing recommendations

`make mango-test` with / without clouseau enabled.

Note that `nose2` will report skipped tests in the `Ran N tests` output. With this change, we see the same number of tests executed when `clouseau` is not enabled:

Before:
```
Ran 373 tests in 11.328s

OK (skipped=146)
```

After:
```
Ran 368 tests in 12.104s

OK (skipped=141)
```

With `clouseau` enabled we see fewer skipped tests after this change (so more overall):

Before:
```
Ran 373 tests in 37.748s

OK (skipped=16)
```

After:
```
Ran 368 tests in 35.741s

OK (skipped=1)
```

## Checklist

### DIFF
--- a/src/mango/test/03-operator-test.py
+++ b/src/mango/test/03-operator-test.py
@@ -14,62 +14,164 @@ import mango
 import unittest
 
 
-class OperatorTests:
-    def assertUserIds(self, user_ids, docs):
-        user_ids_returned = list(d["user_id"] for d in docs)
-        user_ids.sort()
-        user_ids_returned.sort()
-        self.assertEqual(user_ids, user_ids_returned)
+class BaseOperatorTests:
+    class Common(object):
+        def assertUserIds(self, user_ids, docs):
+            user_ids_returned = list(d["user_id"] for d in docs)
+            user_ids.sort()
+            user_ids_returned.sort()
+            self.assertEqual(user_ids, user_ids_returned)
 
-    def test_all(self):
-        docs = self.db.find(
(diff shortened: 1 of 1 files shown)


pgj:
I tested this with and without Clouseau and I can confirm the results.

src/mango/test/03-operator-test.py
```diff
@@ -14,191 +14,180 @@
...
-        for d in docs:
-            self.assertIn("twitter", d)
-
-    # ideally this work be consistent across index types but, alas, it is not
-    @unittest.skipUnless(
-        not mango.has_text_service(),
-        "text indexes do not support range queries across type boundaries",
-    )
+class BaseOperatorTests:
+    class Common(object):
```
pgj, line 18: Why do we need this `Common` subclass?

willholley: This was just to clarify the scope of the common tests - that they should only be used as a mixin to the JSON/text/AllDocs test classes rather than being a standalone test suite.

pgj (approved):
Nice find @willholley, thanks!

PR #4829, opened 2023-11-02
do-not-merge(`mango`): fix derivation of upper bound for the range of `$beginsWith`

pgj:
This is a dummy PR to trigger a full CI run to test the fix for #4810.  The code is derived from #4828.


pgj:
@willholley FYI

pgj:
This PR is no longer needed therefore closing it.

PR #4835, opened 2023-11-09
feat(`dev`): add Clouseau to the developer setup

pgj:
Provide a way to the developers to deploy and launch the Clouseau search module more easily, therefore making it more accessible. This can help with running the Search-based Elixir tests as well as the Mango `text` search tests.  This could allow us to catch more bugs ahead of time and might even inspire more improvements in the area.

The extension is designed in a way to make it simple to integrate with the CI — only the Java environment of the proper version needs to be deployed, everything else could be managed from this repository.  Such as bumping the version of Clouseau or fine-tuning the configuration parameters.

This is an opt-in feature on two levels.  First, one has to tell `./configure --enable-clouseau` to instantiate Clouseau locally. This will create the `clouseau` sub-directory that holds the contents of the Clouseau distribution of the specified version, which currently defaults to 2.22.0 (the latest).  If an older version is needed, use the `--clouseau-version` flag.

Then the `mango-test` and `elixir-search` targets will try to use Clouseau automatically just to make it easy to use in the old way. However, the `dev/run` script will not do the same.  It requires the `--with-clouseau` flag to be passed for that.  That is because the developer may not necessarily want to launch Clouseau ad hoc, even if it is available.

With this, note that `elixir-search` is added to the `check` target. In lack of a configured Clouseau instance, it will become a no-op and a warning is emitted.

## Testing recommendations

There are many ways to try this change.  The conventional approach is to have some flavor of JDK 8 installed, via [`asdf`](https://asdf-vm.com/) for example, configure CouchDB to deploy Clouseau and run the corresponding test suites.

```console
asdf plugin add java
asdf install java zulu-8.74.0.17
export CLOUSEAU_JAVA_HOME=$(asdf where java zulu-8.74.0.17)
./configure --enable-clouseau <other-configure-flag-of-yours>
make elixir-search
make mango-test
```

The `elixir-search` target should not emit a warning (but run the tests) and `mango-test` should have only a single skipped test.

The `dev/run` script could be run individually where Search could be talked to, e.g.:

```console
dev/run --with-clouseau -n 1 -a adm:pass
curl -sS http://127.0.0.1:15984/ | jq -rc '.features'
```

The last command should show something similar to this:

```console
["search","access-ready","partitioned","pluggable-storage-engines","reshard","scheduler"]
```

[...]

PR #4855, opened 2023-11-17
Jenkins: full build with Clouseau included, please

pgj:
This is a technical PR to trigger the full CI run to see if #4835 would work if it was merged.


pgj:
This is not needed any more.

PR #4856, opened 2023-11-17
Jenkins: test the PR with clouseau, please

pgj:
This is a technical PR to trigger Jenkins to see if #4835 works as expected for regular PRs.  The original does not work for this purpose as the corresponding branch is not on `apache/couchdb`.


pgj:
This is not needed any more.

PR #4861, opened 2023-11-19
chore(`dev`): provide better error message on `make` missing before `run`

pgj:
The `./dev/run` script requires a boot script to be compiled which assumes a completed build of the source tree.  If this did not happen, the script fails to work and CouchDB cannot be started up.

The failure of the boot script's compilation by itself does not necessarily help to understand the reason, so enhance it with a suggestion for the user.

## Testing recommendations

The behavior should change when one tries to run the `./dev/run` script without calling `make`.  This can also be simulated by running `make devclean` and `make clean` if the source have already been build before.  This intended to be an improvement for the developer experience, it should not break anything.  Instead, there shall be less confusion around this unhappy case.

### DIFF
--- a/dev/run
+++ b/dev/run
@@ -90,10 +90,10 @@ log.verbose = True
 
 
 def main():
-    ctx = setup()
     try:
+        ctx = setup()
         startup(ctx)
-    except ClouseauError:
+    except StartupError:
         sys.exit(1)
     else:
         if ctx["cmd"]:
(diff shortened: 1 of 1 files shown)


nickva (approved):
+1

I get:
```
failed: Boot script could not be created.  Perhaps you forgot to run `make`?
```

A very nice developer experience improvement. I had run into this many times in the past.

pgj:
Cool, thanks @nickva !

PR #4872, opened 2023-11-25
first attempt at building nouveau in github actions

rnewson:


.github/workflows/nouveau.yml
```diff
@@ -0,0 +1,20 @@
+name: Run Gradle on PRs
+on: pull_request
+jobs:
+  gradle:
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest ]
```
big-r81, line 7: Hey, I think there is no need to exclude Windows here :wink:, see https://github.com/gradle/gradle-build-action#use-the-action-to-setup-gradle

rnewson: I excluded it because it _failed_ when I tried it.

willholley:
If it's a useful reference, [couchdb-helm](https://github.com/apache/couchdb-helm) uses GH actions to test/publish. It attempts to follow the ASF guidance at https://cwiki.apache.org/confluence/display/BUILDS/GitHub+Actions+status which requires any external actions to be included via Git submodules.

PR #4874, opened 2023-11-27
doc: take out w + n params, folks get confused.

janl:


nickva (commented):
These are valid parameters. Wonder if there is a specific confusing point we can clarify perhaps?

Is this mostly about `w=3`? We could add a section about the common pitfalls about it , say if we use `w=3` but you only two or one nodes up, or describe a scenario with how conflicts can still be generated, and difference between 201 and 202 responses for writes.

PR #4908, opened 2023-12-10
fix(`mango`): communicate rows read for global stats collection

pgj:
Besides the execution statistics that could be shown for the user, the number of rows and documents read are tracked separately on the global level.  (Which is more like a bug than a feature, but that is out of the scope of this change.)

With the introduction of covering indexes, these two have been become different, therefore global counting has to be taught about the number of times when no documents were read but rows only.  For `text` indexes, the number of rows and documents should be the same, the distinction makes sense only for `json` indexes.

## Testing recommendations

The changes are covered by the respective portion of the tests:

```shell
make eunit apps=mango
make mango-test
```

The global stats should always report the same values as it is added for the (Mango) execution stats.

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_text.erl
+++ b/src/mango/src/mango_cursor_text.erl
@@ -181,13 +181,14 @@ handle_hit(CAcc0, Sort, Doc) ->
     #cacc{
         limit = Limit,
         skip = Skip,
-        execution_stats = Stats,
+        execution_stats = Stats0,
         documents_seen = Seen
     } = CAcc0,
     CAcc1 = update_bookmark(CAcc0, Sort),
-    Stats1 = mango_execution_stats:incr_docs_examined(Stats),
+    Stats1 = mango_execution_stats:incr_keys_examined(Stats0),
+    Stats = mango_execution_stats:incr_docs_examined(Stats1),
     couch_stats:increment_counter([mango, docs_examined]),
-    CAcc2 = CAcc1#cacc{execution_stats = Stats1},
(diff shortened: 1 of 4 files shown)


iilyak (approved):
+1

chewbranca (approved):

src/mango/src/mango_cursor_text.erl
```diff
@@ -631,10 +619,11 @@ t_execute_more_results(_) ->
     ),
     meck:expect(mango_selector_text, convert, [selector], meck:val(query)),
     meck:expect(mango_selector, match, fun(selector, {doc, _}) -> true end),
-    ?assertEqual({ok, {acc, 6}}, execute(Cursor, fun foo:normal/2, {acc, 0})),
-    ?assertEqual(3, meck:num_calls(couch_stats, increment_counter, 1)),
-    ?assertEqual(3, meck:num_calls(mango_execution_stats, incr_docs_examined, 1)),
-    ?assertEqual(3, meck:num_calls(mango_execution_stats, incr_results_returned, 1)).
+    meck:expect(chttpd_stats, incr_rows, [0], meck:val(ok)),
```
pgj, line 622: @iilyak, @chewbranca please note that `text` search will not modify the `rows` value.  Is this acceptable or shall they also update the value?  If yes, how?

pgj: I have talked about this with @willholley and @mikerhodes offline.  The conclusion was that `rows` should be equal to `reads` for `text` indexes.  The distinction between these two categories make difference only for `json` indexes, especially when covering indexes are in use.

Issue #4922, opened 2023-12-15
Regex containing umlauts in square brackets not matching


muety:
## Description
I might be mistaken, but the following seems like a bug to me. Regexes that contain a German umlaut (or probably any other special character) wrapped between curly braces (like `^foo[ä]b`) won't match as expected, e.g. won't match the word `fooäbar`.

## Steps to Reproduce

Suppose you have the following user document in your database:

```json
{
  "_id": "3f6b99deb2a9d8ba80ac3a8cf3000558",
  "_rev": "1-889fc2635b372fa8b0f436fe6c6d4393",
  "lastname": "Mütsch",
  "firstname": "Ferdinand",
  "created": 1702629477669
}
```

Try to run these queries against the `/_find` endpoint:

**Query 1:**

```json
{
   "selector": {
      "lastname": {
         "$regex": "(?i)^müt"
      }
   }
}
```

**Expected:** returns above document
**Actual:** returns above document :white_check_mark: 

**Query 2:**
```json
{
   "selector": {
      "lastname": {
         "$regex": "(?i)^mü[t]"
      }
   }
}
```

**Expected:** returns above document
**Actual:** returns above document :white_check_mark: 

**Query 3:**
```json
{
   "selector": {
      "lastname": {
         "$regex": "(?i)^m[ü]t"
      }
   }
}
```

**Expected:** returns above document
**Actual:** returns empty result :x: 

According to my tests on Regex101 with PCRE, the last query should match the above document's `lastname` field, but for some reason doesn't. 

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )
[TIP]:  # ( You can paste the output of curl http://YOUR-COUCHDB:5984/ here. )

* CouchDB version used: 3.3.3 (inside Docker with latest `3`-tagged image)
* Browser name and version: Firefox 120.0.1
* Operating system and version: Fedora 38 / whatever OS is used in the Docker image

willholley:
The Erlang [regex syntax](https://www.erlang.org/doc/man/re.html#sect7) can be a little unusual. In this case, I think the issue is that, as described in [Square Brackets and Character Classes](https://www.erlang.org/doc/man/re.html#square-brackets-and-character-classes), you need to set `UTF mode` to get the behaviour you expect.

Changing the query to:

```json
{
   "selector": {
      "lastname": {
         "$regex": "(*UTF8)(?i)^m[ü]t"
      }
   }
}
```

should give you the correct result.

[...]

PR #4925, opened 2023-12-18
fix(docs): Break long lines for better readability within tables

big-r81:
See #4922 for background.

Before PR:

grafik

After PR:

grafik


nickva (approved):
+1

PR #4935, opened 2024-01-01
rfc(SSOT): Add Single Source of Truth (SSOT) configuration system

big-r81:
Initial ideas of a Single Source Of Truth config system for CouchDB.

### DIFF
--- a/src/docs/rfcs/018-ssot-config-system.md
+++ b/src/docs/rfcs/018-ssot-config-system.md
@@ -1,0 +1,128 @@
+---
+name: Formal RFC
+about: Submit a formal Request For Comments for consideration by the team.
+title: 'Single Source Of Truth (SSOT) system for CouchDBs configuration files'
+labels: rfc, discussion
+assignees: 'ronny@apache.org'
+
+---
+
+# Introduction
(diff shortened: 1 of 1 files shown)


rnewson:
Before we get too deep into this I'd like to see us tidy up what we currently have Specifically;

move code defaults to priv config files:

Like `stats_descriptions.cfg`, for each application create a file called `config_defaults.cfg`. This file could look like;

```
#{
  couchdb => #{
    max_dbs_open => #{
      default => 500,
      description => <<"Maximum number of .couch files to open at once. ...">>
    }
  }
}
```

These files would _replace_ the commented out items in `default.ini`. An endpoint could be added to return this data (but not allow them to be modified).

`config:get/2` would be enhanced to retrieve the default from these config files (if present, `undefined` if not) and all uses of `config:get/3` would be removed, so that the only copy of the default value for a given config item is in the `config_defaults.cfg`.

This means `default.ini` is empty except for things like `database_dir` with an actual install-specific value (whether from dev/run or a package install. Arguably it could be eliminated entirely and all of those settings could be in `local.ini`.

PR #4955, opened 2024-01-09
fix(docs): revamp the installation instructions for FreeBSD

pgj:
The documentation on how to have CouchDB installed and enabled as a service on FreeBSD contains a number of outdated information. This change is to bring it to up to date along the following lines:

- The port/package has been renamed to `couchdb3`.
- The configuration options and location for the configuration files for the port have changed.
- Propagate the `service` command to work with the wrapper script instead of accessing the script itself directly.
- Mention the binary package path (installation via `pkg`) which has become quite a standard these days.
- Remove instructions on manually creating the `couchdb` user and group.  There is no need for those any more, it shall be handled by the port/package.

### DIFF
--- a/src/docs/src/install/freebsd.rst
+++ b/src/docs/src/install/freebsd.rst
@@ -16,71 +16,66 @@
 Installation on FreeBSD
 =======================
 
-Installation from ports
-=======================
+Installation
+============
 
-.. code-block:: text
+Use the pre-built binary packages to install CouchDB::
 
-    cd /usr/ports/databases/couchdb
+    pkg install couchdb3
+
+Alternatively, it is possible installing CouchDB from the Ports
+Collection::
+
+    cd /usr/ports/databases/couchdb3
     make install clean
-
-This will install CouchDB from the ports collection.
 
 .. note::
    Be sure to :ref:`create an admin user<config/admins>` before starting
    CouchDB for the first time!
 
-Start script
-------------
+Service Configuration
+=====================
 
-The following options for ``/etc/rc.conf`` or ``/etc/rc.conf.local`` are
-supported by the start script (defaults shown)::
+The port is shipped with a script that integrates CouchDB with
+FreeBSD's `rc.d service framework`_.  The following options for
+``/etc/rc.conf`` or ``/etc/rc.conf.local`` are supported (defaults
+shown)::
 
-    couchdb_enable="NO"
-    couchdb_enablelogs="YES"
-    couchdb_user="couchdb"
+    couchdb3_enable="NO"
+    couchdb3_user="couchdb"
+    couchdb3_erl_flags="-couch_ini /usr/local/libexec/couchdb3/etc/default.ini /usr/local/etc/couchdb3/local.ini"
+    couchdb3_chdir="/var/db/couchdb3"
 
-After enabling the couchdb rc service use the following command to start CouchDB::
+After enabling the ``couchdb3`` service (by setting
+``couchdb3_enable`` to ``"YES"``), use the following command to start
+CouchDB::
 
-    /usr/local/etc/rc.d/couchdb start
+    service couchdb3 start
 
-This script responds to the arguments `start`, `stop`, `status`, `rcvar` etc..
+This script responds to the arguments ``start``, ``stop``, ``status``,
+``rcvar`` etc.  If the service is not yet enabled in ``rc.conf``, use
+``onestart`` to start it up ad-hoc.
 
-The start script will also use settings from the following config files:
+The service will also use settings from the following config files:
 
-- /usr/local/etc/couchdb/default.ini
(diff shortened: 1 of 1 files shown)


big-r81 (approved):

src/docs/src/install/freebsd.rst
```diff
@@ -16,71 +16,93 @@
...
+To change any of these settings, please refrain from editing
+``/etc/passwd`` and instead use ``pw usermod ...`` or ``vipw``. Make
+sure that the user has no shell, but instead uses
+``/usr/sbin/nologin``. The ``*`` in the second field means that this
+user can not login via password authorization. For the details, see
 `man 5 passwd`_.
 
-.. _man 5 passwd: http://linux.die.net/man/5/passwd
+.. _the pre-built binary packages: https://docs.freebsd.org/en/books/handbook/ports/
+.. _rc.d service framework: https://man.freebsd.org/cgi/man.cgi?query=service&apropos=0&sektion=0&manpath=default&arch=default&format=html
```
dch, line 106: we can use https://man.freebsd.org/rc.d instead, much shorter

src/docs/src/install/freebsd.rst
```diff
@@ -16,71 +16,93 @@
...
+.. _in the ports tree: https://github.com/freebsd/freebsd-ports/blob/main/databases/couchdb3/files/pkg-message.in
+.. _man 5 passwd: https://man.freebsd.org/cgi/man.cgi?query=passwd&apropos=0&sektion=0&manpath=default&arch=default&format=html
```
dch, line 108: IMHO we can remove all of the passwd/group related stuff, no? Or are there common circumstances where this will help somebody who uses FreeBSD but doesn't know about groups & passwords yet?

src/docs/src/install/freebsd.rst
```diff
@@ -16,71 +16,93 @@
...
+    # pw groupadd couchdb
+    # pw groupmod couchdb -m couchdb
 
-.. code-block:: text
+Then the user is added to ``/etc/passwd`` and the group to
+``/etc/groups``.  These could be verified in the following way::
 
-    shell#  grep couchdb /etc/passwd
-    couchdb:*:1013:1013:Couchdb, time to relax:/var/lib/couchdb/:/usr/sbin/nologin
+    $ grep couchdb /etc/passwd
```
dch, line 93: oh dear how long has `/var/lib/couchdb` been here and I didn't notice?

IMHO we could remove this entire section unless you feel it's useful.

src/docs/src/install/freebsd.rst
```diff
@@ -16,71 +16,93 @@
...
+This script responds to the arguments `start`, `stop`, `status`,
+`rcvar` etc.  If the service is not yet enabled in ``rc.conf``, use
+``onestart`` to start it up ad-hoc.
 
-The start script will also use settings from the following config files:
+The service will also use settings from the following config files:
 
-- /usr/local/etc/couchdb/default.ini
-- /usr/local/etc/couchdb/local.ini
+- ``/usr/local/libexec/couchdb3/etc/default.ini``
```
dch, line 61: I think these (and other) paths will vary depending on whether this is installed from sources, or via packages/ports.

Either way making this more explicit is a good thing --

`default.ini` should be left read-only, and will be replaced on upgrades and re-installs without warning. Local modifications can go in `local.ini` as usual.

Does that read a bit better?

dch (approved):
Some thoughts in line, nothing blocking. Thanks for the long-overdue updates!

src/docs/src/install/freebsd.rst
pgj: Cool, thanks!

src/docs/src/install/freebsd.rst
pgj: The paragraph about managing the `couchdb` user and group was before, I just tried to update it.  But I do not insist on keeping around.  Personally, I also do not feel it important enough to have it.

I was trying to use `cgit.freebsd.org` for the link but at the time of access, I was not working properly.  But perhaps this should not influence the decision about using it as a canonical reference.

src/docs/src/install/freebsd.rst
pgj: I do not have any personal preference.  I am fine with losing it.

src/docs/src/install/freebsd.rst
pgj: I think the `rc.d` script is shipped only with the version from ports/packages.  Therefore it makes clear that we are not talking about installing the vanilla CouchDB from sources.  That said, I would not worry about the differences in the location of the configuration files for now.

pgj: I used the sentence about `default.ini` in the text though because I agree that it gives a better understanding of its nature.

PR #4956, opened 2024-01-09
Control nouveau.yaml in configure and dev/run

rnewson:
Move the generation/templating of nouveau.yaml to couchdb side to make packaging of nouveau simpler and more consistent with the rest.


nickva (approved):
+1, very nice!

Was able to create a release from a dist tarball and start it with `./nouveau/bin/nouveau server ./etc/nouveau.yaml`

PR #4958, opened 2024-01-10
feat(`mango`): rolling execution statistics

### LINKED ISSUE #4560: Inconsistent execution statistics for Mango queries

pgj:
In case of map-reduce views, the arrival of the `complete` message is not guaranteed for the view callback (at the shard) when a `stop` is issued during the aggregation (at the coordinator).  Due to that, internally collected shard-level statistics may not be fed back to the coordinator which can cause data loss hence inaccuracy in the overall execution statistics.

Address this issue by switching to a "rolling" model where row-level statistics are immediately streamed back to the coordinator.  Support mixed-version cluster upgrades by activating this model only if requested through the map-reduce arguments and the given shard supports that.

Note that this change shares some similarities with @chewbranca's work on Couch Stats Resource Tracker though it stems from a different origin.  In the long run, the goal is to either have it merged or even replaced with that approach to facilitate convergence.  It has a smaller scope hence it is less complicated and it is ready for inclusion.

## Testing recommendations

Running the respective Mango unit and integration test suites shall suffice (which is done by the CI):

```shell
make eunit apps=mango
make mango-test MANGO_TEST_OPTS="15-execution-stats-test"
```

But there is a detailed description in related the ticket (see below) on how to trigger the problem, which might also be used to exercise the code change.

Originates from #4735, forerunner of #4812
Fixes #4560

## Checklist

### DIFF
--- a/src/mango/src/mango_cursor_view.erl
+++ b/src/mango/src/mango_cursor_view.erl
@@ -55,6 +55,15 @@
         covering_index => 'maybe'(#idx{})
     }.
 
+-type mrargs_extra_item() ::
+    {callback, {atom(), atom()}}
+    | {selector, any()}
+    | {callback_args, viewcbargs()}
+    | {ignore_partition_query_limit, boolean()}
+    | {execution_stats_map, boolean()}
+    | {execution_stats_rolling, boolean()}.
+-type mrargs_extra() :: [mrargs_extra_item()].
+
 -spec viewcbargs_new(Selector, Fields, CoveringIndex) -> ViewCBArgs when
     Selector :: selector(),
     Fields :: fields(),
@@ -207,7 +216,9 @@ base_args(#cursor{index = Idx, selector = Selector, fields = Fields} = Cursor) -
             % - Return execution statistics in a map
             {execution_stats_map, true},
             % - Return view rows in a map
-            {view_row_map, true}
+            {view_row_map, true},
+            % - Stream execution statistics
+            {execution_stats_rolling, true}
         ]
     }.
 
@@ -341,6 +352,43 @@ choose_best_index(IndexRanges) ->
     {SelectedIndex, SelectedIndexRanges, _} = hd(SortedIndexRanges),
     {{SelectedIndex, SelectedIndexRanges}, SortedIndexRanges}.
 
+-spec format_stats(RawStats, Options) -> FormattedStats when
+    RawStats :: shard_stats_v2(),
+    Options :: mrargs_extra(),
+    FormattedStats :: shard_stats_v1() | shard_stats_v2().
+format_stats(Stats, Options) when is_list(Options) ->
+    case couch_util:get_value(execution_stats_map, Options, false) of
+        true ->
+            Stats;
+        false ->
+            #{docs_examined := DocsExamined} = Stats,
+            {docs_examined, DocsExamined}
+    end.
+
+-spec submit_stats(Options) -> ok when
+    Options :: mrargs_extra().
+submit_stats(Options) when is_list(Options) ->
+    ShardStats = mango_execution_stats:shard_get_stats(),
+    Stats = format_stats(ShardStats, Options),
+    % Send execution stats in batch (shard-level)
+    ok = rexi:stream2({execution_stats, Stats}).
+
+-spec roll_stats(ViewRow, Options) -> ViewRow when
+    ViewRow :: view_row(),
+    Options :: mrargs_extra().
+roll_stats(ViewRow, Options) when is_list(Options) ->
+    ViewRowMap = couch_util:get_value(view_row_map, Options, false),
+    RollingStats = couch_util:get_value(execution_stats_rolling, Options, false),
+    case ViewRowMap andalso RollingStats of
+        true ->
+            ShardStats = mango_execution_stats:shard_get_stats(),
+            mango_execution_stats:shard_init(),
+            Stats = format_stats(ShardStats, Options),
+            fabric_view_row:set_stats(ViewRow, Stats);
+        false ->
+            ViewRow
+    end.
+
 -spec view_cb
     (Message, #mrargs{}) -> Response when
         Message :: {meta, any()} | {row, row_properties()} | complete,
@@ -382,7 +430,8 @@ view_cb({row, Props}, #mrargs{extra = Options} = Acc) ->
             case match_and_extract_doc(Doc, Selector, Fields) of
                 {match, FinalDoc} ->
                     ViewRow1 = fabric_view_row:set_doc(ViewRow, FinalDoc),
-                    ok = rexi:stream2(ViewRow1),
+                    ViewRow2 = roll_stats(ViewRow1, Options),
+                    ok = rexi:stream2(ViewRow2),
                     set_mango_msg_timestamp();
                 {no_match, undefined} ->
                     maybe_send_mango_ping()
@@ -397,7 +446,8 @@ view_cb({row, Props}, #mrargs{extra = Options} = Acc) ->
             Process(Doc);
         {undefined, _} ->
             % include_docs=false. Use quorum fetch at coordinator
-            ok = rexi:stream2(ViewRow),
+            ViewRow1 = roll_stats(ViewRow, Options),
+            ok = rexi:stream2(ViewRow1),
             set_mango_msg_timestamp();
         {Doc, _} ->
             mango_execution_stats:shard_incr_docs_examined(),
@@ -406,17 +456,7 @@ view_cb({row, Props}, #mrargs{extra = Options} = Acc) ->
     end,
     {ok, Acc};
 view_cb(complete, #mrargs{extra = Options} = Acc) ->
-    ShardStats = mango_execution_stats:shard_get_stats(),
-    Stats =
-        case couch_util:get_value(execution_stats_map, Options, false) of
-            true ->
-                ShardStats;
-            false ->
-                DocsExamined = maps:get(docs_examined, ShardStats),
-                {docs_examined, DocsExamined}
-        end,
-    % Send shard-level execution stats
-    ok = rexi:stream2({execution_stats, Stats}),
+    submit_stats(Options),
     % Finish view output
     ok = rexi:stream_last(complete),
     {ok, Acc};
@@ -472,6 +512,21 @@ maybe_send_mango_ping() ->
 set_mango_msg_timestamp() ->
     put(mango_last_msg_timestamp, os:timestamp()).
 
+-spec add_shard_stats(#execution_stats{}, shard_stats()) -> #execution_stats{}.
+add_shard_stats(Stats0, {docs_examined, DocsExamined}) ->
+    mango_execution_stats:incr_docs_examined(Stats0, DocsExamined);
+add_shard_stats(Stats0, #{} = ShardStats) ->
+    DocsExamined = shard_stats_get(docs_examined, ShardStats),
+    KeysExamined = shard_stats_get(keys_examined, ShardStats),
+    Stats = mango_execution_stats:incr_docs_examined(Stats0, DocsExamined),
+    mango_execution_stats:incr_keys_examined(Stats, KeysExamined).
+
+-spec handle_execution_stats(#cursor{}, shard_stats()) -> {ok, #cursor{}}.
+handle_execution_stats(Cursor0, ShardStats) ->
+    #cursor{execution_stats = Stats} = Cursor0,
+    Cursor = Cursor0#cursor{execution_stats = add_shard_stats(Stats, ShardStats)},
+    {ok, Cursor}.
+
 -spec handle_message(message(), #cursor{}) -> Response when
     Response ::
         {ok, #cursor{}}
@@ -495,20 +550,10 @@ handle_message({row, Props}, Cursor) ->
             couch_log:error("~s :: Error loading doc: ~p", [?MODULE, Error]),
             {ok, Cursor}
     end;
-handle_message({execution_stats, {docs_examined, DocsExamined}}, Cursor0) ->
-    #cursor{execution_stats = Stats} = Cursor0,
-    Cursor = Cursor0#cursor{
-        execution_stats = mango_execution_stats:incr_docs_examined(Stats, DocsExamined)
-    },
-    {ok, Cursor};
+handle_message({execution_stats, {docs_examined, _} = ShardStats}, Cursor0) ->
+    handle_execution_stats(Cursor0, ShardStats);
 handle_message({execution_stats, #{} = ShardStats}, Cursor0) ->
-    DocsExamined = shard_stats_get(docs_examined, ShardStats),
-    KeysExamined = shard_stats_get(keys_examined, ShardStats),
-    #cursor{execution_stats = Stats0} = Cursor0,
-    Stats1 = mango_execution_stats:incr_docs_examined(Stats0, DocsExamined),
-    Stats = mango_execution_stats:incr_keys_examined(Stats1, KeysExamined),
-    Cursor = Cursor0#cursor{execution_stats = Stats},
-    {ok, Cursor};
+    handle_execution_stats(Cursor0, ShardStats);
 handle_message(complete, Cursor) ->
     {ok, Cursor};
 handle_message({error, Reason}, _Cursor) ->
@@ -648,9 +693,14 @@ consider_index_coverage(Index, Fields, #mrargs{include_docs = IncludeDocs0} = Ar
         | {no_match, null, {execution_stats, shard_stats()}}
         | any().
 doc_member_and_extract(Cursor, RowProps) ->
-    Db = Cursor#cursor.db,
-    Opts = Cursor#cursor.opts,
-    ExecutionStats = Cursor#cursor.execution_stats,
+    #cursor{db = Db, opts = Opts, execution_stats = ExecutionStats0} = Cursor,
+    ExecutionStats =
+        case couch_util:get_value(stats, RowProps) of
+            undefined ->
+                ExecutionStats0;
+            ShardStats ->
+                add_shard_stats(ExecutionStats0, ShardStats)
+        end,
     Selector = Cursor#cursor.selector,
     case couch_util:get_value(doc, RowProps) of
         {DocProps} ->
@@ -748,7 +798,8 @@ base_opts_test() ->
             }},
             {ignore_partition_query_limit, true},
             {execution_stats_map, true},
-            {view_row_map, true}
+            {view_row_map, true},
+            {execution_stats_rolling, true}
         ],
     MRArgs =
         #mrargs{
@@ -1093,7 +1144,8 @@ t_execute_ok_all_docs(_) ->
             }},
             {ignore_partition_query_limit, true},
             {execution_stats_map, true},
-            {view_row_map, true}
+            {view_row_map, true},
+            {execution_stats_rolling, true}
         ],
     Args =
         #mrargs{
@@ -1180,7 +1232,8 @@ t_execute_ok_query_view(_) ->
             }},
             {ignore_partition_query_limit, true},
             {execution_stats_map, true},
-            {view_row_map, true}
+            {view_row_map, true},
+            {execution_stats_rolling, true}
         ],
     Args =
         #mrargs{
@@ -1279,7 +1332,8 @@ t_execute_ok_all_docs_with_execution_stats(_) ->
             }},
             {ignore_partition_query_limit, true},
             {execution_stats_map, true},
-            {view_row_map, true}
+            {view_row_map, true},
+            {execution_stats_rolling, true}
         ],
     Args =
         #mrargs{
@@ -1394,6 +1448,8 @@ t_view_cb_row_matching_regular_doc(_) ->
                     fields => all_fields,
                     covering_index => undefined
                 }},
+                {execution_stats_map, true},
+                {execution_stats_rolling, true},
                 {view_row_map, true}
             ]
         },
@@ -1413,6 +1469,8 @@ t_view_cb_row_non_matching_regular_doc(_) ->
                     fields => all_fields,
                     covering_index => undefined
                 }},
+                {execution_stats_map, true},
+                {execution_stats_rolling, true},
                 {view_row_map, true}
             ]
         },
@@ -1432,6 +1490,8 @@ t_view_cb_row_null_doc(_) ->
                     fields => all_fields,
                     covering_index => undefined
                 }},
+                {execution_stats_map, true},
+                {execution_stats_rolling, true},
                 {view_row_map, true}
             ]
         },
@@ -1452,6 +1512,8 @@ t_view_cb_row_missing_doc_triggers_quorum_fetch(_) ->
                     fields => all_fields,
                     covering_index => undefined
                 }},
+                {execution_stats_map, true},
+                {execution_stats_rolling, true},
                 {view_row_map, true}
             ]
         },
@@ -1479,6 +1541,8 @@ t_view_cb_row_matching_covered_doc(_) ->
                     fields => Fields,
                     covering_index => Index
                 }},
+                {execution_stats_map, true},
+                {execution_stats_rolling, true},
                 {view_row_map, true}
             ]
         },
@@ -1503,6 +1567,8 @@ t_view_cb_row_non_matching_covered_doc(_) ->
                     fields => Fields,
                     covering_index => Index
                 }},
+                {execution_stats_map, true},
+                {execution_stats_rolling, true},
                 {view_row_map, true}
             ]
         },
@@ -1638,10 +1704,13 @@ t_handle_message_row_ok_above_limit(_) ->
             user_acc = accumulator,
             user_fun = fun foo:bar/2
         },
-    Row = [{id, id}, {key, key}, {doc, Doc}],
+    ShardStats = #{keys_examined => 2, docs_examined => 3},
+    Row = [{id, id}, {key, key}, {doc, Doc}, {stats, ShardStats}],
     Cursor1 =
         Cursor#cursor{
-            execution_stats = #execution_stats{resultsReturned = 1},
+            execution_stats = #execution_stats{
+                resultsReturned = 1, totalKeysExamined = 2, totalDocsExamined = 3
+            },
             limit = 8,
             user_acc = updated_accumulator,
             bookmark_docid = id,
@@ -1689,12 +1758,15 @@ t_handle_message_row_ok_triggers_quorum_fetch_match(_) ->
             user_acc = accumulator,
             limit = 1
         },
-    Row = [{id, id}, {doc, undefined}],
+    ShardStats = #{keys_examined => 2, docs_examined => 3},
+    Row = [{id, id}, {doc, undefined}, {stats, ShardStats}],
     Cursor1 =
         Cursor#cursor{
             execution_stats =
                 #execution_stats{
                     totalQuorumDocsExamined = 1,
+                    totalKeysExamined = 2,
+                    totalDocsExamined = 3,
                     resultsReturned = 1
                 },
             user_acc = updated_accumulator,
--- a/src/fabric/src/fabric_view_row.erl
+++ b/src/fabric/src/fabric_view_row.erl
@@ -18,9 +18,11 @@
     get_value/1,
     get_doc/1,
     get_worker/1,
+    get_stats/1,
     set_key/2,
     set_doc/2,
     set_worker/2,
+    set_stats/2,
     transform/1
 ]).
 
@@ -91,6 +93,14 @@ set_worker(#view_row{} = Row, Worker) ->
 set_worker({view_row, #{} = Row}, Worker) ->
     {view_row, Row#{worker => Worker}}.
 
+get_stats({view_row, #{stats := Stats}}) ->
+    Stats;
+get_stats({view_row, #{}}) ->
+    undefined.
+
+set_stats({view_row, #{} = Row}, Stats) ->
+    {view_row, Row#{stats => Stats}}.
+
 transform(#view_row{value = {[{reduce_overflow_error, Msg}]}}) ->
     {row, [{key, null}, {id, error}, {value, reduce_overflow_error}, {reason, Msg}]};
 transform(#view_row{key = Key, id = reduced, value = Value}) ->
@@ -109,8 +119,13 @@ transform({view_row, #{} = Row0}) ->
     Value = maps:get(value, Row0, undefined),
     Doc = maps:get(doc, Row0, undefined),
     Worker = maps:get(worker, Row0, undefined),
+    Stats = maps:get(stats, Row0, undefined),
     Row = #view_row{id = Id, key = Key, value = Value, doc = Doc, worker = Worker},
-    transform(Row).
+    {row, Props} = RowProps = transform(Row),
+    case Stats of
+        undefined -> RowProps;
+        #{} -> {row, [{stats, Stats} | Props]}
+    end.
 
 -ifdef(TEST).
 -include_lib("couch/include/couch_eunit.hrl").
--- a/src/mango/test/15-execution-stats-test.py
+++ b/src/mango/test/15-execution-stats-test.py
@@ -73,6 +73,75 @@ class ExecutionStatsTests(mango.UserDocsTests):
         self.assertEqual(resp["execution_stats"]["total_quorum_docs_examined"], 0)
         self.assertEqual(resp["execution_stats"]["results_returned"], 3)
 
+    def test_reporting_consistency(self):
+        cases = [
+            {
+                "title": "with limit",
+                "selector": {"age": {"$lte": 42}},
+                "fields": ["name", "email", "age"],
+                "limit": 3,
+                "total_keys_examined": 4,
+                "total_docs_examined": 4,
+                "results_returned": 3,
+            },
+            {
+                "title": "partial matches",
+                "selector": {"favorites": {"$elemMatch": {"$eq": "Erlang"}}},
+                "fields": ["name", "email", "twitter"],
+                "limit": 200,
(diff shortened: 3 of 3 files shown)


src/mango/src/mango_cursor_view.erl
```diff
@@ -336,19 +348,66 @@ choose_best_index(IndexRanges) ->
...
+maybe_init_stats(Options) ->
+    case couch_util:get_value(execution_stats_rolling, Options, false) of
+        true -> ok;
+        false -> mango_execution_stats:shard_init()
+    end.
+
+-spec roll_stats(Stats, Options) -> ok when
+    Stats :: shard_stats_v2(),
+    Options :: mrargs_extra().
+roll_stats(Stats, Options) ->
```
nickva, line 362: Suggestion: add asserts about the types of args we're dealing with `roll_stats(#{} = Stats, Options) when is_list(Options) -> ...`

src/mango/src/mango_cursor_view.erl
```diff
@@ -336,19 +348,66 @@ choose_best_index(IndexRanges) ->
...
+-spec format_stats(RawStats, Options) -> FormattedStats when
+    RawStats :: shard_stats_v2(),
+    Options :: mrargs_extra(),
+    FormattedStats :: shard_stats_v1() | shard_stats_v2().
+format_stats(Stats, Options) ->
+    case couch_util:get_value(execution_stats_map, Options, false) of
+        true ->
+            Stats;
+        false ->
+            DocsExamined = maps:get(docs_examined, Stats),
```
nickva, line 382: Using in-line map de-structuring is a bit more idiomatic. `#{docs_examined := DocsExamined} = ...`

src/mango/src/mango_execution_stats.erl
```diff
@@ -123,19 +123,19 @@ shard_init() ->
...
-shard_incr_docs_examined() ->
-    incr(docs_examined).
+-spec shard_incr_docs_examined(integer()) -> any().
+shard_incr_docs_examined(N) ->
+    incr(docs_examined, N).
 
--spec incr(atom()) -> any().
-incr(Key) ->
+-spec incr(atom(), integer()) -> any().
+incr(Key, N) ->
```
nickva, line 135: Add `is_integer(N)` guard and possible N>1 or N>=0 depending on our constraints here.

nickva (approved):
+1 lgtm but see a few style suggestions here and there (as inline comments)

src/mango/src/mango_execution_stats.erl
pgj: Good idea!  I used this opportunity as well to add a branch for handling `incr(Key, 0)` specifically to do nothing since that should cause any change.

rnewson:
Sorry, I did mean to send feedback when I saw this ages ago.

I'm a little concerned at the increase in messages here (every `n*q` shard for every `_find` request will send a message for each row it processes?) to workaround a problem that, perhaps, we should instead fix directly?

Specifically, would it not be possible to alter how we implement the end of the workers, so they do get an opportunity to send their final stats while exiting (as their exit reason, say, which I believe we can capture).

nickva:
Would some batching work to accumulate stats for some time (100 msec?), and only send stats intermittently? It might miss some stats at the end but for a long running request it should capture the stats as it goes along without emitting too many messages?

pgj:
@nickva currently, the M/R `complete` message is used as a cue to submit the statistics.  If this is not handled properly, it does not matter if there is batching or not, we will miss some data at the end.  For the details, please see #4735.

nickva:
@pgj the idea is that it would be the difference between not sending any data back if `complete` doesn't run, so for an hour long running request we'd have no stats for the whole hour, or, send most of the data back and only miss the last few msec or so. That is if finalization would be tricky to do like @rnewson suggested. and sending messages for every row is expensive.

pgj:
I believe this does not work for short-lived requests.  Actually, the real cause of this problem is a race condition: workers often die earlier they could touch the `complete` message due to their untimely cancellation.  In such cases, zero counts are observed, therefore the `_find` results are invisible.

But why is the increase in the message count problematic?  @chewbranca's upcoming is much more about the same approach.  Or did I miss something, perhaps I misunderstood that?  Is not there an optimization somewhere down the stack that enables [coalesencing](https://en.wikipedia.org/wiki/Coalescing_(computer_science)) of messages?  The messages would not increase the required bandwith much due to their small size...?

nickva:
I was mainly suggesting a way to have a bound on the uncertainty in the stats, if it the increase in messages is a problem. It could be a time-bound (milliseconds) or doc examined count (if N > 100 emit ...) bound. It's a half-way hack, it's better than getting no stats at all, but maybe less expensive than emitting stats for every row. But, I don't know if it's a performance issue, I think you and @chewbranca did some measurements? 

> In such cases, zero counts are observed, therefore the _find results are invisible.

From what I remember in mango the issue was happening when we have a `limit` specified? With a `limit` we spawn all the workers with the same limit value but we will stop sooner based on when we hit the limit on the coordinator side. So then, some workers won't get a chance to emit a `completed` message and will keep emitting rows, others will emit a `completed` message but the coordinator by that time is already stopping and won't care. When the coordinator stops it will immediately issue a `kill_all` rexi call to terminate all the workers. Is that roughly what was happing?

> Is not there an optimization somewhere down the stack that enables [coalesencing](https://en.wikipedia.org/wiki/Coalescing_(computer_science)) of messages?

I don't know if rexi has message batching. There is buffering if the send channel is blocked, and there is an un-acked message limit. But  you're also right that the lower level dist and tcp channel will probably do the batching depending on how the environment is set up.

rnewson:
If @chewbranca's work would also significantly increase message count in a way that adversely affected performance (or latency of other requests trying to make progress inbetween such messages) I would also want to see that improved before it was merged (and, if it were unavoidably severe, _not_ merged). I'll focus on reviewing the work in hand here and not speculate further about other changes until they are ready for review.

I like nick's middle ground of having the workers send info periodically, we would benefit from seeing those stats rising during the processing of a long request (though I think `_find` unlike `_view` has a short-ish finite timeout by default)

@pgj's point is also relevant, very short requests won't appear at all, and to my mind that lines up with my original point that we should improve how the end of the request is handled. Ideally we would wait for all workers to return their final message (perhaps as their exit reason as already noted), but without forcing the user to wait for more responses than are needed to form a complete response (i.e, the first response from each shard range).

Put simply, it was not anticipated that we would care about the other workers when this was built, the code fires off the jobs and wants to return as soon as it can, killing the remaining processes. A good fix here should tackle that directly.

pgj:
> I was mainly suggesting a way to have a bound on the uncertainty in the stats, if it the increase in messages is a problem.

@nickva I am afraid we cannot allow any uncertainty in the statistics if downstream consumers wanted to build uses cases on top of this mechanism that involves billing.

> I don't know if it's a performance issue, I think you and @chewbranca did some measurements?

@chewbranca might have some performance results on his side, but I have not collected any myself.  Probably I should do some just for curiosity and for including here.

> From what I remember in mango the issue was happening when we have a `limit` specified?

Yes, but if I recall correctly there is always a limit assigned for the `_find` requests.  I believe the emphasis is on the relative distance of the limit and the number of results returned.  It is technically a result of a race condition: if the `stop` signal arrives at the wrong time, either the `completed` message is not emitted or it is not handled in the time.

> Ideally we would wait for all workers to return their final message (perhaps as their exit reason as already noted), but without forcing the user to wait for more responses than are needed to form a complete response (i.e, the first response from each shard range).

This can be done easily: replace the `stop` for an `ok` in `mango_cursor_view:handle_doc/2`.  Actually, this is also case that could be tested for performance.  Maybe it is not that expensive and sending the stats for every row.

> Put simply, it was not anticipated that we would care about the other workers when this was built, the code fires off the jobs and wants to return as soon as it can, killing the remaining processes. A good fix here should tackle that directly.

@rnewson do you have any pointers or ideas where to start with this?  In the previous discussions, you had [some suggestions](https://github.com/apache/couchdb/pull/4735#issuecomment-1715743087), but I could not make them work.

nickva:
> @nickva I am afraid we cannot allow any uncertainty in the statistics if downstream consumers wanted to build uses cases on top of this mechanism that involves billing.

Hmm, if billing is involved then even the current implementation of the PR, will not produce repeatable results unless we're talking about Q=1 databases only. I was under the impression that this is mostly about providing slightly more accurate results (less 0 ones). Maybe this uncertainty is ok, depending on the downstream consumers' requirements, but if billing is involved, it's good to make that clear.

To convince ourselves I quickly patched an old script I had to run a few _find results https://gist.github.com/nickva/31e0f1f1c2a5a651259dc897a1bb5cfa

With 5000 docs we get a wild variety of results:

```
% make && ./dev/run --admin=adm:pass -n 1

(venv3) % ./findlimit.py
 * deleting http://127.0.0.1:15984/db
 * creating http://127.0.0.1:15984/db {'q': '4'}
 * creating 5000 docs with val range 200
 * created docs in 2.0 seconds
 * calling _find with selector: {'x': 50}
 * _find in 0.1 seconds 10 docs {'total_keys_examined': 2311, 'total_docs_examined': 2311, 'total_quorum_docs_examined': 0, 'results_returned': 10, 'execution_time_ms': 52.502}

(venv3) % ./findlimit.py
 * deleting http://127.0.0.1:15984/db
 * creating http://127.0.0.1:15984/db {'q': '4'}
 * creating 5000 docs with val range 200
 * created docs in 2.0 seconds
 * calling _find with selector: {'x': 50}
 * _find in 0.1 seconds 10 docs {'total_keys_examined': 3115, 'total_docs_examined': 3115, 'total_quorum_docs_examined': 0, 'results_returned': 10, 'execution_time_ms': 106.121}

(venv3) % ./findlimit.py
 * deleting http://127.0.0.1:15984/db
 * creating http://127.0.0.1:15984/db {'q': '4'}
 * creating 5000 docs with val range 200
 * created docs in 2.0 seconds
 * calling _find with selector: {'x': 50}
 * _find in 0.1 seconds 10 docs {'total_keys_examined': 3619, 'total_docs_examined': 3619, 'total_quorum_docs_examined': 0, 'results_returned': 10, 'execution_time_ms': 77.16}

(venv3) % ./findlimit.py
 * deleting http://127.0.0.1:15984/db
 * creating http://127.0.0.1:15984/db {'q': '4'}
 * creating 5000 docs with val range 200
 * created docs in 1.9 seconds
 * calling _find with selector: {'x': 50}
 * _find in 0.1 seconds 10 docs {'total_keys_examined': 2651, 'total_docs_examined': 2651, 'total_quorum_docs_examined': 0, 'results_returned': 10, 'execution_time_ms': 58.746}
 
(venv3) % ./findlimit.py
 * deleting http://127.0.0.1:15984/db
 * creating http://127.0.0.1:15984/db {'q': '4'}
 * creating 5000 docs with val range 200
 * created docs in 2.1 seconds
 * calling _find with selector: {'x': 50}
 * _find in 0.1 seconds 10 docs {'total_keys_examined': 4121, 'total_docs_examined': 4121, 'total_quorum_docs_examined': 0, 'results_returned': 10, 'execution_time_ms': 100.792}
```

To be clear, I don't know if it is a performance impact to send all the  stats rows, I was merely mentioning a possible strategy to apply if it is a concern. All this given the assumption there is already some variability in results anyway. 

I am slightly concerned with the synchronous stream finalization strategy: we'd dealing with a streaming protocol during the active phase, only to switch back to a synchronous termination, which involves extra waiting applied to every request (finalize all Q workers, wait for all Q workers to respond, possibly also handle broken, timed out cases, etc). That would add some fixed delay to all the requests (extra 10-20 msec?). Maybe we could do the finalization after we respond the user, but we'd have to see how that interacts with request resource cleanup part.

pgj:
I have done some preliminary [benchmarking with `k6.io`](https://github.com/pgj/couchdb-k6/tree/main/mango/couchdb_4958) on my Apple M1 Max, with 3 nodes (r=2, w=2, q=2), 100K documents, `/{db}/_find` queries of various complexity, 2 minutes warmup and 5 minutes run (where `count` tells many queries completed, the higher is the better).

Here are the results:

- `couchdb/main` (baseline, as of `1f77e2e3`):
  - 1 thread: `min=0.60ms max=8676.29ms  p(95)=6259.13ms p(98)=6631.37ms p(99)=7139.79ms, count=240`
  - 2 threads: `min=0.82ms max=17990.32ms p(95)=13364.63ms p(98)=13960.05ms p(99)=14177.66ms, count=206`

- `pgj/feat/mango/rolling-execution-stats` (rebased to `couchdb/main`):
  - 1 thread: `min=0.82ms max=11053.97ms p(95)=6880.65ms p(98)=7219.82ms  p(99)=7493.13ms,  count=196`
  - 2 threads: `min=0.73ms max=20467.99ms p(95)=15500.12ms p(98)=17401.71ms p(99)=19277.02ms, count=189`

- `couchdb/main` where `stop` is replaced for `ok` in `mango_cursor_view:handle_doc/2`:
  - 1 thread: `min=1.67ms max=15335.05ms p(95)=7943.36ms p(98)=9496.31ms  p(99)=12347.95ms, count=88`
  - 2 threads: `min=0.82ms max=30654.26ms p(95)=16628.83ms p(98)=17278.53ms p(99)=20883.10ms, count=75`

There seem to be a 10-20% performance penalty for the streaming statistics, while waiting for the quorum causes 66%.

pgj:
> Hmm, if billing is involved then even the current implementation of the PR, will not produce repeatable results unless we're talking about Q=1 databases only.

I do not think it should produce repeatable results.  If a query happened to cause the scanning of 2,000 documents once and then 1,500 for another instance is acceptable — that is how the system works.  But returning zero in either case is definitely a miss on the side of accounting.

> To convince ourselves I quickly patched an old script I had to run a few _find results

Great, thanks @nickva!

> To be clear, I don't know if it is a performance impact to send all the stats rows

Unfortunately there is (see my comment above) :-(

> I am slightly concerned with the synchronous stream finalization strategy: we'd dealing with a streaming protocol during the active phase, only to switch back to a synchronous termination, which involves extra waiting applied to every request (finalize all Q workers, wait for all Q workers to respond, possibly also handle broken, timed out cases, etc).

If that is what replacing `stop` for `ok` in `mango_cursor_view:handle_doc/2` means, then the performance is much worse than that of streaming statistics.

rnewson:
noting that "billing" of requests is not a couchdb concern.

pgj:
> noting that "billing" of requests is not a couchdb concern.

Yes, I agree.  But it is CouchDB that could ensure that the data that is going to `chttpd_stats_reporter` could be trusted and provide support for scenarios like billing.

rnewson:
A higher level question occurs. It seems to me you want to account for the cost of the workers that _did not contribute_ to the result (they were killed by the coordinator because a response from another copy was received first), whereas couchdb has an interest in not doing the unnecessary work. Did I misunderstand what the "missing" data is here?

rnewson:
actually, I'm going to assume none of us mean the execution stats of non-contributing workers.

What I think is being proposed here sounds like an attempt to fix the current, and by implication quite weird, execution stats code.

Only a `_find` request would read a view row or fetch a document and _not_ pass this to the client, so the execution stats only apply here, and only because it is useful to compare the number of rows retrieved versus the number of rows (or docs) retrieved to find those rows (which could be a much larger number).

by inference the execution stats have been calculated and transmitted out-of-band relative to the returned rows, and this, if I'm right, is a mistake we should fix.

i.e, I suggest that the workers include the number of rows/docs they examined and discarded with each matching row they return to the coordinator. this is a small amount of extra data to an existing message. For very efficient query executions, it's the number 0 on every row. The coordinator can obviously count how many rows it passed to the client and can sum these new numbers to calculate how many rows/docs were examined. As its the same message, we can't get the wrong answer.

Now I'll wait until the morning to learn how completely wrong I am about the problem.

pgj:
Thanks for sharing your thoughts @rnewson.

Yes, the execution stats are currently passed around in a dedicated `execution_stats` message but only at the end of the transmission.  The other, separate generic `complete` message is the cue for submitting `execution_stats` which may go missing if `complete` is dropped on early termination.

The problem is that this can happen even on a single-node cluster.  Although I can imagine the scenario you described, I am still inclined to believe that the use of `stop` is such a hard stop that it even shuts down the actual worker who actively contributed to the response.  Using `stop` in the view callback is on the coordinator — that is, it already has all the data it needs in order to complete serving the request.  But the worker does not have the chance to say "okay, please find the execution stats in the next, final message" because it is immediately stopped right after `stop` is received.  A remedy to this could be to let the worker hang around for a while to finalize the transmission that way, but it would also sacrifice some of the performance gain we are currently having with `stop`.  This technically leaves us with streaming.

Your suggestion looks similar to the one which is contained in this PR.  However, I notice that you recommended to extend the `row` message (for `mango_cursor_view:handle_message/2`) directly and not to emit fine-grained, per-row `execution_stats` messages.  I can see a case that it works better performance-wise, and I am happy to do a benchmark with that approach as well.

rnewson:
Essentially that,yes. But a bit further. The only execution stat we need at the row level is how many rows were examined but discarded prior to the row we are sending.

nickva:
Thanks for running the performance check, @pgj!

> I do not think it should produce repeatable results. If a query happened to cause the scanning of 2,000 documents once and then 1,500 for another instance is acceptable — that is how the system works. But returning zero in either case is definitely a miss on the side of accounting.

It will depend not just on how many were scanned but also in what order they arrived and how they were merged at the coordinator. The idea is there is some fudge factor there as is. I updated the query script to query the API endpoint multiple times in a row for the same docs:

```
% ./findlimit.py
 * deleting http://127.0.0.1:15984/db
 * creating http://127.0.0.1:15984/db {'q': '16'}
 * creating 10000 docs with val range 100
 * created docs in 6.0 seconds
 * _find in 0.1 seconds 5 docs {'total_keys_examined': 3382, 'total_docs_examined': 3382, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 50.381}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3323, 'total_docs_examined': 3323, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 37.183}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3359, 'total_docs_examined': 3359, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 32.422}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3347, 'total_docs_examined': 3347, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 31.574}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3282, 'total_docs_examined': 3282, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 31.719}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3325, 'total_docs_examined': 3325, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 30.074}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3276, 'total_docs_examined': 3276, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 29.015}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3274, 'total_docs_examined': 3274, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 30.524}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3120, 'total_docs_examined': 3120, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 29.053}
 * _find in 0.0 seconds 5 docs {'total_keys_examined': 3368, 'total_docs_examined': 3368, 'total_quorum_docs_examined': 0, 'results_returned': 5, 'execution_time_ms': 31.881}
 * docs: 3382 - 3120 = 262
```

I like @rnewson's idea to add stats to  the view row. The `view_row` is a record, so we'd have to do a bit tuple wrangling across multiple commits (PRs) to avoid breaking online cluster upgrades, but that's the cleanest solution overall. 

As a bonus, we'd also then avoid odd out-of-order processing of stats like we have here: https://github.com/apache/couchdb/blob/2501fe69b06dd88611b4a3e290f080823476af70/src/fabric/src/fabric_view_all_docs.erl#L260-L274 Where we may accumulate worker stats even though we might not actually have processed the row emitted before.

With per-row stats we may miss some stats emitted by workers at the end from workers which sent the rows already, but they just didn't go towards producing the response. So there is some discrepancy between total work induced in the cluster as a result of the API request vs work which took place on the workers before the emitted rows, if those rows were included in the response.

> For the chttpd_stats_reporter case, we could stream stats directly to chttpd_stats_reporter instead of going through the same framework that execution stats uses, and would be free to complete sending statistics after the response has been returned to the user.

Ah, that's interesting to think about chttpd_stats, @willholley.  That's in chttpd and it takes in request/response objects. But if we wanted to emit online stats for each request as it's being processed (a long running request), we could alter the API there such that each worker would get some set of request bits (path, principal, nonce, request id?...) passed to it (in #mrargs extras), and then the worker can report stats independently without having to shuffle them back to the coordinator. Kind of a larger change, but that way it could account for all the work generated as a side-effect of an API call, even if it didn't make as part of the response.

pgj:
I have implemented the half of @rnewson's suggestions and the code is now sending execution stats along with the view rows only.  That is still a separate message, but it already improved the performance (same configuration as above):

- 1 thread: `min=0.79ms max=7556.06ms  p(95)=6461.86ms p(98)=6609.84ms p(99)=6741.12ms,  count=209`
- 2 threads: `min=0.78ms max=31902.38ms p(95)=13449.74ms p(98)=14130.32ms p(99)=14873.91ms, count=191`

It seems to be only 5-10% off with this change only.

pgj:
There is an updated version now where the stats are sent along with the row data.  Instead of creating a new record type, I switched to using maps and added support for coordinators and workers of the old version.  The benchmark shows good performance (same configuration as earlier):

 - 1 thread: `min=0.77ms max=9711.29ms  p(95)=5879.81ms p(98)=6864.91ms p(99)=7143.41ms,  count=248`
 - 2 threads: `min=0.60ms max=14197.50ms p(95)=13028.72ms p(98)=13755.92ms p(99)=13950.17ms, count=238`

I tested this manually with old worker—new coordinator, and new worker—old coordinator configurations with a random set of `/{db}/_find` queries, and it was working.

Also, the combined message format revealed that in case of limits, there is always an extra key and sometimes and extra document examined.  Previously this information was lost since the corresponding `execution_stats` message could not make its way to the coordinator.

The only question that remains (for me) is whether in `#mrargs.extra`, a separate key should be created for the map-based row messages.  Currently, this is tied to `execution_stats_rolling`.  It obviously depends on maps, but it might make sense to introduce maps with its own dedicated `view_row_map` key e.g., even in a separate PR.  The `execution_stats_rolling` could then depend on the presence of this feature.

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -273,10 +313,24 @@ handle_message({execution_stats, _} = Msg, {_, From}, St) ->
...
+    lists:reverse(insert_row(fwd, Key, R, lists:reverse(List))).
+
+merge_row(fwd, #view_row{} = Row, Rows) ->
     lists:keymerge(#view_row.id, [Row], Rows);
-merge_row(rev, Row, Rows) ->
-    lists:rkeymerge(#view_row.id, [Row], Rows).
+merge_row(rev, #view_row{} = Row, Rows) ->
+    lists:rkeymerge(#view_row.id, [Row], Rows);
+merge_row(Dir, #{} = Row, Rows) ->
+    insert_row(Dir, id, Row, Rows).
```
nickva, line 333: I think only `id` is the key. Might be better to just simply and hard-code the key in insert_row

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -273,10 +313,24 @@ handle_message({execution_stats, _} = Msg, {_, From}, St) ->
     rexi:stream_ack(From),
     {Go, St#collector{user_acc = Acc}}.
 
-merge_row(fwd, Row, Rows) ->
+insert_row(_Dir, _Key, R, []) ->
+    [R];
+insert_row(fwd, Key, R, [H | T] = List) ->
+    V1 = maps:get(Key, R),
+    V2 = maps:get(Key, H),
```
nickva, line 320: I think we can just use `id` here and also do a more common KV match `#{id := RowId} = R`and `#{id := HeadId} = H`. With `id` atom those an be matched directly in the head, whichever one looks better to you.

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -273,10 +313,24 @@ handle_message({execution_stats, _} = Msg, {_, From}, St) ->
...
+    if
+        V1 =< V2 -> [R | List];
+        true -> [H | insert_row(fwd, Key, R, T)]
+    end;
```
nickva, line 324: Nit: Avoid using `if` in Erlang as a rule, usually `case` is preferred. `case V1 =< V2 of true -> ... ; false -... end`

src/fabric/src/fabric_view.erl
```diff
@@ -241,11 +246,63 @@ possibly_embed_doc(
...
+                                        {ok, [{{not_found, _}, Rev}]} ->
+                                            Row#{doc => null};
+                                        Else ->
+                                            Row#{doc => {error, Else}}
+                                    end
+                            end
+                        )
+                    end),
+                    receive
+                        {'DOWN', Ref, process, Pid, Resp} ->
```
nickva, line 293: If we have a utilty function to fetch the doc body only, wonder if we could remove some of the duplication between the two cases.

src/fabric/src/fabric_view.erl
```diff
@@ -349,7 +410,11 @@ transform_row(#view_row{key = Key, id = Id, value = Value, doc = undefined}) ->
 transform_row(#view_row{key = Key, id = _Id, value = _Value, doc = {error, Reason}}) ->
     {row, [{id, error}, {key, Key}, {value, Reason}]};
 transform_row(#view_row{key = Key, id = Id, value = Value, doc = Doc}) ->
-    {row, [{id, Id}, {key, Key}, {value, Value}, {doc, Doc}]}.
+    {row, [{id, Id}, {key, Key}, {value, Value}, {doc, Doc}]};
+transform_row(#{} = Row) ->
+    #{key := Key, id := Id, value := Value, doc := Doc, stats := Stats} = Row,
+    {row, Props} = transform_row(#view_row{key = Key, id = Id, value = Value, doc = Doc}),
+    {row, Props, Stats}.
```
nickva, line 417: Thinking out loud: this feels a slightly awkward, that the callback would sometimes be called with 2 or 3 times. What we if added the `stats` field to the row props and accessed them later or removed them later.  That's just so that the transform calls always return `{row, Props}` only.

nickva (commented):
Fantastic work, @pgj!

I like the new view row map. I wonder if we can go a step further and make all the view rows into maps, even the map/reduce and the other cases. In other words eventually be able to remove the view_row record altogether. But I think it should also be its own PR.

In the separate PR for mrviews, we'd update https://github.com/apache/couchdb/blob/main/src/fabric/src/fabric_rpc.erl#L480-L531. Based an on the `#mrargs.extra` option we'd either use `view_cb` or `view_cb_compact`. One would return a map the other, the record. Replicator might be another places which uses view_row records, so we might to update those as well. Same idea as you did for mango rows.

Since we're trying to return stats with every row, we'd also want to account for all the rows emitted after the last row and until the `complete` message. As is, we'd still need to send `rexi:stream2({execution_stats, Stats}),` just before the `complete` is emitted. What if, while we're changing the view row formats, we also update complete to be `{complete, #{stats => Stats}}` instead of `complete`. Another one to update might be the no_pass message, that could also have an additional map for stats or other things. 

To summarize, I think it's a great change, it's about time we switch away from sending hard-coded records across the cluster. But I think we can take this effort ever further and first have a PR to use map for view rows and complete callbacks. And then, in a separate PR, add rolling stats, which should now be trivial.

pgj:
@nickva thanks for the confirmation and the suggestions!  I will address them in the other PR about introducing `view_row_map`.  I will then rebase this PR to that one and take care of the remainders, if there will be any.

src/fabric/src/fabric_view.erl
```diff
@@ -241,11 +246,63 @@ possibly_embed_doc(
             end;
         _ ->
             Row
+    end;
+possibly_embed_doc(
+    #collector{db_name = DbName, query_args = Args},
+    #{key := _Key, id := _Id, value := Value, doc := _Doc} = Row
```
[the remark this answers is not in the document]
pgj: I guess those names with the underscore prefix were spelled out for documentation.  I do not think they do any harm as the compiler automatically discards them due to the underscore in the beginning.  Maintaining readability could always come in handy.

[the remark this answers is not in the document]
pgj: However...  keeping those pattern matches around would impose additional set of constraints on the use of the function which could reduce its flexibility.  Aw, you are right.  With maps we do not need this anymore.

src/fabric/src/fabric_view.erl
```diff
@@ -241,11 +246,63 @@ possibly_embed_doc(
...
+        _ ->
+            Row
     end.
 
 detach_partition(#view_row{key = {p, _Partition, Key}} = Row) ->
     Row#view_row{key = Key};
-detach_partition(#view_row{} = Row) ->
+detach_partition(#{key := {p, _Partition, Key}} = Row) ->
+    Row#{key => Key};
+detach_partition(Row) ->
```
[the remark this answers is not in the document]
pgj: I do not think this causes problems as missing keys will automatically be considered as `undefined`.

src/fabric/src/fabric_view_all_docs.erl
```diff
@@ -187,6 +187,49 @@ shards(Db, Args) ->
...
+                end;
+            false ->
+                case Row1 of
+                    #view_row{} ->
+                        Row1;
+                    #{} ->
+                        #{id := Id, key := Key} = Row1,
+                        Value = maps:get(value, Row1, null),
+                        Doc = maps:get(doc, Row1, null),
+                        Worker = maps:get(worker, Row1, null),
```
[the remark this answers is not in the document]
pgj: No, that is `undefined`.

nickva (approved):
@pgj nicely done!

Also kudos to @rnewson for catching the performance regression and coming up with a better design for sending rows stats (https://github.com/apache/couchdb/pull/4958#issuecomment-1919033815)

pgj:
Thank you all for your help too!

PR #4954, opened 2024-01-08
chore(Jenkins): enable Clouseau for more of the platform builds

pgj:
The CI container images have been updated and now they include support for running Clouseau with them.  Throw the switch and let Clouseau configured and used to extend the coverage.

PR #4959, opened 2024-01-10
chore(Jenkins): enable Clouseau for more platforms

pgj:
The CI container images have been updated and now they include support for running Clouseau with them.  Throw the switch and let Clouseau configured and used to extend the coverage.

The FreeBSD builder is not yet enabled due to failures that have been fully investigated.

See #4954 for the build logs and preliminary validation.


nickva (approved):
+1 based on results from #4954 (both failures there seem flaky)

PR #4982, opened 2024-02-10
fix flaky password hashing test

rnewson:


test/elixir/test/password_cache_test.exs
```diff
@@ -128,11 +128,10 @@ defmodule PasswordCacheTest do
     user1 = Map.replace!(user1, "password", "new_password")
     save_doc(db_name, user1)
 
-    # Wait for auth cache to notice password change
-    :timer.sleep(5500)
-
     # Slow rejection for wrong password
-    assert_cache(:expect_slow, "user1", "wrong_password", :expect_login_fail)
+    retry_until(fn ->
```
pgj, line 132: Is this an unbounded retry, I wonder?

jaydoane: Based on [this code](https://github.com/apache/couchdb/blob/8f6797b8d35c7fb0e257e5873f01d8f04b9e1031/test/elixir/lib/couch/dbtest.ex#L436) the default timeout appears to be 30 seconds.

rnewson: retry_until is used in lots of places, I should have used it the first time around.

I think we 'just' need to extend the outer timeout, though I don't know how. Though it is puzzling that auth cache invalidation takes even 2.5s (locally) let alone more than 5s.

big-r81: but why does it fails then?

rnewson: unsure. it passes locally.

rnewson: ah, but only if I run it individually (as is my habit when writing new tests). in the suite as a whole it fails (times out at 30s).

big-r81: ah yes, I can confirm this. btw. is there a way to only run a collection of test in elixir too like in eunit?
in the log we got messages like this:
```
[warning] 2024-02-12T14:25:24.415998Z node1@127.0.0.1 <0.796.0> d6c79accfc couch_httpd_auth: Authentication failed for user user1 from 127.0.0.1
```

big-r81 (commented):
an official "post merging" +1 ;-)

PR #4986, opened 2024-02-14
nouveau: jdk 21 compatibility

rnewson:
* upgrade gradle to 8.6
* upgrade spotless to 6.25.0
* rename 'permits' as its a (new) reserved keyword (for sealed packages)


Makefile
```diff
@@ -557,7 +557,7 @@ nouveau-test: nouveau-test-gradle nouveau-test-elixir
 .PHONY: nouveau-test-gradle
 nouveau-test-gradle: couch nouveau
 ifeq ($(with_nouveau), true)
-	@cd nouveau && ./gradlew test
+	@cd nouveau && ./gradlew test --info
```
big-r81, line 560: I think we should port this to `Makefile.win` too.

big-r81 (changes requested):

big-r81 (approved):
+1

PR #5015, opened 2024-03-25
nouveau: fix rexi_DOWN clause

rnewson:
reported in https://github.com/apache/couchdb/issues/5002

the badrecord is because we execute the second clause of handle_message and mess up the internal state (just returning the Counters rather than a state record around it)


big-r81 (approved):
+1

PR #5038, opened 2024-04-25
Fix small detail about conflicts in overview.rst

Sliosh:
Conflicts are not purged while compaction and have to be deleted manually

## Overview

There is a little error in the docs, stating that conflicts could be purged during compaction. This is removed because its not correct.

- [ x] Documentation changes were made in the `src/docs` folder


rnewson (changes requested):
the CI checks reject this as the new line is too long. can you reindent and then I can merge. thanks.

rnewson (approved):

rnewson:
thanks!

Issue #5046, opened 2024-05-03
Add "exp" to mandatory JWT claims


evavica:
## Summary

As I implemented JWT auth for CouchDb, I stumbled upon the "exp" claim being optional and not checked if not configured explicitly.
As long-lived or never expiring JWT tokens are most likely not what you (should) want when working with them, I would expect that claim to be part of standard validation behavior.
Making it optional might lead people to think it's ok to neglect it.

## Desired Behaviour

This claim should be mandatory by default, as JWTs should be short lived following the concept. 
If someone actually wants to make their tokens last forever, they would have to actively make a bad decision rather than passively making it by not carefully reading the documentation, not testing properly or a lack of conceptual understanding of JWT auth.

## Possible Solution

Please consider adding "exp" to the mandatory claims alongside "alg" and "sub".
If it has to stay possible to disable the check for a (good) reason, please add a configuration opportunity to disable the default exp check.

If you do not want to implement this, please emphasize that ignoring "exp" is a bad idea in the documentation. 

## Additional context

Security is not everybody's favorite topic. "Do my tokens really have to be short lived" is a common question beginners have. As a community, we should help pointing people in the right direction by defining sensible defaults and implementing best practices to promote the proper design of things. 

## Acknowledgements
Anyways, thank you for your time and work on this great product!

rnewson:
We used to validate the `exp` claim if present in the token, regardless of whether it was in `required_claims`, but this turned out to be a spec violation, so we changed to ignore the non-required claims.

```
commit 331894a6acb4565c71d800f2e63206101dfbb48c
Author: Jay Doane <jaydoane@apache.org>
Date:   Mon Mar 1 13:48:19 2021 -0800

    Ignore unchecked JWT claims

    Previously, if a JWT claim was present, it was validated regardless of
    whether it was required.

    However, according to the spec [1]:

    "all claims that are not understood by implementations MUST be ignored"

    which we interpret to mean that we should not attempt to validate
    claims we don't require.

    With this change, only claims listed in required checks are validated.

    [1] https://tools.ietf.org/html/rfc7519#section-4
```

I agree with you that `exp` is an important field (along with `nbf`) and there is an argument for it being in the default list of `required_claims`. The current default is an empty list (`alg` and `sub` are required for other reasons, we can't validate without `alg` and without `sub` we wouldn't know which user we'd be authenticating).

evavica:
[...]

PR #5057, opened 2024-05-16
JWT: require valid `exp` claim by default

rnewson:
## Overview

Users of JWT rightly expect tokens to be considered invalid once they expire. It is a surprise to some that this requires a change to the default configuration. In the interest of security we will now require a valid `exp` claim in tokens. Administrators can disable the check by changing `required_claims` back to the empty string.

We do not add `nbf` as a required claim as it seems to not be set often in practice.

## Testing recommendations

N/A

https://github.com/apache/couchdb/issues/5046

## Checklist

### DIFF
--- a/rel/overlay/etc/default.ini
+++ b/rel/overlay/etc/default.ini
@@ -225,7 +225,10 @@ bind_address = 127.0.0.1
 ; List of claims to validate
 ; can be the name of a claim like "exp" or a tuple if the claim requires
 ; a parameter
-;required_claims = exp, {iss, "IssuerNameHere"}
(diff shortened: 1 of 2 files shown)


big-r81 (approved):
+1

big-r81 (changes requested):
As Jan [commented](https://couchdb.slack.com/archives/C01TBE2J197/p1715857416679689), remove the 3.4 changelog entry in this PR and integrate it into #4888.

nickva (approved):

big-r81 (approved):

PR #5063, opened 2024-05-21
Fix duplicate keys in find query

xrobau:
RFC-8259 discourages using duplicate keys in JS/JSON, and the ECMA-262 standard explicitly says that the last key will overwrite the first key.

This probably means these never actually worked!


nickva (approved):
+1

Thank you for your contribution, @xrobau

PR #5136, opened 2024-07-16
print log dir on dev/run startup

rnewson:
print log dir on dev/run startup


big-r81:
Needs an additional `python-black-update`…

big-r81 (approved):

PR #5151, opened 2024-07-26
dev(chore): Use atom for config key `with_spidermonkey`

big-r81:


rnewson (approved):
bonus points if we also fix the typo in "js_enigine"

nickva (approved):
+1

./configure --disable-spidermonkey and ./configure

`./configure --disable-spidermonkey` and `./configure` work. Followed by `make dist`, unpack tgz, `./configure ...` again and `make release`.

Also  +1 to fix my silly typo. Thanks @rnewson for noticing!

big-r81:
> bonus points if we also fix the typo in "js_enigine"

Ah, good catch. Updated.

PR #5209, opened 2024-08-29
enable native access for Lucene

rnewson:
toggle the native access support for Lucene (requires launching with `java -jar` in packages/scripts though)


nickva (approved):
+1

Tested with dev/run then `config:set("nouveau", "enable", "true").`

```
% http post $DB/_nouveau_analyze analyzer=keyword text=ab
HTTP/1.1 200 OK

{
    "tokens": [
        "ab"
    ]
}

```

Issue #5273, opened 2024-10-02
Nouveau range faceting doesn’t work as documented


espy:
## Description

The range faceting syntax in the [Nouveau docs](https://docs.couchdb.org/en/latest/ddocs/nouveau.html#ranges) doesn’t work. It describes the same ranges syntax as the [ddocs search page](https://docs.couchdb.org/en/latest/api/ddoc/search.html#get--db-_design-ddoc-_search-index), but this doesn’t actually work with Nouveau.

I’m not sure whether the Nouveau docs are wrong, or Nouveau is missing the intended implementation. There is a working syntax described elsewhere.

In addition, this other working syntax has confusing defaults for inclusive/exclusive range facets (described further below).

## Steps to Reproduce

In a CouchDB with activated Nouveau, with a running Nouveau server:

1. Have some docs with a number value, eg. `rating: 8.1`
2. Add a Nouveau `double` index for that key:
    ```json
    {
      "_id": "_design/nouveau_example",
      "nouveau": {
        "range": {
          "index": "if (doc.rating) { index('double', 'rating', doc.rating) } }"
        }
      }
    }
    ```
3. Send a ranges query as described in the [Nouveau Docs](https://docs.couchdb.org/en/latest/ddocs/nouveau.html#ranges), but without using `Infinity`, because that also doesn’t work as documented, see [this issue](https://github.com/apache/couchdb/issues/5271):

    Taken from the docs:
    ```
    ?q=*:*&ranges={"price":{"cheap":"[0 TO 100]","expensive":"{100 TO Infinity}"}}
    ```
    …adapted to use our data and `10` instead of `Infinity`:
    ```
    ?q=*:*&ranges={"rating":{"boring":"[0 TO 8]","amazing":"{8 TO 10}"}}
    ```
    As JSON, for readability:
    ```JSON
    {
      "rating": {
        "boring": "[0 TO 8]",
        "amazing": "{8 TO 10}"
      }
    }
    ```
4. This returns:
    ```json
    {
      "error": "query_parse_error",
      "reason": "range values must be lists of objects"
    }
    ```

Turning the query into a list of objects doesn’t work though. What _does_ work is the syntax found in this [Nouveau README.md](https://github.com/apache/couchdb/blob/8d5669173841923a8653acddc2200f6fff8547b8/nouveau/README.md?plain=1#L95):

```
[...]

PR #5276, opened 2024-10-02
fix ranges doc for nouveau

rnewson:
closes: https://github.com/apache/couchdb/issues/5273

### DIFF
--- a/src/docs/src/ddocs/nouveau.rst
+++ b/src/docs/src/ddocs/nouveau.rst
@@ -692,23 +692,17 @@ Ranges
 .. note::
     The ``ranges`` option is only available when making global queries.
 
-The ``range`` facet syntax reuses the standard Lucene syntax for ranges to return counts
-of results that fit into each specified category. Inclusive range queries are denoted by
-brackets (``[``, ``]``). Exclusive range queries are denoted by curly brackets (``{``,
-``}``).
-
-.. note::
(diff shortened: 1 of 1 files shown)


janl (approved):

src/docs/src/ddocs/nouveau.rst
```diff
@@ -692,23 +692,17 @@ Ranges
...
+must have a ``label``, ``min`` and ``max`` value (of type string, double, double
+respectively), and optional ``min_inclusive`` and ``max_inclusive`` properties
+(defaulting to ``true`` if not specified).
 
 *Example of a request that uses faceted search for matching ranges:*
 
 .. code-block:: text
 
-    ?q=*:*&ranges={"price":{"cheap":"[0 TO 100]","expensive":"{100 TO *}"}}
+    ?q=*:*&ranges={"price":[{"label":"cheap","min":0,"max":"100"},{"label":"expensive","min":100}]}
```
espy, line 705: This example is slightly confusing, since anything priced at `100` will fall into both facets and be counted twice. Would suggest:

```
?q=*:*&ranges={"price":[{"label":"cheap","min":0,"max":"100", "max_inclusive": false},{"label":"expensive","min":100 }]}
```
This also provides an example for the booleans.

rnewson: yes, I like that

PR #5280, opened 2024-10-04
fix(`default.ini`): sync suggested fabric timeout settings with the sources

pgj:
The default settings for certain fabric timeout actions do not reflect the reality in the comments of `default.ini`.  Adjust those to match with the constants are laid down in the sources.

Note that apparently search requests are also associated with configurable timeouts, so document those as well.


nickva (approved):

pgj:
Thank you @nickva!

PR #5324, opened 2024-10-28
Add clouseau_rpc_test

jiahuili430:
## Overview

Add tests for clouseau_rpc.erl.

## Testing recommendations
Enable clouseau, and then run `make eunit`
```bash
./configure --dev --disable-spidermonkey
make
make eunit-search
```

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,6 +157,7 @@ escriptize: couch
 check: all
 	@$(MAKE) xref
 	@$(MAKE) eunit
+	@$(MAKE) eunit-search
 	@$(MAKE) mango-test
 	@$(MAKE) elixir
 	@$(MAKE) elixir-search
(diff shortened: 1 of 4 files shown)


rnewson:
I realise this is draft but I have a concern. A lot of this test appears to be asserting the internals of the implementation, rather than testing that it works functionally. The cut-and-paste of records from the implementation, etc. Before you spend much more time on this I think it's worth reassessing the purpose of eunit tests.

rnewson:
the peculiar line in the makefile target that uses find to delete a beam file is a concern also. that is a very unusual way to disable a test (and won't work on Windows, though perhaps you intend a variation when you updated Makefile.win). I don't think it's the right way.

jiahuili430:
Trying to use this test to compare the differences between different Clouseau.
If it doesn't work on Windows, I'll drop this PR.
Thank you for the heads up! @rnewson

PR #5341, opened 2024-11-22
Fix mango tests using custom db name

jiahuili430:
## Overview

When running mango tests, sometimes get `org.apache.lucene.store.NoSuchDirectoryException` error. This is because each test in the same class used the same database name, and in the setUp function always delete and recreate it, which caused a race condition. Using custom database names should solve it.

- Makefile: Add `-F, --failfast` to stop mango tests early on failures
- Add `setUp()` and `teardown()` to create/delete db for each test to  
  avoid `org.apache.lucene.store.NoSuchDirectory Exception` errors.
- Replace `self.db.recreate()` with `super().setUp(db_per_test=True)`

## Testing recommendations

`make mango-test`

  

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -339,7 +339,7 @@ mango-test: devclean all
 		--admin=adm:pass \
 		--no-eval "\
 COUCH_USER=adm COUCH_PASS=pass \
-src/mango/.venv/bin/nose2 -s src/mango/test -c src/mango/unittest.cfg $(MANGO_TEST_OPTS)"
+src/mango/.venv/bin/nose2 -F -s src/mango/test -c src/mango/unittest.cfg $(MANGO_TEST_OPTS)"
 
 
 .PHONY: weatherreport-test
--- a/src/mango/test/16-index-selectors-test.py
+++ b/src/mango/test/16-index-selectors-test.py
@@ -82,7 +82,7 @@ oldschoolddoctext = {
 
 class IndexSelectorJson(mango.DbPerClass):
     def setUp(self):
-        self.db.recreate()
+        super().setUp(db_per_test=True)
         self.db.save_docs(copy.deepcopy(DOCS))
 
     def test_saves_partial_filter_selector_in_index(self):
@@ -189,7 +189,7 @@ class IndexSelectorJson(mango.DbPerClass):
 @unittest.skipUnless(mango.has_text_service(), "requires text service")
 class IndexSelectorText(mango.DbPerClass):
     def setUp(self):
-        self.db.recreate()
+        super().setUp(db_per_test=True)
         self.db.save_docs(copy.deepcopy(DOCS))
 
     def test_saves_partialfilterselector_in_index(self):
--- a/src/mango/test/mango.py
+++ b/src/mango/test/mango.py
@@ -80,10 +80,7 @@ class Concurrently(object):
 
 
 class Database(object):
-    def __init__(
-        self,
-        dbname,
-    ):
+    def __init__(self, dbname):
         self.dbname = dbname
         self.sess = requests.session()
         self.sess.auth = (COUCH_USER, COUCH_PASS)
@@ -107,24 +104,19 @@ class Database(object):
 
     def delete(self):
         r = self.sess.delete(self.url)
+        r.raise_for_status()
 
     def recreate(self):
-        NUM_TRIES = 10
-
-        for k in range(NUM_TRIES):
-            r = self.sess.get(self.url)
-            if r.status_code == 200:
-                db_info = r.json()
-                docs = db_info["doc_count"] + db_info["doc_del_count"]
-                if docs == 0:
-                    # db exists and it is empty -- exit condition is met
-                    return
-                self.delete()
-            self.create()
-            time.sleep(k * 0.1)
-        raise Exception(
-            "Failed to recreate the database after {} tries".format(NUM_TRIES)
-        )
+        r = self.sess.get(self.url)
+        if r.status_code == 200:
+            db_info = r.json()
+            docs = db_info["doc_count"] + db_info["doc_del_count"]
+            if docs == 0:
+                # db is not in use, no need to recreate
+                return
+            self.delete()
+        self.create()
+        self.recreate()
 
     def save_doc(self, doc):
         self.save_docs([doc])
@@ -356,8 +348,21 @@ class DbPerClass(unittest.TestCase):
         if clean_up_dbs():
             klass.db.delete()
 
-    def setUp(self):
-        self.db = self.__class__.db
+    def setUp(self, db_per_test=False, partitioned=False):
+        if db_per_test:
+            self.db = Database(random_db_name())
+            self.db.create(q=1, n=1, partitioned=partitioned)
+            self.db_per_test = db_per_test
+        else:
+            self.db = self.__class__.db
+
+    def tearDown(self):
+        if (
+            hasattr(self, "db_per_test")
+            and self.__getattribute__("db_per_test")
+            and clean_up_dbs()
+        ):
+            self.db.delete()
 
 
 class UserDocsTests(DbPerClass):
--- a/Makefile.win
+++ b/Makefile.win
@@ -310,7 +310,7 @@ mango-test: devclean all
 		--admin=adm:pass \
 		"\
 env COUCH_USER=adm COUCH_PASS=pass \
-src\mango\.venv\Scripts\nose2 -s src\mango\test -c src\mango\unittest.cfg $(MANGO_TEST_OPTS)"
+src\mango\.venv\Scripts\nose2 -F -s src\mango\test -c src\mango\unittest.cfg $(MANGO_TEST_OPTS)"
 
 
 ################################################################################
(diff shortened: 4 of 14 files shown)


src/mango/test/16-index-selectors-test.py
```diff
@@ -189,6 +189,7 @@ def test_uses_partial_index_with_non_indexable_selector(self):
 @unittest.skipUnless(mango.has_text_service(), "requires text service")
 class IndexSelectorText(mango.DbPerClass):
     def setUp(self):
+        self.db = mango.Database(f"{mango.random_db_name()}_{str(self).split()[0]}")
```
iilyak, line 192: There are many tests files which use `self.db.recreate()` did you consider modifying that function?

src/mango/test/16-index-selectors-test.py
```diff
@@ -189,6 +189,7 @@ def test_uses_partial_index_with_non_indexable_selector(self):
...
         self.db.recreate()
```
jaydoane, line 193: There was [recent work](https://github.com/apache/couchdb/commit/eb6a74d5603d4bac1d77fc895946ffbdd99f4b97) done on the `recreate` function, but it didn't seem to address the raciness of reusing db names.

jiahuili430: Thank you for the link!

src/mango/test/16-index-selectors-test.py
jiahuili430: If using the same db name, `0.13` is the shortest time I need to avoid `500` locally.
This approach also increased the overall Mango test time to `60.730s`, compared to `49.936s` using custom database name.

```diff
def recreate(self):
-    NUM_TRIES = 10
+    NUM_TRIES = 3

-    for k in range(NUM_TRIES):
+    for k in range(1, NUM_TRIES):
         r = self.sess.get(self.url)
         if r.status_code == 200:
             db_info = r.json()
             docs = db_info["doc_count"] + db_info["doc_del_count"]
                 if docs == 0:
                     # db exists and it is empty -- exit condition is met
                     return
                 self.delete()
+        time.sleep(k * 0.13)
         self.create()
-        time.sleep(k * 0.1)
     raise Exception(
         "Failed to recreate the database after {} tries".format(NUM_TRIES)
     )
```

iilyak: All time based approaches are flowed. They don't solve the problem. They introduce flakiness. The options I see:

1. restructure all tests so they are not using recreate. They would always start from scratch and create fresh databaase with uniq name. 
2. update dreyfus_index_manager.erl to handle `cleanup` differently
3.  modify clouseau so it handle rapid re-creation better
4. change recreate function so it uses new name on each retry

pgj: I would vote for option 4.

src/mango/test/16-index-selectors-test.py
pgj: Yes, that scope of that commit was only limited to solving the infinite loop of retries with recursion.

src/mango/test/16-index-selectors-test.py
pgj: Virtually, that is still about recreating the database, in my opinion.  But I agree that `create_random` would capture better the adjusted semantics for option 4.

src/mango/test/mango.py
```diff
@@ -107,24 +104,19 @@ def create(self, q=1, n=1, partitioned=False):
 
     def delete(self):
         r = self.sess.delete(self.url)
+        r.raise_for_status()
 
     def recreate(self):
-        NUM_TRIES = 10
-
-        for k in range(NUM_TRIES):
```
jiahuili430, line 114: Tried many times, `k` is only equal to 0 and 1, so I decided to revert back.

iilyak (approved):
+1

I run the `mango test` many times and it is definitely makes mango tests less flaky. 
```
❯ make mango-test
==> snappy (compile)
==> ibrowse (compile)
==> jiffy (compile)
==> mochiweb (compile)
==> meck (compile)
==> recon (compile)
==> proper (compile)
==> couch_epi (compile)
==> config (compile)
==> couch_log (compile)
WARN:  Missing plugins: [pc]
==> b64url (compile)
WARN:  Missing plugins: [pc]
==> exxhash (compile)
==> ets_lru (compile)
==> couch_quickjs (compile)
make[1]: `libquickjs.lto.a' is up to date.
make[1]: `qjsc' is up to date.
==> chttpd (compile)
==> couch (compile)
==> couch_event (compile)
==> mem3 (compile)
==> couch_index (compile)
==> couch_mrview (compile)
==> couch_replicator (compile)
==> couch_pse_tests (compile)
==> couch_stats (compile)
==> couch_peruser (compile)
==> couch_tests (compile)
==> couch_dist (compile)
==> custodian (compile)
==> ddoc_cache (compile)
==> dreyfus (compile)
==> nouveau (compile)
==> fabric (compile)
==> global_changes (compile)
==> ioq (compile)
==> jwtf (compile)
==> ken (compile)
==> mango (compile)
==> rexi (compile)
==> setup (compile)
==> smoosh (compile)
==> weatherreport (compile)
==> couch_prometheus (compile)
==> couch_scanner (compile)
==> rel (compile)
==> couchdb (compile)
==> weatherreport (escriptize)
[ * ] Setup log dir: /Users/iilyak/Code/couchdb/dev/logs ... ok
[ * ] Start node node1 ... ok
[ * ] Exec command  COUCH_USER=adm COUCH_PASS=pass src/mango/.venv/bin/nose2 -F -s src/mango/test -c src/mango/unittest.cfg  ... ......................................................................................................................................................................................................................................s..............................................................................................................................................................
Ran 389 tests in 41.966s

OK (skipped=1)
```

Makefile
```diff
@@ -339,7 +339,7 @@ mango-test: devclean all
 		--admin=adm:pass \
 		--no-eval "\
 COUCH_USER=adm COUCH_PASS=pass \
-src/mango/.venv/bin/nose2 -s src/mango/test -c src/mango/unittest.cfg $(MANGO_TEST_OPTS)"
+src/mango/.venv/bin/nose2 -F -s src/mango/test -c src/mango/unittest.cfg $(MANGO_TEST_OPTS)"
```
pgj, line 342: What does `-F` buy for us?  The Mango tests are quick to run, and sometimes perhaps it is good to see if there are other errors besides the one that failed the whole run.

jiahuili430: While working on this, I even changed `unittest.cfg` to run `16-index-selectors-test.py` only. Because I keep getting 500 errors due to this test file. Different combinations of test cases will result in the same failure message but fail on different test cases. 0.85 sec vs 50 sec to run the entire test suite.

`-F, --failfast` allows us to identify the issue quickly, speed up the feedback loop, and make debugging easier.

pgj: > 0.85 sec vs 50 sec to run the entire test suite.

Sure, I know that running the Mango test suite in the past required around a minute to run.  In my opinion, that is acceptable, since it is still does not take tens of minutes or many hours.

> Different combinations of test cases will result in the same failure message but fail on different test cases.

I am not sure if I got you on this.  If the "fail fast" mode stops at the first failing test, how can it result in the failure message if there are multiple failures?

> `-F, --failfast` allows us to identify the issue quickly, speed up the feedback loop, and make debugging easier.

Note that you could tell the names of the modules or specific test cases on the command line when invoking `nose2`.  This is also exposed in the general `Makefile` of the repository, for the `mango-test` target as the `MANGO_TEST_OPTS` variable.

With the help of this, one could easily limit the current run to the tests that one has to run for keeping the feedback loop short.  With the regards to finding the tests to run, it is okay to run the whole suite once and take the cost of the full run.

jiahuili430: > Same error message, but bailed on different test cases.

The error message is as follows:
```bash
  File "/Users/jiahuili/src/a/src/mango/test/16-index-selectors-test.py", line 214, in test_uses_partial_index_for_query_selector
    docs = self.db.find(selector, use_index="Selected", fields=["_id", "location"])
  File "/Users/jiahuili/src/a/src/mango/test/mango.py", line 317, in find
    r.raise_for_status()
    ~~~~~~~~~~~~~~~~~~^^
  File "/Users/jiahuili/src/a/src/mango/.venv/lib/python3.13/site-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://127.0.0.1:15984/mango_test_480b738790f748d9b5f3c5c17202d501/_find
```

Sometimes it fails on `test_uses_partial_index_for_query_selector` and sometimes it fails on `test_uses_partial_index_with_different_selector`. Comment out some test cases and you'll get different test results.

We can remove the `-F` if it is not needed: https://github.com/apache/couchdb/pull/5348/files

pgj: > Comment out some test cases and you'll get different test results.

That is what one would not want to do.  Commenting test cases in and out is what makes the feedback loop slower.

> We can remove the `-F` if it is not needed: https://github.com/apache/couchdb/pull/5348/files?w=1

I have not had the time to try this new approach, maybe it works.

PR #5369, opened 2024-12-19
fix(chttpd): Always show browser login popup

big-r81:
Always show the browser login popup, no matter
which option of `require_valid_user` or
`require_valid_user_except_up` is set.

### DIFF
--- a/src/chttpd/src/chttpd.erl
+++ b/src/chttpd/src/chttpd.erl
@@ -1235,13 +1235,17 @@ error_headers(#httpd{mochi_req = MochiReq} = Req, 401 = Code, ErrorStr, ReasonSt
                                 undefined ->
                                     {Code, []};
                                 AuthRedirect ->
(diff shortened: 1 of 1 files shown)


big-r81:
## Additional infos

### Abbreviations

`rvu` = `request_valid_user`
`rvuefu` = `request_valid_user_except_for_up` 

### Before #5370 (use auth for live node checks)

| rvu | rvuefu | start dev server |
| - | - | - |
| false | false | false |
| true | false | false |
| false | true | false |
| true | true | false |

### After #5370 

| rvu | rvuefu | start dev server |
| - | - | - |
| false | false | true |
| true | false | true |
| false | true | false |
| true | true | true |

### With this PR 

| rvu | rvuefu | start dev server |
| - | - | - |
| false | false | true |
| true | false | true |
| false | true | true |
| true | true | true |

PR #5378, opened 2025-01-03
chore(ci): Use Bookworm for "Build Release Tarball" stage

big-r81:
Use the latest Debian version for building the release tarball.


nickva (approved):
+1 if CI passes

PR #5381, opened 2025-01-05
docs/replication: change unauthorized example to 401

alxndrsn:
This looks more appropriate...


big-r81 (approved):
+1

Refering to the description above the code block, it makes more sense! Thx!

PR #5385, opened 2025-01-07
introduce acc record

rnewson:
refactor fabric_doc_update to use a record for accumulator, to make https://github.com/apache/couchdb/pull/5371 enhancement easier.

### DIFF
--- a/src/fabric/src/fabric_doc_update.erl
+++ b/src/fabric/src/fabric_doc_update.erl
@@ -17,6 +17,14 @@
 -include_lib("mem3/include/mem3.hrl").
 -include_lib("couch/include/couch_db.hrl").
 
+-record(acc, {
+    waiting_count,
+    doc_count,
+    w,
+    grouped_docs,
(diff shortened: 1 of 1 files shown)


janl (approved):

src/fabric/src/fabric_doc_update.erl
```diff
@@ -58,28 +72,34 @@ go(DbName, AllDocs0, Opts) ->
         rexi_monitor:stop(RexiMon)
     end.
 
-handle_message({rexi_DOWN, _, {_, NodeRef}, _}, _Worker, Acc0) ->
```
pgj, line 61: Should not the code still be able to handle messages with the tuple-shaped accumulators?  Such messages can arrive during upgrade where the cluster has mixed version of nodes.

rnewson: the acc is local, the first clause in handle_message is the message term from other nodes

PR #5412, opened 2025-01-23
Extract the response of `_versions` endpoint

jiahuili430:
## Overview

When running with old version of Clouseau, it will return `{ok, null}`
when we call `clouseau_rpc:version().`. So add `is_binary/1` to prevent
`badarg` errors. Also simplify the code by extracting the response to
`get_versions/0` function.

## Checklist


jaydoane (approved):
Nice!

PR #5432, opened 2025-02-15
Basic rework

AstraLuma:
## Overview

Give the mango system its own conceptual docs, instead of just a pile of API references.

Discussed it a bit in the mailing list.

## Checklist


src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,794 @@
...
+.. distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+.. WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+.. License for the specific language governing permissions and limitations under
+.. the License.
+
+=============
+Mango Queries
+=============
+
+In addition to :ref:`map/reduce views <viewfun>`, CouchDB supports a expressive
```
big-r81, line 17: s/a/an

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,794 @@
...
+query syntax called Mango.
+
+The mango system consists of two major concepts:
+
+* Selectors, which are the queries, and are passed to the mango endpoints
+* Indexes, which are a specialization of design docs used in mango queries
```
big-r81, line 23:
```suggestion
* Selectors, which are the queries, are passed to the mango endpoints and
* Indexes, which are a specialization of design docs used in mango queries
```

AstraLuma:
I'm currently rediscovering how mango and nouveau interact.

I'm kinda just ignoring mango-croustau, partly because I can barely find docs on how to set up croustau, and partly because documenting nouveau indexes is already a bunch of work.

src/docs/src/ddocs/mango.rst
AstraLuma: I personally don't like this as a writing style. Certainly not without punctuation.

AstraLuma:
I'm more-or-less done with the writing here, so I'm calling this ready for review.

I'll create a PR for the `3.4.x` and `fdbmain` branches after editing's finished.

src/docs/src/ddocs/index.rst
```diff
@@ -26,6 +26,7 @@ replications.
 
     ddocs
     views/index
+    mango
```
willholley, line 29: I'm not convinced this documentation should live under the "design documents" documentation. Mango does use design documents under the hood but it's an implementation detail - users shouldn't be manually editing / interpreting them. 

Additionally, the `_find`, `_explain` and `_index` endpoints are distinct from design document endpoints, unlike `_view`/`_search`/`_nouveau` which sit underneath the `<db>/_design/<ddoc_name>` URL root.

willholley: Perhaps this is as simple as breaking out a "querying data" topic from the "design documents" guide.

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,922 @@
...
+of function or result.
+
+.. warning::
+
+    While CouchDB will happily store just about anything JSON, Mango has
+    limitations about what it can work with:
+
+    * Empty field names (``""``) cannot be queried ("One or more conditions is
+      missing a field name.")
+    * Field names starting with ``$`` cannot be queried ("Invalid operator: $")
```
willholley, line 51: fields beginning with `$` can be queried by escaping it with a double slash e.g.
```json
{
  "selector": {
    "\\$foo": "bar"
  }
}
```

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,922 @@
...
+    * Fields at the root of the document starting with ``_`` cannot be queried
```
willholley, line 52: This may be referring to something else, but it is possible to query fields in the root of the document that begin with an underscore. For example:
```json
{
  "selector": {
    "_id":  "foo"
  }
}
```

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,922 @@
...
+    ]
+
+You can create more complex selector expressions by combining operators.
+For best performance, it is best to combine 'combination' or
+'array logical' operators, such as ``$regex``, with an operator
+that defines a contiguous range of keys such as ``$eq``,
+``$gt``, ``$gte``, ``$lt``, ``$lte``, and ``$beginsWith``
+(but not ``$ne``). For more information about creating complex
+selector expressions, see :ref:`creating selector expressions
+<find/expressions>`.
```
willholley, line 89: noting that this note on performance applies only to `json` index types only

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,922 @@
...
+the same syntax).
+
+.. code-block:: javascript
+
+    {
+        "_id": { "$gt": null },
+        "cameras": {
+            "$keyMapMatch": {
+                "$eq": "secondary"
+            }
```
willholley, line 560: this code example looks copy/pasted from the previous one and should show a `$text` operator instead.

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,922 @@
...
+    filter large data sets. They can, however, be used to restrict a
+    :ref:`partial index <find/partial_indexes>`.
+
+.. _find/expressions:
+
+Creating Selector Expressions
+-----------------------------
+
+We have seen examples of combining selector expressions, such as :ref:`using
+explicit $and and $eq operators <find/combination_operators>`.
```
willholley, line 670:
```suggestion
explicit ``$and`` and ``$eq`` operators <find/combination_operators>`.
```

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,922 @@
...
+* **fields**: ``"all_fields"`` or list of objects:
+
+  * **name** (`string`): not blank
+  * **type** (`string`): one of ``"text"``, ``"string"``, ``"number"``, ``"boolean"``
+
+* **default_analyzer** (`string`): Nouveau analyzer to use, defaults to ``"keyword"``
+* **default_field**: boolean or object of ``enabled`` and ``analyzer``
+* **partial_filter_selector** (`object`): A :ref:`selector<find/selectors>`, causing this
+  to be a :ref:`partial index<find/partial_indexes>`
+* **selector** (`object`): A :ref:`selector<find/selectors>`
```
willholley, line 913: This description is the same for text type indexes, so we could duplicate it as another section. [Cloudant's documentation](https://cloud.ibm.com/docs/Cloudant?topic=Cloudant-query#creating-a-type-text-index) has a fuller description around the default values and how Mango translates fields/queries into Lucene.

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,922 @@
...
+=============
+Mango Queries
+=============
+
+In addition to :ref:`map/reduce views <viewfun>`, CouchDB supports an expressive
+query system called Mango.
+
+Mango consists of two major concepts:
+
+* Selectors, which are the queries, and are passed to the mango endpoints
```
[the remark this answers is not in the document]
AstraLuma: Only the `_find` endpoint does this, and while that's an important feature, Selectors also appear in other contexts (namely partial indexes)

src/docs/src/ddocs/mango.rst
AstraLuma: see, this is why these docs needed to be written.

Anything else need escaping?

src/docs/src/ddocs/mango.rst
AstraLuma: You can query the document special fields, like `_id` and `_rev`.

CouchDB can also store things like `_` or `_jungle`, but Mango will refuse to query them.

src/docs/src/ddocs/mango.rst
AstraLuma: Ok, I copied this bit, so I'll just delete it.

src/docs/src/ddocs/index.rst
AstraLuma: I won't say you're wrong, but that sounds like scope creep, and as an external contributor, I would quite like to get this PR finished.

src/docs/src/ddocs/mango.rst
AstraLuma: Oh, yeah, I did copy/paste this and forgot to update the example.

src/docs/src/ddocs/mango.rst
AstraLuma: I'll try it, but links aren't always happy to have code bits in them.

src/docs/src/ddocs/mango.rst
AstraLuma: Ok, I'll update the language to refer to both.

src/docs/src/ddocs/mango.rst
AstraLuma: Yeah, this doesn't work.

janl (commented):
This is an outstanding contribution, thank you very much!

I left a few nitpicks that should be easy to sort out.

On the “where to put this” question, I suggest we rename the “3. Design Documents” section to “3. Querying CouchDB” and sort mango into this as proposed. I suggest we keep the `ddoc/` path and only change the title, so links don’t break.
 
I’m happy to append this last commit (if y’all agree) and if @AstraLuma is busy otherwise.

src/docs/src/ddocs/mango.rst
AstraLuma: Oops, that must have been a mistake in my experimentation. Going back, it looks like you're right.

src/docs/src/ddocs/mango.rst
```diff
@@ -0,0 +1,913 @@
...
+
+.. code-block:: javascript
+
+    {
+        "imdb": {
+            "rating": { "$eq": 8 }
+        }
+    }
+
+An example of the ``$eq`` operator used with full text indexing:
```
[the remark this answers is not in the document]
AstraLuma: I have absolutely no idea how this invokes full text indexing.

I traced it back to the original docs repo, looks like it was copied from Cloudant.

AstraLuma:
Yeah, a simple title change is perfectly doable.

I just don't want a massive refactor bolted onto this already large refactor.

src/docs/src/ddocs/mango.rst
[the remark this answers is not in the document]
janl: heh, let’s take it out here and leave this for the search section further down

janl:
> I just don't want a massive refactor bolted onto this already large refactor.

I think what I propose is fine for what we need. It’ll be better than the status quo due to your contribution. I also don‘t want to put up a big barrier and rather get this merged.

janl (approved):

janl:
@AstraLuma thanks loads for this! I just rebased and resolved conflicts but the kudos is all yours!

AstraLuma:
:tada:

PR #5472, opened 2025-03-11
dev(CI): Temporary exclude Nouveau tests on Windows

big-r81:
Exclude Nouveau tests on Windows, because of flaky and failing tests on Windows CI runs.


nickva (approved):
+1 thanks, Ronny

It's not ideal but at least we'll still have the majority of windows tests running

big-r81:
> +1 thanks, Ronny
> 
> It's not ideal but at least we'll still have the majority of windows tests running

Unfortunately no, we are running only the Mango tests for now. But if CouchDB gets compiled and a Windows installer is created, it's a huge win win ...

PR #5488, opened 2025-03-25
chore(`clouseau`): bump to 2.25.0

pgj:
We have had a couple of new Clouseau releases in the past few months.  So catch up with them and use the latest version as both the default for the `configure` script and implicitly for the CI.

## Testing recommendations

Other than exercising the change in the CI, one could manually verify it.  Configure with Clouseau enabled (UNIX-like systems, including macOS) and be sure to [have Java 8 on the path](https://github.com/apache/couchdb/blob/main/README-DEV.rst#running-clouseau):

```shell
./configure --with-clouseau
```

or on Windows:

```shell
.\configure.ps1 -WithClouseau
```

Note that you may have to remove any older local installation of Clouseau before proceeding:

```shell
rm -rf clouseau
```

Run the Mango tests (implies using Clouseau):

```shell
make mango-test
```

Note that there shall be only a single skipped test in this scenario.  Anything more than that may signal that Clouseau has not been fully configured.

```console
Ran 389 tests in 40.212s

OK (skipped=1)
```

Run the Elixir search tests (implies using Clouseau):

```shell
make elixir-search
```

This should all work with Erlang/OTP 25, 26, and 27.

To verify if logging is still working, you can take a look at the contents of the `dev/logs/clouseau1.log` file after Clouseau has been started up alongside CouchDB, e.g.:

```shell
dev/run -a adm:pass -n 1 --with-clouseau
```

For the full CI run (with all the builders where Clouseau is enabled), please see [`jenkins-clouseau-2.25.0` #12](https://ci-couchdb.apache.org/job/jenkins-cm1/job/FullPlatformMatrix/job/jenkins-clouseau-2.25.0/12/pipeline-graph/).

## Checklist


big-r81:
Hey Gábor, to run the full CI, you need a branch name starting with `jenkins-...` iirc...

pgj:
Yes, you are right.  That is my plan forward, but first I wanted to make sure that it passes the basic tests.

pgj:
I see [this for Windows](https://ci-couchdb.apache.org/job/jenkins-cm1/job/FullPlatformMatrix/job/jenkins-clouseau-2.25.0/1/pipeline-console/log?nodeId=509):

```
[2025-03-25T10:36:18.253Z] <ERROR> get_cookie: can't open cookie fileerl_call: can't create C node c17@WIN-F71NFGBB2HR; 0
[2025-03-25T10:36:18.253Z] [ * ] Check Clouseau node clouseau1 ... failed: Command '['erl_call', '-n', 'node1@127.0.0.1', '-a', "net_adm ping ['clouseau1@127.0.0.1']"]' returned non-zero exit status 1.
```

Could this be an issue specific to the Windows builder?

big-r81:
Is it possible that Clouseau isn't loaded at this stage, so that the "ping" failed?

How does Clouseau gets startet? We only have `$env:CLOUSEAU_JAVA_HOME = "${toolsDir}\${java8Build}"` as var to point to a jdk8. In the normal path, there is a Java21 linked, see [bin/shell.ps1](https://github.com/apache/couchdb-glazier/blob/main/bin/shell.ps1) and [bin/variables.ps1](https://github.com/apache/couchdb-glazier/blob/main/bin/variables.ps1)

pgj:
> Is it possible that Clouseau isn't loaded at this stage, so that the "ping" failed?

I am not sure about that.  To me, the error message above suggests that `erl_call` itself cannot establish connection to the Clouseau node because it was not able to create the missing Erlang cookie.  For what it is worth, [Clouseau always uses the `HOME` environment variable to determine the location of the cookie](https://github.com/cloudant-labs/clouseau/blob/master/src/main/scala/scalang/Node.scala), but [`erl_call` uses the combined value of the `HOMEDRIVE` and `HOMEPATH` on Windows](https://github.com/erlang/otp/blob/7bbd03d2ee872c13e3cc0104b87444f3ead3607b/lib/erl_interface/src/connect/ei_connect.c#L2835-L2852).

> How does Clouseau gets started? We only have `$env:CLOUSEAU_JAVA_HOME = "${toolsDir}\${java8Build}"` as var to point to a jdk8.

I believe that should be sufficient.  This is out of my concern at the moment.

pgj:
Fixing the CI for the FreeBSD, macOS, and Windows workers getting more complicated.  I will just drop those commits from the PR so we could just safely merge the bump itself.

pgj:
The Windows build has been fixed in #5541 so we could now test this Clouseau upgrade everywhere reliably.

nickva (approved):
very nice

+1 if pull request and full jenkins CI passes

pgj:
This is the link for the full Jenkins run: https://ci-couchdb.apache.org/job/jenkins-cm1/job/FullPlatformMatrix/job/jenkins-clouseau-2.25.0/12/pipeline-graph/

PR #5490, opened 2025-03-25
chore(`ci`): enable Clouseau for macOS

pgj:
The macOS worker has Java 8 installed but it could not be leveraged because it was not configured through the `CLOUSEAU_JAVA_HOME` environment variable.  Address this gap and un-break running Clouseau on that platform as well.


nickva (approved):
+1 if full jenkins CI passes

PR #5539, opened 2025-05-12
Extend `remsh` to read cookie from various sources

jiahuili430:
## Overview

We can set cookies using `./dev/run --erlang-cookie=crumbles` or `vm.args`: `-setcookie crumbles`.

If neither of them is set or we set the cookie in both places, then the Erlang node will read the cookie from `~/.erlang.cookie`.

Add this mechanism to the `remsh` script to extend flexibility.

Related PR: https://github.com/apache/couchdb/pull/5531

## Testing recommendations

```bash
# After the change it should pass
./dev/run --admin=adm:pass -n 1 --erlang-cookie=crumbles --no-eval 'dev/remsh'
```

## Checklist

### DIFF
--- a/dev/remsh
+++ b/dev/remsh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not
 # use this file except in compliance with the License. You may obtain a copy of
 # the License at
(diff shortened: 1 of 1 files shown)


dev/remsh
```diff
@@ -11,18 +11,48 @@
...
+
+if [[ -n "$COOKIE_PY" && -z "$COOKIE_VM" ]]; then
+  erl -name "$NAME" -remsh "$NODE" -hidden -setcookie "$COOKIE_PY"
+elif [[ -z "$COOKIE_PY" && $(echo "$COOKIE_VM" | wc -l) -eq 1 ]]; then
+  erl -name "$NAME" -remsh "$NODE" -hidden -setcookie "$COOKIE_VM"
+else
+  erl -name "$NAME" -remsh "$NODE" -hidden
+fi
+
+erl -name "$NAME" -remsh "$NODE" -hidden
```
big-r81, line 58: Is this necessary? We have it in the "else" branch already?

big-r81:
Thanks for looking into it, but wouldn't it be better to have only one [remsh` script](https://github.com/apache/couchdb/blob/main/rel/overlay/bin/remsh), which can be used for 'dev' and 'prod'?

PR #5541, opened 2025-05-14
chore(`ci`): enable Clouseau for Windows

pgj:
Implement the `with_clouseau` flag for the Windows worker.  The `elixir-search` target could be safely run unconditionally, the `Makefile` is going to skip the tests when Clouseau is not present.

At the moment, sharing the Erlang cookie between the CouchDB and Clouseau nodes for orchestrating the tests does not work in the current setup.  By default, we rely on the `.erlang.cookie` file in the user's home directory, but this file could only be opened exclusively.  Work this around by setting the cookie string through the command line.

For the full CI run (as a verification), please see [`jenkins-clouseau-windows` #16](https://ci-couchdb.apache.org/job/jenkins-cm1/job/FullPlatformMatrix/job/jenkins-clouseau-windows/16/pipeline-console/).

### DIFF
--- a/build-aux/Jenkinsfile.full
+++ b/build-aux/Jenkinsfile.full
@@ -134,6 +134,7 @@ meta = [
   'win2022': [
     name: 'Windows 2022',
     spidermonkey_vsn: '128',
+    with_clouseau: true,
     node_label: 'win'
   ]
 ]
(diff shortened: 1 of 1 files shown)


nickva (approved):
+1 nicely done!

PR #5542, opened 2025-05-16
Improve `couch_debug.erl`

jiahuili430:
## Overview

- Fix `mem3:ping/1` function clause error

Error log:

```
** exception error: no function clause matching mem3:ping_nodes(['node2@127.0.0.1']) (src/mem3.erl, line 453)
```

- Improve `couch_debug.erl`
  - Use `module_info(exports)` instead of hardcoding.
  - Export `print_tree/2` as it's already documented in `couch_debug:help/1` function.
  - Add a new line after each resource hoggers snapshot table for better viewing.
  - Add `couch_debug:ping_live_cluster_nodes/0`, and `ping_live_cluster_nodes/1`.

## Checklist


src/couch/src/couch_debug.erl
```diff
@@ -75,29 +79,7 @@
...
-        print_table,
-        print_report,
-        print_report_with_info_width,
-        restart,
-        restart_busy,
-        dead_nodes,
-        ping_nodes,
-        node_events
-    ].
+    lists:uniq([F || {F, _A} <- couch_debug:module_info(exports)]) -- [help, module_info].
```
iilyak, line 82: Not every exported function has help. The idea of help function is to only return functions which can be called as `help(FunName)`. This is intended for higher level tools like admin GUI (not yet exists). The left panel would display list of topics (result of `help()`. call). By selecting one of the items the `help(SelectedItem)` is executed to display the help for that function. This change wold break this idea.

iilyak (commented):
+1 after reverting the `help()` change

iilyak (approved):
+1

PR #5546, opened 2025-05-23
Improve `clouseau_rpc.erl`

jiahuili430:
## Overview

Add `check_service/1` `check_services/0` to check if clouseau services are alive.

## Testing recommendations

Run CouchDB with Clouseau and run the following commands in remsh:
```bash
(node1@127.0.0.1)1> clouseau_rpc:check_services().
[{sup,alive},{main,alive},{analyzer,alive},{cleanup,alive}]
(node1@127.0.0.1)2> clouseau_rpc:check_service(init).
** exception throw: {not_found,<<"no such service: init">>}
     in function  clouseau_rpc:check_service/1 (src/clouseau_rpc.erl, line 363)
(node1@127.0.0.1)3> clouseau_rpc:check_service(main).
{main,alive}
```

Run CouchDB only:
```bash
(node1@127.0.0.1)4> clouseau_rpc:check_services().
** exception throw: {noconnection,<<"Clouseau node is not connected.">>}
     in function  clouseau_rpc:check_services/0 (src/clouseau_rpc.erl, line 378)
```

## Checklist

### DIFF
--- a/src/dreyfus/src/clouseau_rpc.erl
+++ b/src/dreyfus/src/clouseau_rpc.erl
@@ -22,7 +22,7 @@
 -export([delete/2, update/3, cleanup/1, cleanup/2, rename/1]).
 -export([analyze/2, version/0, disk_size/1]).
 -export([set_purge_seq/2, get_purge_seq/1, get_root_dir/0]).
--export([connected/0]).
+-export([connected/0, check_service/1, check_services/0]).
 
 %% string represented as binary
 -type string_as_binary(_Value) :: nonempty_binary().
@@ -31,6 +31,7 @@
 -type path() :: string_as_binary(_).
 
 -type error() :: any().
+-type throw(_Reason) :: no_return().
 
 -type analyzer_name() :: string_as_binary(_).
 
@@ -63,6 +64,8 @@
         | {string_as_binary(fields), {analyzer_fields()}}
         | {string_as_binary(stopwords), [field_name()]}
     ].
+
+-define(TIMEOUT, 200).
 
 -spec open_index(Peer :: pid(), Path :: shard(), Analyzer :: analyzer()) ->
     {ok, indexer_pid()} | error().
@@ -285,6 +288,21 @@ analyze(Analyzer, Text) ->
 version() ->
     rpc({main, clouseau()}, version).
 
+-spec clouseau_major_vsn() -> binary() | throw({atom(), binary()}).
+clouseau_major_vsn() ->
+    case version() of
+        {ok, <<Major, _/binary>>} ->
+            <<Major>>;
+        {'EXIT', noconnection} ->
+            throw({noconnection, <<"Clouseau node is not connected.">>});
+        {ok, null} ->
+            %% Backward compatibility:
+            %% If we run Clouseau from source code, remsh will return
+            %% `{ok, null}` for Clouseau <= 2.25.0.
+            %% See PR: https://github.com/cloudant-labs/clouseau/pull/106
+            <<$2>>
+    end.
+
 -spec connected() -> boolean().
 
 connected() ->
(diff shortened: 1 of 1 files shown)


src/dreyfus/src/clouseau_rpc.erl
```diff
@@ -316,3 +318,25 @@ rpc(Ref, Msg) ->
...
 clouseau() ->
     list_to_atom(config:get("dreyfus", "name", "clouseau@127.0.0.1")).
+
+-spec services() -> ok | error().
+services() ->
+    Services = [sup, main, analyzer, cleanup],
+    case connected() of
+        true ->
+            AliveServices = [is_alive(S) || S <- Services],
+            io:fwrite("Clouseau ervices: ~p~n", [AliveServices]);
```
big-r81, line 328: s/ervices/services

src/dreyfus/src/clouseau_rpc.erl
```diff
@@ -316,3 +326,49 @@ rpc(Ref, Msg) ->
...
 clouseau() ->
     list_to_atom(config:get("dreyfus", "name", "clouseau@127.0.0.1")).
+
+-type service() :: sup | main | analyzer | cleanup.
+-type liveness_status() :: alive | timeout.
+
+-spec clouseau_services(MajorVsn) -> [service()] when
+    MajorVsn :: binary() | null.
+clouseau_services(MajorVsn) ->
+    case MajorVsn of
```
pgj, line 336: Should we have this case distinction given that both branches are the same at the moment?  I think it is fine to receive the major version as a parameter, but I am not sure if that knowledge should be utilized now.

src/dreyfus/src/clouseau_rpc.erl
```diff
@@ -285,6 +287,14 @@ analyze(Analyzer, Text) ->
 version() ->
     rpc({main, clouseau()}, version).
 
+-spec clouseau_major_vsn() -> null | binary() | no_return().
+clouseau_major_vsn() ->
+    case version() of
+        {ok, <<Major, _/binary>>} -> <<Major>>;
+        {ok, null} -> null;
+        {'EXIT', noconnection} -> throw("Clouseau is not connected yet!")
```
pgj, line 295: The word "yet" sounds a bit odd.  It sounds like the reason for getting `noconnection` is because we have not yet configured the connection to Clouseau, while it might be happening because we have lost the connection to Clouseau that we had in the past.  Therefore it may make sense to lose the "yet" here.

src/dreyfus/src/clouseau_rpc.erl
```diff
@@ -285,6 +287,14 @@ analyze(Analyzer, Text) ->
 version() ->
     rpc({main, clouseau()}, version).
 
+-spec clouseau_major_vsn() -> null | binary() | no_return().
+clouseau_major_vsn() ->
+    case version() of
+        {ok, <<Major, _/binary>>} -> <<Major>>;
+        {ok, null} -> null;
```
iilyak, line 294: Is there a case where the response is `null`?

src/dreyfus/src/clouseau_rpc.erl
```diff
@@ -316,3 +326,46 @@ rpc(Ref, Msg) ->
...
+    MajorVsn :: binary() | null.
+clouseau_services(_MajorVsn) ->
+    [sup, main, analyzer, cleanup].
+
+-spec is_valid(Service) -> boolean() when
+    Service :: atom().
+is_valid(Service) when is_atom(Service) ->
+    lists:member(Service, clouseau_services(clouseau_major_vsn())).
+
+-spec check_service(Service) -> {service(), liveness_status()} | no_return() when
```
iilyak, line 343:
```erlang
| throw({not_found, binary()})
```

BTW earlier the error in throw was string, but here it is binary.

src/dreyfus/src/clouseau_rpc.erl
jiahuili430: Thanks to Jay for reminding me of this.
If we run clouseau in dev mode, it will return `null`.
```bash
cd clouseau
mvn scala:run -Dname=clouseau1 -Dlauncher=with-cookie -Dcookie=<cookie>
```

iilyak: > If we run clouseau in dev mode, it will return null.

Please add a comment about it.

pgj: Sorry, what "dev mode" are we talking about here?

jaydoane: > Sorry, what "dev mode" are we talking about here?

When it's run via e.g. `make clouseau1`, and not from a jar file as it does in production.

iilyak: > > If we run clouseau in dev mode, it will return null.
> 
> Please add a comment about it.

Or change return value from `null` to something self explanatory.

```
 {ok, null} -> dev_mode;
 {ok, null} -> edge;
 {ok, null} -> dev;
 {ok, null} -> development;
 {ok, null} -> experimental;
```

pgj: Albeit I acknowledge this is going to stay with us, so the `null` case should be supported somehow.  For that, I like @iilyak's idea.

src/dreyfus/src/clouseau_rpc.erl
jiahuili430: Updated all throwing functions to `throw({atom(), binary()})`.

src/dreyfus/src/clouseau_rpc.erl
```diff
@@ -285,6 +288,21 @@ analyze(Analyzer, Text) ->
...
+        {ok, <<Major, _/binary>>} ->
+            <<Major>>;
+        {'EXIT', noconnection} ->
+            throw({noconnection, <<"Clouseau node is not connected.">>});
+        {ok, null} ->
+            %% Backward compatibility:
+            %% If we run Clouseau from source code, remsh will return
+            %% `{ok, null}` for Clouseau <= 2.25.0.
+            %% Related PR: https://github.com/cloudant-labs/clouseau/pull/106
+            null
```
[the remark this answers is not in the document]
jiahuili430: Updated to `{ok, null} -> <<$2>>` and types, thank you for the review!

PR #5556, opened 2025-05-31
Add guards to `fabric:design_docs/1` to prevent function_clause error

jiahuili430:
## Overview

When a node is down, `fabric:design_docs/1` will return
`{ok, {nodedown, <<"progress not possible">>}}` instead
of a list of design docs. So added some guards for this
function.

In `ddoc_cache_entry_validation_funs.erl`, we cannot use
an empty list because we need to use the VDU function to
validate the document before updating it, so we raise an
error here.

Error log:
```
exit value:#012{
  {function_clause, [
    {
      lists, flatmap_1,
      [#Fun<ddoc_cache_entry_validation_funs.0.82671214>, {nodedown, <<"progress not possible">>}],
      [{file, "lists.erl"}, {line, 1578}]
    },
    {
      ddoc_cache_entry_validation_funs, recover, 1,
      [{file, "src/ddoc_cache_entry_validation_funs.erl"}, {line, 30}]
    },
    {
      couch_db, '-load_validation_funs/1-fun-0-', 1,
      [{file, "src/couch_db.erl"}, {line, 972}]
    }
  ]}
}#012
```

## Checklist

### DIFF
--- a/src/ddoc_cache/src/ddoc_cache_entry_validation_funs.erl
+++ b/src/ddoc_cache/src/ddoc_cache_entry_validation_funs.erl
@@ -26,7 +26,18 @@ ddocid(_) ->
     no_ddocid.
 
 recover(DbName) ->
-    {ok, DDocs} = fabric:design_docs(mem3:dbname(DbName)),
+    %% The VDU function is used to validate documents update before
+    %% storing them in the database.
+    %% Raise an error when invalid instead of returning an empty list.
+    DDocs =
+        case fabric:design_docs(mem3:dbname(DbName)) of
+            {ok, Resp} when is_list(Resp) ->
+                Resp;
+            {ok, Error} ->
+                error(Error);
+            {error, Error} ->
+                error(Error)
+        end,
     Funs = lists:flatmap(
         fun(DDoc) ->
             case couch_doc:get_validate_doc_fun(DbName, DDoc) of
--- a/src/dreyfus/src/dreyfus_fabric_cleanup.erl
+++ b/src/dreyfus/src/dreyfus_fabric_cleanup.erl
@@ -21,7 +21,14 @@
 -export([go/1]).
 
 go(DbName) ->
-    {ok, DesignDocs} = fabric:design_docs(DbName),
+    DesignDocs =
+        case fabric:design_docs(DbName) of
+            {ok, DDocs} when is_list(DDocs) ->
+                DDocs;
+            Else ->
+                couch_log:debug("Invalid design docs: ~p~n", [Else]),
+                []
+        end,
     ActiveSigs = lists:usort(
         lists:flatmap(
             fun active_sigs/1,
--- a/src/nouveau/src/nouveau_fabric_cleanup.erl
+++ b/src/nouveau/src/nouveau_fabric_cleanup.erl
@@ -22,7 +22,14 @@
 -export([go/1]).
 
 go(DbName) ->
-    {ok, DesignDocs} = fabric:design_docs(DbName),
+    DesignDocs =
+        case fabric:design_docs(DbName) of
+            {ok, DDocs} when is_list(DDocs) ->
+                DDocs;
+            Else ->
+                couch_log:debug("Invalid design docs: ~p~n", [Else]),
+                []
+        end,
     ActiveSigs =
         lists:usort(
             lists:flatmap(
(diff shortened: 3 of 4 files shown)


src/ddoc_cache/src/ddoc_cache_entry_validation_funs.erl
```diff
@@ -27,16 +27,21 @@ ddocid(_) ->
...
+                    case couch_doc:get_validate_doc_fun(DbName, DDoc) of
+                        nil -> [];
+                        Fun -> [Fun]
+                    end
+                end,
+                DDocs
+            ),
+            {ok, Funs};
+        false ->
+            throw({error, DDocs})
```
nickva: Raising an error in general seems more preferable unless we have some code somewhere already catching throws for recovery.

src/ddoc_cache/src/ddoc_cache_entry_validation_funs.erl
```diff
@@ -27,16 +27,21 @@ ddocid(_) ->
...
-        fun(DDoc) ->
-            case couch_doc:get_validate_doc_fun(DbName, DDoc) of
-                nil -> [];
-                Fun -> [Fun]
-            end
-        end,
-        DDocs
-    ),
-    {ok, Funs}.
+    case is_list(DDocs) of
```
nickva: I think we'd want to mention in a comment why we're doing this -- `all_docs()` and by extension `design_docs()` have slightly odd behavior that they can return an `{ok, Error}` tuple. Normally we should expect these things to return an `{error, Error}`.

This comes from https://github.com/apache/couchdb/blob/9da7be8e9660caa0e3c78e0cd36e180817bb827b/src/fabric/src/fabric_view_all_docs.erl#L154-L155 and I don't fully understand the reason for it

src/ddoc_cache/src/ddoc_cache_entry_validation_funs.erl
```diff
@@ -27,16 +27,21 @@ ddocid(_) ->
 
 recover(DbName) ->
     {ok, DDocs} = fabric:design_docs(mem3:dbname(DbName)),
```
nickva: For completeness we might as well expect an `{error, Error}` here and do the same raise error or throw we do like in the case of `{ok, Error}`.

That could come from https://github.com/apache/couchdb/blob/9da7be8e9660caa0e3c78e0cd36e180817bb827b/src/fabric/src/fabric_view_all_docs.erl#L43-L44 via https://github.com/apache/couchdb/blob/9da7be8e9660caa0e3c78e0cd36e180817bb827b/src/fabric/src/fabric.erl#L558-L559

nickva (changes requested):
Very nice work! Just a few minor tweaks needed.

It's also worth highlighting that you're doing the right thing here and we're just turning the function clause into a more appropriate error. Unlike we did for ken, we can't just skip the failing ddocs and return `[]` we have to error out.

src/ddoc_cache/src/ddoc_cache_entry_validation_funs.erl
jiahuili430: Updated to `{error, Resp}` here and add case statement to `fabric:design_docs/1`.
If we receive `{ok, DDocs}`, continue running, otherwise, return `[]` instead.

src/ddoc_cache/src/ddoc_cache_entry_validation_funs.erl
jiahuili430: Changed to `error({error, DDocs})` and added some comments about why we can't use `[]`.

src/ddoc_cache/src/ddoc_cache_entry_validation_funs.erl
```diff
@@ -26,7 +26,17 @@ ddocid(_) ->
...
+    %% The VDU function is used to validate documents update before
+    %% storing them in the database.
+    %% Raise an error when invalid instead of returning an empty list.
+    DDocs =
+        case fabric:design_docs(mem3:dbname(DbName)) of
+            {ok, Resp} when is_list(Resp) ->
+                Resp;
+            Else ->
+                couch_log:debug("Invalid design docs: ~p~n", [Else]),
+                error({error, Else})
```
nickva: This will technically raise `{error, {error, Error}}` or `{error, {ok, Error}}` I think? Maybe we can just handle both and just raise the Error part so it's just `error(<<"nodedown">>)`

jiahuili430: Updated, thank you for the review!

PR #5557, opened 2025-06-01
Update reverse-proxies.rst

rev111:
Corrected reverse proxying subdirectories with Caddy. Former description would not strip the subdirectory from the reverse proxy target. Changed documentation accordingly. Added an additional paragraph how stripping the subdirectory should be configured in Caddys configuration file. Source: Spent an afternoon figuring this out after unsuccesful application of this doc to a CouchDB+Caddy setup.

## Overview

Corrected documentation regarding reverse proxying subdirectories with Caddy. Former description would not strip the subdirectory from the reverse proxy target as described. Changed documentation accordingly. Added an additional paragraph how stripping the subdirectory should be configured in Caddys configuration file. Source: Spent an afternoon figuring this out after unsuccesful application of this doc to a CouchDB+Caddy setup.

## Checklist

- [ X] Documentation changes were made in the `src/docs` folder

### DIFF
--- a/src/docs/src/best-practices/reverse-proxies.rst
+++ b/src/docs/src/best-practices/reverse-proxies.rst
@@ -224,7 +224,7 @@ as ``http(s)://domain.com/couchdb/db1/doc1`` are proxied to
 To proxy the URL ``http(s)://domain.com/couchdb`` to
 ``http://localhost:5984`` so that requests appended to the subdirectory, such
 as ``http(s)://domain.com/couchdb/db1/doc1`` are proxied to
-``http://localhost:5984/db1/doc1``, you have to configure a route in Caddy:
+``http://localhost:5984/db1/doc1``, you have to configure a route in Caddy and strip the subdirectory from the target URI with ``uri strip_prefix``:
 
 .. code-block:: text
 
@@ -236,8 +236,6 @@ as ``http(s)://domain.com/couchdb/db1/doc1`` are proxied to
         }
 
     }
-
-
 
 Reverse proxying + load balancing for CouchDB clusters
 ------------------------------------------------------


src/docs/src/best-practices/reverse-proxies.rst
```diff
@@ -221,6 +221,24 @@ as ``http(s)://domain.com/couchdb/db1/doc1`` are proxied to
...
+    domain.com {
+
+        route /couchdb/* {
+            uri strip_prefix /couchdb
+            reverse_proxy localhost:5984        
+        }
+
+    }
+
+
```
big-r81: Please remove these two (out of three) empty lines.

big-r81 (changes requested):
Hello, thank you for your contribution. Only a small syntax change is necessary.

src/docs/src/best-practices/reverse-proxies.rst
rev111: Removed the lines and appended the text a bit. 
Also line 283 in chapter "Authentication with Caddy 2 as a reverse proxy" possibly contains a mistake when stating "This setup leans entirely on nginx performing...". I'm not sure if it is correct, or it should be "caddy" instead of "nginx", or if the paragraph has been erroneously copied from the identical paragraph in the nginx chapter at line 148 and was accidentally left, so I did not change it.

big-r81: Oh, good catch! Looks like a copy/paste error. We would be pleased if you could fix this error at the same time and while you have a working caddy setup, can you test/evaluate the provided config samples if these are correct?

src/docs/src/best-practices/reverse-proxies.rst
```diff
@@ -221,6 +221,22 @@ as ``http(s)://domain.com/couchdb/db1/doc1`` are proxied to
 
     }
 
+To proxy the URL ``http(s)://domain.com/couchdb`` to
+``http://localhost:5984`` so that requests appended to the subdirectory, such
+as ``http(s)://domain.com/couchdb/db1/doc1`` are proxied to
+``http://localhost:5984/db1/doc1``, you have to configure a route in Caddy and strip the subdirectory from the target URI with ``uri strip_prefix``:
```
big-r81, line 227: the linter complains about this long line. please add a line break (max. 90 chars).

PR #5558, opened 2025-06-04
Introduce peer checkpoints so CouchDB can safely remove deleted documents entirely

rnewson:
## Overview

Apache CouchDB retains some information (at minimum, doc id, doc revision tree and a deleted flag) for all deleted documents forever, in order that replication is guaranteed to converge. This is excessively pessimistic and we would like to improve matters.

This PR introduces a number of changes to achieve its goal;

1) database shards (`.couch` files under `shards/` directory) gained an additional header property called `drop_seq`. Once set to a positive, non-negative integer, any deleted document with a lower update sequence is skipped entirely at next compaction.

2) The notion of a `peer checkpoint document`. These are all local docs and their ids must have prefix `_local/peer-checkpoint-'.

3) All indexers (mrview, search, nouveau) and the replicator have been taught to create and update peer checkpoints with the update sequence they have seen at appropriate times (i.e, after they have made every effort to commit the changes they've seen to durable storage).

2) A new endpoint `POST /$dbname/_update_drop_seq` which gathers information about the shards of the database, update sequences from all peer checkpoint documents, and the internal shard sync documents, and computes the `drop_seq` for each shard, and then sends RPC requests to those databases to update the `drop_seq`.

## Testing recommendations

There are some simple tests in the eunit and elixir suites which will be run via the normal Makefile targets.

Additionally there is a stateful property-based test that exercises the code more comprehensively which can be started with `make elixir-cluster`. This will start a 3 node cluster with nouveau server running and perform random permutations of all relevant operations that could alter which deleted documents are dropped (making docs, deleting docs, creating indexes, creating and updating peer checkpoints, splitting shards).

N/A

## Checklist

- [TODO] Documentation changes were made in the `src/docs` folder


src/couch/src/couch_bt_engine.erl
```diff
@@ -809,6 +819,35 @@ set_update_seq(#st{header = Header} = St, UpdateSeq) ->
...
+                ]),
+                needs_commit = true
+            },
+            {ok, increment_update_seq(NewSt)}
+    end.
+
+increment_drop_count(#st{header = Header} = St, Inc) when is_integer(Inc), Inc >= 0 ->
+    CurrentDropCount = get_drop_count(St),
+    NewSt = St#st{
+        header = couch_bt_engine_header:set(Header, [{drop_count, CurrentDropCount + Inc}]),
```
iilyak, line 844: `get_drop_count/1` is implemented in terms of `couch_bt_engine_header:get/2` which uses `undefined` as default value. An attempt to do `undefined + Inc` would cause an exception. Which would terminate the compaction process. I don't know whether we can have a state where the `drop_count` is not present in the header when we are calling the `increment_drop_count`.

rnewson: thanks, have been focused on drop seq itself for too long and neglected upgrade implications. will fix.

src/couch_mrview/src/couch_mrview_index.erl
```diff
@@ -210,7 +215,17 @@ finish_update(State) ->
 
 commit(State) ->
     Header = {State#mrst.sig, couch_mrview_util:make_header(State)},
-    couch_file:write_header(State#mrst.fd, Header).
+    ok = couch_file:sync(State#mrst.fd),
```
iilyak: I understand why you do the sync here before and after. Although it feels like we are breaking the abstraction boundaries (and bypassing ioq). The decision to do `sync` belongs to couch application or more specifically to the storage engine. As we do in [`couch_bt_engine:commit_data/1`](https://github.com/apache/couchdb/blob/91bd2edb6406ddd39d735548774a384b0ec3a46f/src/couch/src/couch_bt_engine.erl#L563C1-L567C33). 
Unfortunately we don't have something like it for indexes. 
So I was thinking maybe we should update `couch_file` to keep the old header checksum in the state. The checksum is a prefix of the header so it is easy to get access to (we need to know the size of the checksum though). However maintaining the Checksum to be able to bypass the IO is just one side of the coin. The other is the need to wrap the write_header in two sync calls to do a commit. This is where I stopped my train of thought. Because we would need to be able to pass the `NeedsCommit` flag.

Maybe we should add a separate call to `couch_file`. We might have `couch_file:write_header/2` which we wouldn't do the sync (no change) and separate `couch_file:commit_header/2` which would do sync calls.

src/fabric/src/fabric_drop_seq.erl
```diff
@@ -0,0 +1,1016 @@
...
+            {ok, ExistingDoc} ->
+                exit(ExistingDoc#doc.revs);
+            {not_found, _Reason} ->
+                exit({0, []});
+            {error, Reason} ->
+                throw({checkpoint_fetch_failure, Reason})
+        end
+    end),
+    receive
+        {'DOWN', OpenRef, _, _, Revs} ->
```
iilyak: The shape of this message is the same as the one on line 530. You probably want to check the shape of Revs. Otherwise the `{checkpoint_fetch_failure, Reason}` would end up in the term.

```erlang
17> {_, OpenRef} = spawn_monitor(fun() -> throw(fff) end).
{<0.112.0>,#Ref<0.2499101716.2367684609.212273>}
=ERROR REPORT==== 18-Jun-2025::12:09:20.981098 ===
Error in process <0.112.0> with exit value:
{{nocatch,fff},[{shell,apply_fun,3,[{file,"shell.erl"},{line,914}]}]}

18> flush().
Shell got {'DOWN',#Ref<0.2499101716.2367684609.212273>,process,<0.112.0>,
                  {{nocatch,fff},
                   [{shell,apply_fun,3,[{file,"shell.erl"},{line,914}]}]}}
```

src/fabric/src/fabric_drop_seq.erl
```diff
@@ -0,0 +1,1016 @@
...
+        ]
+    ).
+
+cleanup_peer_checkpoint_docs(DbName, SubType, KeepSigs) when
+    is_binary(DbName), is_binary(SubType), is_list(KeepSigs)
+->
+    MrArgs = #mrargs{
+        view_type = map,
+        include_docs = true,
+        start_key = <<?LOCAL_DOC_PREFIX, "peer-checkpoint-", SubType/binary, "-">>,
```
iilyak: I know we have many places in CouchDB codebase which do something like this. However it makes it harder to read for newcomers who are often unfamiliar with the fact that we rely on ASCII ordering.

Would you mind adding a define or helper function (with inline) so it is self documenting. I didn't finish reading the PR, maybe there is a better place to define these helpers.

```erlang
% approach using define
-define LOWEST_KEY(SubType, <<?LOCAL_DOC_PREFIX, "peer-checkpoint-", SubType/binary, "-">>).

-define HIGHEST_KEY(SubType, <<?LOCAL_DOC_PREFIX, "peer-checkpoint-", SubType/binary, ".">>).

...
start_key = ?LOWEST_KEY(SubType),
end_key = ?HIGHEST_KEY(SubType),
```

```erlang
% approach with inline function
-compile({inline, [
    start_key/1,
    end_key/1
]}).
start_key(SubType) -> 
    <<?LOCAL_DOC_PREFIX, "peer-checkpoint-", SubType/binary, "-">>.

end_key(Partition) -> 
    <<?LOCAL_DOC_PREFIX, "peer-checkpoint-", SubType/binary, ".">>.

...
start_key = start_key(SubType),
end_key = end_key(SubType),

```

src/couch/src/couch_db_split.erl
```diff
@@ -319,17 +319,28 @@ copy_meta(#state{source_db = SourceDb, targets = Targets} = State) ->
...
+            {ok, Db4} = couch_db_engine:set_drop_seq(Db3, Uuid, DropSeq),
+            {ok, Db5} = couch_db_engine:increment_drop_count(Db4, DC),
+            {Acc#{K => T#target{db = Db5}}, DCRest}
         end,
+        {#{}, spread(DropCount, map_size(Targets))},
         Targets
     ),
     State#state{targets = Targets1}.
 
+spread(Amount, N) when is_integer(Amount), Amount >= 0, is_integer(N),  N > 0 ->
```
iilyak: Really nice.

src/fabric/src/fabric_drop_seq.erl
```diff
@@ -0,0 +1,1016 @@
...
+    PeerCheckpoints2 = crossref(PeerCheckpoints1, ShardSyncHistory),
+    ShardSyncCheckpoints = latest_shard_sync_checkpoints(ShardSyncHistory),
+    maps:merge_with(fun merge_peers/3, PeerCheckpoints2, ShardSyncCheckpoints).
+
+handle_set_drop_seq_reply(ok, Worker, {Results0, Waiting}) ->
+    DropSeq = proplists:get_value(drop_seq, Worker#shard.opts),
+    [B, E] = Worker#shard.range,
+    BHex = couch_util:to_hex(<<B:32/integer>>),
+    EHex = couch_util:to_hex(<<E:32/integer>>),
+    Range = list_to_binary([BHex, "-", EHex]),
```
iilyak: There is an optimized version of `to_hex` which avoids extra conversion.

```erlang
BHex = couch_util:to_hex_bin(<<B:32/integer>>),
EHex = couch_util:to_hex_bin(<<E:32/integer>>),
<<BHex/binary,"-",EHex/binary>>,
```

src/fabric/src/fabric_drop_seq.erl
```diff
@@ -0,0 +1,1016 @@
...
+        ),
+        shard_sync_history => maps:merge(
+            maps:get(shard_sync_history, Info), maps:get(shard_sync_history, Acc)
+        )
+    }.
+
+merge_peers(_Key, {Uuid1, Val1}, {Uuid2, Val2}) when
+    is_binary(Uuid1), is_binary(Uuid2), is_integer(Val1), is_integer(Val2)
+->
+    true = uuids_match([Uuid1, Uuid2]),
```
iilyak: the `merge_peers` is used in many `maps:merge_with` invocations. All of them would crash if `uuids_match/2` result is not `true`. Is it intentional (like an assertion)?

src/couch_mrview/src/couch_mrview_index.erl
rnewson: the "pluggable storage engine" abstraction has only ever applied to primary data. And, at this stage, I think we would remove the abstraction rather than extend it to indexes, so I don't agree that "The decision to do sync belongs to couch application or more specifically to the storage engine".

I'd prefer not to have the file:sync the .mrview files at all but it seems the simplest way to be as sure as we realistically can be that the index updates are durable (and thus allowing us to permanently drop deleted documents that have already been applied to it). 

In nouveau this is achieved differently as index commits there have to be infrequent for performance reasons. I just don't see any easy way to do the same approach for mrview.

src/fabric/src/fabric_drop_seq.erl
rnewson: it is an assertion, yes. (to guard against bugs in my code if I had a mix up, the uuids serve as a validation that I'm talking about the right shard file)

src/fabric/src/fabric_drop_seq.erl
rnewson: yes, I like that.

src/couch_mrview/src/couch_mrview_index.erl
iilyak: >I'd prefer not to have the file:sync the .mrview files at all but it seems the simplest way to be as sure as we realistically can be that the index updates are durable

Adding `couch_file:commit_header/2` seems easy and it wouldn't bypass `ioq`.

src/fabric/src/fabric_drop_seq.erl
iilyak: Could you add comment. 

```
% Assert we are talking about the right shards file...
% ...
true = uuids_match([Uuid1, Uuid2]),
```

or maybe use `?assert` macro.

```erlang
-include_lib("stdlib/include/assert.hrl").

merge_peers(_Key, {Uuid1, Val1}, {Uuid2, Val2}) when
    is_binary(Uuid1), is_binary(Uuid2), is_integer(Val1), is_integer(Val2)
->
    ?assert(uuids_match([Uuid1, Uuid2]), "UUIDs belong to different shard files"),
```

nouveau/src/main/java/org/apache/couchdb/nouveau/NouveauApplicationConfiguration.java
```diff
@@ -24,7 +24,7 @@ public class NouveauApplicationConfiguration extends Configuration {
     @Min(10)
     private int maxIndexesOpen = 10;
 
-    @Min(10)
+    @Min(5)
```
janl, line 27: does this have to match the integer in the next line?

src/couch/src/couch_bt_engine.erl
```diff
@@ -809,6 +819,35 @@ set_update_seq(#st{header = Header} = St, UpdateSeq) ->
...
+set_drop_seq(#st{header = Header} = St, ExpectedUuidPrefix, NewDropSeq) when
+    is_binary(ExpectedUuidPrefix), is_integer(NewDropSeq), NewDropSeq >= 0
+->
+    CurrentDropSeq = get_drop_seq(St),
+    Uuid = get_uuid(St),
+    ActualUuidPrefix = binary:part(Uuid, 0, byte_size(ExpectedUuidPrefix)),
+    if
+        ExpectedUuidPrefix /= ActualUuidPrefix ->
+            {error, uuid_mismatch};
+        NewDropSeq < CurrentDropSeq ->
```
janl, line 829: this implies we are okay with setting drop_seq to the same value again, in which case we can save ourselves writing a new header maybe?

src/couch/src/couch_db.erl
```diff
@@ -796,6 +807,10 @@ set_security(_, _) ->
 set_user_ctx(#db{} = Db, UserCtx) ->
     {ok, Db#db{user_ctx = UserCtx}}.
 
+set_drop_seq(#db{main_pid = Pid} = Db, UuidPrefix, DropSeq) ->
+    check_is_admin(Db),
```
janl, line 811: does this inherit its user context from the _calculate_drop_seq HTTP API call? Do we need to adjust this if get here via ken?

test/elixir/test/drop_seq_test.exs
```diff
@@ -0,0 +1,248 @@
...
+          range: r
+        })
+        assert resp.status_code == 201
+      end)
+      wait_for_reshard_jobs_to_complete()
+    end
+
+    after_doc_deletion_fn = fn ->
+      split_all_shard_ranges(db_name)
+      wait_for_reshard_jobs_to_complete()
```
janl, line 129: dupe or belt and suspenders?

src/couch/src/couch_bt_engine.erl
rnewson: we call `upgrade` on the header as we open so drop_seq/count get filled in as `0`. Experimentally I have confirmed this with 1) on `main`, create a database 2) switch to this branch 3) `GET /dbname` shows `drop_count` of `0` (and not a crash from trying to do `undefined + undefined`.

src/couch_mrview/src/couch_mrview_index.erl
rnewson: I don't understand your bypassing ioq point or how moving this to a new function in couch_file would prevent it.

nouveau/src/main/java/org/apache/couchdb/nouveau/NouveauApplicationConfiguration.java
rnewson: No, it doesn't. 10 is the default, but @Min prevented me setting it lower than. I made this change to speed up testing as nouveau peer checkpoints only advance after the commit

src/couch/src/couch_bt_engine.erl
rnewson: it's so you can call _update_drop_seq repeatedly without getting an error (if you make no other changes that would increment NewDropSeq). I think it might be ok to make this an error but I don't see much need to do so.

src/couch/src/couch_db.erl
rnewson: ken will not call `/dbname/_update_drop_seq`, only users will. ken will update indexes and those updates will update peer checkpoint docs, but that's all.

test/elixir/test/drop_seq_test.exs
rnewson: no, split_all_shard_ranges calls _reshard/jobs and those jobs run async. wait_for_reshard_jobs_to_complete then waits for the reshard jobs to complete. ;)

src/couch_mrview/src/couch_mrview_index.erl
iilyak: Here is the implementation of `write_header`](https://github.com/apache/couchdb/blob/4adf245b99d5ac2d8ff7f5f49821c4911c6d7688/src/couch/src/couch_file.erl#L435C1-L440C69)

```erlang
write_header(Fd, Data) ->
    Bin = ?term_to_bin(Data),
    Checksum = generate_checksum(Bin),
    % now we assemble the final header binary and write to disk
    FinalBin = <<Checksum/binary, Bin/binary>>,
    ioq:call(Fd, {write_header, FinalBin}, erlang:get(io_priority)).
```

As you can see the call reaches the `couch_file`'s `gen_server` via `ioq`.


The [`sync`](https://github.com/apache/couchdb/blob/4adf245b99d5ac2d8ff7f5f49821c4911c6d7688/src/couch/src/couch_file.erl#L301C1-L307C9) doesn't go via `ioq`

```erlang
sync(Fd) ->
    case gen_server:call(Fd, sync, infinity) of
        ok ->
            ok;
        {error, Reason} ->
            erlang:error({fsync_error, Reason})
    end.
```

Here is what I propose

```erlang
commit_header(Fd, Data) ->
    Bin = ?term_to_bin(Data),
    Checksum = generate_checksum(Bin),
    % now we assemble the final header binary and write to disk
    FinalBin = <<Checksum/binary, Bin/binary>>,
    ioq:call(Fd, {commit_header, FinalBin}, erlang:get(io_priority)).

...

handle_call({commit_header, _Bin}, From, #file{fd = Fd} = File) ->
    sync(Fd),
    Result = handle_call({write_header, Bin}, From, File),
    sync(Fd),
    Result;
```

iilyak: This implementation replaces three calls to gen_server with a single call and ensure durable write of the header.

src/fabric/src/fabric_drop_seq.erl
```diff
@@ -0,0 +1,1020 @@
...
+        #{},
+        ShardSyncHistory
+    ).
+
+%% A shard may have been split since a peer saw it.
+-spec substitute_splits([#shard{}], uuid_map(), peer_checkpoints()) -> peer_checkpoints().
+substitute_splits(Shards, UuidMap, PeerCheckpoints) ->
+    maps:fold(
+        fun({[PS, PE], Node}, {Uuid, Seq}, Acc) ->
+            ShardsInRange = [
```
iilyak, line 400: We could re-use this functionality from mem3.

```erlang
%% mem3.erl

filter_shards_by_range(Range, Shards) ->
    mem3_shards:filter_shards_by_range(Range, Shards)

%% mem3_shards.erl

-export([filter_shards_by_range/2]).
```

src/fabric/src/fabric_drop_seq.erl
```diff
@@ -0,0 +1,1020 @@
...
+                        Acc1,
+                        Others
+                    );
+                true ->
+                    maps:fold(
+                        fun({R, _S, T}, History, Acc2) ->
+                            case
+                                lists:search(
+                                    fun({SU, SS, _TU, _TS}) ->
+                                        uuids_match([Uuid, SU]) andalso SS =< Seq
```
iilyak: Minor optimization. It would be faster if you would swap the order of checks.

```erlang
SS =< Seq andalso uuids_match([Uuid, SU])
```

src/fabric/src/fabric_drop_seq.erl
rnewson: hm, I don't think it's the same calculation. for splits I want to find the ranges that are subsets of the peer checkpoints range, it wouldn't be enough just for them to overlap.

src/fabric/src/fabric_drop_seq.erl
rnewson: that's a good idea. it's the `SS =< Seq` condition that matters ("find the most recent checkpoint that includes the seq I care about"), the uuid check is a safeguard.

src/couch_mrview/src/couch_mrview_index.erl
rnewson: hm, ok. I think that's a separate PR to this effort though, as couch_bt_engine:commit_data does the same thing as I do here.

iilyak: https://github.com/apache/couchdb/compare/introduce-commit_header?expand=1

I am waiting for CI.... to open PR

iilyak: The https://github.com/apache/couchdb/pull/5576 is ready for review

rnewson:
noting we may need a downgrade option as this work adds two items to the db_header (drop_seq and drop_count)

src/couch/src/couch_bt_engine.erl
janl: ah, I didn’t mean to raise an error, but to not write a new header if the drop seqs are the same because presumably that value already made it to a header the last time we were here. but no bother if edge case.

src/couch/src/couch_db.erl
janl: ah boo I meant smoosh for when we no longer require folks to call the endpoint manually

rnewson:
closing in favor of approach in https://github.com/apache/couchdb/pull/5655/. might revisit later if needed.

